Issue #13, opened 2018-01-18
"The parser"


nikomatsakis:
It'd be worth covering where the lexer is, how the parser works, conventions around error recovery.

Michael-F-Bryan:
I'd be interested in helping out with the parser section. I haven't made any contributions to the Rust parser specifically, but I've played around with `syn` and proc macros and also done a fair amount of parsing for other projects (toy projects, custom file formats at work, etc).

Do you know if there's anyone with experience in the parser that I'd be able to bounce questions off?

nikomatsakis:
@Michael-F-Bryan you can certainly bounce questions off of me; I would say the main *owner* of the parser is @petrochenkov

nikomatsakis:
Copying some stuff I wrote in #21 that seems relevant:

> What things would someone need to know when either working on the parser or trying to use the parser

Hmm, good question. I imagine it'd be good to look at a few PRs that modified the parser to see what they had to do. Actually, it might be worth just including a link to those PRs in the guide to help people get oriented.

rust-lang/rust#45047 is such a PR, and here is a [link to some of the relevant content](https://github.com/rust-lang/rust/pull/45047/files#diff-da9d34ca1d0f6beee2838cf02e07345cR5197). 

Some things immediately jump out as worth explaining:

- What is this `eat` function? 
    - It consumes a token, and queues up the next token as lookahead
- How do you handle identifiers and keywords?
    - Actually, this PR didn't have to do anything around that. I'd have to find a good example (cc @petrochenkov -- thoughts?)
- What about constructing spans?
    - I have to bring this back into cache, but in general there is this `last_span` thing that stores the span of the last token, and many things [follow the idiom of saving the "lo" point of the span, parsing some stuff, then extracting the "hi" point and combining them](https://github.com/rust-lang/rust/blob/834674fa1d97f0e83baf99e1da5e24b82d41f63c/src/libsyntax/parse/parser.rs#L1114-L1117). This would be used to make a span that encompasses, for example, an entire `trait` definition (the lo point would come from the `trait` keyword, but the end point comes after having parsed a bunch of other things).
- What is this [crazy pretty printer](https://github.com/rust-lang/rust/pull/45047/files#diff-c82811a081b5fb219ba68dd88b49d0b1R1384)?
   - It dumps out the AST in human readable form. 
   - We have tests that will parse each file in run-pass, pretty print it, and then test that the resulting pretty printed file type checks, so you have to implement this.

[...]

PR #21, opened 2018-01-20
WIP: Started working on the Parser

Michael-F-Bryan:
This is the first instalment towards #13. I'm planning to do a quick high-level overview of the parser and `libsyntax` and mentioning the major players (`Parser`, `ParseSess`, etc) and what they do.

### DIFF
--- a/src/the-parser.md
+++ b/src/the-parser.md
@@ -1,1 +1,14 @@
-# The parser
+# The Parser
+
+The parser is responsible for converting raw Rust source code into a structured
+form which is easier for the compiler to work with, usually called an *Abstract
+Syntax Tree*. The bulk of the parser lives in the [libsyntax] crate.
+
+The parsing process is made up of roughly 3 stages,
+
+- lexical analysis - turn a stream of characters into a stream of token trees
+- macro expansion - run `proc-macros` and expand `macro_rules` macros
+- parsing - turn the token trees into an AST
+
+
+[libsyntax]: https://github.com/rust-lang/rust/tree/master/src/libsyntax
--- a/src/the-parser.md
+++ b/src/the-parser.md
@@ -1,14 +1,42 @@
 # The Parser
 
 The parser is responsible for converting raw Rust source code into a structured
-form which is easier for the compiler to work with, usually called an *Abstract
-Syntax Tree*. The bulk of the parser lives in the [libsyntax] crate.
+form which is easier for the compiler to work with, usually called an [*Abstract
+Syntax Tree*][ast]. An AST mirrors the structure of a Rust program in memory, 
+using a `Span` to link a particular AST node back to its source text.
 
-The parsing process is made up of roughly 3 stages,
+The bulk of the parser lives in the [libsyntax] crate.
+
+Like most parsers, the parsing process is composed of two main steps,
 
 - lexical analysis - turn a stream of characters into a stream of token trees
-- macro expansion - run `proc-macros` and expand `macro_rules` macros
 - parsing - turn the token trees into an AST
 
+The `syntax` crate contains several main players,
+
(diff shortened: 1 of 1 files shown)


src/the-parser.md
```diff
@@ -1 +1,14 @@
...
+
+The parser is responsible for converting raw Rust source code into a structured
+form which is easier for the compiler to work with, usually called an *Abstract
+Syntax Tree*. The bulk of the parser lives in the [libsyntax] crate.
+
+The parsing process is made up of roughly 3 stages,
+
+- lexical analysis - turn a stream of characters into a stream of token trees
+- macro expansion - run `proc-macros` and expand `macro_rules` macros
+- parsing - turn the token trees into an AST
```
nikomatsakis: Here I think I would link to the relevant chapter of the guide -- oh, wait, do we not have a chapter for the AST? We should add one =)

src/the-parser.md
```diff
@@ -1 +1,14 @@
-# The parser
+# The Parser
+
+The parser is responsible for converting raw Rust source code into a structured
+form which is easier for the compiler to work with, usually called an *Abstract
+Syntax Tree*. The bulk of the parser lives in the [libsyntax] crate.
```
nikomatsakis: Link to wikipedia? Maybe pointless?

src/the-parser.md
Michael-F-Bryan: I was thinking the AST would fall under the parser section, maybe as a short sub-chapter.

Michael-F-Bryan:
I just added another commit which points out a couple of (what I perceive to be) the key players in `libsyntax`. I wasn't really sure what to link to when mentioning the specific modules/types, so I linked to their source files.

What things would someone need to know when either working on the parser or trying to use the parser (e.g. because you're using `rustc` as a library)? Also, are there any easy-ish issues relating to the parser I can work on to get a better feel for `libsyntax` and how it is used?

nikomatsakis:
>  I wasn't really sure what to link to when mentioning the specific modules/types, so I linked to their source files.

Yeah, that's tricky. For now let's do that. Eventually rustdoc would probably be better.

> What things would someone need to know when either working on the parser or trying to use the parser

Hmm, good question. I imagine it'd be good to look at a few PRs that modified the parser to see what they had to do. Actually, it might be worth just including a link to those PRs in the guide to help people get oriented.

rust-lang/rust#45047 is such a PR, and here is a [link to some of the relevant content](https://github.com/rust-lang/rust/pull/45047/files#diff-da9d34ca1d0f6beee2838cf02e07345cR5197). 

Some things immediately jump out as worth explaining:

- What is this `eat` function? 
    - It consumes a token, and queues up the next token as lookahead
- How do you handle identifiers and keywords?
    - Actually, this PR didn't have to do anything around that. I'd have to find a good example (cc @petrochenkov -- thoughts?)
- What about constructing spans?
    - I have to bring this back into cache, but in general there is this `last_span` thing that stores the span of the last token, and many things [follow the idiom of saving the "lo" point of the span, parsing some stuff, then extracting the "hi" point and combining them](https://github.com/rust-lang/rust/blob/834674fa1d97f0e83baf99e1da5e24b82d41f63c/src/libsyntax/parse/parser.rs#L1114-L1117). This would be used to make a span that encompasses, for example, an entire `trait` definition (the lo point would come from the `trait` keyword, but the end point comes after having parsed a bunch of other things).
- What is this [crazy pretty printer](https://github.com/rust-lang/rust/pull/45047/files#diff-c82811a081b5fb219ba68dd88b49d0b1R1384)?
   - It dumps out the AST in human readable form. 
   - We have tests that will parse each file in run-pass, pretty print it, and then test that the resulting pretty printed file type checks, so you have to implement this.

Issue #16, opened 2018-01-18
"name resolution"


nikomatsakis:
We should be talking about the `librustc_resolve` crate works here. Some topics of interest:

- Handling scopes within the crate
   - What is a rib?
   - Where do ribs get added / removed, what happens if I need to introduce a binder?
   - How does `pub use` and stuff like that work
   - Globs and ambiguity
- How do names from external crates fit in?

nikomatsakis:
@vorner sounds good to me! One of the things I've been wrestling with, as an aside, is just how much detail this guide should go into. I don't think it should replace "in-code" comments -- that is, we don't want the material to become outdated very regularly. We'll have to feel our way.

To that end, I was thinking that perhaps a good place to start would be to dive into the code and try to add comments explaining how it works. This will of course be kind of "bottom up" -- at some point then we should be able to decide what of that material should go into guide, which should go into the code itself.

Anyway, let me know how things are going. I'll try to check in from time to time.

vorner:
My idea was to do the comments in code (if I find places where the code isn't self-documenting and find them missing) and then try to write a high level picture with very little details. I guess it'll need some more iteration later on, and details can always be added.

nikomatsakis:
@vorner sounds perfect

mark-i-m:
Ping @vorner ?

Did you wish to continue iterating on this chapter?

mark-i-m:
Ping @vorner ?

vorner:
> Did you wish to continue iterating on this chapter?

Sorry, I was busy and missed the previous ping :-|.

Yes, I still want to have a second go through the code, put comments on there and polish the text in the chapter and possibly add some more details.

But if I'm taking too long, I'd definitely understand if someone else wants to take it over (though it seems there's plenty of other work available).

mark-i-m:
No worries!

I just was making sure you were still interested :)

demurgos:
Thank you for your work on this guide, it's really helpful.
I am currently trying to understand how name resolution works with cycles between types so it would be great if this was expanded. I think it relates to the following parts:

> There are some exceptions to this. Items are bit tricky, because they can be used even before encountered ‒ therefore every block needs to be first scanned for items to fill in its Rib.
>
> Other, even more problematic ones, are imports which need recursive fixed-point resolution and macros, that need to be resolved and expanded before the rest of the code can be processed.

mark-i-m:
Did you ever find out anything about this?

[...]

PR #22, opened 2018-01-21
The first approximation of name resolution

vorner:
The first attempt to write something useful about the name resolution.
As the TODO section says, his is not finished thing, but it might
hopefully be useful to someone already.

This is part of #16, but this wouldn't be complete enough to call that one done. My next plan is to do a more thorough read through the code than just a fast scan, annotating the undocumented items with comments (what is the convention, is it OK to put doc comments on private items?) and then try to improve this.

Apart from the possible usefulness for others, I already wrote this incomplete thing to make sure I don't forget what I've already learned ‒ I don't know when I'll have the time to do another pass (which will take longer than the first scan).

Few questions:
* Should this go into master in the current form, so others (be it authors of the other chapters ‒ I think could have used some other chapters if they existed already, or people trying to get their head around the compiler landscape) can have something to start, or should it be kept separately so any inaccuracies don't confuse them?
* As I wrote this only after a scan of the code, I have to ask: is there something completely wrong, or is my impression more or less based on reality?
* Is there an easy way to generate docs for just the one crate? I currently get `use of unstable library feature 'rustc_private': mainly needed for compiler internals (see issue #27812)` from libfmt_macros. I compile with nightly, so it is a bit strange.

### DIFF
--- a/src/name-resolution.md
+++ b/src/name-resolution.md
@@ -1,1 +1,90 @@
 # Name resolution
+
+The name resolution is a separate pass in the compiler. Its input is the syntax
+tree, produced by parsing input files. It produces links from all the names in
+the source to relevant places where the name was introduced. It also generates
+helpful error messages, like typo suggestions or traits to import.
+
+The name resolution lives in the `librustc_resolve` crate, with the meat in
+`lib.rs` and some helpers or symbol-type specific logic in the other modules.
+
+## Namespaces
+
+Different kind of symbols live in different namespaces ‒ eg. types don't
+clash with variables. This usually doesn't happen, because variables start with
+lower-case letter while types with upper case one, but this is only a
+convention. This is legal Rust code that'll compile (with warnings):
+
+```rust
+type x = u32;
+let x: x = 1;
+let y: x = 2; // See? x is still a type here.
+```
+
+To cope with this, and with slightly different scoping rules for these
+namespaces, the resolver keeps them separated and builds separate structures for
+them.
+
+In other words, when the code talks about namespaces, it doesn't mean the module
+hierarchy, it's types vs. values vs. macros.
+
+## Scopes and ribs
+
+A name is visible only in certain area in the source code. This forms a
+hierarchical structure, but not necessarily a simple one ‒ if one scope is part
+of another, it doesn't mean the name visible in the outer one is also visible in
+the inner one, or that it refers to the same thing.
+
+To cope with that, the compiler introduces the concept of Ribs. This is
+abstraction of a scope. Every time the set of visible names potentially changes,
+a new rib is pushed onto a stack. The places where this can happen includes for
+example:
+
+* The obvious places ‒ curly braces enclosing a block, function boundaries,
+  modules.
+* Introducing a let binding ‒ this can shadow another binding with the same
+  name.
+* Macro expansion border ‒ to cope with macro hygiene.
+
+When searching for a name, the stack of ribs is traversed from the innermost
+outwards. This helps to find the closest meaning of the name (the one not
+shadowed by anything else). The transition to outer rib may also change the
+rules what names are usable ‒ if there are nested functions (not closures), the
+inner one can't access parameters and local bindings of the outer one, even
+though they should be visible by ordinary scoping rules.
+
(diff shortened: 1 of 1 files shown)


nikomatsakis:
@vorner 

> what is the convention, is it OK to put doc comments on private items?)

yes that is ok

> Should this go into master in the current form

yep, why wait?

src/name-resolution.md
```diff
@@ -1 +1,90 @@
...
+let x: x = 1;
+let y: x = 2; // See? x is still a type here.
+```
+
+To cope with this, and with slightly different scoping rules for these
+namespaces, the resolver keeps them separated and builds separate structures for
+them.
+
+In other words, when the code talks about namespaces, it doesn't mean the module
+hierarchy, it's types vs. values vs. macros.
```
nikomatsakis: Seems good. I think we want to avoid trying to teach too must Rust here, but some bits are no doubt unavoidable. This note strikes a good balance.

src/name-resolution.md
```diff
@@ -1 +1,90 @@
...
+* Introducing a let binding ‒ this can shadow another binding with the same
+  name.
+* Macro expansion border ‒ to cope with macro hygiene.
+
+When searching for a name, the stack of ribs is traversed from the innermost
+outwards. This helps to find the closest meaning of the name (the one not
+shadowed by anything else). The transition to outer rib may also change the
+rules what names are usable ‒ if there are nested functions (not closures), the
+inner one can't access parameters and local bindings of the outer one, even
+though they should be visible by ordinary scoping rules.
```
nikomatsakis: I think here it'd be helpful to have an example -- just show some rust code and the set of ribs in scope at some particular point. No need to dive into fancy features, just a generic impl + function will have plenty of ribs to get the point across.

src/name-resolution.md
```diff
@@ -1 +1,90 @@
...
+
+## Overall strategy
+
+To perform the name resolution of the whole crate, the syntax tree is traversed
+top-down and every encountered name is resolved. This works for most kinds of
+names, because at the point of use of a name it is already introduced in the Rib
+hierarchy.
+
+There are some exceptions to this. Items are bit tricky, because they can be
+used even before encountered ‒ therefore every bock needs to be first scanned
```
nikomatsakis: Nit: s/bock/block/?

src/name-resolution.md
```diff
@@ -1 +1,90 @@
...
+resolution and macros, that need to be resolved and expanded before the rest of
+the code can be processed.
+
+Therefore, the resolution is performed in multiple stages.
+
+## TODO:
+
+This is a result of the first pass of learning the code. It is definitely
+incomplete and not detailed enough. It also might be inaccurate in places.
+Still, it probably provides useful first guidepost to what happens in there.
```
nikomatsakis: This seems great. It's already hitting on some of the tensions of this book, of course -- I would love it if we had a way to link into the code, so that e.g. we could show some code that is setting up ribs, and then be notified if we got an error later. But I think it's good as is.

nikomatsakis:
ping @vorner -- think you'll have a chance to address feedback here?

vorner:
>  think you'll have a chance to address feedback here?

I hope to get to it in few hours today. I didn't have the time during the week.

nikomatsakis:
@vorner seems good, did you want to squash?

src/name-resolution.md
```diff
@@ -1 +1,107 @@
...
+To cope with that, the compiler introduces the concept of Ribs. This is
+abstraction of a scope. Every time the set of visible names potentially changes,
+a new rib is pushed onto a stack. The places where this can happen includes for
+example:
+
+* The obvious places ‒ curly braces enclosing a block, function boundaries,
+  modules.
+* Introducing a let binding ‒ this can shadow another binding with the same
+  name.
+* Macro expansion border ‒ to cope with macro hygiene.
```
mark-i-m: I'm guessing pattern matching and globals too?

mark-i-m: Also, do generics (`impl<T>`) and lifetimes(`impl<'a>`) follow this rule?

PR #23, opened 2018-01-23
Copy contents of README.md from librustc/hir

theotherphil:
Literally just copy-pasting (s/README/chapter/) @nikomatsakis's README from librustc.

Not sure if it's worth landing this as-is or if we should wait until I've added some more content.

### DIFF
--- a/src/hir-lowering.md
+++ b/src/hir-lowering.md
@@ -1,1 +1,119 @@
 # HIR lowering
+
+The HIR -- "High-level IR" -- is the primary IR used in most of
+rustc. It is a desugared version of the "abstract syntax tree" (AST)
+that is generated after parsing, macro expansion, and name resolution
+have completed. Many parts of HIR resemble Rust surface syntax quite
+closely, with the exception that some of Rust's expression forms have
+been desugared away (as an example, `for` loops are converted into a
+`loop` and do not appear in the HIR).
+
+This chapter covers the main concepts of the HIR.
+
+### Out-of-band storage and the `Crate` type
+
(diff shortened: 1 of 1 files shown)


nikomatsakis:
I say let's add it! I'd love to stand this book up as far as possible as a useful thing to look at.

src/hir-lowering.md
```diff
@@ -1 +1,119 @@
...
+For example, the contents of individual items (e.g., modules,
+functions, traits, impls, etc) in the HIR are not immediately
+accessible in the parents. So, for example, if had a module item `foo`
+containing a function `bar()`:
+
+```
+mod foo {
+  fn bar() { }
+}
+```
```
nikomatsakis: I wonder if we can start testing these code samples. That might .. actually be a way for us to test links too? Gonna take a bit of work to make that easy I suppose.

theotherphil: A potentially useful starting point: https://github.com/budziq/rust-skeptic/blob/master/README.md

Michael-F-Bryan: Now version `0.1.0` of `mdbook`, we might be able to use my [mdbook-test] backend. It's essentially gets a copy of the document from `mdbook` and then runs `rust-skeptic` over it. 

I found it worked pretty well locally but the CI tests were quite flaky and would fail for unknown reasons. It also needs to be updated to pull `mdbook` from crates.io instead of the git repository.

[mdbook-test]: https://github.com/Michael-F-Bryan/mdbook-test

PR #26, opened 2018-01-25
Start macro expansion chapter

mark-i-m:
I just went through this code to implement `?` macro repetition, so I thought I could take a stab at the chapter :smile:


mark-i-m:
@nikomatsakis I don't really know anything about hygiene, proc macros, or custom derive, but I added a bit about macros-by-example, and left TODOs for the rest...

src/macro-expansion.md
```diff
@@ -1 +1,66 @@
 # Macro expansion
+
+Macro expansion happens during parsing. `rustc` has two parsers, in fact: the
+normal Rust parser, and the macro parser. During the parsing phase, the normal
+Rust parser will call into the macro parser when it encounters a macro. The
```
nikomatsakis: can you be more precise about what a reference to a macro is? e.g. ,do you mean a macro invocation, like `foo!(...)`?

src/macro-expansion.md
```diff
@@ -1 +1,66 @@
...
+macro parser, in turn, may call back out to the Rust parser when it needs to
+bind a metavariable (e.g. `$my_expr`). There are a few aspects of this system to
```
nikomatsakis: here, you mean when the macro is trying to parse the contents of the macro invocation against one of the macro arms?

src/macro-expansion.md
nikomatsakis: Also, is it really called from the *parser*? I thought there was a second phase that came after parsing, but maybe I'm going to learn something here =)

src/macro-expansion.md
```diff
@@ -1 +1,66 @@
...
+be explained. The code for macro expansion is in `src/libsyntax/ext/tt/`.
+
+### The macro parser
+
+Basically, the macro parser is like an NFA-based regex parser. It uses an
+algorithm similar in spirit to the [Earley parsing
+algorithm](https://en.wikipedia.org/wiki/Earley_parser). The macro parser is
+defined in `src/libsyntax/ext/tt/macro_parser.rs`.
```
nikomatsakis: can we make links into GH here (master branch)? this at least allows us to detect if those links rot

src/macro-expansion.md
```diff
@@ -1 +1,66 @@
 # Macro expansion
+
+Macro expansion happens during parsing. `rustc` has two parsers, in fact: the
+normal Rust parser, and the macro parser. During the parsing phase, the normal
+Rust parser will call into the macro parser when it encounters a macro. The
+macro parser, in turn, may call back out to the Rust parser when it needs to
+bind a metavariable (e.g. `$my_expr`). There are a few aspects of this system to
+be explained. The code for macro expansion is in `src/libsyntax/ext/tt/`.
```
nikomatsakis: can we make links into GH here (master branch)? this at least allows us to detect if those links rot

src/macro-expansion.md
```diff
@@ -1 +1,66 @@
...
+
+### The macro parser
```
nikomatsakis: as a meta-comment, I think it's a good idea to start out with some kind of concrete example and walk it through. For example:

Imagine we have a macro

```rust
macro_rules! foo {
    ($metavariable:tt) => { ... }
}
```

now you can reference this example from the text below

src/macro-expansion.md
```diff
@@ -1 +1,66 @@
...
+
+Basically, the macro parser is like an NFA-based regex parser. It uses an
+algorithm similar in spirit to the [Earley parsing
+algorithm](https://en.wikipedia.org/wiki/Earley_parser). The macro parser is
+defined in `src/libsyntax/ext/tt/macro_parser.rs`.
+
+In a traditional NFA-based parser, one common approach is to have some pattern
+which we are trying to match an input against. Moreover, we may try to capture
+some portion of the input and bind it to variable in the pattern. For example:
+suppose we have a pattern (borrowing Rust macro syntax) such as `a $b:ident a`
```
nikomatsakis: e.g., this example would reference back to your running example.

src/macro-expansion.md
```diff
@@ -1 +1,66 @@
...
+token. Given an input `a foo a`, the _metavariable_ `$b` would bind to the
+`ident` `foo`. On the other hand, an input `a foo b` would be rejected as a
+parse failure because the pattern `a <ident> a` cannot match `a foo b` (or as
+the compiler would put it, "no rules expected token `b`").
+
+The macro parser does pretty much exactly that with one exception: in order to
+parse different types of metavariables, such as `ident`, `block`, `expr`, etc.,
+the macro parser must sometimes call back to the normal Rust parser.
+
+Interestingly, both definitions and invokations of macros are parsed using the
```
nikomatsakis: Spelling: invo*c*ations

src/macro-expansion.md
```diff
@@ -1 +1,66 @@
...
+parse macro _definitions_ is in `src/libsyntax/ext/tt/macro_rules.rs`. It
+defines the pattern for matching for a macro definition as `$( $lhs:tt =>
+$rhs:tt );+`. In other words, a `macro_rules` defintion should have in its body
+at least one occurence of a token tree followed by `=>` followed by another
+token tree. When the compiler comes to a `macro_rules` definition, it uses this
+pattern to match the two token trees per rule in the definition of the macro
+_using the macro parser itself_.
+
+When the compiler comes to a macro invokation, it needs to parse that
+invokation. This is also known as _macro expansion_. The same NFA-based macro
```
nikomatsakis: Spelling: invocation

src/macro-expansion.md
```diff
@@ -1 +1,66 @@
...
+parser is used that is described above. Notably, the "pattern" (or _matcher_)
+used is the first token tree extracted from the rules of the macro _definition_.
```
nikomatsakis: this is where the running example would be really handy

src/macro-expansion.md
```diff
@@ -1 +1,66 @@
...
+In other words, given some pattern described by the _definition_ of the macro,
+we want to match the contents of the _invokation_ of the macro.
```
nikomatsakis: Spelling: invocation

src/macro-expansion.md
```diff
@@ -1 +1,66 @@
...
+
+The algorithm is exactly the same, but when the macro parser comes to a place in
+the current matcher where it needs to match a _non-terminal_ (i.e. a
+metavariable), it calls back to the normal Rust parser to get the contents of
+that non-terminal. Then, the macro parser proceeds in parsing as normal.
+
+For more information about the macro parser's implementation, see the comments
+in `src/libsyntax/ext/tt/macro_parser.rs`.
```
nikomatsakis: link to repo

federicomenaquintero:
BTW, there's a very interesting discussion about hygiene and proc-macro in https://github.com/rust-lang/rust/issues/45934

src/macro-expansion.md
mark-i-m: Let me verify that :)

mark-i-m: @nikomatsakis 

Ok, so it looks like
- the macro parser is called from `macro_rules::compile` https://github.com/rust-lang/rust/blob/ca9cf3594ab25d2809ac576dfc9defb8e87b45b8/src/libsyntax/ext/tt/macro_rules.rs#L184 which transforms a macro invocation into a syntax extension.
- `macro_rules::compile` is called from `librustc_resolve/macros.rs` and `librustc_resolve/build_reduced_graph.rs`... which I'm guessing is doing name resolution? Does this run after the parser?

mark-i-m:
@nikomatsakis I updated the chapter (a lot). I think I have addressed your comments. Let me know. Thanks!

---

Also, copying this here, because the comment above is "outdated":


> Also, is it really called from the parser? I thought there was a second phase that came after parsing, but maybe I'm going to learn something here =)


Ok, so it looks like
- the macro parser is called from `macro_rules::compile` https://github.com/rust-lang/rust/blob/ca9cf3594ab25d2809ac576dfc9defb8e87b45b8/src/libsyntax/ext/tt/macro_rules.rs#L184 which transforms a macro invocation into a syntax extension.
- `macro_rules::compile` is called from `librustc_resolve/macros.rs` and `librustc_resolve/build_reduced_graph.rs`... which I'm guessing is doing name resolution? Does this run after the parser?

mark-i-m:
@federicomenaquintero Would you be interested in filling in some of the TODOs? I want to learn how they all work, but I don't have the bandwidth in the near future...

nikomatsakis:
> Does this run after the parser?

Yes, it does

src/macro-expansion.md
```diff
@@ -1 +1,154 @@
...
+    (print twice $mvar:ident) => {
+        println!("{}", $mvar);
+        println!("{}", $mvar);
+    }
+}
+```
+
+`$mvar` is called a _metavariable_. Unlike normal variables, rather than binding
+to a value in a computation, a metavariable binds _at compile time_ to a tree of
+_tokens_. A _token_ zero or more symbols that together have some meaning. For
```
nikomatsakis: > A _token_ zero or more symbols that together have some meaning.

This sentence is not grammatical and I'm not quite sure how to fix it. =) In particular, I don't think of a token as "zero or more symbols" (and it's sort of unclear to me what you mean by symbol, which in parsing terminology is often used to mean the union of token and nonterminal).

I think I would maybe say something like this:

"A _token_ is a single "unit" of the grammar, such as an identifier (e.g., `print`) or punctuation (e.g., `=>`). Token trees resulting from paired parentheses-like characters (`(...)`, `[...]`, and `{...}`) -- they include the open and close and all the tokens in between (we do require that parentheses-like characters be balanced)."

but it doesn't seem like the best either :)

src/macro-expansion.md
```diff
@@ -1 +1,154 @@
...
+
+Whenever we refer to the "example _invocation_", we mean the following snippet:
+
+```rust
+printer!(print foo); // Assume `foo` is a variable defined somewhere else...
+```
+
+The process of expanding the macro invocation into the syntax tree
+`println!("{}", foo)` and then expanding that into a call to `Display::fmt` is
+called _macro expansion_, it is the topic of this chapter.
```
nikomatsakis: Nit: the word `it` is not needed here

src/macro-expansion.md
```diff
@@ -1 +1,154 @@
...
+- `tts` is a stream of tokens. The macro parser's job is to consume the raw
+  stream of tokens and output a binding of metavariables to corresponding token
+  trees.
+- `ms` a _matcher_. This is a sequence of token trees that we want to match
+  `tts` against.
+
+In the analogy of a regex parser, `tts` is the input and we are matching it
+against the pattern `ms`. Using our examples, `tts` could be the stream of
+tokens containing the inside of the example invocation `print foo`, while `ms`
+might be the sequence of token (trees) `print $mvar:ident`.
```
nikomatsakis: tying back to the example is 💯

mark-i-m:
@nikomatsakis I updated the paragraph on tokens, as you suggested... I am wondering

1. Is this a term we should add to the glossary?
2. Should the parsing chapter go into more detail about lexing and parsing or rely on external sources for the basics? If it went into more detail, maybe we can just assume that the reader knows about tokens from that chapter and omit the paragraph from this chapter altogether?

Michael-F-Bryan:
> Should the parsing chapter go into more detail about lexing and parsing or rely on external sources for the basics? 

I was asking myself this exact question when I wrote the start of the parser chapter. Should we add a small note up the top saying we assume people know how a basic recursive descent parser works and what tokenizing/lexical analysis is? The idea being this is a book about `rustc` internals, not a book an introduction to parsing. 

There is already loads of good quality material on basic parsers on the internet, a couple paragraphs at the top of the chapter probably wouldn't be able to do it justice.

mark-i-m:
I agree that we shouldn't try to teach parsing here, but given that I don't expect most people to know basic parsing, I worry that it would discourage contributions... Perhaps we can

-

mark-i-m:
Erg... Sorry, I fat-fingered the "close and comment" button.... Updated my post above

federicomenaquintero:
> @federicomenaquintero Would you be interested in filling in some of the TODOs? I want to learn how they all work, but I don't have the bandwidth in the near future...

Yes, I'll see what I can do.

Michael-F-Bryan:
> What do you think?

Sounds like a good idea. We could say something like this:

> Rust syntax is specified by a grammar (link) which is essentially a list of rules where each rule specifies how a piece of the language is written (e.g. a crate contains multiple items, an item may be a function declaration, a function has a bunch of statements, a statement is a ...), with each rule being written in terms of other rules or terminals (the base case, typically tokens). 
>
> Generally speaking, for each grammar rule there will be one parser method. In this way we can translate a token stream into an AST by recursively calling the appropriate method.

It's essentially recursive descent 101,  but you could tie all of this back to `rustc` by inspecting a sample code snippet (e.g. an `if` statement) and then showing what would be called when parsing it.

nikomatsakis:
> I agree that we shouldn't try to teach parsing here, but given that I don't expect most people to know basic parsing, I worry that it would discourage contributions... Perhaps we can

I think I agree with both of you. I don't think we want a lot of introductory material; a few links don't hurt, but not too much. But I think there's a third way, though it may take some iteration to get there: To some extent, I think you can serve both audiences by doing a kind of "walk through" of the code. 

In other words, e.g. to explain tokenizing, we might point to the token data structure and give some source showing how it would be divided into tokens (we can always link to wikipedia or something too). This way, if you know what a token is, you learn about the *Rust-specific parts of it*. If you don't know what a token is, you can just understand it as this Rsut data structure and later learn about the more general form. 

Similarly I imagine we can say something like "Rust has a recursive-descent parser" (where we link to wikipedia) and then walk through how it would parse some small example, showing a few key functions (eg., the one that parses a type). If you're not familiar with recursive descent, this will basically give you the idea, but if you *are*, then you'll learn about the names of key concepts in the code.

PR #28, opened 2018-01-29
add query + incremental section and restructure a bit

nikomatsakis:


nikomatsakis:
cc @michaelwoerister -- this includes a rough description of incremental which is probably wrong in many of the particulars

PR #39, opened 2018-02-01
Various minor improvements to text (typos, grammar, style)

alexreg:


alexreg:
Incidentally, I saw the *Macro expansion* section just got added. Nice work. I hope the hygiene section gets written soon, as I could really use that. :-)

src/how-to-build-and-run.md
```diff
@@ -70,15 +70,15 @@ Once you've created a config.toml, you are now ready to run
 probably the best "go to" command for building a local rust:
 
 ```
-./x.py build -i --stage 1 src/libstd
+./x.py build -i –stage 1 src/libstd
```
mark-i-m: Umm...

src/how-to-build-and-run.md
```diff
@@ -70,15 +70,15 @@ Once you've created a config.toml, you are now ready to run
...
 ```
 
 What this command will do is the following:
 
 - Using the beta compiler (also called stage 0), it will build the
   standard library and rustc from the `src` directory. The resulting
   compiler is called the "stage 1" compiler.
-  - During this build, the `-i` (or `--incremental`) switch enables incremental
+  - During this build, the `-i` (or `–incremental`) switch enables incremental
```
mark-i-m: Same here

src/how-to-build-and-run.md
```diff
@@ -129,9 +129,9 @@ LLVM version: 4.0
...
-  - `./x.py build --stage 1` -- builds everything using the stage 1 compiler, not just up to libstd
-  - `./x.py build` -- builds the stage2 compiler
+  - `./x.py clean` – clean up the build directory (`rm -rf build` works too, but then you have to rebuild LLVM)
+  - `./x.py build –stage 1` – builds everything using the stage 1 compiler, not just up to libstd
+  - `./x.py build` – builds the stage2 compiler
 - Running tests (see the section [running tests](./running-tests.html) for more details):
-  - `./x.py test --stage 1 src/libstd` -- runs the `#[test]` tests from libstd
-  - `./x.py test --stage 1 src/test/run-pass` -- runs the `run-pass` test suite
+  - `./x.py test –stage 1 src/libstd` – runs the `#[test]` tests from libstd
+  - `./x.py test –stage 1 src/test/run-pass` – runs the `run-pass` test suite
```
mark-i-m: And all of these...

src/macro-expansion.md
```diff
@@ -132,8 +132,8 @@ algorithm is exactly the same, but when the macro parser comes to a place in the
 current matcher where it needs to match a _non-terminal_ (e.g. `$mvar:ident`),
 it calls back to the normal Rust parser to get the contents of that
 non-terminal. In this case, the Rust parser would look for an `ident` token,
-which it finds (`foo`) and returns to the macro parser. Then, the macro parser
-proceeds in parsing as normal. Also, note that exactly one of the matchers from
+which it finds (`foo`), and returns to the macro parser. Then, the macro parser
```
mark-i-m: "returns to the macro parser" is doesn't have a subject and a verb, so I think there should be no comma.

src/macro-expansion.md
```diff
@@ -132,8 +132,8 @@ algorithm is exactly the same, but when the macro parser comes to a place in the
...
+proceeds in parsing as normal. Also, note that only one of the matchers from
```
mark-i-m: I think `exactly` is more precise. `Only` could be taken as 0 or 1.

alexreg: The problem is with the "otherwise" clause here. Indeed, "exactly" is more precise", but it doesn't make sense when you have the "otherwise", since the "otherwise" clause is not predicated on there being zero matches.

mark-i-m: Ah, I see.

How about changing it to: if there is more than one match, the parse is ambiguous. If there are no matches at all, there is a syntax error.

alexreg: Sounds good!

src/macro-expansion.md
alexreg: Yeah, I think you're right actually. The `foo` in brackets just broke up the flow a bit. Will revert.

src/how-to-build-and-run.md
alexreg: Overzealous find & replace. :-P

alexreg:
Should all be in order now, @mark-i-m.

src/how-to-build-and-run.md
```diff
@@ -129,9 +129,9 @@ LLVM version: 4.0
 Here are a few other useful x.py commands. We'll cover some of them in detail in other sections:
 
 - Building things:
-  - `./x.py clean` -- clean up the build directory (`rm -rf build` works too, but then you have to rebuild LLVM)
-  - `./x.py build --stage 1` -- builds everything using the stage 1 compiler, not just up to libstd
-  - `./x.py build` -- builds the stage2 compiler
+  - `./x.py clean` – clean up the build directory (`rm -rf build` works too, but then you have to rebuild LLVM)
+  - `./x.py build –stage 1` – builds everything using the stage 1 compiler, not just up to libstd
```
mark-i-m: Missed one :P

alexreg: Done.

PR #40, opened 2018-02-03
Adding ICH to the glossary.

avborhanian:


mark-i-m:
Thanks! Would it be possible add an extra sentence like "ICHs are used for __"?

avborhanian:
Sure, how about

ICHs are used as fingerprints for things such as HIR and crate metadata, to see if things hashed are equal or not.

mark-i-m:
That's great! Thanks :)

I would even add just a bit more to that: "ICHs are used as fingerprints for things such as HIR and crate metadata, to check if changes have been made. This is useful in incremental compilation to see if part of a crate has changed and should be recompiled."

mark-i-m:
(correct me if I'm wrong... I don't know much about incr. comp. :p)

avborhanian:
Nah, that sounds perfect. I'll update the commit with that extra line.-

mark-i-m:
Thanks! r=me when Travis I'm happy :)

PR #44, opened 2018-02-05
Add DAG to glossary

phansch:
The DAG is mentioned on the [incremental compilation](https://rust-lang-nursery.github.io/rustc-guide/incremental-compilation.html#the-basic-algorithm) page, I guess it makes sense to have it in the glossary, too?


mark-i-m:
That's a good idea 👍 

A couple of comments:
- Please move the entry after `cx`, so it stays alphabetically sorted.
- Instead of "to index which queries execute which other queries", perhaps "to keep track of dependencies between queries" might be clearer?

phansch:
Thank you, I fixed those 👍

PR #46, opened 2018-02-07
Explain how to dump HIR

theotherphil:


src/hir.md
```diff
@@ -10,6 +10,14 @@ been desugared away (as an example, `for` loops are converted into a
 
 This chapter covers the main concepts of the HIR.
 
+You can view the HIR representation of your code by passing the
+`-Zunpretty=hir-tree` flag to rustc, for example by setting the `RUSTFLAGS`
+environment variable:
+
+```
+RUSTFLAGS=-Zunpretty=hir-tree cargo build
+```
```
nikomatsakis: Nit: `cargo rustc -- -Zunpretty=hir-tree` would be easier in any case.

nikomatsakis:
Good idea to mention it.

src/hir.md
theotherphil: Ah, I didn't realise you could do this.

theotherphil: Ah, I didn't realise you could do this.

Issue #48, opened 2018-02-09
Glossary


mark-i-m:
Are there any other terms that should be added to the glossary? Please make suggestions.

- [x] filemap (#35)
- [x] soundness (#54)
- [x] completeness (#54)
- [x] skolemization (#67)
- [x] sigil (#57)
- [x] LTO, ThinLTO, FatLTO (#56)
- [x] TLS (#56)
- [x] LLVM (#56)
- [x] IR (#56)
- [x] Fat pointer (#112)
- [x] region (#57)
- [x] HIR map (#63)
- [x] CTFE (compile-time function evaluation)... This just needs a link to the const eval/miri sections (#152)
- [x] rib (see name resolution chapter) (#152)
- [x] upvar (I believe an "upvar" is just a variable captured by a closure) (#152)
- [x] early-bound lifetime (#281)
- [x] interned (#281)
- [x] item (#281)
- [x] late-bound lifetime (#281)
- [x] lang item (#281)
- [x] memoized (#281)
- [x] string interning (#281)
- [x] drop glue (#281)
- [x] late-bound regions/lifetimes (#281)
- [ ] ~gensym~
- [ ] landing pad
- [ ] ~skolemization leak~

phansch:
One thing I have been wondering is if 'soundness' means something special in the compiler context or if it is just a different word for correctness. There are [lot's of issues](https://github.com/rust-lang/rust/search?q=soundness&type=Issues&utf8=%E2%9C%93) mentioning 'soundness bugs' and 'soundness fixes'. So maybe it makes sense to add this too or explain it somewhere?

mark-i-m:
Yes, soundness is a technical term in programming languages literature. You will usually read about proving "soundness" and "completeness".

Roughly, IIRC, soundness means that if a program type-checks, it is type-safe; i.e. I can never (in safe rust) force a value into a variable if the wrong type. Completeness means that every type-safe program also type-checks. Having both is very hard, and usually soundness is more important.

samWson:
I've found a good definition of _Fat Pointers_ in the Programming Rust book. I can enter that in the glossary this weekend if it's OK with you team.

steveklabnik:
We've been using "double pointer" or "wide pointer" as well; might be worth considering as "double pointer" at least is significantly more descriptive, IMO.

rkruppe:
[...]

PR #57, opened 2018-02-15
Add a few more terms to the glossary

mark-i-m:
cc #48


mark-i-m:
I would appreciate some confirmation on these definitions.

PR #60, opened 2018-02-18
Fix Github Pages path.

ehuss:
The newer version of mdBook uses a different path for the html output.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,6 @@ deploy:
   provider: pages
   skip-cleanup: true
   github-token: $GITHUB_TOKEN
-  local-dir: book
+  local-dir: book/html
   on:
     branch: master


.travis.yml
```diff
@@ -16,6 +16,6 @@ deploy:
   provider: pages
   skip-cleanup: true
   github-token: $GITHUB_TOKEN
-  local-dir: book
+  local-dir: book/html
```
mark-i-m: Hmm... Maybe it should be just `html` instead? I'm judging by https://github.com/rust-lang-nursery/rustc-guide/tree/gh-pages

Also, does this mean we have to go around fixing _every link in the book_?

mark-i-m: Hmm... no, it might be `book/html` after all. But I think we will need to go fix all of the links...

mark-i-m: Hmm... no, after looking at the travis docs, I think I'm wrong all together...

Issue #10, opened 2018-01-18
"Using the compiler testing framework"


nikomatsakis:
This section should describe how to compiler is tested. Substantial work has been done but there remains more to document:

- [x] discuss bors, travis, platform emulation (#66 )
- [ ] crater: what is it and why is it useful? How do you request a crater run and when should you to do so? (#151)
- [ ] maybe talk about "perf runs" too (#151)
- [x] rustdoc tests (cc @QuietMisdreavus)
- [x] talk about unit tests from libs and things (#62)
- [x] Doctests: How to run the doc tests in the standard library.  Other doc testing steps like linkchecker. (#62)
- [x] distcheck (#62)
- [x] mention that tests also get run for individual packages, such as bootstrap, cargo, rls, rustfmt, miri, clippy, cargotest. (#62)

mark-i-m:
See #47 and #53

mark-i-m:
@QuietMisdreavus #47 has landed :)

PR #62, opened 2018-02-23
Add some more information about the various tests that get run.

ehuss:
CC #10

### DIFF
--- a/src/tests/intro.md
+++ b/src/tests/intro.md
@@ -1,17 +1,18 @@
-# Using the compiler testing framework
+# The compiler testing framework
 
-The compiler has an extensive testing framework, masterminded by the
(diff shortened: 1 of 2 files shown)


mark-i-m:
I saw the TODOs. Do you mean for this to be WIP?

ehuss:
> I saw the TODOs. Do you mean for this to be WIP?

Not exactly.  I wanted to block out where some of the rest of #10 could be filled out.  I'm going to start looking into the infrastructure more to maybe fill out that section in a separate PR, but if someone else more knowledgeable gets to it, that would be great!  As for Crater, I think someone else who actually uses it will need to fill it out.

mark-i-m:
@ehuss Sounds good! Thanks for the PR :)

Issue #50, opened 2018-02-09
Code index


mark-i-m:
One thing that could be helpful is an index of important data structures, functions, methods, traits, etc in the compiler. I propose we add a chapter after the glossary with a big table (sorted alphabetically by item name) that looks something like this:

Item |  Kind  | Short description | Chapter | Declaration
--------|------|-----------------------------|--------------------|-------------------
`foo` | function | used by the parser to keep track bar | [Parsing](the-parser.html) | [src/libsyntax/foo.rs](https://github.com/rust-lang/rust/blob/master/src/libsyntax/abi.rs)

TODO:
- [x] `TyCtxt` (#55)
- [x] `Ty` (#85)
- [x] `DefId`, `HirId`, `NodeId`, `BodyId` (#153 )
- [x] `CodeMap` (#69)
- [x] `StringReader` (#72)
- [x] `ParseSess` (#68)
- [x] `Crate` (#76)
- [x] `TokenStream` (#153)
- [x] `Rib` (#153)
- [x] `ParamEnv` (#153)
- [x] `TraitRef` (#153)
- [x] `Span` (#139)
- [x] `P<T>` (as mentioned in #182) (#275)
- [ ] `FileMap`
- [ ] `Parser`

mark-i-m:
A good first step on this would be to crawl the existing chapters and look for data structures that are explicitly mentioned.

mark-i-m:
I'm going to go ahead and close this issue, as it doesn't seem to be useful any more...

PR #72, opened 2018-03-03
Add info about StringReader for #50

vishalsodani:


vishalsodani:
@mark-i-m I don't have much understanding how the compiler works. I found this [walkthrough](https://github.com/gchp/gchp.ie/blob/master/_drafts/rust-compiler-walkthrough-parsing.markdown). Should I add this line 

> This is the interface which consumes the raw source input and converts it to a series of Tokens & Spans in a TokenStream.

mark-i-m:
@vishalsodani I think that looks good. Perhaps we could modify it as such:

> This is the lexer used during parsing. It consumes characters from the raw source code being compiled and produces a series of tokens for use by the rest of the parser.

PR #86, opened 2018-03-13
Add the rustdoc readme

mark-i-m:
cc @QuietMisdreavus @Michael-F-Bryan 

After this merges, I hope to remove the README from librustdoc so that we can have everything in one place.

### DIFF
--- a/src/appendix-code-index.md
+++ b/src/appendix-code-index.md
@@ -8,6 +8,7 @@ Item            |  Kind    | Short description           | Chapter            | 
 ----------------|----------|-----------------------------|--------------------|-------------------
 `CodeMap` | struct | The CodeMap maps the AST nodes to their source code | [The parser] | [src/libsyntax/codemap.rs](https://github.com/rust-lang/rust/blob/master/src/libsyntax/codemap.rs)
 `CompileState` | struct | State that is passed to a callback at each compiler pass | [The Rustc Driver] | [src/librustc_driver/driver.rs](https://github.com/rust-lang/rust/blob/master/src/librustc_driver/driver.rs)
+`DocContext` | struct | A state container used by rustdoc when crawling through a crate to gather its documentation | [Rustdoc] | [src/librustdoc/core.rs](https://github.com/rust-lang/rust/blob/master/src/librustdoc/core.rs)
 `ast::Crate` | struct | Syntax-level representation of a parsed crate | [The parser] | [src/librustc/hir/mod.rs](https://github.com/rust-lang/rust/blob/master/src/libsyntax/ast.rs)
 `hir::Crate` | struct | More abstract, compiler-friendly form of a crate's AST | [The Hir] | [src/librustc/hir/mod.rs](https://github.com/rust-lang/rust/blob/master/src/librustc/hir/mod.rs)
 `ParseSess` | struct | This struct contains information about a parsing session | [the Parser] | [src/libsyntax/parse/mod.rs](https://github.com/rust-lang/rust/blob/master/src/libsyntax/parse/mod.rs)
--- a/src/appendix-code-index.md
+++ b/src/appendix-code-index.md
@@ -21,3 +21,4 @@ Item            |  Kind    | Short description           | Chapter            | 
 [The parser]: the-parser.html
 [The Rustc Driver]: rustc-driver.html
 [The `ty` modules]: ty.html
+[Rustdoc]: rustdoc.html
--- a/src/rustdoc.md
+++ b/src/rustdoc.md
@@ -116,9 +116,9 @@ Here is current (as of this writing) list of passes:
   markdown to like it. This is necessary because the convention for writing
   documentation is to provide a space between the `///` or `//!` marker and the
   text, and stripping that leading space will make the text easier to parse by
-  the Markdown parser. (In my experience it's less necessary now that we have a
-  Commonmark-compliant parser, since it doesn't have a problem with headers
-  that have a space before the `##` that marks the heading.)
+  the Markdown parser. (In the past, the markdown parser used was not Commonmark-
+  compliant, which caused annoyances with extra whitespace but this seems to be 
+  less of an issue today.)
 - `strip-priv-imports` strips all private import statements (`use`, `extern
   crate`) from a crate. This is necessary because rustdoc will handle *public*
   imports by either inlining the item's documentation to the module or creating
--- a/src/rustdoc.md
+++ b/src/rustdoc.md
@@ -5,9 +5,17 @@ library. This chapter is about how it works. (A new implementation is also [unde
 
 [under way]: https://github.com/steveklabnik/rustdoc
 
-Rustdoc is implemented entirely within the crate `librustdoc`. After partially compiling a crate to
-get its AST (technically the HIR map) from rustc, librustdoc performs two major steps past that to
-render a set of documentation:
+Rustdoc is implemented entirely within the crate [`librustdoc`][rd]. It runs
+the compiler up to the point where we have an internal representation of a
+crate (HIR) and the ability to run some queries about the types of items. [HIR]
+and [queries] are discussed in the linked chapters.
+
+[HIR]: ./hir.html
+[queries]: ./query.html
+[rd]: https://github.com/rust-lang/rust/tree/master/src/librustdoc
+
+`librustdoc` performs two major steps after that to render a set of
+documentation:
 
 * "Clean" the AST into a form that's more suited to creating documentation (and slightly more
   resistant to churn in the compiler).
@@ -16,9 +24,11 @@ render a set of documentation:
 Naturally, there's more than just this, and those descriptions simplify out lots of details, but
 that's the high-level overview.
 
-(Side note: this is a library crate! The `rustdoc` binary is crated using the project in
-`src/tools/rustdoc`. Note that literally all that does is call the `main()` that's in this crate's
+(Side note: `librustdoc` is a library crate! The `rustdoc` binary is crated using the project in
+[`src/tools/rustdoc`][bin]. Note that literally all that does is call the `main()` that's in this crate's
 `lib.rs`, though.)
+
+[bin]: https://github.com/rust-lang/rust/tree/master/src/tools/rustdoc
 
 ## Cheat sheet
 
@@ -87,12 +97,31 @@ These do things like combine the separate "attributes" into a single string and 
 whitespace to make the document easier on the markdown parser, or drop items that are not public or
 deliberately hidden with `#[doc(hidden)]`. These are all implemented in the `passes/` directory, one
 file per pass. By default, all of these passes are run on a crate, but the ones regarding dropping
-private/hidden items can be bypassed by passing `--document-private-items` to rustdoc.
+private/hidden items can be bypassed by passing `--document-private-items` to rustdoc. Note that
+unlike the previous set of AST transformations, the passes happen on the _cleaned_ crate.
 
 (Strictly speaking, you can fine-tune the passes run and even add your own, but [we're trying to
 deprecate that][44136]. If you need finer-grain control over these passes, please let us know!)
 
 [44136]: https://github.com/rust-lang/rust/issues/44136
+
+Here is current (as of this writing) list of passes:
+
+- `collapse-docs` is necessary because each line of a doc comment is given as a
+  separate doc attribute, and this will combine them into a single string with
+  line breaks between each attribute.
+- `unindent-comments` is necessary because the convention for writing
+  documentation is to provide a space between the `///` or `//!` marker and the
+  text, and stripping that leading space will make the text easier to parse by
+  the Markdown parser. (In my experience it's less necessary now that we have a
+  Commonmark-compliant parser, since it doesn't have a problem with headers
+  that have a space before the `##` that marks the heading.)
+- `strip-priv-imports` is necessary because rustdoc will handle *public*
+  imports by either inlining the item's documentation to the module or creating
+  a "Reexports" section with the import in it. The pass ensures that all of
+  these imports are actually relevant to documentation.
+- `strip-hidden` and `strip-private` also remove items that are not relevant
+  for public documentation.
 
 ## From clean to crate
 
@@ -159,6 +188,9 @@ crate walk to grab *just* the hand-written documentation. Combined with the afor
 handing them off to the libtest test runner. One notable location in `test.rs` is the function
 `make_test`, which is where hand-written doctests get transformed into something that can be
 executed.
+
+Some extra reading about `make_test` can be found
+[here](https://quietmisdreavus.net/code/2018/02/23/how-the-doctests-get-made/).
 
 ## Dotting i's and crossing t's
 
--- a/src/rustdoc.md
+++ b/src/rustdoc.md
@@ -87,7 +87,7 @@ The other major thing that happens in `clean/mod.rs` is the collection of doc co
 `#[doc=""]` attributes into a separate field of the Attributes struct, present on anything that gets
 hand-written documentation. This makes it easier to collect this documentation later in the process.
 
-The primary output of this process is a clean::Crate with a tree of Items which describe the
+The primary output of this process is a `clean::Crate` with a tree of Items which describe the
 publicly-documentable items in the target crate.
 
 ### Hot potato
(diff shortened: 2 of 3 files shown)


mark-i-m:
I should note that (except for the first sentence) this is a straight copy/paste of what @QuietMisdreavus wrote up.

mark-i-m:
cc #2

src/rustdoc.md
```diff
@@ -0,0 +1,177 @@
+# The walking tour of rustdoc
+
+Rustdoc actually uses the rustc internals directly. It lives in-tree with the compiler and standard
+library. This chapter is about how it works. (A new implementation is also [under way], though).
+
+[under way]: https://github.com/steveklabnik/rustdoc
+
+Rustdoc is implemented entirely within the crate `librustdoc`. After partially compiling a crate to
+get its AST (technically the HIR map) from rustc, librustdoc performs two major steps past that to
```
QuietMisdreavus: > partially compiling a crate to get its AST (technically the HIR map)

Is this a valid thing to say? The complete story of what librustdoc does is that it runs "phase 3" of the compilation process to get a `TyCtxt` and query the compiler about type information that it can't get from the source code. But saying that it compiles the crate to its HIR felt like a simpler/smoother thing to say in this context.

src/rustdoc.md
```diff
@@ -0,0 +1,177 @@
...
+calls out to `html/markdown.rs` which interfaces with the Markdown parser. This is exposed as a
+series of types that wrap a string of Markdown, and implement `fmt::Display` to emit HTML text. It
+takes special care to enable certain features like footnotes and tables and add syntax highlighting
+to Rust code blocks (via `html/highlight.rs`) before running the Markdown parser. There's also a
+function in here (`find_testable_code`) that specifically scans for Rust code blocks so the
+test-runner code can find all the doctests in the crate.
+
+### From soup to nuts
+
+(alternate title: ["An unbroken thread that stretches from those first `Cell`s to us"][video])
```
QuietMisdreavus: When i first wrote this, i waffled about whether or not to include this reference. My eventual reasoning was that READMEs weren't as formal a context as proper crate documentation, so it was okay to sneak this pun in. Now that the document's being moved here, is this worth keeping?

src/rustdoc.md
```diff
@@ -0,0 +1,177 @@
...
+to `html/markdown.rs`, optionally including a mode which inserts a Table of Contents to the output
+HTML.
+
+For the latter, rustdoc runs a similar partial-compilation to get relevant documentation in
+`test.rs`, but instead of going through the full clean and render process, it runs a much simpler
+crate walk to grab *just* the hand-written documentation. Combined with the aforementioned
+"`find_testable_code`" in `html/markdown.rs`, it builds up a collection of tests to run before
+handing them off to the libtest test runner. One notable location in `test.rs` is the function
+`make_test`, which is where hand-written doctests get transformed into something that can be
+executed.
```
QuietMisdreavus: I have [a longer writeup about `make_test`](https://quietmisdreavus.net/code/2018/02/23/how-the-doctests-get-made/) if y'all feel it's worth linking here.

src/rustdoc.md
Michael-F-Bryan: What about saying something along the lines of "it runs the compiler up to the point where we have an internal representation of a crate (HIR) and the ability to query the type system (via `TyCtxt`)"?

I'm not sure referring to the `TyCtxt` as "the type system" is actually correct, but it helps give you the general idea.

src/rustdoc.md
Michael-F-Bryan: Its probably a good idea to mention it. I just skimmed through your article and learner a lot about how rustdoc generates tests. It's really quite smart.

src/rustdoc.md
Michael-F-Bryan: I like the light-hearted title :grin: 

It's kinda hard to put a simple name to this section, so I'm not sure what I'd replace it with if we were to change it. We're essentially trying to exfiltrate information from the guts of the compiler without breaking things due to lifetimes and the various other restrictions.

src/rustdoc.md
```diff
@@ -0,0 +1,177 @@
+# The walking tour of rustdoc
+
+Rustdoc actually uses the rustc internals directly. It lives in-tree with the compiler and standard
+library. This chapter is about how it works. (A new implementation is also [under way], though).
+
+[under way]: https://github.com/steveklabnik/rustdoc
+
+Rustdoc is implemented entirely within the crate `librustdoc`. After partially compiling a crate to
```
Michael-F-Bryan: Link to the crate root on GitHub?

src/rustdoc.md
```diff
@@ -0,0 +1,177 @@
...
+* The Markdown renderer is loaded up in `html/markdown.rs`, including functions for extracting
+  doctests from a given block of Markdown.
+* The tests on rustdoc *output* are located in `src/test/rustdoc`, where they're handled by the test
+  runner of rustbuild and the supplementary script `src/etc/htmldocck.py`.
+* Tests on search index generation are located in `src/test/rustdoc-js`, as a series of JavaScript
+  files that encode queries on the standard library search index and expected results.
+
+## From crate to clean
+
+In `core.rs` are two central items: the `DocContext` struct, and the `run_core` function. The latter
```
Michael-F-Bryan: It may be worth adding `DocContext` to the code index at the end of the guide.

src/rustdoc.md
```diff
@@ -0,0 +1,177 @@
...
+pass is mainly to get a few intermediate wrappers around the HIR types and to process visibility
+and inlining. This is where `#[doc(inline)]`, `#[doc(no_inline)]`, and `#[doc(hidden)]` are
+processed, as well as the logic for whether a `pub use` should get the full page or a "Reexport"
+line in the module page.
+
+The other major thing that happens in `clean/mod.rs` is the collection of doc comments and
+`#[doc=""]` attributes into a separate field of the Attributes struct, present on anything that gets
+hand-written documentation. This makes it easier to collect this documentation later in the process.
+
+The primary output of this process is a clean::Crate with a tree of Items which describe the
```
Michael-F-Bryan: Should this be `clean::Crate`?

src/rustdoc.md
```diff
@@ -0,0 +1,177 @@
...
+publicly-documentable items in the target crate.
+
+### Hot potato
+
+Before moving on to the next major step, a few important "passes" occur over the documentation.
+These do things like combine the separate "attributes" into a single string and strip leading
```
Michael-F-Bryan: If we're listing the various rustdoc passes, would it be worth using bullet points? E.g.

- collate and clean doc comments
- remove `[doc(hidden)]` items
- ...

src/rustdoc.md
```diff
@@ -0,0 +1,177 @@
...
+whitespace to make the document easier on the markdown parser, or drop items that are not public or
+deliberately hidden with `#[doc(hidden)]`. These are all implemented in the `passes/` directory, one
+file per pass. By default, all of these passes are run on a crate, but the ones regarding dropping
+private/hidden items can be bypassed by passing `--document-private-items` to rustdoc.
```
Michael-F-Bryan: Ooh, that's cool! I wonder if you can use it from `cargo`.

src/rustdoc.md
```diff
@@ -0,0 +1,177 @@
...
+
+(Strictly speaking, you can fine-tune the passes run and even add your own, but [we're trying to
+deprecate that][44136]. If you need finer-grain control over these passes, please let us know!)
+
+[44136]: https://github.com/rust-lang/rust/issues/44136
+
+## From clean to crate
+
+This is where the "second phase" in rustdoc begins. This phase primarily lives in the `html/`
+folder, and it all starts with `run()` in `html/render.rs`. This code is responsible for setting up
```
Michael-F-Bryan: Do you reckon this file would be noteworthy enough to warrant a link to its page on GitHub?

src/rustdoc.md
QuietMisdreavus: Not directly. You'd have to use `$RUSTDOCFLAGS` or `cargo rustdoc` to add additional arguments to rustdoc.

src/rustdoc.md
QuietMisdreavus: Here's the current list of passes for rustdoc, as given by a kinda-recent nightly:

```console
$ rustdoc +nightly --passes list
WARNING: the 'passes' flag is considered deprecated
WARNING: please see https://github.com/rust-lang/rust/issues/44136
Available passes for running rustdoc:
        strip-hidden - strips all doc(hidden) items from the output
   unindent-comments - removes excess indentation on comments in order for markdown to like it
       collapse-docs - concatenates all document attributes into one document attribute
       strip-private - strips all private items from a crate which cannot be seen externally, implies strip-priv-imports
  strip-priv-imports - strips all private import statements (`use`, `extern crate`) from a crate
   propagate-doc-cfg - propagates `#[doc(cfg(...))]` to child items

Default passes for rustdoc:
        strip-hidden
       strip-private
       collapse-docs
   unindent-comments
   propagate-doc-cfg
```

`collapse-docs` is necessary because each line of a doc comment is given as a separate doc attribute, and this will combine them into a single string with line breaks between each attribute. `unindent-comments` is necessary because the convention for writing documentation is to provide a space between the `///` or `//!` marker and the text, and stripping that leading space will make the text easier to parse by the Markdown parser. (In my experience it's less necessary now that we have a Commonmark-compliant parser, since it doesn't have a problem with headers that have a space before the `##` that marks the heading.) `strip-priv-imports` is necessary because rustdoc will handle *public* imports by either inlining the item's documentation to the module or creating a "Reexports" section with the import in it. The pass ensures that all of these imports are actually relevant to documentation. `strip-hidden` and `strip-private` also remove items that are not relevant for public documentation.

Passing `--document-private-items` disables all the passes and only runs `collapse-docs` and `unindent-comments`, which means that everything in the crate gets documented, not just the public items.

src/rustdoc.md
```diff
@@ -0,0 +1,177 @@
...
+
+The primary output of this process is a clean::Crate with a tree of Items which describe the
+publicly-documentable items in the target crate.
+
+### Hot potato
+
+Before moving on to the next major step, a few important "passes" occur over the documentation.
+These do things like combine the separate "attributes" into a single string and strip leading
+whitespace to make the document easier on the markdown parser, or drop items that are not public or
+deliberately hidden with `#[doc(hidden)]`. These are all implemented in the `passes/` directory, one
```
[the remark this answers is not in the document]
QuietMisdreavus: The key difference between the "I lied a littler earlier" and the "Hot potato" paragraphs is that of timing: The `visit_ast.rs` processing happens *before* crate cleaning, but the passes run *after* it (but before page rendering).

src/rustdoc.md
QuietMisdreavus: To be fair, the entire document is full of my usual style of "vague headings that only kinda relate to their section via some extended metaphor" blog-post writing. The specific link to the video is because of the `Cell` pun - `DocContext` and `SharedContext` and `Cache` are full of `RefCell`s because of how they're passed around. "From soup to nuts" is a variation of the idea that the thing being discussed stretches across the whole documentation process, and won out because it fits the rhythm i established with "From crate to clean" and "From clean to crate". Whether or not to keep the video link is up to the accumulated style of the rest of guide, which is an editorial decision i'm not prepared to make. `>_>`

mark-i-m: I personally like the titles :)

We can change the later if others want a more formal tone...

src/rustdoc.md
mark-i-m: Also, totally unrelated, but that gives me an idea for how to build a simple REPL for rust :P

src/rustdoc.md
mark-i-m: No, I believe it is this one: https://github.com/rust-lang/rust/blob/master/src/librustdoc/clean/mod.rs#L132

src/rustdoc.md
mark-i-m: Hmm... it would be a bit weird IMHO since there are a lot of other files mentioned, and none of them got special treatment...

src/rustdoc.md
mark-i-m: I added this list (verbatim from your comment) to the chapter, but I think we should avoid documenting some of these hidden cmd line flags since they seem unstable...

mark-i-m:
@Michael-F-Bryan @QuietMisdreavus I think I addressed all of the comments. Let me know if I missed anything.

src/rustdoc.md
Michael-F-Bryan: Oh no. I meant there should be back ticks around the word so it gets styles as code.

Michael-F-Bryan:
Looks good to me. @QuietMisdreavus, do you think this is good to merge?

src/appendix-code-index.md
```diff
@@ -8,6 +8,7 @@ Item            |  Kind    | Short description           | Chapter            |
 ----------------|----------|-----------------------------|--------------------|-------------------
 `CodeMap` | struct | The CodeMap maps the AST nodes to their source code | [The parser] | [src/libsyntax/codemap.rs](https://github.com/rust-lang/rust/blob/master/src/libsyntax/codemap.rs)
 `CompileState` | struct | State that is passed to a callback at each compiler pass | [The Rustc Driver] | [src/librustc_driver/driver.rs](https://github.com/rust-lang/rust/blob/master/src/librustc_driver/driver.rs)
+`DocContext` | struct | A state container used by rustdoc when crawling through a crate to gather its documentation | [Rustdoc] | [src/librustdoc/core.rs](https://github.com/rust-lang/rust/blob/master/src/librustdoc/core.rs)
```
QuietMisdreavus: `[Rustdoc]` doesn't have a link target.

src/rustdoc.md
```diff
@@ -87,13 +97,32 @@ These do things like combine the separate "attributes" into a single string and
...
+unlike the previous set of AST transformations, the passes happen on the _cleaned_ crate.
 
 (Strictly speaking, you can fine-tune the passes run and even add your own, but [we're trying to
 deprecate that][44136]. If you need finer-grain control over these passes, please let us know!)
 
 [44136]: https://github.com/rust-lang/rust/issues/44136
 
+Here is current (as of this writing) list of passes:
+
+- `collapse-docs` is necessary because each line of a doc comment is given as a
```
QuietMisdreavus: The structure of this list a little jarring because it's introduced as "here are the passes" but it's written as "why are they here". Specifically, starting off with "(pass) is necessary..." is awkward because you haven't even described what the pass *does* yet.

src/rustdoc.md
QuietMisdreavus: Yes, this should get backticks since the rest of the document uses those around other type names.

mark-i-m:
@QuietMisdreavus I think I have addressed your comments. I added a bit to each bullet to describe what it does first. Let me know what you think :)

src/rustdoc.md
```diff
@@ -107,21 +107,26 @@ deprecate that][44136]. If you need finer-grain control over these passes, pleas
...
   separate doc attribute, and this will combine them into a single string with
   line breaks between each attribute.
-- `unindent-comments` is necessary because the convention for writing
+- `unindent-comments` removes excess indentation on comments in order for
+  markdown to like it. This is necessary because the convention for writing
   documentation is to provide a space between the `///` or `//!` marker and the
   text, and stripping that leading space will make the text easier to parse by
   the Markdown parser. (In my experience it's less necessary now that we have a
   Commonmark-compliant parser, since it doesn't have a problem with headers
   that have a space before the `##` that marks the heading.)
```
QuietMisdreavus: I'm not sure this parenthetical is worth keeping? It's definitely not adding as-is, since the rest of the document has so far avoided first-person. Maybe having a historical note that the previous Markdown parser had a problem with spaces before a heading would be useful, but i'm less confident about the remark about it being "less necessary" since i'm not sure that Pulldown is totally okay with everything having an indent on it.

mark-i-m: Better?

QuietMisdreavus:
Looks good to me! Sorry about the back-and-forth there at the end. This has my :+1: now.

mark-i-m:
> Sorry about the back-and-forth there at the end.

Not at all! Thanks for taking the time :smile:

PR #93, opened 2018-03-16
Add mdbook search.

ehuss:
May want to wait for 0.1.6 to be released since there is a minor CSS issue, but I thought this was nifty!

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -6,3 +6,5 @@ description = "A guide to developing rustc "
 [output.html]
 
 [output.linkcheck]
+
+[output.html.search]
--- a/ci/install.sh
+++ b/ci/install.sh
@@ -20,4 +20,4 @@ function cargo_install() {
 }
 
 cargo_install mdbook 0.1.5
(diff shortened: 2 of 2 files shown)


mark-i-m:
Hmm... was waiting for 0.1.6, but I am beginning to kind of want this... should we still wait?

mark-i-m:
@Michael-F-Bryan any thoughts?

Michael-F-Bryan:
Well the CSS issues aren't super critical, they're more of an annoyance than anything else (rust-lang-nursery/mdBook#647) and are mostly solved by rust-lang-nursery/mdBook#648. I just want to test that PR locally before hitting merge, after which we'll probably do another release.

I don't see why we can't bump the `mdbook` version now, then bump it again when `mdbook` does another release. That search function is really really useful!

mark-i-m:
@ehuss could you rebase?

ehuss:
Updated!

mark-i-m:
Thanks!

PR #105, opened 2018-03-28
Update links

dlrobertson:
- `method-lookup`: Added link for UFCS.
 - `traits-bibliography`: Content expects a link for `[Theresa Swift]`, but there was none.

### DIFF
--- a/src/method-lookup.md
+++ b/src/method-lookup.md
@@ -12,7 +12,7 @@ the form:
 receiver.method(...)
 ```
 
-into a more explicit UFCS form:
+into a more explicit [UFCS] form:
 
 ```rust
 Trait::method(ADJ(receiver), ...) // for a trait call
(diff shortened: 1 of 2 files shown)


src/method-lookup.md
```diff
@@ -37,6 +37,7 @@ probe phase produces a "pick" (`probe::Pick`), which is designed to be
 cacheable across method-call sites. Therefore, it does not include
 inference variables or other information.
 
+[UFCS]: https://en.wikipedia.org/wiki/Uniform_Function_Call_Syntax
```
mark-i-m: Perhaps this might be a better link: https://github.com/rust-lang/rfcs/blob/master/text/0132-ufcs.md

mark-i-m:
Thanks!

dlrobertson:
Good suggestion. Updated.

Issue #115, opened 2018-04-17
Expand descriptions of the various compilation stages


frewsxcv:
This is what we currently have: https://github.com/rust-lang-nursery/rustc-guide/blob/46864ca9714707f18e658afba9d2ea62afbe2374/src/how-to-build-and-run.md#running-xpy-and-building-a-stage1-compiler

There's a relevant thread on reddit about what these stages all mean and I think the descriptions in [this answer](https://www.reddit.com/r/rust/comments/8covpo/why_does_rustc_compile_itself_three_times_during/dxgnj6m/) could be incorporated into this section in the rustc guide

ehsanmok:
I am interested in this as my very first PR. I'll prepare it over this weekend.

mark-i-m:
Thanks @frewsxcv and @ehsanmok :)

PR #117, opened 2018-04-22
Expand descriptions of compilation stages

ehsanmok:
[#115](https://github.com/rust-lang-nursery/rustc-guide/issues/115)

### DIFF
--- a/src/how-to-build-and-run.md
+++ b/src/how-to-build-and-run.md
@@ -2,7 +2,7 @@
 
 The compiler is built using a tool called `x.py`. You will need to
 have Python installed to run it. But before we get to that, if you're going to
-be hacking on rustc, you'll want to tweak the configuration of the compiler.
+be hacking on `rustc`, you'll want to tweak the configuration of the compiler.
 The default configuration is oriented towards running the compiler as a user,
 not a developer.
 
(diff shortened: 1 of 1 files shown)


src/how-to-build-and-run.md
```diff
@@ -55,13 +55,22 @@ compiler to compile the newer version. In particular, the newer version of the
...
-  download this binary from the internet.
-- **Stage 1:** the code in your clone is then compiled with the stage
-  0 compiler to produce the stage 1 compiler.
-- **Stage 2:** the code in your clone is then compiled with the stage
-  1 compiler *again* to produce the stage 2 compiler (i.e. it builds
-  itself).
+- **Stage 0:** the stage0 compiler can be your existing
+  (perhaps older version of)
+  Rust compiler, the current _beta_ compiler or you may download the binary
+  from the internet
```
mark-i-m: Nit: no period at end of sentence

Also, this seems a bit strange... how can you guarantee that the compiler already installed _can_ compile the nightly compiler?

src/how-to-build-and-run.md
```diff
@@ -55,13 +55,22 @@ compiler to compile the newer version. In particular, the newer version of the
...
+- **Stage 1:** the code in your clone (for new version)
+  is then compiled with the stage0
+  compiler to produce the stage1 compiler.
+  However, it was built with an older compiler (stage0),
+  so to optimize the stage1 compiler we go to next stage
```
mark-i-m: Nit: no period at end of sentence.

mark-i-m:
Thanks :)

src/how-to-build-and-run.md
ehsanmok: AFAIK the incremental bootstrapping process goes like this. Assume you have a language A and it's compiler "CA"
1) You extend A slightly to obtain B
2) You extend the compiler CA (to CB) so that it compiles B, but *without* using the added new features
3) Use CA to compile CB, then you (possibly after rewriting CB with the added features) compile CB with CB ...

[This](https://softwareengineering.stackexchange.com/questions/105313/how-could-the-first-c-compiler-be-written-in-c) helped me understand it *not only for the first compiler* you write for a new language but also how the bootstrap continues after that.

PR #127, opened 2018-05-08
update to mdbook 0.1.7

mark-i-m:


mark-i-m:
We might also need to drop travis caches...

Michael-F-Bryan:
Just to be sure, I blew away the travis caches for `master`. It's annoying because installing `mdbook` always takes forever in CI, but that's a once-off cost.

PR #128, opened 2018-05-10
Deleted <a> tag which was messing up the associated types page

sunjay:
This tag is extra and must have been added mistakenly. Results in the entire rest of the page getting underlined on mouse over for no reason.

### DIFF
--- a/src/appendix-background.md
+++ b/src/appendix-background.md
@@ -4,7 +4,7 @@ This section covers a numbers of common compiler terms that arise in
 this guide. We try to give the general definition while providing some
 Rust-specific context.
 
-<a name=cfg>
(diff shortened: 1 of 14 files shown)


sunjay:
Good catch! It turns out we do this everywhere for exactly the reasons you listed. The reason we never caught it before is because the `<a>` usually appears before a heading, so when the browser attempts to fix the unclosed tag, it doesn't add all of that in.

I went ahead and closed all of the `<a>` tags so we never run into a problem again in the future on any of the pages.

mark-i-m:
Thanks @sunjay !

mark-i-m:
r=me when travis passes :)

PR #129, opened 2018-05-10
Issue84 Add note about ./x.py build

rajcspsg:


Michael-F-Bryan:
Looks good to me :+1:

There isn't really much more to say about `./x.py check` seeing as it's quite similar to `cargo check`, which most people are already familiar with.

PR #135, opened 2018-05-20
Define HIR more specifically

alexcameron89:
IR is a foreign acronym to me, so having it fully expressed in the beginning as Intermediate Representation helps me comprehend the subject.

I also verified that the link works correctly.


alexcameron89:
Thanks @mark-i-m! It's passing now.

PR #136, opened 2018-05-21
Add type inference example

phansch:
This should make the chapter a bit more approachable, as it doesn't
start out with a reference to the HM type inference algorithm.

### DIFF
--- a/src/type-inference.md
+++ b/src/type-inference.md
@@ -1,5 +1,21 @@
 # Type inference
 
+Type inference is the process of automatic detection of the type of an
+expression.
+
+It is what allows Rust to work with fewer or no type annotations,
+making things easier for users:
+
+```rust,ignore
+fn main() {
+    let mut things = vec![];
+    things.push("thing")
+}
+```
+
+Here, `things` is *inferenced* to be `&str` because that's the value we push
+into `things`.
+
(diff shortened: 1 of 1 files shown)


phansch:
Thanks! I'm doing the same right now to potentially fix https://github.com/rust-lang/rust/issues/50548 at some point in the future :)

As for `inferenced` vs `inferred`, I'm not a native English speaker and `inferenced` seems to be used everywhere else in the guide. There's some discussion on [english.stackexchange.org](https://english.stackexchange.com/questions/123403/whats-the-difference-between-inferencing-and-inferring) which suggest that `inferencing` is used in linguistics, which could apply here as well, I guess.

alexcameron89:
Oh very interesting, thanks for sharing!

src/type-inference.md
```diff
@@ -1,5 +1,21 @@
...
+making things easier for users:
+
+```rust,ignore
+fn main() {
+    let mut things = vec![];
+    things.push("thing")
+}
+```
+
+Here, `things` is *inferenced* to be `&str` because that's the value we push
```
mark-i-m: Nit: "things is inferenced to be `&str`" -> "the type of things is ..."

phansch:
Fixed :+1:

Issue #14, opened 2018-01-18
"Diagnostics: reporting errors and lints"


nikomatsakis:
We should talk first about:

- Guidelines for emitting errors
    - How to find an error code and add an extended error
    - Conventions are how the error should be formatted (lowercase letter, etc)
        - link or maybe just inline [the RFC with guidelines on how extended errors should look](https://github.com/rust-lang/rfcs/blob/master/text/1567-long-error-codes-explanation-normalization.md)
    - when to use `span_label` etc
- How to emit lints
    - When to use a lint
    - If you have a tcx
    - Before you have a tcx
    - Some tips here at the end of [Rustc bug fix procedure](https://forge.rust-lang.org/rustc-bug-fix-procedure.html)

EDIT(mark-i-m): See https://github.com/rust-lang/rustc-dev-guide/issues/14#issuecomment-514932385

cc @estebank

Karrq:
Looks to me like this issue is currently missing the following:


Errors:
- How to find an error code and add an extended error
- Conventions are how the error should be formatted (lowercase letter, etc) 
- when to use `span_label` etc

Lints:
Nothing? Maybe the section about removing lints present in the [rustc Bug Fix Procedure](https://forge.rust-lang.org/rustc-bug-fix-procedure.html) could be added.

phansch:
It might also be nice to link to the [lint naming recommendations](https://rust-lang.github.io/rfcs/0344-conventions-galore.html#lints) from RFC0344

rylev:
Triage: closing. As we now have sections on diagnostics, this issue is a bit too general. I'm closing this in favor of more specific issues.

PR #139, opened 2018-05-22
Add info about emitting lints and errors

mark-i-m:
Since I was just working on this, I thought I would write something up.

We should wait to merge this until https://github.com/rust-lang/rust/pull/48309 merges.

r? @estebank @Manishearth 

cc #14 #50


src/diag.md
```diff
@@ -0,0 +1,177 @@
...
+[sptosnip]: https://doc.rust-lang.org/nightly/nightly-rustc/syntax/codemap/struct.CodeMap.html#method.span_to_snippet
+
+## Error messages
+
+The [`rustc_errors`][errors] crate defines most of the utilities used for
+reporting errors.
+
+[errors]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_errors/index.html
+
+Most "session"-like types in the compiler (e.g. [`Session`][session]) have
```
Manishearth: I think it's fine to just say Session and ParseSess here.

src/diag.md
```diff
@@ -0,0 +1,177 @@
...
+
+[spanerr]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc/session/struct.Session.html#method.span_err
+[strspanerr]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc/session/struct.Session.html#method.struct_span_err
+[diagbuild]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_errors/diagnostic_builder/struct.DiagnosticBuilder.html
+[emit]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_errors/diagnostic_builder/struct.DiagnosticBuilder.html#method.emit
+
+For example, to add a help message to an error, one might do:
+
+```rust,ignore
+let snip = sess.codemap().span_to_snippet(sp);
```
Manishearth: Needs an `unwrap_or`, span_to_snippet returns an Option

src/diag.md
```diff
@@ -0,0 +1,177 @@
...
+                        err.emit();
+                    }
+                }
+            }
+        }
+    }
+}
+```
+
+### Edition Lints
```
Manishearth: "Edition-gated lints", there are more kinds of edition lints.

src/diag.md
```diff
@@ -0,0 +1,177 @@
...
+For example, to add a help message to an error, one might do:
+
+```rust,ignore
+let snip = sess.codemap().span_to_snippet(sp);
+
+sess.struct_span_err(sp, "oh no! this is an error!")
+    .span_suggestion(other_sp, "try using a qux here", format!("qux {}", snip))
+    .emit();
+```
+
```
Manishearth: Could you also include some stuff on suggestion applicabilities here?

src/diag.md
```diff
@@ -0,0 +1,177 @@
...
+error[E0999]: oh no! this is an error!
+ --> mycode.rs:3:5
+  |
+3 |     sad()
+  |     ^ help: try using a qux here: `qux sad()`
+
+error: aborting due to previous error
+
+For more information about this error, try `rustc --explain E0999`.
+```
```
estebank: Add explanation: 

In some cases, like when the suggestion spans multiple lines or when there are multiple suggestions, the suggestions are displayed on its own:

```console
error[E0999]: oh no! this is an error!
 --> mycode.rs:3:5
  |
3 |     sad()
  |     ^
help: try using a qux here:
  |
3 |     qux sad()
  |     ^^^

error: aborting due to previous error

For more information about this error, try `rustc --explain E0999`.
```

src/diag.md
estebank: I don't want to block too much on this, once this PR is merged it can be expanded more iteratively.

src/diag.md
mark-i-m: > // add code to verify that `sp` is not a macro if this is a consideration <...expand explanation of this later...>

@estebank I didn't quite understand what you meant by this.

----

I think will just create a new section for suggestions since they will play a big role in the edition.

estebank: No problem, you can forego that comment. It is a caveat that we'll need to add at some point, as suggestions that do not verify wether the current context is mixing macros can lead to very incorrect suggestions (like in #39450). This is not a problem under normal circumstances, but adding a check for this in the documentation could go a long way. Don't need to add it now though.

mark-i-m:
@Manishearth @estebank Thanks for the feedback! I think I addressed all the comments.

mark-i-m:
https://github.com/rust-lang/rust/pull/48309 merged

PR #142, opened 2018-05-23
Fixes #141

### LINKED ISSUE #141: describe --bless in the testing section
I don't think we talk about the new `--bless` option in the testing section, do we? You can pass it to `./x.py test` and it will update the UI reference files. This replaces the old script we used to run. 

We should also talk about rustfix.

oli-obk:
cc @nikomatsakis

### DIFF
--- a/src/tests/adding.md
+++ b/src/tests/adding.md
@@ -273,7 +273,7 @@ If you have changed the compiler's output intentionally, or you are
 making a new test, you can pass `--bless` to the test subcommand. E.g.
 if some tests in `src/test/ui` are failing, you can run
 
-```
+```text
 ./x.py test --stage 1 src/test/ui --bless
 ```
 
--- a/src/tests/adding.md
+++ b/src/tests/adding.md
@@ -150,7 +150,8 @@ source.
 
 * `run-rustfix` for UI tests, indicates that the test produces
   structured suggestions, which are then applied and the final
-  source is compiled again.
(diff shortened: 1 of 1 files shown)


mark-i-m:
Actually, could you also add a bit on the run-rustfix header command and .fixed files?

oli-obk:
I have never used that personally and I'm not sure if it already works correctly (for some definition of correctly)

cc @killercup

mark-i-m:
Ping @oli-obk

src/tests/adding.md
```diff
@@ -148,6 +148,9 @@ exhaustive. Header commands can generally be found by browsing the
 `TestProps` structure found in [`header.rs`] from the compiletest
 source.
 
+* `run-rustfix` for UI tests, indicates that the test produces
+  structured suggestions, which are then applied and the final
```
mark-i-m: Does it get compiled again? I think the output of rustfix is just compared against the .fixed file.

killercup: Yes, [it gets compiled again](https://github.com/rust-lang/rust/blob/b4463d788bfd30b622a87a0e6f8e9271b9102e50/src/tools/compiletest/src/runtest.rs#L2634-L2649)

mark-i-m: Ah, then @oli-obk, could you just add that the expected fixed code should be in .fixed? Then r=me

Issue #137, opened 2018-05-21
tracking issue for documenting chalk and the trait system 


nikomatsakis:
We have some great docs on the trait system already (if I do say so myself 😉) but there's lots of room for improvement. This issue exists to track ideas and try to stay organized.

- [x] Update rustc guide with new rules for GATs (cc @tmandry, @scalexm)

mark-i-m:
Might be good to link to this talk by @sunjay somewhere: https://www.youtube.com/watch?v=rZqS4bLPL24

It's a great intro :)

mark-i-m:
This thread doesn't seem to be very useful, so I will close it to keep the tracker clean.

PR #145, opened 2018-05-26
Explain what each domain goal means

tmandry:
I felt the introduction to domain goals would be gentler with some examples and a brief discussion of each. Feedback wanted!


tmandry:
cc #137

PR #147, opened 2018-05-31
Add lowering rules for GATs

tmandry:
Add the lowering rules from https://github.com/rust-lang-nursery/chalk/issues/12#issuecomment-377258347 that were added to Chalk in https://github.com/rust-lang-nursery/chalk/pull/134/.

cc #137, @scalexm, @nikomatsakis

### DIFF
--- a/src/traits-lowering-rules.md
+++ b/src/traits-lowering-rules.md
@@ -132,8 +132,6 @@ to be **well-formed**:
 
 ```text
 // Rule WellFormed-TraitRef
-//
-// For each where clause WC:
 forall<Self, P1..Pn> {
   WellFormed(Self: Trait<P1..Pn>) :- Implemented(Self: Trait<P1..Pn>) && WellFormed(WC)
 }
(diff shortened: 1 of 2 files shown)


src/traits-lowering-rules.md
```diff
@@ -198,38 +196,70 @@ in detail in the [section on associated types](./traits-associated-types.html),
...
-show that the impl is well-formed, and hence we can rely on them
+the `Bounds` declared on the associated type must have been proven to hold
+to show that the impl is well-formed, and hence we can rely on them
 elsewhere.
 
 ```text
-// XXX how exactly should we set this up? Have to be careful;
-// presumably this has to be a kind of `FromEnv` setup.
+// Rule Implied-Bound-From-AssocTy
+forall<Self, P1..Pn, Pn+1..Pm> {
```
scalexm: Could you add: « for each `Bound` in the set of `Bounds` » or something like this, to emphasize the fact that we output one `FromEnv` rule at a time

src/traits-lowering-rules.md
```diff
@@ -270,25 +300,26 @@ Given an impl that contains:
...
+and our where clause `WC1` on the trait associated type from above, we
+produce the following rule:
 
 ```text
 // Rule Normalize-From-Impl
 forall<P0..Pm> {
   forall<Pn+1..Pm> {
     Normalize(<A0 as Trait<A1..An>>::AssocType<Pn+1..Pm> -> T) :-
-      WC && WC1
+      WC_impl && WC1
```
scalexm: `WC_impl` can actually be replaced with `Implemented(A0: Trait)` (and that’s what we do in chalk)

tmandry: Done. Do you think it's more or less confusing to keep `WC_impl` around then? :)

scalexm: Let’s keep it, so that it looks like a general enough impl.

tmandry: Sounds good. I added a note about `WC_impl` being unused just so it doesn't seem like a mistake.

PR #154, opened 2018-06-05
Add bit about ctags

mark-i-m:
cc @nikomatsakis


mark-i-m:
This is pretty much verbatim what Niko write, so I will go ahead and merge...

PR #155, opened 2018-06-07
Add chalk rules for type defs

scalexm:
Small sub section about `WellFormed` and `FromEnv` predicates for types.

### DIFF
--- a/src/traits/lowering-rules.md
+++ b/src/traits/lowering-rules.md
@@ -174,6 +174,80 @@ we must show that `WellFormed(TraitRef)`. This in turn justifies the
 implied bounds rules that allow us to extend the set of `FromEnv`
 items.
 
+## Lowering type definitions
+
+We also want to have some rules which define when a type is well-formed.
+For example, given this type:
+
+```rust,ignore
+struct Set<K> where K: Hash { ... }
+```
+
+then `Set<i32>` is well-formed because `i32` implements `Hash`, but
+`Set<NotHash>` would not be well-formed. Basically, a type is well-formed
+if its parameters verify the where clauses written on the type definition.
+
+Hence, for every type definition:
+
+```rust, ignore
+struct Type<P1..Pn> where WC { ... }
+```
+
+we produce the following rule:
+
+```text
+// Rule WellFormed-Type
+forall<P1..Pn> {
+  WellFormed(Type<P1..Pn>) :- WC
+}
+```
+
+Note that we use `struct` for defining a type, but this should be understood
+as a general type definition (it could be e.g. a generic `enum`).
+
+Conversely, we define rules which say that if we assume that a type is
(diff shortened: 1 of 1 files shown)


scalexm:
cc @nikomatsakis

src/traits/lowering-rules.md
```diff
@@ -174,6 +174,80 @@ we must show that `WellFormed(TraitRef)`. This in turn justifies the
...
+forall<P1..Pn> {
+  WellFormed(Type<P1..Pn>) :- WC
+}
+```
+
+Note that we use `struct` for defining a type, but this should be understood
+as a general type definition (it could be e.g. a generic `enum`).
+
+Conversely, we define rules which say that if we assume that a type is
+well-formed, we can also assume that its where clauses hold. That is,
```
alexcameron89: Nit: "its" -> "it's"

theotherphil: “Its” is correct here.

src/traits/lowering-rules.md
```diff
@@ -174,6 +174,80 @@ we must show that `WellFormed(TraitRef)`. This in turn justifies the
...
+fn foo<K>(collection: Set<K>, x: K, y: K) {
+    // `x` and `y` can be equalized even if we did not explicitly write
+    // `where K: Eq`
+    if x == y {
+        ...
+    }
+}
+```
+
+in the `foo` function, we assume that `Set<K>` is well-formed, i.e. we have
```
mark-i-m: Nit: capitalize "in"

src/traits/lowering-rules.md
```diff
@@ -174,6 +174,80 @@ we must show that `WellFormed(TraitRef)`. This in turn justifies the
...
+
+```text
+forall<K> {
+  FromEnv(K: Eq) :- FromEnv(K: Hash)
+}
+```
+
+which means that we finally get `FromEnv(K: Eq)` and then can compare `x`
+and `y` without needing an explicit where clause.
+
```
mark-i-m: And I'm guessing that if `K: !Hash`, then somewhere down the line we won't be able to prove `WellFormed(Set<K>)`, right?

src/traits/lowering-rules.md
scalexm: (I’ll fix in a few hours)

src/traits/lowering-rules.md
scalexm: Yes, basically it’s the `foo` caller’s responsibility to prove that the arguments they’re passing to `foo` are well formed. So for example trying to call `foo` with an argument of type `Set<NotHash>` (where `NotHash: !Hash`) will result in an error because of the reason you cited.

I think it should be good to explain this in the (empty) « implied bounds » chapter of this guide, unfortunately I don’t have a lot of time this days.

PR #164, opened 2018-07-06
existential type resolution

oli-obk:
cc @nikomatsakis


nikomatsakis:
These notes seem incomplete but better than nothing. If travis is happy, I'm inclined to merge.

Issue #122, opened 2018-05-05
Finish chapters on macro expansion?


alexreg:
Could we get someone to do this? I'm not sure who would be most appropriate. @jseyfried knows it better than most, but he's been absent lately.

mark-i-m:
I believe @petrochenkov and @nrc are also really knowledgeable about the hygiene parts.

alexreg:
Yep, you're right, @petrochenkov is for sure. Haven't interacted with @nrc about it, but I'll trust you on it!

mark-i-m:
@federicomenaquintero had also mentioned in #15 that they might try to write something.

mark-i-m:
I'm going to go ahead and close this as a duplicate of #15. But thanks for filing the issue @alexreg !

Issue #15, opened 2018-01-18
"macro expansion"


nikomatsakis:
Talk about where macro expansion happens. There is lots of other stuff we should cover here eventually:

- Hygiene
- Custom derive and procedural macros and how those work

cc @jseyfried -- can you help on elaborating the list of things that belong here?

mark-i-m:
Started a simple chapter on macro expansion here: #26 

I don't know much about hygiene, custom derive, or procedural macros, but I can write about normal `macro_rules` expansion...

mark-i-m:
cc @federicomenaquintero

mark-i-m:
Ping @federicomenaquintero :)

Are you still interested in writing something up about hygiene or proc macros?

federicomenaquintero:
I am!  Apologies for taking so long; I've been busy with librsvg.

Now that I'm porting gnome-class to syn 0.12, I'm getting a much more practical view of hygiene.  I'll write something soon.

mark-i-m:
Thanks for the update!

mark-i-m:
@federicomenaquintero ping

(sorry to bug you again)

PR #167, opened 2018-07-09
Add the small amount i know about hygiene

mark-i-m:
cc #15 #122 @alexreg

r? @petrochenkov


mark-i-m:
Oh, I always forget. Thanks :)

mark-i-m:
I think there are a few other places where this needs to be fixed too.

PR #173, opened 2018-07-24
Add testing chapter

djrenren:
This is just a deblog-ified version of blog.
http://blog.jrenner.net/rust/testing/2018/07/19/test-in-2018.html


src/SUMMARY.md
```diff
@@ -17,6 +17,7 @@
     - [Incremental compilation](./incremental-compilation.md)
     - [Debugging and Testing](./incrcomp-debugging.md)
 - [The parser](./the-parser.md)
+- [Testing](./testing.md)
```
nikomatsakis: Nit: Testing is very vague. Maybe we should call this "How `#[test]` is implemented" or something like that.

djrenren:
Should be all set!

djrenren:
@nikomatsakis -- *poke*

PR #191, opened 2018-08-31
document keep-stage1 and also try to clarify

nikomatsakis:
This documents the `--keep-stage 1` workflow (thanks @Mark-Simulacrum !) and tries to clarify a bit how staging works.

cc @sunjay -- the new language is intended to clear up some of the confusion you had around why building `--stage 1 src/libstd` does more than build the standard library. What do you think?

### DIFF
--- a/src/how-to-build-and-run.md
+++ b/src/how-to-build-and-run.md
@@ -166,11 +166,12 @@ LLVM version: 4.0
 There are two workflows that are useful for faster builders of the
 compiler.
 
-**Check, check, and check again.** The first workflow, which is useful when doing
-simple refactorings, is to run `./x.py check` continuously. Here you
-are just checking that the compiler can **build**, but often that is
-all you need (e.g., when renaming a method). You can then run `./x.py build`
-when you acqtually need to run tests.
+**Check, check, and check again.** The first workflow, which is useful
+when doing simple refactorings, is to run `./x.py check`
+continuously. Here you are just checking that the compiler can
+**build**, but often that is all you need (e.g., when renaming a
+method). You can then run `./x.py build` when you acqtually need to
+run tests.
 
 In fact, it is eomtimes useful to put off tests even when you are not
 100% sure the code will work. You can then keep building up
@@ -192,7 +193,8 @@ don't work (but that is easily detected and fixed).
 The sequence of commands you want is as follows:
 
 - Initial build: `./x.py build -i --stage 1 src/libstd`
-  - As [documented above](#command), this will build a functional stage1 compiler
+  - As [documented above](#command), this will build a functional
+    stage1 compiler
 - Subsequent builds: `./x.py build -i --stage 1 src/libstd --keep-stage 1`
   - Note that we added the `--keep-stage 1` flag here
   
--- a/src/how-to-build-and-run.md
+++ b/src/how-to-build-and-run.md
@@ -223,7 +223,8 @@ using `--keep-stage 1`** -- for example, strange [ICEs](appendix/glossary.html) 
 panics. In that case, you should simply remove the `--keep-stage 1`
 from the command and rebuild.  That ought to fix the problem.
 
-Note: you can also use `--keep-stage 1` when running tests. Something like this:
+You can also use `--keep-stage 1` when running tests. Something like
+this:
 
 - Initial test run: `./x.py test -i --stage 1 src/test/ui`
 - Subsequent test run: `./x.py test -i --stage 1 src/test/ui --keep-stage 1`
--- a/src/how-to-build-and-run.md
+++ b/src/how-to-build-and-run.md
@@ -53,24 +53,21 @@ One thing to keep in mind is that `rustc` is a _bootstrapping_ compiler. That
 is, since `rustc` is written in Rust, we need to use an older version of the
 compiler to compile the newer version. In particular, the newer version of the
 compiler, `libstd`, and other tooling may use some unstable features
-internally. The result is the compiling `rustc` is done in stages.
+internally. The result is the compiling `rustc` is done in stages:
 
-- **Stage 0:** the stage0 compiler can be your existing
-  (perhaps older version of)
-  Rust compiler, the current _beta_ compiler or you may download the binary
-  from the internet.
-- **Stage 1:** the code in your clone (for new version)
-  is then compiled with the stage0
-  compiler to produce the stage1 compiler.
-  However, it was built with an older compiler (stage0),
-  so to optimize the stage1 compiler we go to next stage.
-- **Stage 2:** we rebuild our stage1 compiler with itself
-  to produce the stage2 compiler (i.e. it builds
-  itself) to have all the _latest optimizations_.
-- _(Optional)_ **Stage 3**: to sanity check of our new compiler,
-  we can build it again
-  with stage2 compiler which must be identical to itself,
-  unless something has broken.
+- **Stage 0:** the stage0 compiler is usually the current _beta_ compiler
+  (`x.py` will download it for you); you can configure `x.py` to use something
+  else, though.
+- **Stage 1:** the code in your clone (for new version) is then
+  compiled with the stage0 compiler to produce the stage1 compiler.
+  However, it was built with an older compiler (stage0), so to
+  optimize the stage1 compiler we go to next stage.
+- **Stage 2:** we rebuild our stage1 compiler with itself to produce
+  the stage2 compiler (i.e. it builds itself) to have all the _latest
+  optimizations_.
+- _(Optional)_ **Stage 3**: to sanity check of our new compiler, we
+  can build it again with stage2 compiler which must be identical to
+  itself, unless something has broken.
 
 For hacking, often building the stage 1 compiler is enough, but for
 final testing and release, the stage 2 compiler is used.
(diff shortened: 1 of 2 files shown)


sunjay:
LGTM. Thanks for adding all of this to the guide! 😁🎉

src/how-to-build-and-run.md
```diff
@@ -140,6 +151,61 @@ release: 1.25.0-dev
...
+### Suggested workflows for faster builds of the compiler
+
+There are two workflows that are useful for faster builders of the
+compiler.
+
+**Check, check, and check again.** The first workflow, which is useful when doing
+simple refactorings, is to run `./x.py check` continuously. Here you
+are just checking that the compiler can **build**, but often that is
+all you need (e.g., when renaming a method). You can then run `./x.py build`
+when you acqtually need to run tests.
```
ehuss: acqtually > actually

src/how-to-build-and-run.md
```diff
@@ -140,6 +151,61 @@ release: 1.25.0-dev
...
+
+In fact, it is eomtimes useful to put off tests even when you are not
```
ehuss: eomtimes > sometimes

nikomatsakis:
@ehuss I added a paragraph about that

ehuss:
Oh, that makes sense.  TYVM!

src/how-to-build-and-run.md
```diff
@@ -53,24 +53,30 @@ One thing to keep in mind is that `rustc` is a _bootstrapping_ compiler. That
...
+  - (In theory, the stage1 compiler is functionally identical to the
+    stage2 compiler, but in practice there are subtle differences. In
+    particular, the stage1 compiler itself was built by stage0 and
+    hence not by the source in your working directory: this means that
+    the symbol names used in the compiler source may not match the
+    symbol names that would have been made by the stage1 compiler.
+    This can be important when using dynamic linking (e.g., with
+    derives. Sometimes this means that some tests don't work when run
+    with stage1.)
+- **Stage 2:** we rebuild our stage1 compiler with itself to produce
```
Mark-Simulacrum: It might be good to note that you'll never (aside from full-bootstrap) see stage2 compiler artifacts" since they are copied from stage1.

src/how-to-build-and-run.md
```diff
@@ -53,24 +53,30 @@ One thing to keep in mind is that `rustc` is a _bootstrapping_ compiler. That
...
+  the stage2 compiler (i.e. it builds itself) to have all the _latest
+  optimizations_.
+- _(Optional)_ **Stage 3**: to sanity check of our new compiler, we
```
Mark-Simulacrum: This is actually stage 2 artifacts - which should be identical to the artifacts produced by the stage 1 artifacts. Staging is a bit confusing here.

src/how-to-build-and-run.md
```diff
@@ -88,27 +96,39 @@ probably the best "go to" command for building a local rust:
...
-What this command will do is the following:
+This may *look* like it only builds libstd, but that is not the case.
+What this command does is the following:
+
+- Build libstd using the stage0 compiler (using incremental)
+- Build librustc using the stage0 compiler (using incremental)
+  - This produces the stage1 compiler
+- Build libstd using the stage1 compiler (cannot use incremental)
+
+This final product (stage1 compiler + libs build using that compiler)
```
Mark-Simulacrum: s/build/built

src/how-to-build-and-run.md
```diff
@@ -140,6 +160,63 @@ release: 1.25.0-dev
...
+**The TL;DR is that you might get weird behavior from a compile when
+using `--keep-stage 1`** -- for example, strange ICEs or other
+panics. In that case, you should simply remove the `--keep-stage 1`
+from the command and rebuild.  That ought to fix the problem.
+
+Note: you can also use `--keep-stage 1` when running tests. Something like this:
+
+- Initial test run: `./x.py test -i --stage 1 src/test/ui`
+- Subsequent test run: `./x.py test -i --stage 1 src/test/ui --keep-stage 1`
+
```
Mark-Simulacrum: I think it's good to note that the `--keep-stage` flag can be passed multiple times.

Furthermore, if you want guaranteed to work and optimal build times when changing just the compiler, the workflow I would suggest is: `./x.py build -i --stage 2 --keep-stage 0`; the `--stage 2` is optional. The effect of this command is that when the compiler is changed, we just rebuild it and you have a compiler that is now both fully capable (i.e., proc macros, sysroot fully populated) and you did not need to rebuild std at all. If, however, you wanted to have your changes take effect on the std that compiler uses, then that won't happen with this. But that's relatively rare.

src/how-to-build-and-run.md
```diff
@@ -140,6 +160,63 @@ release: 1.25.0-dev
...
+method). You can then run `./x.py build` when you actually need to
+run tests.
+
+In fact, it is sometimes useful to put off tests even when you are not
+100% sure the code will work. You can then keep building up
+refactoring commits and only run the tests at some later time. You can
+then use `git bisect` to track down **precisely** which commit caused
+the problem. A nice side-effect of this style is that you are left
+with a fairly fine-grained set of commits at the end, all of which
+build and pass testes. This often helps reviewing.
```
dwijnand: s/testes/tests/

src/how-to-build-and-run.md
```diff
@@ -140,6 +160,63 @@ release: 1.25.0-dev
...
+The effect of `--keep-stage1` is that we just *assume* that the old
+standard library can be re-used. If you are editing the compiler, this
+is almost always true: you haven't changed the standard library, after
+all.  But sometimes, it's not true: for example, if you are editing
+the "metadata" part of the compiler, which controls how the compiler
+encodes types and other states into the `rlib` files, or if you are editing
+things that wind up in the metadata (such as the definition of the MIR).
+
+**The TL;DR is that you might get weird behavior from a compile when
+using `--keep-stage 1`** -- for example, strange ICEs or other
```
dwijnand: ICEs?

mark-i-m: Internal Compiler Error

See https://rust-lang-nursery.github.io/rustc-guide/appendix/glossary.html ;)

src/how-to-build-and-run.md
nikomatsakis: I'm not sure I follow. Are you saying we just misuse the term stage3 to mean "building stdlib with the stage2 compiler"?

src/how-to-build-and-run.md
nikomatsakis: > and you did not need to rebuild std at all

I don't think such a compiler can possibly be guaranteed to work. I guess what you mean is "if you want a compiler that avoids the stage1 limitations"?

nikomatsakis: I decided not to add this text because I still don't really understand what `--keep-stage 0` even means. :)  Does it mean that we copy the stdlib from the beta?

I guess in that case maybe we *always* want to suggest it, unless you are editing the stdlib?

nikomatsakis: cc @Mark-Simulacrum

nikomatsakis:
I'm pulling out [this conversation](https://github.com/rust-lang-nursery/rustc-guide/pull/191#discussion_r214495962) about `--keep-stage 0` etc since GitHub wants to hide it. @Mark-Simulacrum can you clarify?

>  I decided not to add this text because I still don't really understand what --keep-stage 0 even means. :) Does it mean that we copy the stdlib from the beta? I guess in that case maybe we always want to suggest it, unless you are editing the stdlib?

Mark-Simulacrum:
I think there's a few workflows to consider:

Modifying std, with the intent of trying out some other user-space (i.e., not compiler-space) code. I believe the fastest way to do this today, when iterating, is: `x.py build --stage 1 --keep-stage 0 src/libstd`.
   * This will build the compiler once and only once
   * That compiler will then build a standard library and be mostly usable -- can build most code, but no proc macros currently
   * The compiler produced is `build/<build triple>/stage1/bin/rustc`

Modifying std, with intent to see results in the compiler. If you don't care about userspace -- which includes tests -- then `x.py build --stage 1 src/libstd --keep-stage 1` will be the fastest way to start and iterate. This will rebuild std, test, and the compiler, but it won't rebuild std more than once. If you do care about userspace seeing the std changes, you'll need to either remove the `--keep-stage 1` argument (and pay for two std rebuilds), or, if you're willing to invest more time upfront, then you can run `x.py build --stage 2 --keep-stage 0` instead. This will build a fully working compiler, building std, test, and rustc once each time you change std, and produces a compiler capable of supporting proc macros.

Modifying the compiler:

If you need proc macros, you'll need to invest more time up-front. Use `x.py build --keep-stage 0 src/librustc_codegen_llvm`. This will, on each run, attempt to build std/test/rustc with the stage1/bin/rustc compiler. Since we're only changing rustc, this is quite fast.

If you don't need proc macros, and aren't changing metadata or other ABI details, you can invest less time up-front. Use `x.py build --keep-stage 1 src/libstd`. This will build everything as per normal but skips the std build in stage 1, just reusing "old" artifacts.

---------

In general, any command run with `--keep-stage N` should have a preceding `--stage N` run with other arguments being the same, otherwise you will likely fail to produce a working compiler due to the artifacts being stale or nonexistent.

I'm happy to clarify further, not sure if I've properly explained or even explained in the wrong direction -- let me know what would be best.

nikomatsakis:
My inclination is to merge this as is and link to @Mark-Simulacrum for follow-up. I don't think I have the time to try and write up more detailed notes here (and I also still feel sort of unclear about what anything other than `--keep-stage 1` actually does)

Issue #175, opened 2018-07-30
Explain how to benchmark the compiler


orium:

mark-i-m:
@nnethercote @wesleywiser Would either of you be able to write something for this?

nnethercote:
https://github.com/rust-lang-nursery/rustc-perf/blob/master/collector/README.md pretty much has this information. Maybe the guide should just point to that document?

rylev:
Triage: Closing. This section already exists. It's not perfect, but we have other issues like #1150 that provide tracking for more specific issues.

PR #193, opened 2018-08-31
add a section about profiling with perf

nikomatsakis:
cc @wesleywiser -- maybe we want to elaborate on the `-Zself-profile` output?

cc @nnnethercote -- we should add a section about using callgrind


mark-i-m:
cc @nnethercote
cc #175

src/profiling/with_perf.md
```diff
@@ -0,0 +1,294 @@
+# Profiling with perf
+
+sThis is a guide for how to profile rustc with perf.
```
mark-i-m: s/sThis/This/

mark-i-m: Perhaps add a link to one of the following?
- https://perf.wiki.kernel.org/index.php/Main_Page
- http://www.brendangregg.com/perf.html

src/profiling/with_perf.md
```diff
@@ -0,0 +1,294 @@
...
+## Initial steps
+
+- Get a clean checkout of rust-lang/master, or whatever it is you want to profile.
+- Set the following settings in your `config.toml`:
+  - `debuginfo-lines = true`
+  - `use-jemalloc = false` -- lets you do memory use profiling with valgrind 
+  - leave everything else the defaults
+- Run `./x.py build` to get a full build
+- Make a rustup toolchain (let's call it `rust-prof`) pointing to that result
+  - `rustup toolchain link` XXX
```
mark-i-m: Sorry, what is `XXX` here?

src/profiling/with_perf.md
```diff
@@ -0,0 +1,294 @@
...
+
+### The basics
+
+The basic `perf` command is this:
+
+```
+perf record -F99 --call-graph dwarf XXX
+```
+
+The `-F99` tells perf to sample at 99 Hz, which avoids generating too
```
mark-i-m: Why 99?

src/profiling/with_perf.md
```diff
@@ -0,0 +1,294 @@
...
+
+```
+perf record -F99 --call-graph dwarf cargo +rust-prof rustc
+```
+
+to run `cargo`. But there are some things to be aware of:
+
+- You probably don't want to profile the time spend building
+  dependencies. So something like `cargo build; cargo clean -p $C` may
+  be helpful (where `$C` is the crate name)
```
mark-i-m: Is the suggestion that one should run this before running perf so that we only get the one crate? Also, is there any measurable overhead of using cargo as opposed to using rustc directly?

src/profiling/with_perf.md
```diff
@@ -0,0 +1,294 @@
...
+```bash
+touch src/lib.rs
+CARGO_INCREMENTAL=0 perf record -F99 --call-graph dwarf cargo rustc --profile check --lib
+```
+
+Note that final command: it's a doozy! It uses the `cargo rustc`
+command, which executes rustc with (potentially) additional options;
+the `--profile check` and `--lib` options specify that we are doing a
+`cargo check` execution, and that this is a library (not an
+execution).
```
mark-i-m: s/not an execution/not a binary/ ?

src/profiling/with_perf.md
```diff
@@ -0,0 +1,294 @@
...
+
+**A note of caution.** Each of the rustc-perf tests is its own special
+  snowflake. In particular, some of them are not libraries, in which
+  case you would want to do `touch src/main.rs` and avoid passing
+  `--lib`. I'm not sure how best to tell which test is which to be
+  honest.
+
+### Gathering NLL data
+
+If you want to profile an NLL run, you can just pass extra options to the `cargo rustc` command. The actual perf site just uses `-Zborrowck=mir`, which we can simulate like so:
```
mark-i-m: "simulate"?

src/profiling/with_perf.md
```diff
@@ -0,0 +1,294 @@
...
+## Analyzing a perf profile with `perf focus`
+
+Once you've gathered a perf profile, we want to get some information
+about it. For this, I personally use [perf focus][pf]. It's a kind of
+simple but useful tool that lets you answer queries like:
+
+- "how much time was spent in function F" (no matter where it was called from)
+- "how much time was spent in function F when it was called from G"
+- "how much time was spent in function F *excluding* time spent in G"
+- "what fns does F call and how much time does it spend in them"
```
mark-i-m: nit: s/fns/functions/ for consistency

src/profiling/with_perf.md
```diff
@@ -0,0 +1,294 @@
...
+To understand how it works, you have to know just a bit about
+perf. Basically, perf works by *sampling* your process on a regular
+basis (or whenever some event occurs). For each sample, perf gathers a
+backtrace. `perf focus` lets you write a regular expression that tests
+which fns appear in that backtrace, and then tells you which
+percentage of samples had a backtrace that met the regular
+expression. It's probably easiest to explain by walking through how I
+would analyze NLL performance.
+
+## Installing `perf-focus`
```
mark-i-m: Perhaps make this a third-level header? `###`

src/profiling/with_perf.md
```diff
@@ -0,0 +1,294 @@
...
+
+You can install perf-focus using `cargo install`:
+
+```
+cargo install perf-focus
+```
+
+## Example: How much time is spent in MIR borrowck?
```
mark-i-m: Perhaps make this a third-level header? `###`

src/profiling/with_perf.md
```diff
@@ -0,0 +1,294 @@
...
+```
+
+This will pipe the output from `perf script` through `c++filt` and
+should mostly convert those names into a more friendly format. The
+`--from-stdin` flag to `perf focus` tells it to get its data from
+stdin, rather than executing `perf focus`. We should make this more
+convenient (at worst, maybe add a `c++filt` option to `perf focus`, or
+just always use it -- it's pretty harmless).
+
+## Example: How much time does MIR borrowck spend solving traits?
```
mark-i-m: Perhaps make this a third-level header? `###`

src/profiling/with_perf.md
```diff
@@ -0,0 +1,294 @@
...
+turns out the answer is "almost never" -- only 12 samples fit that
+description (if you ever see *no* samples, that often indicates your
+query is messed up).
+
+If you're curious, you can find out exactly which samples by using the
+`--print-match` option. This will print out the full backtrace for
+each sample. The `|` at the front of the line indicates the part that
+the regular expression matched.
+
+## Example: Where does MIR borrowck spend its time?
```
mark-i-m: Perhaps make this a third-level header? `###`

src/profiling/with_perf.md
```diff
@@ -0,0 +1,294 @@
...
+Matches    : 228
+Not Matches: 542
+Percentage : 29%
+```
+
+The `'{do_mir_borrowck}'` argument is called the **matcher**. It
+specifies the test to be applied on the backtrace. In this case, the
+`{X}` indicates that there must be *some* function on the backtrace
+that meets the regular expression `X`. In this case, that regex is
+just the name of the fn we want (in fact, it's a subset of the name;
```
mark-i-m: nit: s/fn/function/

Also, elsewhere.

src/profiling/with_perf.md
```diff
@@ -0,0 +1,294 @@
...
+the full name includes a bunch of other stuff, like the module
+path). In this mode, perf-focus just prints out the percentage of
+samples where `do_mir_borrowck` was on the stack: in this case, 29%.
+
+**A note about c++filt.** To get the data from `perf`, `perf focus`
+  currently executes `perf script` (perhaps there is a better
+  way...). I've sometimes found that `perf script` outputs C++ mangled
+  names. This is annoying. You can tell by running `perf script |
+  head` yourself -- if you see named like `5rustc6middle` instead of
```
mark-i-m: s/named/names/

src/profiling/with_perf.md
```diff
@@ -0,0 +1,294 @@
+# Profiling with perf
+
+sThis is a guide for how to profile rustc with perf.
+
+## Initial steps
+
+- Get a clean checkout of rust-lang/master, or whatever it is you want to profile.
+- Set the following settings in your `config.toml`:
+  - `debuginfo-lines = true`
+  - `use-jemalloc = false` -- lets you do memory use profiling with valgrind
```
mark-i-m: nit: we have been trying to move away from `--` and towards em-dash. Also, elsewhere.

src/profiling/with_perf.md
```diff
@@ -0,0 +1,294 @@
...
+noisy and includes random stuff like the innards of
+malloc. `--tree-max-depth` can be useful too, it just limits how many
+levels we print.
+
+For each line, we display the percent of time in that function
+altogether ("total") and the percent of time spent in **just that
+function and not some callee of that function** (self). Usually
+"total" is the more interesting number, but not always.
+
+### Absolute vs relative percentages
```
mark-i-m: The name "Absolute vs relative" is a bit confusing because both types are relative to something. I'm not sure what the best way to clarify this is... perhaps "total vs sample"?

src/profiling/with_perf.md
```diff
@@ -0,0 +1,294 @@
...
+execution**. This is useful to help you keep perspective -- often as
+we drill down to find hot spots, we can lose sight of the fact that,
+in terms of overall program execution, this "hot spot" is actually not
+important. It also ensures that percentages between different queries
+are easily compared against one another.
+
+That said, sometimes it's useful to get relative percentages, so `perf
+focus` offers a `--relative` option. In this case, the percentages are
+listed only for samples that match (vs all samples). So for example we
+could find out get our percentages relative to the borrowck itself
```
mark-i-m: s/find out get/get/

src/profiling/with_perf.md
nikomatsakis: probably me just not remembering that crazy syntax

src/profiling/with_perf.md
nikomatsakis: no particular reason, just a number that works "ok"

src/profiling/with_perf.md
nikomatsakis: I've never observed any overhead

src/profiling/with_perf.md
nikomatsakis: oh? doesn't markdown change `--` to em-dash? em-dashes are annoying to type :)

src/profiling/with_perf.md
nikomatsakis: I'll just write "Relative percentages"

src/profiling/with_perf.md
nikomatsakis: I added a note to that effect

src/profiling/with_perf.md
nikomatsakis: I guess there is some HTML entity for them...

src/profiling/with_perf.md
mark-i-m: :stuck_out_tongue: Perhaps link to `how-to-build-and-run.html#creating-a-rustup-toolchain`

src/profiling/with_perf.md
mark-i-m: Ah, ok.

src/profiling/with_perf.md
nikomatsakis: OK, I resorted to copy-and-paste :)

mark-i-m: At least on GitHub, they are not changed. I find them annoying too, but we've already changed them elsewhere. I usually copy/paste too.

Thanks :)

src/profiling/with_perf.md
```diff
@@ -1,17 +1,18 @@
...
+  to profile.
 - Set the following settings in your `config.toml`:
   - `debuginfo-lines = true`
-  - `use-jemalloc = false` -- lets you do memory use profiling with valgrind 
+  - `use-jemalloc = false` — lets you do memory use profiling with valgrind 
   - leave everything else the defaults
 - Run `./x.py build` to get a full build
 - Make a rustup toolchain (let's call it `rust-prof`) pointing to that result
-  - `rustup toolchain link` XXX
+  - `rustup toolchain link <path-to-toolchain>`
```
mark-i-m: I think this is missing the toolchain name, right? Perhaps we can just link to `how-to-build-and-run.html#creating-a-rustup-toolchain`?

src/profiling/with_perf.md
```diff
@@ -110,11 +113,12 @@ can be helpful; it is covered below.
...
 ### Gathering NLL data
 
-If you want to profile an NLL run, you can just pass extra options to the `cargo rustc` command. The actual perf site just uses `-Zborrowck=mir`, which we can simulate like so:
+If you want to profile an NLL run, you can just pass extra options to
+the `cargo rustc` command, like so:
 
 ```bash
 touch src/lib.rs
-CARGO_INCREMENTAL=0 perf record -F99 --call-graph dwarf cargo rustc --profile check --lib -- -Zborrowck=mir
+CARGO_INCREMENTAL=0 perf record -F99 --call-graph dwarf cargo rustc --profile check --lib — -Zborrowck=mir
```
mark-i-m: This one actually should be `--` :P

mark-i-m:
also, I think you are missing ` ```console ` in a few places, causing travis to fail.

Thanks again!

src/profiling/with_perf.md
```diff
@@ -0,0 +1,299 @@
...
+CARGO_INCREMENTAL=0 cargo +rust-prof check 
+```
+
+Next: we want record the execution time for *just* the clap-rs crate,
+running cargo check. I tend to use `cargo rustc` for this, since it
+also allows me to add explicit flags, which we'll do later on.
+
+```bash
+touch src/lib.rs
+CARGO_INCREMENTAL=0 perf record -F99 --call-graph dwarf cargo rustc --profile check --lib
```
wesleywiser: I noticed here you stopped saying `cargo +rust-prof`. Is that intentional?

wesleywiser:
I just had that one small question but I'm really excited to see this get written! Great work!! 👍

src/profiling/with_perf.md
```diff
@@ -0,0 +1,299 @@
...
+
+The basic `perf` command is this:
+
+```
+perf record -F99 --call-graph dwarf XXX
+```
+
+The `-F99` tells perf to sample at 99 Hz, which avoids generating too
+much data for longer runs (why 99 Hz you ask? No particular reason, it
+just seems to work well for me). The `--call-graph dwarf` tells perf
```
nnethercote: The usual reason given for 99 Hz is that it's unlikely to be in lockstep with any kind of periodic activity. In comparison, 100 Hz could be, and that could lead to skewed results.

mark-i-m: Ah, that makes sense. Thanks :)

src/profiling/with_perf.md
```diff
@@ -0,0 +1,299 @@
...
+
+### The basics
+
+The basic `perf` command is this:
+
+```
+perf record -F99 --call-graph dwarf XXX
+```
+
+The `-F99` tells perf to sample at 99 Hz, which avoids generating too
```
nnethercote: Everything about doing the `perf` recording -- from here to the end of the "Gathering NLL data" section -- is doing things the hard way! `rustc-perf` is able to do all this stuff for you, specifically the [`collector profile perf-record`](https://github.com/rust-lang-nursery/rustc-perf/blob/master/collector/README.md#profiling-local-builds) command.

It already uses `-F99` and `--call-graph dwarf`, it does dependency stuff for you and only measures the compilation of the final crate, you don't need to distinguish between library and non-library crates, you can specify what kind of "run" you want (Clean, Nll, BaseIncr, CleanIncr, PatchedIncrs, All, or any combination of the above), and also what kind of "build" (Check, Debug, Opt, All, or a combination).

nnethercote:
Thank you for writing this. My comment above about doing `perf record` the easy way is my main comment about this.

src/profiling/with_perf.md
```diff
@@ -0,0 +1,299 @@
...
+  honest.
+
+### Gathering NLL data
+
+If you want to profile an NLL run, you can just pass extra options to
+the `cargo rustc` command, like so:
+
+```bash
+touch src/lib.rs
+CARGO_INCREMENTAL=0 perf record -F99 --call-graph dwarf cargo rustc --profile check --lib — -Zborrowck=mir
```
wesleywiser: There's an em dash in here `--lib *—* -Zborrowck=mir`

nnethercote:
Reading this inspired me to improve the [rustc-perf docs](https://github.com/rust-lang-nursery/rustc-perf/tree/master/collector) in https://github.com/rust-lang-nursery/rustc-perf/pull/284.

nnethercote:
Yes, that's what I suggested in https://github.com/rust-lang-nursery/rustc-guide/pull/193#pullrequestreview-151583216.

src/profiling/with_perf.md
nikomatsakis: it was, because I said that we should call the toolchain `rust-prof`, but I've rewritten it now to be more clear.

src/profiling/with_perf.md
nikomatsakis: Ah, good to know.

nikomatsakis: Learn something new =)

nikomatsakis:
Added a link to the "easy way to do things" -- going to merge this I think, but @nnethercote perhaps you want to double check that I got it right?

nnethercote:
> @nnethercote perhaps you want to double check that I got it right?

Looks good to me, thanks.

PR #197, opened 2018-09-10
Make max line length 100

mark-i-m:
r? @nikomatsakis 

cc @alexreg, since you were involved in the initial choice of 80c

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ cache:
 - cargo
 before_install:
 - shopt -s globstar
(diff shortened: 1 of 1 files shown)


alexreg:
👍 from me

mark-i-m:
Niko added this change to one of his PRs. Should merge soon.

alexreg:
Are all existing lines going to be reformatted to be max 100 chars?

mark-i-m:
I hope not. That would be a very hard diff to verify.

nagisa:
in rustc we simply try to follow the style in the file. 100 is the limit
but it doesn't mean everything must be as close to the limit as possible.

PR #200, opened 2018-09-11
Fix typos

phansch:
These were fixed using [codespell](https://github.com/codespell-project/codespell).


src/traits/goals-and-clauses.md
```diff
@@ -196,7 +196,7 @@ Implemented(Foo: Send) :-
 ```
 
 As you can probably imagine, proving that `Option<Box<Foo>>: Send` is
-going to wind up circularly requiring us to prove that `Foo: Send`
+going to wind up circular requiring us to prove that `Foo: Send`
```
mark-i-m: I think circularly is correct.

phansch: indeed, rebased :+1:

PR #212, opened 2018-10-19
Write implied bounds chapter

scalexm:
Complete overview of the implementation of implied bounds. It references the chapter on WF checks, which I'm also currently writing.

### DIFF
--- a/src/traits/implied-bounds.md
+++ b/src/traits/implied-bounds.md
@@ -1,9 +1,500 @@
 # Implied Bounds
 
-*to be written*
+Implied bounds remove the need to repeat where clauses written on
+a type declaration or a trait declaration. For example, say we have the
+following type declaration:
+```rust,ignore
+struct HashSet<K: Hash> {
+    ...
+}
+```
 
-Cover:
+then everywhere we use `HashSet<K>` as an "input" type, that is appearing in
+the receiver type of an `impl` or in the arguments of a function, we don't
+want to have to repeat the `where K: Hash` bound, as in:
 
-- Why the `FromEnv` setup etc is the way it is
-- Perhaps move some of the material from 'lowering rules' in to here
-- Show various examples where you could go wrong
+```rust,ignore
+// I don't want to have to repeat `where K: Hash` here.
+impl<K> HashSet<K> {
+    ...
+}
+
+// Same here.
+fn loud_insert<K>(set: &mut HashSet<K>, item: K) {
+    println!("inserting!");
+    set.insert(item);
+}
+```
+
+Note that in the `loud_insert` example, `HashSet<K>` is not the type of an
+argument of the `loud_insert` function, it only *appears* in the argument type
+`&mut HashSet<K>`.
+
+The rationale for applying implied bounds to input types is that, for example,
+in order to call the `loud_insert` function above, the programmer must have
+*produced* the type `HashSet<K>` already, hence the compiler already verified
+that `HashSet<K>` was well-formed, i.e. that `K` effectively implemented
+`Hash`, as in the following example:
+
+```rust,ignore
+fn main() {
+    // I am producing a value of type `HashSet<i32>`.
+    // If `i32` was not `Hash`, the compiler would report an error here.
+    let set: HashSet<i32> = HashSet::new();
+    loud_insert(&mut set, 5);
+}
+```
+
+hence we don't want to repeat where clauses for input types because that would
+sort of duplicate the work of the programmer, having to verify that their types
+are well-formed both when calling the function and when using them in the
+arguments of their function. The same reasoning applies when using an `impl`.
+
+Similarly, given the following trait declaration:
+```rust,ignore
+trait Copy where Self: Clone {
+    ...
+}
+```
+
+then everywhere we bound over `SomeType: Copy`, we would like to be able to
+use the fact that `SomeType: Clone` without having to write it explicitly,
+as in:
+```rust,ignore
+fn loud_clone<T: Clone>(x: T) {
+    println!("cloning!");
+    x.clone();
+}
+
+fn fun_with_copy<T: Copy>(x: T) {
+    println!("will clone a `Copy` type soon...");
+
+    // I'm using `loud_clone<T: Clone>` with `T: Copy`, I know this
+    // implies `T: Clone` so I don't want to have to write it explicitly.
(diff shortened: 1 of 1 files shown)


mark-i-m:
Thanks! I'll take a look soon :)

src/traits/implied-bounds.md
```diff
@@ -1,9 +1,500 @@
...
+should not.
+
+## Implied bounds coming from the environment
+
+So the naive approach does not work. What we need to do is to somehow decouple
+implied bounds from impls. Suppose we know that a type `SomeType<...>`
+implements `Bar` and we want to deduce that `SomeType<...>` must also implement
+`Foo`.
+
+There are two possibilities: first one, we have enough information about
```
mark-i-m:
```suggestion
There are two possibilities: first, we have enough information about
```

src/traits/implied-bounds.md
```diff
@@ -1,9 +1,500 @@
...
+Second possibility: we do not know enough about `SomeType<...>` in order to
+find a `Bar` impl which covers it, for example if `SomeType<...>` is just
+a type parameter in a function:
+```rust,ignore
+fn foo<T: Bar>() {
+    // We'd like to deduce `Implemented(T: Foo)`.
+}
+```
+
+that is, the information that `T` implements `Bar` here comes from the
```
mark-i-m:
```suggestion
That is, the information that `T` implements `Bar` here comes from the
```

src/traits/implied-bounds.md
```diff
@@ -1,9 +1,500 @@
...
+*environment*. The environment is the set of things that we assume to be true
+when we type check some Rust declaration. In that case, what we assume is that
+`T: Bar`. Then at that point, we might authorize ourselves to have some kind
+of  "local" implied bound reasoning which would say
+`Implemented(T: Foo) :- Implemented(T: Bar)`. This reasoning would
+only be done within our `foo` function in order to avoid the earlier
+problem where we had a global clause.
+
+We can apply these local reasonings everywhere we can have an environment
+-- i.e. when we can write where clauses -- that is inside impls,
```
mark-i-m:
```suggestion
-- i.e. when we can write where clauses -- that is, inside impls,
```

src/traits/implied-bounds.md
```diff
@@ -1,9 +1,500 @@
...
+trait declarations and type declarations.
```
mark-i-m:
```suggestion
trait declarations, and type declarations.
```

src/traits/implied-bounds.md
```diff
@@ -1,9 +1,500 @@
...
+```text
+forall<T> { Implemented(T: A) :- FromEnv(T: A). }
+
+forall<T> { Implemented(T: B) :- FromEnv(T: B). }
+forall<T> { FromEnv(T: A) :- FromEnv(T: B). }
+
+forall<T> { Implemented(T: C) :- FromEnv(T: C). }
+forall<T> { FromEnv(T: C) :- FromEnv(T: C). }
+```
+So these clauses are defined globally (that is they are available from
```
mark-i-m:
```suggestion
So these clauses are defined globally (that is, they are available from
```

src/traits/implied-bounds.md
```diff
@@ -1,9 +1,500 @@
...
+    // clauses on the `Foo` trait.
+}
+
+impl Bar for Y {
+    // This impl is legal: we have a `Foo` impl for `Y`.
+}
+```
+We must define what "legal" and "illegal" mean. For this, we introduce another
+predicate: `WellFormed(Type: Trait)`. We say that the trait reference
+`Type: Trait` is well-formed is `Type` meets the bounds written on the
```
mark-i-m:
```suggestion
`Type: Trait` is well-formed if `Type` meets the bounds written on the
```

src/traits/implied-bounds.md
```diff
@@ -1,9 +1,500 @@
...
+Now we would like to write the following impl:
+```rust,ignore
+impl Foo for i32 {
+    type Item = i32;
+}
+```
+The `Foo` trait definition and the `impl Foo for i32` are perfectly valid
+Rust: we're kind of recursively using our `Foo` impl in order to show that
+the associated value indeed implements `Foo`, but that's ok. But if we
+translates this to our well-formedness setting, the compiler proof process
```
mark-i-m:
```suggestion
translate this to our well-formedness setting, the compiler proof process
```

src/traits/implied-bounds.md
```diff
@@ -1,9 +1,500 @@
...
+```
+
+hence we don't want to repeat where clauses for input types because that would
+sort of duplicate the work of the programmer, having to verify that their types
+are well-formed both when calling the function and when using them in the
+arguments of their function. The same reasoning applies when using an `impl`.
+
+Similarly, given the following trait declaration:
+```rust,ignore
+trait Copy where Self: Clone {
```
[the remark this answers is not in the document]
scalexm: Yes, so that it is more clear that `Copy: Clone` is just a where clause like another one and that the special syntax don't play any role here.

src/traits/implied-bounds.md
```diff
@@ -1,9 +1,500 @@
...
+    loud_clone(x);
+}
+```
+
+The rationale for implied bounds for traits is that if a type implement `Copy`,
+that is if there exists an `impl Copy` for that type, there *ought* to exist
+an `impl Clone` for that type, otherwise the compiler would have reported an
+error in the first place. So again, if we were forced to repeat the additionnal
+`where SomeType: Clone` everywhere whereas we already know that
+`SomeType: Copy` hold, we would kind of duplicate the verification work.
```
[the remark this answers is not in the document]
scalexm: Actually for both: the programmer has to repeat bounds to be sure that the compiler will be able to verify something it has already verified :)

It is indeed an ergonomics feature, but what I was trying to explain is why it doesn't even make sense to repeat the bounds, beyond the ergonomics aspect :)

src/traits/implied-bounds.md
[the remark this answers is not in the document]
mark-i-m: Perhaps to make that clear, we could add a note like

```suggestion
trait Copy where Self: Clone { // desugared version of `Copy: Clone`
```

mark-i-m:
Thanks :)

PR #215, opened 2018-10-22
Clarify why 'static is free

### LINKED ISSUE #185: Clarify that `'static` is only a "free" lifetime parameter because of coercion
See [this section](https://rust-lang-nursery.github.io/rustc-guide/traits/canonicalization.html#canonicalizing-the-query), which is otherwise confusing.

This is my understanding, at least; someone correct me if I'm wrong.

mark-i-m:
@alexreg @nikomatsakis Does this look right?

Fix #185


alexreg:
Looks good! Thanks.

PR #217, opened 2018-10-24
Clean up skolemization with placeholder

csmoe:
Closes #211 
r? @mark-i-m

PR #219, opened 2018-10-25
Traits chapter cleanup

tmandry:
This does a small reorganization of the traits (new-style) chapter. IMO it makes sense to group the logic things together. Before, canonicalized queries were kind of in the middle of all the "logic" topics.

Here's what the chapter sidebar looks like now:
screen shot 2018-10-26 at 11 06 03 am

I also did some fly-by additions, and edits to break up the text. The edits make the text easier to scan/read by my eye, but this is extremely subjective, so let me know if anyone has objections.

This work is part of https://github.com/rust-lang/rust/issues/55100.


mark-i-m:
Thanks @tmandry! I will take a look soon.

Also, a heads up: this might end up having merge conflicts with #217

src/SUMMARY.md
```diff
@@ -38,12 +38,12 @@
       - [Goals and clauses](./traits/goals-and-clauses.md)
       - [Equality and associated types](./traits/associated-types.md)
       - [Implied bounds](./traits/implied-bounds.md)
+      - [Well-formedness checking](./traits/wf.md)
```
scalexm: Could you put this chapter just after "lowering rules"? It's intended to be a "complete reference" chapter and I think it's better for this one to read it after being comfortable with the full set of lowering rules.

tmandry: done.

PR #222, opened 2018-10-29
Write well-formedness checking chapter

scalexm:
This covers types, traits and trait impls.

cc @nikomatsakis @tmandry

### DIFF
--- a/src/traits/wf.md
+++ b/src/traits/wf.md
@@ -233,22 +233,32 @@ impl<T> Foo<T> for f32 {
 }
 ```
 
-So where clauses on associated types work *exactly* like where clauses on
-trait methods: in an impl, we must substitute the parameters from the traits
-with values provided by the impl, we may omit them if we don't need them, but
-we cannot add new where clauses.
+> So in Rust, where clauses on associated types work *exactly* like where
+> clauses on trait methods: in an impl, we must substitute the parameters from
+> the traits with values provided by the impl, we may omit them if we don't
+> need them, but we cannot add new where clauses.
 
 Now let's see the generated goal for this general impl:
 ```text
 forall<P1...> {
-    if (FromEnv(WC_impl), FromEnv(InputTypes(SomeType<A2...>))) {
-        WellFormed(SomeType<A2...>: Trait<A1...>) &&
-            WellFormed(InputTypes(WC_impl)) &&
+    // Well-formedness of types appearing in the impl
+    if (FromEnv(WC_impl), FromEnv(InputTypes(SomeType<A2...>: Trait<A1...>))) {
+        WellFormed(InputTypes(WC_impl)) &&
 
             forall<P2...> {
                 if (FromEnv(WC_assoc)) {
-                    WellFormed(SomeValue<A3...>: Bounds_assoc) &&
                         WellFormed(InputTypes(SomeValue<A3...>))
+                }
+            }
+    }
+
+    // Implied bounds checking
+    if (FromEnv(WC_impl), FromEnv(InputTypes(SomeType<A2...>: Trait<A1...>))) {
+        WellFormed(SomeType<A2...>: Trait<A1...>) &&
+
+            forall<P2...> {
+                if (FromEnv(WC_assoc)) {
+                    WellFormed(SomeValue<A3...>: Bounds_assoc)
                 }
             }
     }
@@ -257,25 +267,32 @@ forall<P1...> {
 
 Here is the most complex goal. As always, first, assuming that
 the various where clauses hold, we prove that every type appearing in the impl
-is well-formed, ***except*** types appearing in the receiver type
-`SomeType<A2...>`. Instead, we *assume* that those types are well-formed
-(hence the `if (FromEnv(InputTypes(SomeType<A2...>)))` condition). This is
+is well-formed, ***except*** types appearing in the impl header
+`SomeType<A2...>: Trait<A1...>`. Instead, we *assume* that those types are
+well-formed
+(hence the `if (FromEnv(InputTypes(SomeType<A2...>: Trait<A1...>)))`
+conditions). This is
 part of the implied bounds proposal, so that we can rely on the bounds
-written on the definition of the `SomeType<A2...>` type (and that we don't
+written on the definition of e.g. the `SomeType<A2...>` type (and that we don't
 need to repeat those bounds).
+> Note that we don't need to check well-formedness of types appearing in
+> `WC_assoc` because we already did that in the trait decl (they are just
+> repeated with some substitutions of values which we already assume to be
+> well-formed)
 
-Next, assuming that the where clauses on the impl `WC_impl` hold and that the
-input types of `SomeType<A2...>` are well-formed, we prove that
+Next, still assuming that the where clauses on the impl `WC_impl` hold and that
+the input types of `SomeType<A2...>` are well-formed, we prove that
 `WellFormed(SomeType<A2...>: Trait<A1...>)` hold. That is, we want to prove
 that `SomeType<A2...>` verify all the where clauses that might transitively
 come from the `Trait` definition (see
 [this subsection](./implied-bounds.md#co-inductiveness-of-wellformed)).
 
-Lastly, assuming that the where clauses on the associated type `WC_assoc` hold,
+Lastly, assuming in addition that the where clauses on the associated type
+`WC_assoc` hold,
 we prove that `WellFormed(SomeValue<A3...>: Bounds_assoc)` hold. Again, we are
 not only proving `Implemented(SomeValue<A3...>: Bounds_assoc)`, but also
-all the facts that might transitively come from `Bounds_assoc`. This is because
-we allow the use of implied bounds on associated types: if we have
+all the facts that might transitively come from `Bounds_assoc`. We must do this
+because we allow the use of implied bounds on associated types: if we have
 `FromEnv(SomeType: Trait)` in our environment, the lowering rules
 chapter indicates that we are able to deduce
 `FromEnv(<SomeType as Trait>::Assoc: Bounds_assoc)` without knowing what the
@@ -283,6 +300,12 @@ precise value of `<SomeType as Trait>::Assoc` is.
 
 Some examples for the generated goal:
 ```rust,ignore
+// Trait Program Clauses
+
+// These are program clauses that come from the trait definitions below
+// and that the trait solver can use for its reasonings. I'm just restating
+// them here so that we have them in mind.
+
 trait Copy { }
 // This is a program clause that comes from the trait definition above
 // and that the trait solver can use for its reasonings. I'm just restating
@@ -303,6 +326,8 @@ trait Complete where Self: Partial { }
 //     Implemented(Self: Complete) &&
 //     WellFormed(Self: Partial).
 // ```
+
+// Impl WF Goals
 
 impl<T> Partial for T where T: Complete { }
 // The generated goal is:
--- a/src/traits/wf.md
+++ b/src/traits/wf.md
@@ -27,18 +27,18 @@ In addition to the notations introduced in the chapter about
 lowering rules, we'll introduce another notation: when checking WF of a
 declaration, we'll often have to prove that all types that appear are
 well-formed, except type parameters that we always assume to be WF. Hence,
-we'll use the following notation: for a type `SomeType<...>`, we denote
-`InputTypes(SomeType<...>)` the set of all non-parameter types appearing in
-`SomeType<...>`, including `SomeType<...>` itself.
+we'll use the following notation: for a type `SomeType<...>`, we define
+`InputTypes(SomeType<...>)` to be the set of all non-parameter types appearing
+in `SomeType<...>`, including `SomeType<...>` itself.
 
 Examples:
 * `InputTypes((u32, f32)) = [u32, f32, (u32, f32)]`
-* `InputTypes(Box<T>) = [Box<T>]`
+* `InputTypes(Box<T>) = [Box<T>]` (assuming that `T` is a type parameter)
 * `InputTypes(Box<Box<T>>) = [Box<T>, Box<Box<T>>]`
 
-We may naturally extend the `InputTypes` notation to where clauses, for example
-`InputTypes(A0: Trait<A1,...,An>)` is the union of `InputTypes(A0)`,
-`InputTypes(A1)`, ..., `InputTypes(An)`.
+We also extend the `InputTypes` notation to where clauses in the natural way.
+So, for example `InputTypes(A0: Trait<A1,...,An>)` is the union of
+`InputTypes(A0)`, `InputTypes(A1)`, ..., `InputTypes(An)`.
 
 # Type definitions
 
@@ -51,7 +51,7 @@ struct Type<P...> where WC_type {
 }
 ```
 
-we generate the following goal:
+we generate the following goal, which represents its well-formedness condition:
 ```text
 forall<P...> {
     if (FromEnv(WC_type)) {
@@ -63,7 +63,7 @@ forall<P...> {
 }
 ```
 
-which in English gives: assuming that the where clauses defined on the type
+which in English states: assuming that the where clauses defined on the type
 hold, prove that every type appearing in the type definition is well-formed.
 
 Some examples:
@@ -284,7 +284,7 @@ Next, still assuming that the where clauses on the impl `WC_impl` hold and that
 the input types of `SomeType<A2...>` are well-formed, we prove that
 `WellFormed(SomeType<A2...>: Trait<A1...>)` hold. That is, we want to prove
 that `SomeType<A2...>` verify all the where clauses that might transitively
-come from the `Trait` definition (see
+be required by the `Trait` definition (see
 [this subsection](./implied-bounds.md#co-inductiveness-of-wellformed)).
 
 Lastly, assuming in addition that the where clauses on the associated type
--- a/src/traits/wf.md
+++ b/src/traits/wf.md
@@ -1,11 +1,269 @@
 # Well-formedness checking
 
-This chapter is mostly *to be written*. WF checking, in short, has the
-job of checking that the various declarations in a Rust program are
-well-formed. This is the basis for implied bounds, and partly for that
-reason, this checking can be surprisingly subtle! (For example, we
+WF checking has the job of checking that the various declarations in a Rust
+program are well-formed. This is the basis for implied bounds, and partly for
+that reason, this checking can be surprisingly subtle! For example, we
 have to be sure that each impl proves the WF conditions declared on
-the trait.)
+the trait.
 
+For each declaration in a Rust program, we will generate a logical goal and try
+to prove it using the lowered rules we described in the
+[lowering rules](./lowering-rules.md) chapter. If we are able to prove it, we
+say that the construct is well-formed. If not, we report an error to the user.
 
+Well-formedness checking happens in the [`src/rules/wf.rs`][wf] module in
+chalk. After you have read this chapter, you may find useful to see an
+extended set of examples in the [`src/rules/wf/test.rs`][wf_test] submodule.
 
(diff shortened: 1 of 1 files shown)


src/traits/wf.md
```diff
@@ -1,11 +1,440 @@
...
+
+Some examples to illustrate that:
+```rust,ignore
+trait Foo<T> {
+    type Assoc where T: Clone;
+}
+
+struct OnlyClone<T: Clone> { ... }
+
+impl<T> Foo<Option<T>> for () {
```
mark-i-m: Perhaps make the type parameter `U` instead of `T` for this impl, so that when you quote the bound `T: Clone`, is unambiguous with `Option<U>: Clone`.

src/traits/wf.md
```diff
@@ -1,11 +1,440 @@
...
+parameters from the trait are substituted with values provided by the impl
+(see example below). You cannot add new where clauses. You may omit to write
+the where clauses if you want to emphasize the fact that you are actually not
+relying on them.
+
+Some examples to illustrate that:
+```rust,ignore
+trait Foo<T> {
+    type Assoc where T: Clone;
+}
```
mark-i-m: What exactly does this mean? What's the difference between this and the following:

```rust
trait Foo<T: Clone> {
    type Assoc;
}
```

src/traits/wf.md
```diff
@@ -1,11 +1,440 @@
...
+    type Assoc = OnlyClone<Option<T>> where Option<T>: Clone;
+    //                                ^^^^^^^^^^^^^^^^^^^^^^
+    //                                this where clause does not exist
+    //                                on the original trait decl: illegal
+}
+```
+
+So where clauses on associated types work *exactly* like where clauses on
+trait methods: in an impl, we must substitute the parameters from the traits
+with values provided by the impl, we may omit them if we don't need them, and
```
mark-i-m:
```suggestion
with values provided by the impl, we may omit them if we don't need them, but
```

src/traits/wf.md
```diff
@@ -1,11 +1,440 @@
...
+we cannot add new where clauses.
+
+Now let's see the generated goal for this general impl:
+```text
+forall<P1...> {
+    if (FromEnv(WC_impl), FromEnv(InputTypes(SomeType<A2...>))) {
```
mark-i-m: Should `FromEnv(InputTypes(SomeType<A2...>))` be `FromEnv(WellFormed(InputTypes(SomeType<A2...>)))`?

Same in the paragraph below.

src/traits/wf.md
```diff
@@ -1,11 +1,440 @@
...
+                if (FromEnv(WC_assoc)) {
+                    WellFormed(SomeValue<A3...>: Bounds_assoc) &&
+                        WellFormed(InputTypes(SomeValue<A3...>))
+                }
+            }
+    }
+}
+```
+
+Here is the most complex goal. As always, a first thing is that assuming that
```
mark-i-m:
```suggestion
Here is the most complex goal. As always, first, assuming that
```

mark-i-m: Also, what part of the goal does this correspond to? Is it `WellFormed(InputTypes(WC_impl))`? If so, could we reorder the two lines in the goal so they match the order discussed in the paragraphs.

src/traits/wf.md
```diff
@@ -1,11 +1,440 @@
...
+we allow the use of implied bounds on associated types: if we have
+`FromEnv(SomeType: Trait)` in our environment, the lowering rules
+chapter indicates that we are able to deduce
+`FromEnv(<SomeType as Trait>::Assoc: Bounds_assoc)` without knowing what the
+precise value of `<SomeType as Trait>::Assoc` is.
+
+Some examples for the generated goal:
+```rust,ignore
+trait Copy { }
+// `WellFormed(Self: Copy) :- Implemented(Self: Copy).`
```
mark-i-m: It might be worth noting that this is not a goal but a fact that we can use to prove our goals.

src/traits/wf.md
```diff
@@ -1,11 +1,440 @@
...
+impl<T> Debug for Box<T> where T: Debug { ... }
+
+trait PointerFamily {
+    type Pointer<T>: Debug where T: Debug;
+}
+// `WellFormed(Self: PointerFamily) :- Implemented(Self: PointerFamily).`
+
+struct BoxFamily;
+
+impl PointerFamily for CowFamily {
```
mark-i-m:
```suggestion
impl PointerFamily for BoxFamily {
```

src/traits/wf.md
```diff
@@ -1,11 +1,440 @@
...
+    type Pointer<T> = Box<T> where T: Debug;
+}
+// The generated goal is:
+// ```
+// forall<T> {
+//     WellFormed(CowFamily: PointerFamily) &&
```
mark-i-m:
```suggestion
//     WellFormed(BoxFamily: PointerFamily) &&
```

src/traits/wf.md
```diff
@@ -1,11 +1,440 @@
...
+//
+//     if (FromEnv(T: Debug)) {
+//         WellFormed(Box<T>: Debug) &&
+//             WellFormed(Box<T>)
+//     }
+// }
+// ```
+// `WellFormed(CowFamily: PointerFamily)` amounts to proving
```
mark-i-m:
```suggestion
// `WellFormed(BoxFamily: PointerFamily)` amounts to proving
```

src/traits/wf.md
```diff
@@ -1,11 +1,440 @@
...
+// `Implemented(CowFamily: PointerFamily)`, which is ok thanks to our impl.
```
mark-i-m:
```suggestion
// `Implemented(BoxFamily: PointerFamily)`, which is ok thanks to our impl.
```

mark-i-m:
Also, travis is failing.

src/traits/wf.md
```diff
@@ -1,11 +1,440 @@
...
+    if (FromEnv(WC_type)) {
+        WellFormed(InputTypes(WC_type)) &&
+            WellFormed(InputTypes(A1)) &&
+            ...
+            WellFormed(InputTypes(An))
+    }
+}
+```
+
+which in English gives: assuming that the where clauses defined on the type
```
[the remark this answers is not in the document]
scalexm: Yes, as you can see in the other declarations, we always retrieve all types appearing and check their well-formedness. It is also the case within functions even if I did not write about functions yet, but it's talked about a bit in the intro of the implied bounds chapter.

src/traits/wf.md
scalexm: Mmh so the notation is a bit weird, but it introduced in the lowering rules chapter I think: `InputTypes(T)` is a set of types, say `[A1, A2,..., An]`, then `FromEnv(InputTypes(T))` is just
`[FromEnv(A1), FromEnv(A2)..., FromEnv(An)]`. It's like when we write `FromEnv(WC)` whereas
`WC` might denote a set of where clauses.

src/traits/wf.md
scalexm: This does not really correspond to a part actually because it spreads all through the goal, since there are nested where clauses etc, i.e. when checking well-formedness of the input types of the associated type value, we can assume that the where clauses defined on the assoc type hold.

So this corresponds to both `WellFormed(InputTypes(WC_impl))` and `WellFormed(InputTypes(SomeValue<A3...>))` but we don't have the same assumptions in our environment for these two goals.

Not sure how to best format this, if you have any suggestions... :)

src/traits/wf.md
scalexm: But also maybe that `trait Foo<T>` example was a bit constrived, to be fair.

src/traits/wf.md
```diff
@@ -1,11 +1,444 @@
...
+all the facts that might transitively come from `Bounds_assoc`. This is because
+we allow the use of implied bounds on associated types: if we have
+`FromEnv(SomeType: Trait)` in our environment, the lowering rules
+chapter indicates that we are able to deduce
+`FromEnv(<SomeType as Trait>::Assoc: Bounds_assoc)` without knowing what the
+precise value of `<SomeType as Trait>::Assoc` is.
+
+Some examples for the generated goal:
+```rust,ignore
+trait Copy { }
```
tmandry: I found the organization of this slightly hard to parse at first, how about something like...

```suggestion
// Trait Program Clauses

// These are program clauses that come from the trait definitions below
// and that the trait solver can use for its reasonings. I'm just restating
// them here so that we have them in mind.

trait Copy { }
```

src/traits/wf.md
```diff
@@ -1,11 +1,444 @@
...
+// ```
+
+trait Complete where Self: Partial { }
+// ```
+// WellFormed(Self: Complete) :-
+//     Implemented(Self: Complete) &&
+//     WellFormed(Self: Partial).
+// ```
+
+impl<T> Partial for T where T: Complete { }
```
tmandry:
```suggestion
// Impl WF Goals

impl<T> Partial for T where T: Complete { }
```

src/traits/wf.md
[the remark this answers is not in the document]
mark-i-m: Got it :+1: I was just checking my understanding.

src/traits/wf.md
```diff
@@ -1,11 +1,444 @@
...
+```text
+forall<P1...> {
+    if (FromEnv(WC_impl), FromEnv(InputTypes(SomeType<A2...>))) {
+        WellFormed(SomeType<A2...>: Trait<A1...>) &&
+            WellFormed(InputTypes(WC_impl)) &&
+
+            forall<P2...> {
+                if (FromEnv(WC_assoc)) {
+                    WellFormed(SomeValue<A3...>: Bounds_assoc) &&
+                        WellFormed(InputTypes(SomeValue<A3...>))
```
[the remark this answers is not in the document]
scalexm: There is indeed something missing but not exactly what you’re pointing :)

I stated that you could assume that the receiver type and its input types are well-formed.

This is correct but not complete: actually you can assume `FromEnv(InputTypes(SomeType<A2...>: Trait<A1...>))`. We can do that because everywhere else we check that all types appearing in where clauses are WF. This is what we currently do in chalk, I just forgot about it.

Then, we already checked  `WC_assoc` in the trait decl. In the impl, they are basically repeated, except that we substituted type parameters from the trait with the values provided by the impl, that is `A1...`. But since we already assume that the `A1...` are WF, no need to check again.

src/traits/wf.md
```diff
@@ -1,11 +1,444 @@
...
+part of the implied bounds proposal, so that we can rely on the bounds
+written on the definition of the `SomeType<A2...>` type (and that we don't
+need to repeat those bounds).
+
+Next, assuming that the where clauses on the impl `WC_impl` hold and that the
+input types of `SomeType<A2...>` are well-formed, we prove that
+`WellFormed(SomeType<A2...>: Trait<A1...>)` hold. That is, we want to prove
+that `SomeType<A2...>` verify all the where clauses that might transitively
+come from the `Trait` definition (see
+[this subsection](./implied-bounds.md#co-inductiveness-of-wellformed)).
```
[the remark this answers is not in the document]
scalexm: Actually I was trying to factor out all the `WellFormed(InputTypes(...))` in the previous paragraph (« assuming that the various where clauses hold, we prove that every type appearing in the impl is WF»)

src/traits/wf.md
```diff
@@ -1,11 +1,444 @@
...
+
+Lastly, assuming that the where clauses on the associated type `WC_assoc` hold,
+we prove that `WellFormed(SomeValue<A3...>: Bounds_assoc)` hold. Again, we are
+not only proving `Implemented(SomeValue<A3...>: Bounds_assoc)`, but also
+all the facts that might transitively come from `Bounds_assoc`. This is because
```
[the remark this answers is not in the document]
scalexm: Mmh so we’re not only using the where clauses in fact, we can also rely on the various program clauses that don’t come from the environment but rather from the other impls,

scalexm:
@tmandry Thanks, I’ll fix what we discussed in the comments soon (on my phone now).

src/traits/wf.md
mark-i-m: Hmm... that does help. I guess my question was more of a GATs question, then. In the `Foo<T>` example, is it possible for the caller to use `T: !Clone`? The bound is on an assoc type, rather than being in `WC_trait`, but it seems to be apply to any use of the trait.

src/traits/wf.md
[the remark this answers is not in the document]
tmandry: Oh, I missed that, sorry. Maybe update the goal clauses to match the order that they are mentioned in the text?

Also, it does say something to the effect of "assuming that the various where clauses hold" twice, this might be slightly confusing.

src/traits/wf.md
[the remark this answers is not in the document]
tmandry: Good point.

src/traits/wf.md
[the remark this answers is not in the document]
scalexm: Yeah I think when writing that paragraph I wasn’t entierely convinced by my formulation either :)

I’ll separate the if goals in two parts yes, it should be clearer.

src/traits/wf.md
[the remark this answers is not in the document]
tmandry: > Then, we already checked WC_assoc in the trait decl. In the impl, they are basically repeated, except that we substituted type parameters from the trait with the values provided by the impl, that is A1.... But since we already assume that the A1... are WF, no need to check again.

Yeah, on second thought this seems clear, though it may be worth saying. Also, the text might want to mention (in the paragraphs above) _where_ the `WC_assoc` is verified to be equivalent to the trait decl. i.e. is it during lowering in chalk?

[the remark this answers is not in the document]
scalexm: In chalk we have an easy solution for that: we just don’t write where clauses on associated type values, they are like auto-repeated from the trait :)

I don’t know yet where we’ll do that in rustc, especially because if we want to really mimic where clauses on trait methods, then it’s a bit more complex than just « omit it or write it »: on trait methods, you can actually write "new" where clauses provided that they are weaker than (= implied by) the original ones, for example you can do that:
```rust
trait Foo {
    fn foo<T> where T: Copy;
}

impl Foo for i32 {
    fn foo<T> where T: Clone { ... }
}
```

Of course this is useless from a semantic POV, it just serves as a documentation that you are not using the full power of the original where clause. But to check that the where clauses you wrote are indeed weaker, you’d need to use the trait solver of course. So I decided to juste sidestep this in the text as this is anecdotic I think.

scalexm:
Ok @tmandry should be good now. Just waiting for @nikomatsakis to proofread it and we should be ok.

src/traits/wf.md
```diff
@@ -1,11 +1,469 @@
...
+[lowering rules](./lowering-rules.md) chapter. If we are able to prove it, we
+say that the construct is well-formed. If not, we report an error to the user.
 
+Well-formedness checking happens in the [`src/rules/wf.rs`][wf] module in
+chalk. After you have read this chapter, you may find useful to see an
+extended set of examples in the [`src/rules/wf/test.rs`][wf_test] submodule.
 
+The new-style WF checking has not been implemented in rustc yet.
+
+[wf]: https://github.com/rust-lang-nursery/chalk/blob/master/src/rules/wf.rs
```
nikomatsakis: Maybe we should link to the chalk rustdocs at <http://rust-lang-nursery.github.io/chalk/doc/chalk/rules/wf/index.html> instead? Not sure.

src/traits/wf.md
```diff
@@ -1,11 +1,469 @@
...
+[wf_test]: https://github.com/rust-lang-nursery/chalk/blob/master/src/rules/wf/test.rs
+
+We give here a complete reference of the generated goals for each Rust
+declaration.
+
+In addition to the notations introduced in the chapter about
+lowering rules, we'll introduce another notation: when checking WF of a
+declaration, we'll often have to prove that all types that appear are
+well-formed, except type parameters that we always assume to be WF. Hence,
+we'll use the following notation: for a type `SomeType<...>`, we denote
```
nikomatsakis: Nit: Instead of "we denote" maybe "we define `InputTypes(SomeType<..>)` to be the set of all non-parameter types..."

src/traits/wf.md
```diff
@@ -1,11 +1,469 @@
...
+`InputTypes(SomeType<...>)` the set of all non-parameter types appearing in
+`SomeType<...>`, including `SomeType<...>` itself.
+
+Examples:
+* `InputTypes((u32, f32)) = [u32, f32, (u32, f32)]`
+* `InputTypes(Box<T>) = [Box<T>]`
+* `InputTypes(Box<Box<T>>) = [Box<T>, Box<Box<T>>]`
+
+We may naturally extend the `InputTypes` notation to where clauses, for example
```
nikomatsakis: Nit: "We also extend the `InputTypes` notation to where clauses in the natural way. So, for example, `InputTypes(A0: ...)`..."

src/traits/wf.md
```diff
@@ -1,11 +1,469 @@
...
+Given a general type definition:
+```rust,ignore
+struct Type<P...> where WC_type {
+    field1: A1,
+    ...
+    fieldn: An,
+}
+```
+
+we generate the following goal:
```
nikomatsakis: Nit: we generate the following goal, which represents its well-formedness condition:

src/traits/wf.md
```diff
@@ -1,11 +1,469 @@
...
+lowering rules, we'll introduce another notation: when checking WF of a
+declaration, we'll often have to prove that all types that appear are
+well-formed, except type parameters that we always assume to be WF. Hence,
+we'll use the following notation: for a type `SomeType<...>`, we denote
+`InputTypes(SomeType<...>)` the set of all non-parameter types appearing in
+`SomeType<...>`, including `SomeType<...>` itself.
+
+Examples:
+* `InputTypes((u32, f32)) = [u32, f32, (u32, f32)]`
+* `InputTypes(Box<T>) = [Box<T>]`
```
nikomatsakis: I would add something like ", assuming that `T` is a type parameter"

nikomatsakis: Remind me btw why we pushed the "expansion" into the lowering step, versus defining a proper `WellFormed(T)` goal? (And then having `struct Box<T>` get to assume that `WellFormed(T)` in its "environment"?)

src/traits/wf.md
```diff
@@ -1,11 +1,469 @@
...
+    if (FromEnv(WC_type)) {
+        WellFormed(InputTypes(WC_type)) &&
+            WellFormed(InputTypes(A1)) &&
+            ...
+            WellFormed(InputTypes(An))
+    }
+}
+```
+
+which in English gives: assuming that the where clauses defined on the type
```
nikomatsakis: s/gives/states/

src/traits/wf.md
```diff
@@ -1,11 +1,469 @@
...
+> Note that we don't need to check well-formedness of types appearing in
+> `WC_assoc` because we already did that in the trait decl (they are just
+> repeated with some substitutions of values which we already assume to be
+> well-formed)
+
+Next, still assuming that the where clauses on the impl `WC_impl` hold and that
+the input types of `SomeType<A2...>` are well-formed, we prove that
+`WellFormed(SomeType<A2...>: Trait<A1...>)` hold. That is, we want to prove
+that `SomeType<A2...>` verify all the where clauses that might transitively
+come from the `Trait` definition (see
```
nikomatsakis: Nit: s/come/be required by/

src/traits/wf.md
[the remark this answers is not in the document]
nikomatsakis: I think the idea would be that -- to prove the impl is well-formed -- we have to prove that the where-clauses from the trait imply the where-clauses from the impl. i.e., assuming the where-clauses from the trait hold, we show we can prove the where-clauses from the impl -- but I agree it can wait to be added.

src/traits/wf.md
scalexm: I think that the code is still a good documentation for that chapter as there are multiple subtle things happening.

PR #226, opened 2018-11-02
Document MIR building somewhat

oli-obk:
r? @nikomatsakis 

Most of these things are already pretty well documented in https://github.com/rust-lang/rust/blob/e8009885794e6108313683087fc6bfcd55145395/src/librustc_mir/build/expr/mod.rs#L11 so I created more of an overview

Is this what you had in mind? I can also add examples to demonstrate the things explained in text.

### DIFF
--- a/src/mir/construction.md
+++ b/src/mir/construction.md
@@ -64,8 +64,8 @@ There are essentially four kinds of representations one might want of a value:
 * `Operand` is an argument to e.g. a `+` operation or a function call
 * a temporary variable containing a copy of the value
 
-Since we start out with lowering the function body to an `Rvalue` so we can create an
-assignment to `RETURN_PLACE`, that `Rvalue` lowering will in turn trigger lowering to
+We start out with lowering the function body to an `Rvalue` so we can create an
+assignment to `RETURN_PLACE`, This `Rvalue` lowering will in turn trigger lowering to
 `Operand` for its arguments (if any). `Operand` lowering either produces a `const`
 operand, or moves/copies out of a `Place`, thus triggering a `Place` lowering. An
 expression being lowered to a `Place` can in turn trigger a temporary to be created
--- a/src/mir/construction.md
+++ b/src/mir/construction.md
@@ -112,7 +112,7 @@ again). Instead there are `Rvalue`s for binary and unary operators and index ope
 These `Rvalue`s later get codegened to llvm primitive operations or llvm intrinsics.
 
 Operators on all other types get lowered to a function call to their `impl` of the
-Operator's corresponding trait.
+operator's corresponding trait.
 
 Irrelevant of the lowering kind, the arguments to the operator are lowered to `Operand`s.
 This means all arguments are either constants, or refer to an already existing value
--- a/src/mir/construction.md
+++ b/src/mir/construction.md
@@ -114,7 +114,7 @@ These `Rvalue`s later get codegened to llvm primitive operations or llvm intrins
 Operators on all other types get lowered to a function call to their `impl` of the
 operator's corresponding trait.
 
-Irrelevant of the lowering kind, the arguments to the operator are lowered to `Operand`s.
+Regardless of the lowering kind, the arguments to the operator are lowered to `Operand`s.
 This means all arguments are either constants, or refer to an already existing value
 somewhere in a local or static.
 
--- a/src/mir/construction.md
+++ b/src/mir/construction.md
@@ -90,7 +90,7 @@ basic block onto which the statements should be appended.
 
 There are essentially four kinds of representations one might want of an expression:
 
-* `Place` refers to a (or a part of) preexisting memory location (local, static, promoted)
+* `Place` refers to a (or part of a) preexisting memory location (local, static, promoted)
 * `Rvalue` is something that can be assigned to a `Place`
 * `Operand` is an argument to e.g. a `+` operation or a function call
 * a temporary variable containing a copy of the value
--- a/src/mir/construction.md
+++ b/src/mir/construction.md
@@ -140,7 +140,7 @@ to a new temporary variable.
 
 ## Aggregate construction
 
-Aggregate values of any kind are built via `Rvalue::Aggregate`. All fields are
+Aggregate values of any kind (e.g. structs or tuples) are built via `Rvalue::Aggregate`. All fields are
 lowered to `Operator`s. This is essentially equivalent to one assignment
 statement per aggregate field plus an assignment to the discriminant in the
 case of `enum`s.
--- a/src/mir/construction.md
+++ b/src/mir/construction.md
@@ -126,7 +126,7 @@ In [MIR] there is no difference between method calls and function calls anymore.
 `if` conditions and `match` statements for `enum`s without variants with fields are
 lowered to `TerminatorKind::SwitchInt`. Each possible value (so `0` and `1` for `if`
 conditions) has a corresponding `BasicBlock` to which the code continues.
-The argument being branched on is (again) an `Operand` representing the value of 
+The argument being branched on is (again) an `Operand` representing the value of
 the if condition.
 
 ### Pattern matching
@@ -138,7 +138,8 @@ to a new temporary variable.
 
 ## Aggregate construction
 
-Aggregate values of any kind (e.g. structs or tuples) are built via `Rvalue::Aggregate`. All fields are
+Aggregate values of any kind (e.g. structs or tuples) are built via `Rvalue::Aggregate`.
+All fields are
 lowered to `Operator`s. This is essentially equivalent to one assignment
 statement per aggregate field plus an assignment to the discriminant in the
 case of `enum`s.
--- a/src/mir/construction.md
+++ b/src/mir/construction.md
@@ -1,1 +1,111 @@
 # MIR construction
+
+The lowering of [HIR] to [MIR] occurs for the following (probably incomplete)
+list of items:
+
+* Function and Closure bodies
+* Initializers of `static` and `const` items
+* Initializers of enum discriminants
+* Glue and Shims of any kind
+    * Tuple struct initializer functions
+    * Drop code (the `Drop::drop` function is not called directly)
+    * Drop implementations of types without an explicit `Drop` implementation
+
+The lowering is triggered by calling the `mir_built` query. The entire lowering
+code lives in `src/librustc_mir/build`. There is an intermediate representation
+between [HIR] and [MIR] called the `HAIR` that is only used during the lowering.
+The `HAIR` has datatypes that mirror the [HIR] datatypes, but instead of e.g. `-x`
+being a `hair::ExprKind::Neg(hair::Expr)` it is a `hair::ExprKind::Neg(hir::Expr)`.
+
+This shallowness enables the `HAIR` to represent all datatypes that [HIR] has, but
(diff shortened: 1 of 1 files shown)


src/mir/construction.md
```diff
@@ -1 +1,111 @@
...
+The lowering is triggered by calling the `mir_built` query. The entire lowering
+code lives in `src/librustc_mir/build`. There is an intermediate representation
+between [HIR] and [MIR] called the `HAIR` that is only used during the lowering.
+The `HAIR` has datatypes that mirror the [HIR] datatypes, but instead of e.g. `-x`
+being a `hair::ExprKind::Neg(hair::Expr)` it is a `hair::ExprKind::Neg(hir::Expr)`.
+
+This shallowness enables the `HAIR` to represent all datatypes that [HIR] has, but
+without having to create an in-memory copy of the entire [HIR]. The `HAIR` also
+does a few simplifications, e.g. method calls and function calls have been merged
+into a single variant.
```
nikomatsakis: The most important thing about the HAIR is that all "adjustments" and other implicit bits of syntax are made explicit, I think. Can we mention that?

src/mir/construction.md
```diff
@@ -1 +1,111 @@
...
+* Initializers of `static` and `const` items
+* Initializers of enum discriminants
+* Glue and Shims of any kind
+    * Tuple struct initializer functions
+    * Drop code (the `Drop::drop` function is not called directly)
+    * Drop implementations of types without an explicit `Drop` implementation
+
+The lowering is triggered by calling the `mir_built` query. The entire lowering
+code lives in `src/librustc_mir/build`. There is an intermediate representation
+between [HIR] and [MIR] called the `HAIR` that is only used during the lowering.
```
nikomatsakis: Link `HAIR` to <https://doc.rust-lang.org/nightly/nightly-rustc/rustc_mir/hair/index.html>?

src/mir/construction.md
nikomatsakis: It might be worth linking to this module too <https://doc.rust-lang.org/nightly/nightly-rustc/rustc_mir/hair/cx/expr/index.html>, as that is where HAIR construction takes place. Kind of annoying: it's hard to get a good stable-ish URL to a trait impl.

src/mir/construction.md
```diff
@@ -1 +1,111 @@
...
+* Function and Closure bodies
+* Initializers of `static` and `const` items
+* Initializers of enum discriminants
+* Glue and Shims of any kind
+    * Tuple struct initializer functions
+    * Drop code (the `Drop::drop` function is not called directly)
+    * Drop implementations of types without an explicit `Drop` implementation
+
+The lowering is triggered by calling the `mir_built` query. The entire lowering
+code lives in `src/librustc_mir/build`. There is an intermediate representation
```
nikomatsakis: It's best to link to the rustdoc for whatever you reference, in my opinion. That way the linkchecker can tell us when the lints are wrong. Here, I would probably link to the provider for the `mir_built` query at <https://doc.rust-lang.org/nightly/nightly-rustc/rustc_mir/transform/fn.mir_built.html>.

src/mir/construction.md
```diff
@@ -1 +1,111 @@
...
+variable.
+
+With this initialization out of the way, the lowering triggers a recursive call
+to a function that generates the MIR for the body (a `Block` expression) and
+writes the result into the `RETURN_PLACE`.
+
+## `unpack!` all the things
+
+One important thing of note is the `unpack!` macro, which accompanies all recursive
+calls. The macro ensures, that you get the result of the recursive call while updating
```
nikomatsakis: Nit: "The macro ensures, that" => "The macro ensures that"

src/mir/construction.md
```diff
@@ -1 +1,111 @@
...
+field accesses that read the fields from the argument and writes the value to the binding
+variable.
+
+With this initialization out of the way, the lowering triggers a recursive call
+to a function that generates the MIR for the body (a `Block` expression) and
+writes the result into the `RETURN_PLACE`.
+
+## `unpack!` all the things
+
+One important thing of note is the `unpack!` macro, which accompanies all recursive
```
nikomatsakis: I would probably explain back up the purpose of this macro a bit more:

Functions that generate MIR tend to fall into one of two patterns. First, if the function generates only statements, then it will take a basic block as argument onto which those statements should be appended. It can then return a result as normal:

```rust
fn generate_some_mir(&mut self, block: BasicBlock) -> ResultType {
   ...
}
```

But there are other functions that may generate new basic blocks as well. For example, lowering an expression like `if foo { 22 } else { 44 }` requires generating a small "diamond-shaped graph". In this case, the functions take a basic block where their code starts and return a (potentially) new basic block where the code generation ends. The `BlockAnd` type is used to represent this:

```rust
fn generate_more_mir(&mut self, block: BasicBlock) -> BlockAnd<ResultType> {
    ...
}
```

When you invoke these functions, it is common to have a local variable `block` that is effectively a "cursor". It represents the point at which we are adding new MIR. When you invoke `generate_more_mir`, you want to update this cursor. You can do this manually, but it's tedious:

```rust
let mut block;
let v = match self.generate_more_mir(..) {
    BlockAnd { block: new_block, value: v } => {
        block = new_block;
        v
    }
};
```

For this reason, we offer a macro that lets you write `let v = unpack!(block = self.generate_more_mir(...))`. It simply extracts the new block and overwrites the variable `block` that you named in the `unpack!`.
MIR functions that generate statements always take as argument a basic block onto which the statements should be appended.

src/mir/construction.md
```diff
@@ -1 +1,111 @@
...
+  the basic block has changed due to the inserted assertion (assertions terminate
+  blocks and jump either to a panic block or a newly created block, the latter being
+  the one you get back).
+
+The `unpack!` macro will call the recursive function you pass it, return the `Rvalue`
+and update the basic block by mutating the basic block variable you pass to it.
+
+## Lowering expressions into the desired MIR
+
+There are essentially four kinds of representations one might want of a value:
```
nikomatsakis: I would be careful with the phrase "value" -- I think we should be moving towards a terminology of expressions appearing in a "place context" (where they name some memory location) and "value context" (where we extracting the value). Hence I would maybe say, "Depending on the context in which an expression appears, we may want to lower it into one of four different sorts of things:" or something.

src/mir/construction.md
```diff
@@ -1 +1,111 @@
...
+
+* `Place` refers to a (or a part of) preexisting memory location (local, static, promoted)
+* `Rvalue` is something that can be assigned to a `Place`
+* `Operand` is an argument to e.g. a `+` operation or a function call
+* a temporary variable containing a copy of the value
```
nikomatsakis: It'd be great if we can link to some of the code here. The "handover" between those different bits of code is really sorta confusing and worth trying to document but I wont' force you to do it in this PR =) 

(e.g., it'd be nice to document the "categorization" stuff etc)

src/mir/construction.md
nikomatsakis: Another example: HAIR converts things like overloaded operators into fn calls

src/mir/construction.md
oli-obk: I addressed everything but this.

> The "handover" between those different bits of code is really sorta confusing and worth trying to document

I discussed this a little bit below in the form of an abstract example. I could add some graphics:

digraph G {
    Operand -> Constant
    Operand -> Place
    Place -> Projection
    Projection -> Place
    Place -> Local
    Rvalue -> Operand
    Assignment -> Place
    Assignment -> Operand
}


Or more verbose

![digraph G {
    Operand -> Constant
    Operand -> Place [taillabel="read"]
    Place -> Projection
    Projection -> Place
    Place -> LocalId
    Rvalue -> "Rvalue\nAggregate"
    Rvalue -> "Rvalue\nBinaryOp"
    Rvalue -> "Rvalue\nUnaryOp"
    Rvalue -> "Rvalue\n..."
    "Rvalue\nAggregate" -> Operand [headlabel="*"]
    "Rvalue\nBinaryOp" -> Operand [headlabel="2"]
    "Rvalue\nUnaryOp" -> Operand
    "Statement\nAssignment" -> Place [taillabel="write"]
    "Statement\nAssignment" -> Rvalue
    Statement -> "Statement\nAssignment"
    Statement -> "Statement\n..."
    Block -> Statement [headlabel="*"]
    Block -> Terminator
    Terminator -> "Terminator\nSwitchInt"
    "Terminator\nSwitchInt" -> Operand
    "Terminator\nSwitchInt" -> Constant [headlabel="*"]
    "Terminator\nSwitchInt" -> BlockId [headlabel="*"]
    Terminator -> "Terminator\n..."
    Mir -> Block [headlabel="*"]
    Mir -> Local [headlabel="*"]
}](https://user-images.githubusercontent.com/332036/47997756-42640900-e0fd-11e8-8446-957fa8c7ceef.png)

src/mir/construction.md
```diff
@@ -1 +1,150 @@
...
+It simply extracts the new block and overwrites the
+variable `block` that you named in the `unpack!`.
+MIR functions that generate statements always take as argument a
+basic block onto which the statements should be appended.
+
+## Lowering expressions into the desired MIR
+
+There are essentially four kinds of representations one might want of an expression:
+
+* `Place` refers to a (or a part of) preexisting memory location (local, static, promoted)
```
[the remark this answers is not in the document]
oli-obk: We have some docs about them in https://github.com/rust-rfcs/const-eval/blob/master/promotion.md but that's it.

The glossary does mention them, but the text is basically "TODO"

src/mir/construction.md
```diff
@@ -1 +1,150 @@
...
+
+Method calls are lowered to the same `TerminatorKind` that function calls are.
+In [MIR] there is no difference between method calls and function calls anymore.
+
+## Conditions
+
+`if` conditions and `match` statements for `enum`s without variants with fields are
+lowered to `TerminatorKind::SwitchInt`. Each possible value (so `0` and `1` for `if`
+conditions) has a corresponding `BasicBlock` to which the code continues.
+The argument being branched on is again an `Operand`.
```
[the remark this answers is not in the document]
oli-obk: Indeed

src/mir/construction.md
oli-obk: I wrote them in dot format (the dot source is in the comment source, which I guess is only visible with edit powers to the comment) and used graphviz (or specifically https://dreampuf.github.io/GraphvizOnline/) to generate the pngs. I'd prefer svgs (which can also be generated), but I can't include them in github comments.

When including them in this repo I'd like to include the .dot and .svg files and make the html page point to the svgs. The dot files are important for being able to edit the images (by regenerating them from scratch after editing the dot source)

I think the big one should probably not be inline, but rather mentioned like 
[Click here for a more detailed graph](https://user-images.githubusercontent.com/332036/47997290-98d04800-e0fb-11e8-9843-ce637b0230a8.png)

oli-obk: I mean if we feel fancy we can also generate such graphs with rustdoc directly when documenting crates, but it's most likely going to be less readable than the above, due to everything being connected to everything in some way.

mark-i-m:
It looks like you need to change a bunch of snipets to ```rust,ignore

oli-obk:
works now. Also locally checked that the images show up correctly in the generated html

mark-i-m:
Thank you!

PR #228, opened 2018-11-04
Add traits SLG solver section

tmandry:
This is mostly just an adapted version of Niko's blog post.

From here we need to add a subsection on negative reasoning, and possibly one on how the solver handles regions. Also one on using the solver via the `Context` and related traits.

cc @nikomatsakis @scalexm

### DIFF
--- a/src/traits/slg.md
+++ b/src/traits/slg.md
@@ -16,10 +16,10 @@ a set of region constraints, which we'll ignore in this introduction.
 
 There are often many, or even infinitely many, solutions to a query. For
 example, say we want to prove that `exists<T> { Vec<T>: Debug }` for _some_
-type `?T`. Our solver should be capable of iterating over each answer one at
-a time, say `?T = u32`, then `?T = i32`, and so on, rather than iterating
-over every type in the type system. If we need more answers, we can request
-more until we are done. This is similar to how Prolog works.
+type `?T`. Our solver should be capable of yielding one answer at a time, say
+`?T = u32`, then `?T = i32`, and so on, rather than iterating over every type
+in the type system. If we need more answers, we can request more until we are
+done. This is similar to how Prolog works.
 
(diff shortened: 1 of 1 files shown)


scalexm:
I don't have any objections.

src/traits/slg.md
```diff
@@ -1,3 +1,283 @@
...
+    A0 = [?0 = u32]
+  Strand:
+    (Vec<?U>: Debug) :- (?U: Debug)
+    (Rc<?V>: Debug) :- (?V: Debug)
+```
+
+Note that I am writing out the answer A0 as a substitution that can be
+applied to the table goal; actually, in the code, the goals for each
+X-clause are also represented as substitutions, but in this exposition
+I've chosen to write them as full goals, following NFTD.
```
mark-i-m: NFTD?

src/traits/slg.md
```diff
@@ -1,3 +1,283 @@
...
+   
+Since we now have an answer, `ensure_answer(T1, A0)` will return `Ok`
+to the table T0, indicating that answer A0 is available. T0 now has
+the job of incorporating that result into its active strand. It does
+this in two ways. First, it creates a new strand that is looking for
+the next possible answer of T1. Next, it incorpoates the answer from
```
mark-i-m: We create this new strand so that if anyone asks for `T0, A1`, we have a strand that can compute it, right?

tmandry: Yes. In particular, the new strand "remembers" that more answers from `T1` (starting with `T1, A1` should be incorporated into `T0`.

tmandry:
Waiting on https://github.com/rust-lang-nursery/chalk/pull/190 to be merged so I can link to some methods.

mark-i-m:
@tmandry r=me :) Should I wait for Niko?

Issue #227, opened 2018-11-04
Can't find "use-jemalloc" in config.toml.


kenta7777:
[how-to-build-and-run.md](https://github.com/rust-lang-nursery/rustc-guide/blob/607a58f8d2f56068ac1cf60c550d54f3ee6d90c6/src/how-to-build-and-run.md) tells us to open up to config.toml file and change the following settings.
`use-jemalloc = false`
However, the statement cannot be found in the file.

ehuss:
This was changed just a day ago to `jemalloc` in rust-lang/rust#55238.

kenta7777:
@ehuss Thank you for your comment. Then, should I send a PR for this revision?

mark-i-m:
That would be great!

kenta7777:
@mark-i-m I'll send a PR.

kenta7777:
Thanks!

PR #229, opened 2018-11-06
revised a jemalloc statement in how-to-build-and-run.md

kenta7777:
I revised a jemalloc statement concerned with creating config.toml.

### DIFF
--- a/src/how-to-build-and-run.md
+++ b/src/how-to-build-and-run.md
@@ -41,10 +41,6 @@ debuginfo = true
 
 # Gives you line numbers for backtraces.
 debuginfo-lines = true
-
-# Link the compiler against `jemalloc`, where on Linux and OSX it should
-# override the default allocator for rustc and LLVM.
-jemalloc = false
 ```
 
 ### What is x.py?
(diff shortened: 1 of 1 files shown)


kenta7777:
Fixes #227.

mark-i-m:
Thanks! r=me when Travis passes

ehuss:
The default is off, right?  Is there any reason to even mention jemalloc?

mark-i-m:
It's a fair question. This was originally included by Niko, so my inclination is to keep it unless you feel strongly about it...

ehuss:
> It's a fair question. This was originally included by Niko, so my inclination is to keep it unless you feel strongly about it...

It was included because the default was "on", and turning it off can help with development.  Now that it is off by default, I don't see a reason to mention it (there are many other settings in the config that aren't mentioned because the defaults are fine).  Also, this PR removes the original comment (which was unique to this guide) that explains why you would turn it off, and replaces it with a comment that is confusing to me (the way I read it it, it implies that it is *adding* jemalloc) — the other comments explain what it does, this one explains what it *doesn't* do.

mark-i-m:
I see what you mean. So @kenta7777 could you actually just remove these lines altogether?

kenta7777:
@mark-i-m Certainly. I'll remove these lines.

Issue #180, opened 2018-08-10
document MIR-based borrowck


nikomatsakis:
So much to talk about here!

Off the top of my head:

- Dataflow
  - MovePathIndex and that whole setup
- The MIR typeck
- Liveness
- Regions and SCC
- The outlives constraint

mark-i-m:
@nikomatsakis I think almost everything here is in the guide or in the TODOs. Do you want to keep this open?

mark-i-m:
I'm going to go ahead and close this, as much of it is already documented. Please feel free to reopen or file a new issue if something is missing.

PR #234, opened 2018-11-09
Fill out the borrowck chapter a bit more

mark-i-m:
I'm planning to play around with the borrow checker a bit for a class project, so I've learning a bit more.

Please let me know if anything is inaccurate or incomplete.

r? @nikomatsakis @pnkfelix


mark-i-m:
cc #180

src/borrow_check/region_inference.md
```diff
@@ -21,45 +21,49 @@ The MIR-based region analysis consists of two major functions:
...
+[`compute_regions`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_mir/borrow_check/nll/fn.compute_regions.html
+[`RegionInferenceContext`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_mir/borrow_check/nll/region_infer/struct.RegionInferenceContext.html
+[`solve`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_mir/borrow_check/nll/region_infer/struct.RegionInferenceContext.html#method.solve
 [NLL RFC]: http://rust-lang.github.io/rfcs/2094-nll.html
+[MIR type checker]: ./type_check.md
 
 ## Universal regions
 
-*to be written* – explain the `UniversalRegions` type
+The [`UnversalRegions`] type represents a collection of _unversal_ regions
```
nikomatsakis: Nit: "Universal" is misspelled as "unversal" in few places here. I think I would explain this through an example. Basically say something about how it computes information about the various named lifetimes in scope and their relationships to one another e.g., for the following function:

```rust
fn foo<'a>(x: &'a u32) { .. }
```

we would create a "universal region" `'a` as well as `'static`.  (I really want to rename this, but anyway.)

That's probably good enough to start. The various categories of lifetimes around closures are interesting but maybe not needed here.

src/borrow_check/region_inference.md
```diff
@@ -21,45 +21,49 @@ The MIR-based region analysis consists of two major functions:
...
+## Region variables
 
-## The MIR type-check
-
-## Representing the "values" of a region variable
-
-The value of a region can be thought of as a **set**; we call the
-domain of this set a `RegionElement`. In the code, the value for all
-regions is maintained in
+The value of a region can be thought of as a **set** of points in the MIR where
```
nikomatsakis: This isn't really complete. It's actually a set of points + a set of universal regions.

src/borrow_check/region_inference.md
```diff
@@ -83,11 +87,86 @@ The kinds of region elements are as follows:
...
+need to take care of.
+
+The [`RegionInferenceContext`] type contains all of the information needed to
+do inference, including the universal regions from `replace_regions_in_mir` and
+the constraints computed for each region. It is constructed just after we
+compute the liveness constraints.
+
+Here are some of the fields of the struct:
+
+- `constraints`: contains all the outlives constraints.
```
nikomatsakis: can we link to the rustdoc here, or otherwise try to prevent bitrot?

src/borrow_check/region_inference.md
```diff
@@ -83,11 +87,86 @@ The kinds of region elements are as follows:
...
+
+## Closures
+
+When we are checking the type tests and universal regions, we may come across a
+constraint that we can't prove yet if we are in a closure body! However, the
+necessary constraints may actually hold (we just don't know it yet). Thus, if
+we are inside a closure, we just collect all the constraints we can't prove yet
+and return them. Later, when we are borrow check the MIR node that created the
+closure, we can also check that these constraints hold. At that time, if we
+can't prove they hold, we report an error.
```
nikomatsakis: This seems fine, but really needs some examples

src/borrow_check/region_inference.md
```diff
@@ -83,11 +87,86 @@ The kinds of region elements are as follows:
...
+  universal regions. For example, if we have a where clause that `'a: 'b`, that
+  relation is assumed to be true while borrow checking the implementation (it
+  is checked at the caller), so `universal_region_relations` would contain `'a:
+  'b`.
+- `type_tests`: contains some constraints on types that we must check after
+  inference (e.g. `T: 'a`).
+- `closure_bounds_mapping`: used for propagating region constraints from
+  closures back out to the creater of the closure.
+
+TODO: should we discuss any of the others fields? What about the SCCs?
```
nikomatsakis: I feel like overall this discussion is a bit too focused on the "low-level details" and not much about the high-level process, but it seems like a good start.

src/borrow_check/region_inference.md
nikomatsakis: maybe not worth it, I have some thoughts on how to revisit this section

src/borrow_check/region_inference.md
```diff
@@ -83,11 +87,86 @@ The kinds of region elements are as follows:
...
+
+When we are checking the type tests and universal regions, we may come across a
+constraint that we can't prove yet if we are in a closure body! However, the
+necessary constraints may actually hold (we just don't know it yet). Thus, if
+we are inside a closure, we just collect all the constraints we can't prove yet
+and return them. Later, when we are borrow check the MIR node that created the
+closure, we can also check that these constraints hold. At that time, if we
+can't prove they hold, we report an error.
+
 ## Causal tracking
```
nikomatsakis: This is actually outdated, we don't do this anymore, we can remove this section.

src/borrow_check/region_inference.md
mark-i-m: But adding a universal region is adding a set, no?

Also, the universal regions are the `end('foo)` elements, correct?

mark-i-m:
Thanks Niko! I will take a look at your comments and update accordingly. It might be a little while though...

src/borrow_check/region_inference.md
mark-i-m: I updated this sentence. Let me know what you think.

> I'm trying to remember if there is some point where the distinction is important.

Perhaps something around placeholders?

> I'm also more and more trying to move into the polonius mindset, which is sort of different.

Would it be better to stay in the "NLL" mindset here for consistency and add a new chapter on Polonius?

src/borrow_check/region_inference.md
mark-i-m: TBH, I didn't dig into this part of the code, so I'm not really sure _how_ it computes that info. I did add the other things you mentioned, though, and a note to add more.

src/borrow_check/region_inference.md
mark-i-m: I tried to rework this section a bit to factor out the details from the meat. I also added an example.

I am a bit fuzzy on this part, so please let me know if you spot any problems. In particular, the example is a bit hand-wavey in some places...

PR #243, opened 2018-11-27
Mention that tests need a Python-enabled gdb.

vext01:
As mentioned here:
https://github.com/rust-lang/rust/issues/52452#issuecomment-440354255

It would be good is the rust compiler guide mentioned that the compiler tests need gdb with Python support.

Are there any other test dependencies that we should mention? If there are a few, perhaps we'd even want a subsection listing them.

Thanks

### DIFF
--- a/src/tests/running.md
+++ b/src/tests/running.md
@@ -19,6 +19,10 @@ timestamp file for every test can be found under `build/ARCH/test/`.
 To force-rerun a test (e.g. in case the test runner fails to notice
 a change) you can simply remove the timestamp file.
 
+Note that some tests require a Python-enabled gdb. If you are building
+gdb from source, you need to configure with
+`--with-python=<path-to-python-binary>`.
+
 ## Running a subset of the test suites
 
 When working on a specific PR, you will usually want to run a smaller
--- a/src/tests/running.md
+++ b/src/tests/running.md
@@ -19,8 +19,10 @@ timestamp file for every test can be found under `build/ARCH/test/`.
 To force-rerun a test (e.g. in case the test runner fails to notice
 a change) you can simply remove the timestamp file.
 
-Note that some tests require a Python-enabled gdb. If you are building
-gdb from source, you need to configure with
+Note that some tests require a Python-enabled gdb. You can test if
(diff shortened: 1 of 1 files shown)


eddyb:
We should be detecting this. cc @michaelwoerister @tromey @Mark-Simulacrum

vext01:
> We should be detecting this

I think that's also a good idea.

vext01:
The failure isn't mine, I don't think.

Mark-Simulacrum:
We should point users to the config.toml, not `configure` -- `configure` is primarily intended for automated scripts, whereas the user should edit `config.toml` manually (copying fields from config.toml.example).

vext01:
> We should point users to the config.toml, not configure -- configure is primarily intended for automated scripts

I'm talking about the configure script distributed with the gdb source code.

src/tests/running.md
```diff
@@ -19,6 +19,10 @@ timestamp file for every test can be found under `build/ARCH/test/`.
 To force-rerun a test (e.g. in case the test runner fails to notice
 a change) you can simply remove the timestamp file.
 
+Note that some tests require a Python-enabled gdb. If you are building
+gdb from source, you need to configure with
+`--with-python=<path-to-python-binary>`.
```
mark-i-m: Perhaps it might be good to mention how to check if your gdb is python-enabled?

mark-i-m:
Also, I restarted the build. It looked like a transient failure...

tromey:
> We should be detecting this

I filed https://github.com/rust-lang/rust/issues/56346 for this.

vext01:
The last commit addresses @mark-i-m's comment.

src/tests/running.md
shepmaster: > type some Python code

Not all of us know what valid Python code looks like.

vext01:
Another transient test failure

src/tests/running.md
mark-i-m: Perhaps give an example...

"type some python code (e.g. `print('hello gdb')`)"

mark-i-m:
Thanks!

PR #247, opened 2018-12-01
fix issues link

iximeow:
The `issues` link in the root has linkrotted since this is no longer `rust-lang-nursery`, just updating it.

I also found another two links that aren't quite right but still redirect correctly:
`rls` in `src/diag.md` - rls has also moved out of nursery
`Github repository` in `src/about-this-guide.md` - that's a link to this repo but under `rust-lang-nursery`

Should I update those links as well? I'm not sure if github redirects remain valid forever and I don't want to go just changing stuff around willy nilly

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ The guide can be useful today, but it has a lot of work still go.
 
 If you'd like to help improve the guide, we'd love to have you! You can find
 plenty of issues on the [issue
-tracker](https://github.com/rust-lang-nursery/rustc-guide/issue). Just post a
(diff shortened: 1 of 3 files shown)


mark-i-m:
> Should I update those links as well? I'm not sure if github redirects remain valid forever and I don't want to go just changing stuff around willy nilly

Yes, please change them :+1: 


Thanks for finding these!

iximeow:
the deed is done! just amended the commit.

mark-i-m:
r=me after CI passes :)

PR #250, opened 2018-12-05
Add some documentation about updating LLVM

alexcrichton:
I was talking with @tromey about LLDB/clang updates and I think it's about time we document all this!


mark-i-m:
@alexcrichton Do you have any plans to write more about LLVM or codegen? It would be very welcome.

tromey:
Thanks for doing this.  It looks good to me.

alexcrichton:
@mark-i-m I don't have plans just yet, but perhaps in the future!

alexcrichton:
Updated!

PR #260, opened 2019-01-04
Mention old incr. comp. design doc

Xanewok:
That's the first document that I stumbled upon, which actually explained what _red/green_ mean and what is the red/green algorithm (except for this doc).
Does it make sense to include that? I liked the graphs, more high-level overview and the motivation, but wouldn't like to post links to something that might be obsolete by now.

r? @nikomatsakis


nikomatsakis:
Seems fine, if you found it helpful.

PR #263, opened 2019-01-18
Reorganize the book into 3 parts

mark-i-m:
The guide is quite long and contains _tons_ of useful info, but it at the moment, it lacks flow IMHO.

I propose we organize the guide as two parts: (1) Useful info for contributing, and (2) How the compiler works.

(1) is where we would move general knowledge chapters like "about the compiler team", "building", "debugging", "stabilization process", etc. This section would serve as a reference for stuff that is useful regardless of what you are working on.

(2) is where everything else would live. This section is about how the compiler works.

r? @nikomatsakis

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -4,7 +4,7 @@
 
 ---
 
-- [Part 1 Intro](./part-1-intro.md)
+- [Part 1: Building, debugging, and contributing to Rustc](./part-1-intro.md)
 - [About the compiler team](./compiler-team.md)
 - [How to build the compiler and run what you built](./how-to-build-and-run.md)
     - [Build and Install distribution artifacts](./build-install-distribution-artifacts.md)
@@ -22,7 +22,7 @@
 
 ---
 
-- [Part 2 Intro](./part-2-intro.md)
+- [Part 2: How rustc works](./part-2-intro.md)
 - [High-level overview of the compiler source](./high-level-overview.md)
 - [The Rustc Driver](./rustc-driver.md)
     - [Rustdoc](./rustdoc.md)
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -1,21 +1,28 @@
 # Summary
 
-- [About this guide](./about-this-guide.md)
+[About this guide](./about-this-guide.md)
+
+---
+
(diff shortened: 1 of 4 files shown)


src/SUMMARY.md
```diff
@@ -1,21 +1,28 @@
...
+- [Walkthrough: a typical contribution](./walkthrough.md)
+- [Stabilizing Features](./stabilization_guide.md)
 - [Debugging the Compiler](./compiler-debugging.md)
 - [Profiling the compiler](./profiling.md)
     - [with the linux perf tool](./profiling/with_perf.md)
+- [Coding conventions](./conventions.md)
+
+---
+
+- [Part 2 Intro](./part-2-intro.md)
```
nikomatsakis: I think we should name this something like "Part 2: How the compiler works", so you can see at a glance what it's about

src/SUMMARY.md
```diff
@@ -1,21 +1,28 @@
 # Summary
 
-- [About this guide](./about-this-guide.md)
+[About this guide](./about-this-guide.md)
+
+---
+
+- [Part 1 Intro](./part-1-intro.md)
```
nikomatsakis: I think we should name this something like

Part 1: Building, debugging, and contributing to Rustc

although that seems a bit long to me =)

src/SUMMARY.md
mark-i-m: Good idea :+1: 

Done.

src/SUMMARY.md
mark-i-m: Same.

nikomatsakis:
maybe a pithier title for part 1 might just be "contribution guide"

mark-i-m:
I feel like "contribution guide" is to general (isn't the whole thing a contribution guide?). My original title was "General knowledge", which also felt off somehow... Perhaps "General Guidelines" is right?

nikomatsakis:
maybe keep it as is -- the only problem would be if it looks weird when rendered

mark-i-m:
It looks like a network timeout. I will restart the build.

> maybe keep it as is -- the only problem would be if it looks weird when rendered

Ok. I believe the title should just wrap properly. We already have some long titles like "How to build the compiler and run what you built"...

PR #264, opened 2019-01-19
fix related miri variables

rchaser53:
related: https://github.com/rust-lang/rust/issues/57469

fix variable name like below
 - Value => ConstValue
 - PrimVal => Scalar
 - Value::ByVal => ConstValue::Scalar
 - Value::ByValPair => ConstValue::ScalarPair


rchaser53:
this is the first PR to rustc-guide for me, so i anxious about this pr.
the ci is failed, but i just replaced the words. i guess my commit isn't related the ci failed.
should i need to fix something?

oli-obk:
Cycling CI (the failure seems to be a network timeout)

PR #272, opened 2019-01-23
Organize and finish debugging chapters

### LINKED ISSUE #11: "Debugging the compiler"
This section should talk about the various tips and tricks for debugging the compiler, as well as perhaps some amount of "debug log conventions". It doesn't necessarily have to be a *super* long section, but here are some of the things I can think of:

- [x] How to use `RUST_LOG` and `config.toml` first and foremost (overlaps with https://github.com/rust-lang-nursery/rustc-guide/issues/8)
- [x] Some of the handy `-Zflags`:
    - [x] example: `-Ztreat-err-as-bug`, which causes errors to be reported as bugs. Useful when you are debugging an error that should not be happening because you can use `RUST_BACKTRACE=1` then to get a backtrace of where the error was reported; also avoids dumping out more logs than you need.
    - [x] example: `-Zunpretty=hir-tree`, which will dump out the HIR
    - [x] example: `-Zverbose`, which dumps out extra 
    - [x] see `-Zhelp` for more
- [x] Conventions for logs ("loosely followed")
    - [x] Use `{:?}` most of the time in debug logs
    - [x] I tend to use `debug!("foo(...)")` at the start of a function `foo` and `debug!("foo: ...")` or logs within that function (- nikomatsakis)
- [x] Options to try and correlate node-ids and def-ids with the source
    - [x] e.g., `--pretty expanded,identified`
- [x] Options to dump the MIR
    - [x] `-Zdump-mir` and `-Zdump-mir-graphviz`

mark-i-m:
closes #11 

This adds the final few points from the checklist in #11. It also separates out topic-specific debugging advice into `Debugging` subchapters and leaves general debugging advice in the main debugging chapter.


nikomatsakis:
@mark-i-m sorry I've been very, very slow! I skimmed this material and it looked good. It needs rebase but I'm inclined to merge, we can always fixup mistakes later

mark-i-m:
Thanks @nikomatsakis! No worries :)

PR #273, opened 2019-01-23
Update Conventions chapters, add chapter on dependencies

### LINKED ISSUE #12: "Conventions used in the compiler"
Some of this material exists already scattered about, and/or maybe this overlaps with the glossary, but I think it'd be helpful to try and capture some common conventions.

Here are some things I'm thinking about:

- [x] How to format code (#47)
    - [x] Um... kind of a mess. Hopefully we'll move to rustfmt.
- [x] Naming conventions (maybe this should just be in the glossary, actually?)
    - [x] the `cx` suffix -- short for "context"
    - [x] `'tcx` and `'gcx`  -- mostly direct elsewhere for detailed explanation?
- [x] When to add/avoid adding dependencies from crates.io? What is the vetting process?
    - [x] Whitelisting?
- [ ] What else?

mark-i-m:
close #12

### DIFF
--- a/src/crates-io.md
+++ b/src/crates-io.md
@@ -1,0 +1,23 @@
+# crates.io Dependencies
+
+The rust compiler supports building with some dependencies from `crates.io`.
+For example, `log` and `env_logger` come from `crates.io`.
+
+In general, you should avoid adding dependencies to the compiler for several
+reasons:
+
+- The dependency may not be high quality or well-maintained, whereas we want
+  the compiler to be high-quality.
+- The dependency may not be using a compatible license.
+- The dependency may have transitive dependencies that have one of the above
+  problems.
+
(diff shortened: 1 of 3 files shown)


src/crates-io.md
```diff
@@ -0,0 +1,23 @@
...
+In general, you should avoid adding dependencies to the compiler for several
+reasons:
+
+- The dependency may not be high quality or well-maintained, whereas we want
+  the compiler to be high-quality.
+- The dependency may not be using a compatible license.
+- The dependency may have transitive dependencies that have one of the above
+  problems.
+
+TODO: what is the vetting process?
```
nikomatsakis: This is a good question for the meta WG to address, I guess. It seems like a quick FCP *might* make sense.

PR #274, opened 2019-01-23
Add link to sunjay talk

mark-i-m:
r? @nikomatsakis 

cc @sunjay


sunjay:
Thanks for thinking to include this! 

It may be worth noting two things somewhere near the link:

1. The talk only gives a broad overview of the problem and the solution (it was presented about halfway through my work)
2. The talk was in June of 2018 and some things may have changed (though nothing has at the time of this comment)

mark-i-m:
Thanks @sunjay! Updated.

PR #278, opened 2019-02-06
subchapter with information about `--error-format json`

zackmdavis:


mark-i-m:
Ping @zackmdavis. Do you have any thoughts on how to proceed?

mark-i-m:
@zackmdavis Ping

mark-i-m:
Ping @zackmdavis Do you have any thoughts on how to proceed?

zackmdavis:
@mark-i-m Hi. I can revise the PR to use `console` (thanks, I didn't know about that!) and make it a subsection rather than a subchapter, but not this minute. (I _do_ want to catch up on rustc work, but I've been regretfully super-busy and generally emotionally trashed for a few weeks or months.) I do think the info here is more useful to compiler developers than users, on the grounds that (a) I wrote this mostly for myself while trying to understand rust-lang/rust#53934, and (b) the JSON output is mostly useful to tool developers (RLS and friends), actual end-users will just use the tools. (But maybe tool developers _should_ count as "users" because they're not hacking on rustc itself?)

In general, I think I would favor a less rigorous review process for documentation rather than code (such that reviewers should be more inclined to just merge stuff that they're not 100% enthusiastic about), because natural-language prose isn't "brittle" the way code is (an English paragraph with a typo is typically still readable, a program with a typo typically doesn't compile); even flawed contributions can create a lot of marginal value without making the document an unmaintainable mess that no one knows how to fix?

mark-i-m:
Hi @zackmdavis. Thanks for the response! Please don't feel rushed; I just want to make sure that issues/PRs don't go stale.

In general, I agree with you and do try to be lenient. My concern mainly stems from the fact that if we put too much stuff in the guide, it quickly becomes outdated and hard to maintain. However, given your explanation, I think it seems reasonable to add this.

I can just merge this and then move it to a subchapter in another PR.

PR #281, opened 2019-02-16
Add various terms to the glossary

varkor:
I think this adds all the requested terms in https://github.com/rust-lang/rustc-guide/issues/48 so far. They're a little short, but some context is better than none, and they can be extended in the future.


mark-i-m:
cc @alexreg who had wanted some of these defs :smile:

alexreg:
Good work. A few things I would suggest:

* memoise -> memoize, memoisation -> memoization (I'm British too, but I think US English is standard here).
* Would it be helpful to distinguish early- and late-bound lifetimes in terms of "can bound at definition site" vs. "must be bound at call site"? This is how I understood it from the [old blog post](http://smallcultfollowing.com/babysteps/blog/2013/10/29/intermingled-parameter-lists/), and it resonated with me.

varkor:
> * Would it be helpful to distinguish early- and late-bound lifetimes in terms of "can bound at definition site" vs. "must be bound at call site"?

@alexreg: feel free to make those changes. I just thought I'd add something quickly, just to make sure there was something.

alexreg:
@varkor Okay ta. Just checking they make sense to you.

mark-i-m:
@alexreg A PR would be welcome (also lol, I'm America, and I like US English :stuck_out_tongue: ).

PR #282, opened 2019-02-16
Add a (short) chapter on `Kind`

varkor:
This chapter briefly describes the use of `Kind`, `UnpackedKind` and `Subst`.

This has been delayed long enough that I think it's worth getting *something* in the guide, just to provide some explanation. We can always extend this in the future with more examples.

Fixes https://github.com/rust-lang/rustc-guide/issues/97.

### DIFF
--- a/src/kinds.md
+++ b/src/kinds.md
@@ -34,7 +34,7 @@ the context is clear). As mentioned earlier, substition lists store raw
 `Kind`s, so before dealing with them, it is preferable to convert them to
 `UnpackedKind`s first. This is done by calling the `.unpack()` method.
 
-```rust
+```rust,ignore
 // An example of unpacking and packing a kind.
 fn deal_with_kind<'tcx>(kind: Kind<'tcx>) -> Kind<'tcx> {
     // Unpack a raw `Kind` to deal with it safely.
--- a/src/kinds.md
+++ b/src/kinds.md
@@ -1,31 +1,49 @@
 # Kinds
-A `ty::subst::Kind<'tcx>` represents some entity in the type system: currently
-either a type (`Ty<'tcx>`) or a lifetime (`ty::Region<'tcx>`), though in the
-future this will also include constants (`ty::Const<'tcx>`) to facilitate the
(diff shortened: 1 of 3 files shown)


src/kinds.md
```diff
@@ -0,0 +1,49 @@
...
+## `UnpackedKind`
+As `Kind` itself is not type-safe, the `UnpackedKind` enum provides a more
+convenient and safe interface for dealing with kinds. An `UnpackedKind` can
+be converted to a raw `Kind` using `Kind::from()` (or simply `.into()` when
+the context is clear). As mentioned earlier, substition lists store raw
+`Kind`s, so before dealing with them, it is preferable to convert them to
+`UnpackedKind`s first. This is done by calling the `.unpack()` method.
+
+```rust
+// An example of unpacking and packing a kind.
```
oli-obk: you need imports here in order to make the example compile and pass travis

mark-i-m: Can you make this example compile and pass without importing the whole compiler? Usually, I just add `rust,ignore`...

src/kinds.md
```diff
@@ -0,0 +1,49 @@
...
+instantiation of the parameters, for example `HashMap<i32, u32>`, would be
+represented by the substitution `&'tcx [tcx.types.i32, tcx.types.u32]`.
+
+`Subst` provides various convenience methods to instantiant substitutions
+given item definitions, which should generally be used rather than explicitly
+constructing such substitution slices.
+
+## `Kind`
+The actual `Kind` struct is optimised for space, storing the type, lifetime or
+const as an interned pointer containing a mask identifying its kind (in the
```
eddyb: The lowest 2 bits are the "tag", a (bit) "mask" is what you use to "mask out" bits other than the ones you're interested in (e.g. `let (tag, ptr) = (x & 0b11, x & !0b11);` uses `0b11` to get the lowest 2 bits and `!0b11` for everything else).

varkor: Opened https://github.com/rust-lang/rustc-guide/pull/287 to fix.

PR #287, opened 2019-03-14
Fix mask -> tag terminology

varkor:
Addresses https://github.com/rust-lang/rustc-guide/pull/282#discussion_r265550258.

### DIFF
--- a/src/kinds.md
+++ b/src/kinds.md
@@ -21,7 +21,7 @@ constructing such substitution slices.
 
 ## `Kind`
 The actual `Kind` struct is optimised for space, storing the type, lifetime or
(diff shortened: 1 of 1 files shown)


mark-i-m:
Hmm... it looks like a bunch of links have broken overnight... I will look into it this weekend.

mark-i-m:
@scalexm Do you know why all of the chalk links broke?

Havvy:
If I had to guess, it's because chalk got moved out of the nursery into @rust-lang/chalk.

It looks like that project should have been under a `Pin<&'links>`.

mark-i-m:
I opened #288 to fix the links, but I'm pretty confident that this fix is ok without travis :P

PR #288, opened 2019-03-15
Fix links

mark-i-m:
This should fix the builds.
- chalk moved to rust-lang, so all of the links broke
- @Zoxc changed the rustc_driver to use the new rustc_interface

r? @Zoxc

PR #289, opened 2019-03-16
Add a note about Assembly test suite

denzp:
Follow up for Assembly test suite introduced in https://github.com/rust-lang/rust/pull/58791.

cc @mark-i-m


mark-i-m:
@denzp Thanks for the PR! Looks good to me :)

I would like to merge #288 first so that we can actually use CI again, but other than that r=me. :+1:

PR #290, opened 2019-03-19
Emphasize use of most specific module for RUST_LOG

mark-i-m:
r? @eddyb since they made the suggestion in https://github.com/rust-lang/rustc-guide/issues/11#issuecomment-463398054

### DIFF
--- a/src/compiler-debugging.md
+++ b/src/compiler-debugging.md
@@ -165,7 +165,7 @@ then appear in standard error.
 **Note that unless you use a very strict filter, the logger will emit a lot of
 output, so use the most specific module(s) you can (comma-separated if
 multiple)**. It's typically a good idea to pipe standard error to a file and
-look at the log output with a text editor. You should also .
+look at the log output with a text editor.
 
 So to put it together.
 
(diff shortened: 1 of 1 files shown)


src/compiler-debugging.md
```diff
@@ -162,9 +162,10 @@ your log filter, e.g. to get the logs for a specific module, you can run the
 compiler as `RUST_LOG=module::path rustc my-file.rs`. All `debug!` output will
 then appear in standard error.
 
-Note that unless you use a very strict filter, the logger will emit a *lot*
-of output - so it's typically a good idea to pipe standard error to a file
-and look at the log output with a text editor.
+**Note that unless you use a very strict filter, the logger will emit a lot of
+output, so use the most specific module(s) you can (comma-separated if
+multiple)**. It's typically a good idea to pipe standard error to a file and
+look at the log output with a text editor. You should also .
```
oli-obk: The "You should also" sentence is missing its actual content ;)

mark-i-m: Ooops. Good catch. Copy/paste error.

PR #298, opened 2019-04-16
Update chalk with new organization

mark-i-m:


nikomatsakis:
This all looks good. I am going to merge it -- I suspect once the crate reorg is done, we'll want to revisit this text in any case.

nikomatsakis:
Oh, I see the build is still broken..?

mark-i-m:
Thanks @nikomatsakis!

There were also some broken links from a change to `MovePathIndex`, which I fixed in #299. I will rerun the build, it should be green when it finishes.

PR #301, opened 2019-04-17
Added rust phases diagram and explanation

igaray:
* [x] CoC read
* [x] Included @ehuss 's diagram of rustc compilation phases and artifacts
* [x] Added elaboration on stage naming convention
* [x] Added content from https://github.com/rust-lang/rust/issues/57963


spastorino:
The diagram was made by @ehuss. I was trying to ask in the other issue if you were comfortable by adding this to rustc-guides.

ehuss:
Sure, you're free to use it.  Here's the source file if you want to edit it: https://gist.github.com/ehuss/70eafa935cf21f0d67533790dcafc18e  That can be edited in https://www.draw.io/.  Just keep in mind that this is a simplification.  If there are non-host targets, its a bit different (described [here](https://github.com/rust-lang/rust/issues/57963#issuecomment-458429280)), and passing flags like `--keep-stage` will change what happens.  Also, rustdoc can be built at different stages depending on various things.

igaray:
I'll add in those caveats, thank you

src/how-to-build-and-run.md
```diff
@@ -84,6 +85,53 @@ compiling `rustc` is done in stages:
...
+    ...
+Building rustdoc for stage1
+    ...
+```
+
+Note that the convention `x.py` uses is that
+
+- A "stage N artifact" is an artifact that is _produced_ by the stage N compiler.
+- The "stage (N+1) compiler" is assembled from "stage N artifacts".
+- A `--stage N` flag means build with stage N.
```
mark-i-m:
```suggestion
- A `--stage N` flag means build _with_ stage N.
```

src/how-to-build-and-run.md
```diff
@@ -84,6 +85,53 @@ compiling `rustc` is done in stages:
...
+
+So "`stage0 std artifacts`" are in fact the output of the downloaded `stage0`
+compiler, and when it announces that it is "building stage1 std artifacts" it
+has moved on to the next bootstrapping phase.
+
+An deeper look into `x.py`'s phases can be seen here:
+
+<img alt="A diagram of the rustc compilation phases" src="img/rustc_stages.svg" class="center" />
```
mark-i-m:
```suggestion
![A diagram of the rustc compilation phases][img/rustc_stages.svg]
```

I think this will only render properly if we use markdown-style images (not sure though)...

src/how-to-build-and-run.md
```diff
@@ -84,6 +85,53 @@ compiling `rustc` is done in stages:
   can build the libraries with the stage2 compiler. The result ought
   to be identical to before, unless something has broken.
 
+Thus when running `x.py` you will see output such as:
+
+```
```
mark-i-m:
```suggestion
```txt
```

This will satisfy the CI, I think.

src/how-to-build-and-run.md
igaray: This change didn't work for me. I looked up what the rust book does for images, since this is the first image in the rustc-guide, and copied that.

src/how-to-build-and-run.md
igaray: Thanks for this.

src/how-to-build-and-run.md
mark-i-m: Hmm... interesting. That’s good to know.

src/how-to-build-and-run.md
mark-i-m: Np. You should be able to apply the suggestion directly from github if that makes things easier for you :)

igaray:
Update: pushed some more content and almost done but still missing some important info and editing. Will r= when done.

igaray:
r=mark-i-im

@jonas-schievink @dwijnand @mark-i-m @ehuss @Mark-Simulacrum
Please review the incorporated text, as they are based mostly off of your comments. I tried hard to be clear, non-repetitive and most of all, factually correct. I know the diff is large so if you have the opportunity maybe fetching the branch and reviewing the rendered markdown locally would be easier. Thank you for your time.

src/how-to-build-and-run.md
```diff
@@ -43,54 +43,237 @@ debuginfo = true
...
+  `test` and `rustc`, this stage0 compiler uses the freshly compiled `std`.
+  There are two concepts at play here: a compiler (with its set of dependencies)
+  and its 'target' or 'object' libraries (`std`, `test`, and `rustc`).
+  Both are staged, but in a staggered manner.
 - **Stage 1:** the code in your clone (for new version) is then
   compiled with the stage0 compiler to produce the stage1 compiler.
   However, it was built with an older compiler (stage0), so to
-  optimize the stage1 compiler we go to next stage.
-  - (In theory, the stage1 compiler is functionally identical to the
+  optimize the stage1 compiler we go to next the stage.
```
spastorino: I guess you wanted to say ...

```suggestion
  optimize the stage1 compiler we go to the next stage.
```

src/how-to-build-and-run.md
```diff
@@ -43,54 +43,237 @@ debuginfo = true
...
+This `stage2/bin/rustc` compiler is shipped to end-users, along with the
+`stage 1 {std,test,rustc}` artifacts.
+
+If you want to learn more about `x.py`, read its README.md
+[here](https://github.com/rust-lang/rust/blob/master/src/bootstrap/README.md).
+
 #### Build Flags
 
-There are other flags you can pass to the build portion of x.py that can be
+There are other flags you can pass to the build command of `x.py` that can be
```
spastorino: What about ...

```suggestion
There are other flags you can pass to `x.py` build command that can be
```

src/how-to-build-and-run.md
```diff
@@ -300,9 +484,10 @@ in other sections:
 
 ### ctags
 
-One of the challenges with rustc is that the RLS can't handle it, since it's a bootstrapping
-compiler. This makes code navigation difficult. One solution is to use `ctags`. The following
-script can be used to set it up: [https://github.com/nikomatsakis/rust-etags][etags].
+One of the challenges with `rustc` is that the RLS can't handle it, since it's a
+bootstrapping compiler. This makes code navigation difficult. One solution is to
+use `ctags`. The following script can be used to set it up:
+[https://github.com/nikomatsakis/rust-etags][etags].
```
spastorino: btw @nikomatsakis do you think this is necessary?. I used to have also a ctags config file for Rust but seems like ctags handles that automatically now. Is this still needed?. Anyway, this is a thing for a different PR.

src/how-to-build-and-run.md
ehuss: I think in this case it is parentheses:

```
![A diagram of the rustc compilation phases](img/rustc_stages.svg)
```

But you have to be careful with images because mdbook has a bug (I have a fix at https://github.com/rust-lang-nursery/mdBook/pull/871).  In this case I don't think it matters much either way.

src/how-to-build-and-run.md
```diff
@@ -43,54 +43,237 @@ debuginfo = true
...
+will later be uplifted to stage1_.
+
+Every time any of the main artifacts (`std`, `test`, `rustc`) are compiled, two
+steps are performed.
+When `std` is compiled by a stage N compiler, that `std` will be linked to
+programs built by the stage N compiler (including test and `rustc` built later
+on). It will also be used by the stage (N+1) compiler to link against itself.
+This is somewhat intuitive if one thinks of the stage (N+1) compiler as "just"
+another program we are building with the stage N compiler. In some ways, `rustc`
+(the binary, not the `rustbuild` step) could be thought of as one of the few
```
ehuss: I'm not sure I understand this sentence.  `rustc` links to libstd/libcore?

src/how-to-build-and-run.md
```diff
@@ -43,54 +43,237 @@ debuginfo = true
...
+`sysroot`).
+
+This `std` is pretty much necessary for any useful work with the compiler.
+Specifically, it's used as the `std` for programs compiled by the newly compiled
+compiler (so when you compile `fn main() { }` it is linked to the last `std`
+compiled with `x.py build --stage 1 src/libstd`).
+
+The `rustc` generated by the stage0 compiler is linked to the freshly-built
+`libstd`, which means that for the most part only `std` needs to be cfg-gated,
+so that `rustc` can use featured added to std immediately after their addition,
```
ehuss:
```suggestion
so that `rustc` can use features added to std immediately after their addition,
```

src/how-to-build-and-run.md
nikomatsakis: I don't know, did somebody add ctags support for Rust at some point?

nikomatsakis: I don't see it in [this list](http://ctags.sourceforge.net/languages.html), but who knows. That web page is **sourceforge** so I feel like there is no way it could have been updated in the last 2 decades.

spastorino: Yeah, that website seems very old. There's even a link to [Freshmeat](http://freshmeat.net/) :joy:.
My Linux distro uses https://ctags.io/, the source code is located https://github.com/universal-ctags/ctags and from this https://github.com/universal-ctags/ctags/pull/206 seems like support was added in 2015.

mark-i-m: Anyone who uses exuberant-ctags will be missing the definition. To my knowledge universal-ctags doesn’t  come on any ubuntu package list, so I didn’t know about it until recently when tmandry pointed it out on urlo.

spastorino: @mark-i-m good to know, that explains then. I'm using Archlinux and the ctags version provided is universal-ctags.

mark-i-m: It would definitely be worth it to open another PR to add info about universal-ctags. It is far superior to exuberant-ctags + config file.

igaray: https://github.com/rust-lang/rustc-guide/pull/314

src/how-to-build-and-run.md
mark-i-m: Yes, but IIUC, bootstrap treats libstd/libcore as "just another crate" that we are linking to our program. In other words, one can think of the compiler as a `no_core` program that cross-compiles and links with its own versions of the standard library.

mark-i-m:
@igaray Sorry for the delay. Can you rebase? After that I think we can merge. This PR has already undergone more scrutiny than most.

Thanks for your patience!

igaray:
Rebased but chalk links are broken again. Will rebase again after https://github.com/rust-lang/rustc-guide/pull/322 is merged

PR #302, opened 2019-04-18
Update BodyId description

rasendubi:
It is now a newtype'd HirId, not NodeId.

See https://github.com/rust-lang/rust/pull/58167.


rasendubi:
No problem! I'm just following the guide for the last few days and fixing things along the way ;)

PR #304, opened 2019-04-20
[canonicalization] fix result canonicalization example

rasendubi:


src/traits/canonicalization.md
```diff
@@ -177,8 +177,8 @@ The result would be as follows:
 ```text
 Canonical(QR) = for<T, L> {
     certainty: Proven,
-    var_values: [Vec<?0>, '?1, ?2]
-    region_constraints: [?2: '?1],
+    var_values: [Vec<?0>, '?1, ?0]
```
rasendubi: This is the change I am not exactly sure about and would like to get reviewed.

From all that I understand, this `?2` ought to be `?0` (because we canonicalizing the same variable `?E`).
`for<T, L>` also suggests that there should be only two canonical variables.

cc original author @nikomatsakis

mark-i-m: Yes, that sounds right, but I will defer to an expert like @nikomatsakis @tmandry or @scalexm ...

mark-i-m: Ping @scalexm?

nikomatsakis: This is correct

PR #305, opened 2019-04-21
Add documentation for two-phase borrows

matthewjasper:


src/borrow_check/two_phase_borrows.md
```diff
@@ -0,0 +1,95 @@
...
+let temp8 = x;
+std::ops::AddAssign::add_assign(temp7, temp8);
+```
+
+Whether a borrow can be two-phase is tracked by a flag on the [`AutoBorrow`]
+after type checking, which is then [converted] to a [`BorrowKind`] during MIR
+construction.
+
+Each two-phase borrow is assigned to a temporary that is only used once. As
+such we can add define:
```
mark-i-m:
```suggestion
such we can define:
```

mark-i-m:
Thanks :)

PR #307, opened 2019-04-25
Broken chalk links

igaray:
Fixed some links broken by https://github.com/rust-lang/chalk/pull/215/

### DIFF
--- a/src/traits/lowering-rules.md
+++ b/src/traits/lowering-rules.md
@@ -28,10 +28,10 @@ comment like so:
     // Rule Foo-Bar-Baz
 
 The reference implementation of these rules is to be found in
-[`chalk/chalk-rules/src/clauses.rs`][chalk_rules]. They are also ported in
+[`chalk/chalk-solve/src/clauses.rs`][chalk_rules]. They are also ported in
 rustc in the [`librustc_traits`][librustc_traits] crate.
 
-[chalk_rules]: https://github.com/rust-lang/chalk/blob/master/chalk-rules/src/clauses.rs
+[chalk_rules]: https://github.com/rust-lang/chalk/blob/master/chalk-solve/src/clauses.rs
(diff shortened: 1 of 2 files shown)


spastorino:
I'd only put the Fix broken chalk links commit here. BTW there's a typo in that commit message `Fied` -> `Fixed`

spastorino:
Now looks good /cc @nikomatsakis

igaray:
Travis still failed :(

igaray:
mdbook test is the culprit this time, I didn't know about that.

igaray:
CI toolchain hasn't yet updated to the rustc version which accepts the two-phase borrows in the two-phase borrow documentation examples.

spastorino:
Yes, your PR is unrelated and it's good to go.

mark-i-m:
Thanks @igaray! The CI failures are not from this PR they are from the last PR, which added the two-phase borrows chapter. If you don’t mind fixing them, that would be appreciated greatly. Basically all of the code blocks with error messages need to be annotated to tell mdbook test that they aren’t rust code: ` ```txt `.

igaray:
@mark-i-m Roger, will do. Although the offending code https://github.com/rust-lang/rustc-guide/pull/305/files#diff-5311fe548c2eb1bbb4b4ab3ca2151676R18 does in fact compile with the latest nightly (rustc 1.36.0-nightly (e305df184 2019-04-24)) and `mdbook test` passes. Given that the snippet is in fact correct code, should I still add the `ignore` marker to the snippet?

mark-i-m:
Oh. I guess I confused myself and misread your comment. I'm sorry.

In that case, I think we can just merge. I think the right solution is just for rustc-guide to use the nightly compiler in its CI. After all, it is describing the nightly state of the compiler...

nikomatsakis:
❤️ thanks all

PR #325, opened 2019-05-30
rustdoc: Remove note about deprecated pass configuration

phansch:
It looks like this note refers to https://github.com/rust-lang/rust/pull/50669
and https://github.com/rust-lang/rust/issues/44136 which leads me to believe
that it can be safely removed.

cc @QuietMisdreavus

### DIFF
--- a/src/rustdoc.md
+++ b/src/rustdoc.md
@@ -111,12 +111,6 @@ regarding dropping private/hidden items can be bypassed by passing
 `--document-private-items` to rustdoc. Note that unlike the previous set of AST
 transformations, the passes happen on the _cleaned_ crate.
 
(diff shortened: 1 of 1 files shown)


QuietMisdreavus:
The `--passes` flag still exists, though. It's still possible to pass something like `--no-defaults --passes collect-trait-impls` or something like that, but rustdoc will emit a deprecation warning for it. I feel like the note could still be kept, just reworded.

mark-i-m:
@phansch Ping :)

phansch:
I'm going to have to close this PR as I currently don't have enough time to work on it

mark-i-m:
Ok, thanks for the update...

PR #330, opened 2019-06-04
Indentation of Summary

mdsimmo:
Make Part 1/Part 2 top level with everything else underneath them.
Thus, the "parts" of the book are more obvious.

There has been no change to the structure of the book.


mark-i-m:
Oh, this didn't work the last time I tried it. I'm glad it works now.

My only concern is that having everything indented an extra line might make it harder to read.

It looks like this on my screen:
image

Seems acceptable to me...

@spastorino Thoughts?

spastorino:
Seems good to me :+1:

PR #333, opened 2019-06-07
Debuginfo level

mark-i-m:
r? @spastorino 

I'm not really sure what github is doing but it made me open another PR...

### DIFF
--- a/src/how-to-build-and-run.md
+++ b/src/how-to-build-and-run.md
@@ -36,9 +36,10 @@ debug-assertions = true
 # performance perhaps (less inlining) but it's worth it.
 codegen-units = 0
 
-# I always enable full debuginfo, however the line debuginfo (which is a more important part
-# and enables e.g. line numbers in backtraces) is enabled by `debuginfo-level = 1` too.
-debuginfo-level = 2
+# This enables full debuginfo (`debuginfo-level = 2`). The line debuginfo (which is a
+# more important part and enables e.g. line numbers in backtraces) is also enabled by
+# `debuginfo-level = 1`.
+debug = true
 ```
 
 ### What is `x.py`?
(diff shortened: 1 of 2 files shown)


src/how-to-build-and-run.md
```diff
@@ -36,9 +36,10 @@ debug-assertions = true
 # performance perhaps (less inlining) but it's worth it.
 codegen-units = 0
 
-# I always enable full debuginfo, however the line debuginfo (which is a more important part
-# and enables e.g. line numbers in backtraces) is enabled by `debuginfo-level = 1` too.
-debuginfo-level = 2
+# This enables full debuginfo (`debuginfo-level = 2`). The line debuginfo (which is a
+# more important part and enables e.g. line numbers in backtraces) is also enabled by
+# `debuginfo-level = 1`.
+debug = true
```
petrochenkov: Note that `debuginfo-level = 2` slows down the build, and `debug = true` slows down the build even more because it enables debug assertions, while `debuginfo-level = 1` is pretty much free.

mark-i-m: Hmm... that's good to know. I updated the PR. Thanks :)

spastorino: :+1: to me now, @petrochenkov ?

mark-i-m:
Failures are all spurious time outs...

eddyb:
I was surprised by this, but [at least it's not as bad as it used to be](https://github.com/rust-lang/rust/blob/49c68bd53f90e375bfb3cbba8c1c67a9e0adb9c0/config.toml.example#L287-L297).

I would be happier with us telling people to set `debug = true` if it defaulted to `debuginfo-level = 1` not `2`, and maybe had a different name but I don't know which one of them is more controversial.

PR #335, opened 2019-06-08
Changes to config.toml require a clean

fintelia:


fintelia:
`./x.py clean` doesn't rebuilt LLVM which means that you'll wait for a rebuild cycle and then discover that config changes still haven't taken effect.

mark-i-m:
Ah, I didn't know that. Could you add that as well please?

mark-i-m:
Thanks @fintelia !

PR #349, opened 2019-06-25
Add bibligraphy appendix

mark-i-m:
r? @spastorino

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -92,4 +92,5 @@
 [Appendix B: Background material](./appendix/background.md)
 [Appendix C: Glossary](./appendix/glossary.md)
 [Appendix D: Code Index](./appendix/code-index.md)
+[Appendix E: Bibliography](./appendix/bibliography.md)
 [](./important-links.md)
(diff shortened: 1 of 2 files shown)


Centril:
Hmm... This should perhaps go into the reference instead but we can have it here as a temporary location.

(More broadly, the rustc guide is starting to become more of the better reference than the actual reference...)

mark-i-m:
> (More broadly, the rustc guide is starting to become more of the better reference than the actual reference...)

In what sense? The rustc-guide is about the compiler, but I can't imagine it would be a very good language spec. Am I misunderstanding you?

Centril:
@mark-i-m In the sense that the rustc-guide has, in places, more formal specification of the language (albeit with many implementation details) than the reference. This is a reflection of the sorry state the reference is in :)

spastorino:
:+1:

mark-i-m:
@Centril I see what you mean, though I would be careful with that line of reasoning, since the guide describes the implementation, so it may over-specify in places (that is, the specification here may be less abstract than what the language actually guarantees).

mark-i-m:
Failure is spurious. Merging.

PR #380, opened 2019-07-04
Update pass modes for ui tests

### LINKED ISSUE #379: Update explanation about test annotations
This page [suggests using `compile-pass` as a test annotation](https://github.com/rust-lang/rustc-guide/blob/76be9f99d706d7ac4d10104f44483dd78d6d947a/src/tests/adding.md#tests-that-do-not-result-in-compile-errors). But it will be deprecated in https://github.com/rust-lang/rust/issues/62277. We should update this doc too.

JohnTitor:
Fixes #379 

cc @petrochenkov

### DIFF
--- a/src/tests/adding.md
+++ b/src/tests/adding.md
@@ -171,8 +171,8 @@ source.
   argument is the first version to ignore.  If no second argument is
   given, all subsequent versions are ignored; otherwise, the second
   argument is the last version to ignore.
-* `compile-pass` for UI tests, indicates that the test is
-  supposed to compile, as opposed to the default where the test is
+* `build-pass` for UI tests, indicates that the test is supposed to
+  successfully compile and link, as opposed to the default where the test is
   supposed to error out.
 * `compile-flags` passes extra command-line args to the compiler,
   e.g. `compile-flags -g` which forces debuginfo to be enabled.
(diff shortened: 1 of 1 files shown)


src/tests/adding.md
```diff
@@ -279,7 +279,9 @@ can also make UI tests where compilation is expected to succeed, and
 you can even run the resulting program. Just add one of the following
 [header commands](#header_commands):
 
-- `// compile-pass` – compilation should succeed but do
+- `// check-pass` - compilation should succeed but skip codegen
+  (which is expensive and isn't supposed to fail in most cases)
+- `// build-pass` – compilation should succeed but do
```
petrochenkov:
```suggestion
- `// build-pass` – compilation and linking should succeed but do
```

src/tests/adding.md
```diff
@@ -171,8 +171,8 @@ source.
   argument is the first version to ignore.  If no second argument is
   given, all subsequent versions are ignored; otherwise, the second
   argument is the last version to ignore.
-* `compile-pass` for UI tests, indicates that the test is
-  supposed to compile, as opposed to the default where the test is
+* `build-pass` for UI tests, indicates that the test is supposed to
+  compile, as opposed to the default where the test is
```
petrochenkov:
```suggestion
  successfully compile and link, as opposed to the default where the test is
```

src/tests/adding.md
JohnTitor: squashed and committed

PR #382, opened 2019-07-04
Fix broken links

JohnTitor:
CI is failed due to broken links so I fixed them.

r? @mark-i-m


petrochenkov:
I'm going to merge this now so that rustc-guide submodule could be updated in https://github.com/rust-lang/rust/pull/62369.

Issue #385, opened 2019-07-12
429 Too Many Requests


mark-i-m:
Recently a bunch of CI failures have occurred with this error, mostly from YouTube. Not sure what the solution is. I'm also not sure if this is Travis or YouTube...

@Mark-Simulacrum @pietroalbini do you know if this would be mitigated by moving to Azure?

Mark-Simulacrum:
Do you have links to some of these CI failures? Are you intentionally contacting YouTube?

shepmaster:
Is the link checker still active?

mark-i-m:
https://travis-ci.com/rust-lang/rustc-guide/builds/118875063

https://travis-ci.com/rust-lang/rustc-guide/builds/119000366

https://travis-ci.com/rust-lang/rustc-guide/builds/118874560

There are a few more I think.

We link to the compiler lecture series videos on YT. These failures are from the linkchecker attempting to contact YT and failing with 429.

pietroalbini:
Hmm, does linkchecker set a browser user agent?

mark-i-m:
No idea... cc @Michael-F-Bryan

Michael-F-Bryan:
I don't believe we set a user agent. The validator [just calls `reqwest::get()`](https://github.com/Michael-F-Bryan/mdbook-linkcheck/blob/6ac54352d029ec9288d3503d6c27afc62836653a/src/validation.rs#L36) without doing anything fancy.

mark-i-m:
For now, I am just going to blacklist yt feom the linkchecker because CI has been down for a while.

mark-i-m:
We haven't seen this in a while, so I am closing...

camelid:
This has been occurring (spuriously) again. Can we ignore `429 Too Many Requests` errors in linkcheck? (See #885 for some discussion.)

Issue #249, opened 2018-12-03
ACM Library links keep causing CI failures


mark-i-m:
Annoying, but maybe tolerable...

If this keeps happening, maybe we can just blacklist those two or three links?

mark-i-m:
This hasn't recurred since December, so I will close it. Feel free to reopen if it happens again.

PR #388, opened 2019-07-17
Blacklist some commonly problematic links

mark-i-m:
I think for now, the proper policy to follow is to blacklist links that fail often, especially when they are unlikely to change often.

We may also want to include a warning about unchecked links... or we could just let people find them for us...

cc #385 #249

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -9,4 +9,4 @@ description = "A guide to developing rustc"
 
 [output.linkcheck]
 follow-web-links = true
-exclude = [ "crates\\.io", "gcc\\.godbolt\\.org" ]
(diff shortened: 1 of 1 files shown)


Michael-F-Bryan:
> We may also want to include a warning about unchecked links... or we could just let people find them for us...

@mark-i-m, would some sort of "treat broken links as warnings instead of errors" flag in `mdbook-linkcheck` help here? I'm going through and rewriting a bunch of internals to make it cleaner, faster, and do caching, so a small config update wouldn't be hard to slip in.

mark-i-m:
@Michael-F-Bryan I think it would be helpful to be able to not fail the build for certain types of errors. For example, a 404 should definitely fail the build, whereas a time out, 429 (too many requests), or 50x (internal error) should not fail the build.

PR #400, opened 2019-07-25
Fix broken links

guilhermeleobas:
This PR fixes a couple of broken links found on `construction.md` file.


mark-i-m:
I don't think these are actually broken. mdbook will translate them to the appropriate paths.

But thanks anyway :)

Issue #355, opened 2019-06-29
Transcribe the Compiler Lecture How Salsa Works


spastorino:
Video: https://www.youtube.com/watch?v=_muY4HjSqVw

This issue has been assigned to @Karrq via [this comment](https://github.com/rust-lang/rustc-guide/issues/355#issuecomment-514781388).

PR #401, opened 2019-07-26
Compiler Lecture: How Salsa works

Karrq:
#355 

cc @spastorino 

Some questions left regarding content (see footnotes).

Other doubts (from me):
* Should this be put in another section in the summary, and group other compiler lectures there?
* Comparing this to the other chapter that talk about incremental computation this chapter is pretty high-level and talks about a feature external to the compiler (AFAIK it hasn't been integrated in the compiler yet), does it belong here, in the rustc-guide?

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -31,6 +31,7 @@
         - [The Query Evaluation Model in Detail](./queries/query-evaluation-model-in-detail.md)
         - [Incremental compilation](./queries/incremental-compilation.md)
         - [Incremental compilation In Detail](./queries/incremental-compilation-in-detail.md)
+        - [How Salsa works](./compiler_lectures/how_salsa_works.md)
         - [Debugging and Testing](./incrcomp-debugging.md)
     - [The parser](./the-parser.md)
     - [`#[test]` Implementation](./test-implementation.md)
--- a/src/compiler_lectures/how_salsa_works.md
+++ b/src/compiler_lectures/how_salsa_works.md
@@ -8,10 +8,10 @@ Salsa is a library for incremental recomputation, this means reusing computation
 
 The objectives of Salsa are:
  * Provide that functionality in an automatic way, so reusing old computations is done automatically by the library
- * Doing so in a "sound", or "correct", way, therefore leading the same results as if it had been done from scratch
+ * Doing so in a "sound", or "correct", way, therefore leading to the same results as if it had been done from scratch
 
-Salsa actual model is much richer, allowing many kinds of inputs and many different outputs.
-For example, integrating Salsa with an IDE could mean that the inputs could be the manifest (`Cargo.toml`), entire source files (`foo.rs`), snippets and so on; the outputs of such integration could range from a binary executable, to lints, types (for example, if a user selects a certain variable and wishes to see it's type), completitions etcetera.
+Salsa's actual model is much richer, allowing many kinds of inputs and many different outputs.
+For example, integrating Salsa with an IDE could mean that the inputs could be the manifest (`Cargo.toml`), entire source files (`foo.rs`), snippets and so on; the outputs of such an integration could range from a binary executable, to lints, types (for example, if a user selects a certain variable and wishes to see its type), completions, etc.
 
 ## How does it work?
 
@@ -23,11 +23,11 @@ For example, there might be a function `ast(x: Path) -> AST`. The produced `AST`
 
 This means that when you try to compute with the library, Salsa is going to compute various derived values, and eventually read the input and produce the result for the asked computation.
 
-Salsa is going to track, in the course of computing, which inputs were accessed, which derived values, and this is going to be used later to determine what's going to happen when the inputs change: are the derived values still valid? 
+In the course of computing, Salsa tracks which inputs were accessed and which values are derived. This information is used to determine what's going to happen when the inputs change: are the derived values still valid? 
 
-This doesn't mean necessarily that each computation downstream from the input is going to be checked, since that is going to be costly. Salsa only needs to check each downstream computation until it finds one that isn't changed, therefore it won't need to check the other derived computations, since they wouldn't need to change.
+This doesn't necessarily mean that each computation downstream from the input is going to be checked, which could be costly. Salsa only needs to check each downstream computation until it finds one that isn't changed. At that point, it won't check other derived computations since they wouldn't need to change.
 
-It's is helpful to think about this as a graph with nodes. Each derived value has a dependency on some other values, that could be base or derived. Base values don't have a dependency.
+It's is helpful to think about this as a graph with nodes. Each derived value has a dependency on other values, which could themselves be either base or derived. Base values don't have a dependency.
 
 ```ignore
 I <- A <- C ...
@@ -35,7 +35,7 @@ I <- A <- C ...
 J <- B <--+
 ```
 
-When an input `I` changes, the derived value `A` could have changed, but another derived value `B` , which doesn't depend neither on `I`, nor does it depends on `A`, nor does it depends on any other derived value from `A` or `I`, is not subject to change.
+When an input `I` changes, the derived value `A` could change. The derived value `B` , which does not depend on `I`, `A`, or any value derived from `A` or `I`, is not subject to change.
 Therefore, Salsa can reuse the computation done for `B` in the past, without having to compute it again.
 
 The computation could also terminate early. Keeping the same graph as before, say that input `I` has changed in some way (and input `J` hasn't) but, when computing `A` again, it's found that `A` hasn't changed from the previous computation. This leads to an "early termination", because there's no need to check if `C` needs to change, since both `C` direct inputs, `A` and `B`, haven't changed.
@@ -46,8 +46,8 @@ A query is some value that Salsa can access in the course of computation.
 Each query can have a number of keys (from 0 to many), and all queries have a result, akin to functions.
 0-key queries are called "input" queries.
 ### Database
-The database is basically the context for the entire computation, it's gonna store all Salsa's internal state, all intermidiate values for each query, and anything else that the computation might need.
-The database to know all the queries that the library is going to do before it can be build, but they don't need to be specified in the same place.
+The database is basically the context for the entire computation, it's gonna store all Salsa's internal state, all intermediate values for each query, and anything else that the computation might need.
+The database must know all the queries that the library is going to do before it can be built, but they don't need to be specified in the same place.
 
 After the database is formed, it can be accessed with queries that are basically going to work like functions.
 Since each query is going to be stored in the database, when a query is invoked N times, it's going to return N **cloned** results, without having to recompute the query (unless the input has changed in such a way that it warrants recomputation).
@@ -55,7 +55,7 @@ Since each query is going to be stored in the database, when a query is invoked 
 For each input query (0-key), there's going to be a "set" method, that allows to change the output of such query, and trigger previous memoized values to be potentially invalidated.
 
 ### Query Groups
-A query group is a set of queries which have been defined together as a unit. The dabase is formed by combining query groups.
+A query group is a set of queries which have been defined together as a unit. The database is formed by combining query groups.
 Query groups are akin to "Salsa modules" [^EN2].
 
 A set of queries in a query group are just a set of methods in a trait.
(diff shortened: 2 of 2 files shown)


spastorino:
Would be great if @nikomatsakis can also review this :)

src/compiler_lectures/how_salsa_works.md
```diff
@@ -0,0 +1,169 @@
...
+
+This chapter is based on the explanation given by Niko Matsakis in this [video](https://www.youtube.com/watch?v=_muY4HjSqVw) about [Salsa](https://github.com/salsa-rs/salsa).
+
+## What is Salsa?
+
+Salsa is a library for incremental recomputation, this means reusing computation that has already been done in the past to increase the efficiency of future computations.
+
+The objectives of Salsa are:
+ * Provide that functionality in an automatic way, so reusing old computations is done automatically by the library
+ * Doing so in a "sound", or "correct", way, therefore leading the same results as if it had been done from scratch
```
shepmaster:
```suggestion
 * Doing so in a "sound", or "correct", way, therefore leading to the same results as if it had been done from scratch
```

src/compiler_lectures/how_salsa_works.md
```diff
@@ -0,0 +1,169 @@
...
+
+Salsa actual model is much richer, allowing many kinds of inputs and many different outputs.
```
shepmaster:
```suggestion
Salsa's actual model is much richer, allowing many kinds of inputs and many different outputs.
```

src/compiler_lectures/how_salsa_works.md
```diff
@@ -0,0 +1,169 @@
...
+For example, integrating Salsa with an IDE could mean that the inputs could be the manifest (`Cargo.toml`), entire source files (`foo.rs`), snippets and so on; the outputs of such integration could range from a binary executable, to lints, types (for example, if a user selects a certain variable and wishes to see it's type), completitions etcetera.
```
shepmaster:
```suggestion
For example, integrating Salsa with an IDE could mean that the inputs could be the manifest (`Cargo.toml`), entire source files (`foo.rs`), snippets and so on; the outputs of such an integration could range from a binary executable, to lints, types (for example, if a user selects a certain variable and wishes to see its type), completions, etc.
```

src/compiler_lectures/how_salsa_works.md
```diff
@@ -0,0 +1,169 @@
...
+
+The first thing that Salsa has to do is identify the "base inputs" [^EN1].
+
+Then Salsa has to also identify intermidiate, "derived" values, which are something that the library produces, but, for each derived value there's a "pure" function that computes the derived value.
+
+For example, there might be a function `ast(x: Path) -> AST`. The produced `AST` isn't a final value, it's an intermidiate value that the library would use for the computation.
+
+This means that when you try to compute with the library, Salsa is going to compute various derived values, and eventually read the input and produce the result for the asked computation.
+
+Salsa is going to track, in the course of computing, which inputs were accessed, which derived values, and this is going to be used later to determine what's going to happen when the inputs change: are the derived values still valid?
```
shepmaster:
```suggestion
In the course of computing, Salsa tracks which inputs were accessed and which values are derived. This information is used to determine what's going to happen when the inputs change: are the derived values still valid? 
```

src/compiler_lectures/how_salsa_works.md
```diff
@@ -0,0 +1,169 @@
...
+
+Then Salsa has to also identify intermidiate, "derived" values, which are something that the library produces, but, for each derived value there's a "pure" function that computes the derived value.
+
+For example, there might be a function `ast(x: Path) -> AST`. The produced `AST` isn't a final value, it's an intermidiate value that the library would use for the computation.
+
+This means that when you try to compute with the library, Salsa is going to compute various derived values, and eventually read the input and produce the result for the asked computation.
+
+Salsa is going to track, in the course of computing, which inputs were accessed, which derived values, and this is going to be used later to determine what's going to happen when the inputs change: are the derived values still valid? 
+
+This doesn't mean necessarily that each computation downstream from the input is going to be checked, since that is going to be costly. Salsa only needs to check each downstream computation until it finds one that isn't changed, therefore it won't need to check the other derived computations, since they wouldn't need to change.
```
shepmaster:
```suggestion
This doesn't necessarily mean that each computation downstream from the input is going to be checked, which could be costly. Salsa only needs to check each downstream computation until it finds one that isn't changed. At that point, it won't check other derived computations since they wouldn't need to change.
```

src/compiler_lectures/how_salsa_works.md
```diff
@@ -0,0 +1,169 @@
...
+
+It's is helpful to think about this as a graph with nodes. Each derived value has a dependency on some other values, that could be base or derived. Base values don't have a dependency.
```
shepmaster:
```suggestion
It's is helpful to think about this as a graph with nodes. Each derived value has a dependency on other values, which could themselves be either base or derived. Base values don't have a dependency.
```

src/compiler_lectures/how_salsa_works.md
```diff
@@ -0,0 +1,169 @@
...
+
+```ignore
+I <- A <- C ...
+          |
+J <- B <--+
+```
+
+When an input `I` changes, the derived value `A` could have changed, but another derived value `B` , which doesn't depend neither on `I`, nor does it depends on `A`, nor does it depends on any other derived value from `A` or `I`, is not subject to change.
```
shepmaster:
```suggestion
When an input `I` changes, the derived value `A` could change. The derived value `B` , which does not depend on `I`, `A`, or any value derived from `A` or `I`, is not subject to change.
```

src/compiler_lectures/how_salsa_works.md
```diff
@@ -0,0 +1,169 @@
...
+
+The computation could also terminate early. Keeping the same graph as before, say that input `I` has changed in some way (and input `J` hasn't) but, when computing `A` again, it's found that `A` hasn't changed from the previous computation. This leads to an "early termination", because there's no need to check if `C` needs to change, since both `C` direct inputs, `A` and `B`, haven't changed.
+
+## Key Salsa concepts
+### Query
+A query is some value that Salsa can access in the course of computation.
+Each query can have a number of keys (from 0 to many), and all queries have a result, akin to functions.
+0-key queries are called "input" queries.
+### Database
+The database is basically the context for the entire computation, it's gonna store all Salsa's internal state, all intermidiate values for each query, and anything else that the computation might need.
```
shepmaster:
```suggestion
The database is basically the context for the entire computation, it's gonna store all Salsa's internal state, all intermediate values for each query, and anything else that the computation might need.
```

src/compiler_lectures/how_salsa_works.md
```diff
@@ -0,0 +1,169 @@
...
+The database to know all the queries that the library is going to do before it can be build, but they don't need to be specified in the same place.
```
shepmaster:
```suggestion
The database must know all the queries that the library is going to do before it can be built, but they don't need to be specified in the same place.
```

src/compiler_lectures/how_salsa_works.md
```diff
@@ -0,0 +1,169 @@
...
+
+After the database is formed, it can be accessed with queries that are basically going to work like functions.
+Since each query is going to be stored in the database, when a query is invoked N times, it's going to return N **cloned** results, without having to recompute the query (unless the input has changed in such a way that it warrants recomputation).
+
+For each input query (0-key), there's going to be a "set" method, that allows to change the output of such query, and trigger previous memoized values to be potentially invalidated.
```
shepmaster: All of the "going to be" phrasing makes it sound like these are features that Salsa could have in the future, but doesn't currently. I'd review the text to phrase it in the present tense.

src/compiler_lectures/how_salsa_works.md
```diff
@@ -0,0 +1,169 @@
...
+
+### Query Groups
+A query group is a set of queries which have been defined together as a unit. The dabase is formed by combining query groups.
```
shepmaster:
```suggestion
A query group is a set of queries which have been defined together as a unit. The database is formed by combining query groups.
```

src/compiler_lectures/how_salsa_works.md
```diff
@@ -0,0 +1,169 @@
...
+///produce this tree as output, 
+///and produce a bunch of intermidiate stuff
+///that Salsa also uses.
+///One of these things is a "StorageStruct", whose name we have specified in the attribute.
+#[salsa::query_group(InputsStorage)]
+pub trait Inputs {
+    //! This query group is a bunch of **input** queries, that do not rely on any derived input
+
+    ///This attribute (`#[salsa::input]`) indicates that this query is a base input, 
+    ///therfore `set_manifest` is going to be auto-generated
```
shepmaster:
```suggestion
    ///therefore `set_manifest` is going to be auto-generated
```

src/compiler_lectures/how_salsa_works.md
```diff
@@ -0,0 +1,169 @@
...
+    #[salsa::input]
+    fn manifest(&self) -> Manifest;
+    
+    #[salsa::input]
+    fn source_text(&self, name: String) -> String;
+}
+```
+
+To create a **derived** query group, one must specify which other query group does this one depends on, by specifying it as a supertrait, as seen in the following example:
```
shepmaster:
```suggestion
To create a **derived** query group, one must specify which other query groups this one depends on by specifying them as supertraits, as seen in the following example:
```

src/compiler_lectures/how_salsa_works.md
Karrq: If you are talking about this section specifically, it's because this refers to something that's going to happen when the library is used... If talking about the whole article: the text is based of a video done a while ago, some stuff might not had yet been implemented by that point. I don't know what's in salsa now so I can't really tell if something should be on the present tense. Either way, I decided to rewrite this section a bit with present tense as you suggested

mark-i-m:
@Karrq sorry for the delay. I will try to get to this soon.

src/compiler_lectures/how_salsa_works.md
```diff
@@ -0,0 +1,169 @@
...
+ * Doing so in a "sound", or "correct", way, therefore leading to the same results as if it had been done from scratch
+
+Salsa's actual model is much richer, allowing many kinds of inputs and many different outputs.
+For example, integrating Salsa with an IDE could mean that the inputs could be the manifest (`Cargo.toml`), entire source files (`foo.rs`), snippets and so on; the outputs of such an integration could range from a binary executable, to lints, types (for example, if a user selects a certain variable and wishes to see its type), completions, etc.
+
+## How does it work?
+
+The first thing that Salsa has to do is identify the "base inputs" [^EN1].
+
+Then Salsa has to also identify intermidiate, "derived" values, which are something that the library produces, but, for each derived value there's a "pure" function that computes the derived value.
```
mark-i-m:
```suggestion
Then Salsa has to also identify intermediate, "derived" values, which are something that the library produces, but, for each derived value there's a "pure" function that computes the derived value.
```

src/SUMMARY.md
```diff
@@ -31,6 +31,7 @@
         - [The Query Evaluation Model in Detail](./queries/query-evaluation-model-in-detail.md)
         - [Incremental compilation](./queries/incremental-compilation.md)
         - [Incremental compilation In Detail](./queries/incremental-compilation-in-detail.md)
+        - [How Salsa works](./compiler_lectures/how_salsa_works.md)
```
mark-i-m: Perhaps move this after the debugging subchapter?

src/compiler_lectures/how_salsa_works.md
```diff
@@ -0,0 +1,169 @@
+# How Salsa works
+
+This chapter is based on the explanation given by Niko Matsakis in this [video](https://www.youtube.com/watch?v=_muY4HjSqVw) about [Salsa](https://github.com/salsa-rs/salsa).
+
+## What is Salsa?
+
+Salsa is a library for incremental recomputation, this means reusing computation that has already been done in the past to increase the efficiency of future computations.
```
mark-i-m: Perhaps add a note that it is not used directly in rustc, but it is used extensively for rust-analyzer and may be integrated into the compiler in the future

mark-i-m:
Oh, also, could you wrap the lines to 100 chars? That will appease the CI.

src/compiler_lectures/how_salsa_works.md
spastorino: Exactly what @mark-i-m is mentioning here is what I was going to ask :).

spastorino:
This is good enough, I saw some parts that mention something about inaudible parts of the video, I'm not sure what you plan to do with those. If we can get rid of those with something that still makes sense I think it would be better.

PR #402, opened 2019-07-31
Update section on "existential type" to "opaque type"

varkor:
Updates the guide as per the changes in https://github.com/rust-lang/rfcs/pull/2515.


Centril:
cc https://github.com/rust-lang/rust/issues/63063

src/opaque-types-type-alias-impl-trait.md
```diff
@@ -0,0 +1,54 @@
...
+fn foo() -> Foo {
+    Struct
+}
+```
+
+Any other "defining use site" needs to produce the exact same type.
+
+## Defining use site(s)
+
+Currently only the return value of a function inside can
```
Arnavion: `a function inside can` -> `a function can` ?

src/opaque-types-type-alias-impl-trait.md
```diff
@@ -0,0 +1,54 @@
...
+comprehend what the type system is doing"* has decided to disallow children
+of opaque types to be defining use sites.
+
+### Associated opaque types
+
+Associated opaque types can be defined by any other associated item
+on the same trait `impl` or a child of these associated items. For instance:
+
+```rust,ignore
+trait Baz {
```
Arnavion: I think this is meant to be a trait impl, not a trait definition. ie

```rust
trait Baz { type Foo; fn foo() -> Self::Foo; }

impl Baz for Quux { type Foo = impl Bar; fn foo() -> Self::Foo { ... } }
```

mark-i-m:
Ping @varkor there are some comments waiting for you

varkor:
Addressed the comments.

mark-i-m:
Thanks!

src/opaque-types-type-alias-impl-trait.md
```diff
@@ -0,0 +1,62 @@
...
+
+This declares an opaque type named `Foo`, of which the only information is that
+it implements `Bar`. Therefore, any of `Bar`'s interface can be used on a `Foo`,
+but nothing else (regardless of whether it implements any other traits).
+
+Since there needs to be a concrete background type, you can currently
+express that type by using the opaque type in a "defining use site".
+
+```rust,ignore
+struct Struct;
```
kpp: Shouldn't it be `struct Foo;` ?

varkor: Yes it should! Feel free to open a PR with a fix :)

kpp: Then the whole example should be

```rust
fn foo() -> impl Bar {
    struct Foo;
    impl Bar for Foo { /* stuff */ }
    Foo
}
```

?

varkor: Sorry, I misread earlier. Really, this code snippet should be read in the context of the previous one. But that's confusing, so maybe we should add the definition of `Foo` here too. So the complete snippet would be:
```rust
type Foo = impl Bar;

struct Struct;

impl Bar for Struct { /* stuff */ }

fn foo() -> Foo {
    Struct
}
```

PR #409, opened 2019-08-01
run-pass dir is gone

### LINKED ISSUE #405: Update rustc-guide according to removal of run-pass
After @pnkfelix change https://github.com/rust-lang/rust/pull/53994, the CI of rustc-guide is giving errors because of links to run-pass github directory that does not exist anymore. But more than that, would be great to fix the mentions of run-pass in the guide.

tshepang:
Fixes #405


tshepang:
References remain to `run-fail`, so please have a look if I missed anything important.

PR #414, opened 2019-08-05
no problem was mentioned before then

tshepang:


spastorino:
What about using `which commits are causing problems`.

tshepang:
A "git bisect" would imply there is a problem, not problem*s*. Beyond that, it's not a pretty sentence, and I don't feel that `git bisect` needs to be mentioned at that point... feels too specific.

tshepang:
Re-reading the sentence, am not even sure my commit improves things, so feel free to close

spastorino:
What I thought you wanted to improve was the fact that saying a problem means that there's just one problem but could be more than one. If you keep coding without testing and at the end you see failures and run git blame there could be multiple problems.

If it makes sense to you go with something like that, otherwise, yeah we can close it I think.

PR #416, opened 2019-08-05
Hack around timeouts

mark-i-m:
r? @spastorino 

Basically we check if the error message is for a timeout ("error while fetching ...") and if all of the errors are timeouts we just pass. This extremely brittle, but it should work until we figure out something better...


spastorino:
I think it's good enough for now

PR #422, opened 2019-08-13
document how to use system llvm

matklad:


matklad:
Btw, I feel like there should be *some* caveats to this as, iirc, we ship patches to llvm. If someone more knowledgeable can expand on this, it would be great.

mark-i-m:
My understanding is that it works in the most common cases, especially if you are only making changes to the compiler front-end (before LLVM). However, if you need to test debuginfo or codegen stuff, you should use the official rust LLVM backend.

src/how-to-build-and-run.md
```diff
@@ -469,6 +469,18 @@ You can also use `--keep-stage 1` when running tests. Something like this:
...
+### Building with system LLVM
+
+By default, LLVM is built from source, and that can take significant amount of time.
+An alternative is to use LLVM already installed on your computer.
+
+This is specified in the `target` section of `config.toml`:
+
+```toml
+[target.x86_64-unknown-linux-gnu]
+llvm-config = "/path/to/llvm/llvm-7.0.1/bin/llvm-config"
```
Centril: Common paths for e.g. ubuntu and popular dists would be great :)

mark-i-m: @Centril Could you open an issue for that? Otherwise, this is likely to just be forgotten. Alternately, could you open a PR?

Centril: @mark-i-m Well I don't know what the common paths are. :)

mark-i-m: On my system it is `/usr/bin/llvm-config-7`

I will add this to my next PR.

PR #424, opened 2019-08-21
Explain our stability attributes

oli-obk:
r? @centril

### DIFF
--- a/src/appendix/bibliography.md
+++ b/src/appendix/bibliography.md
@@ -61,7 +61,7 @@ Rust, as well as publications about Rust.
 * [Reenix: Implementing a Unix-Like Operating System in
   Rust](http://scialex.github.io/reenix.pdf). Undergrad paper by Alex
   Light.
-* [Evaluation of performance and productivity metrics of potential programming languages in the HPC environment](http://octarineparrot.com/assets/mrfloya-thesis-ba.pdf).
+* [Evaluation of performance and productivity metrics of potential programming languages in the HPC environment](https://github.com/1wilkens/thesis-ba).
   Bachelor's thesis by Florian Wilkens. Compares C, Go and Rust.
 * [Nom, a byte oriented, streaming, zero copy, parser combinators library
   in Rust](http://spw15.langsec.org/papers/couprie-nom.pdf). By
@@ -81,7 +81,7 @@ Rust, as well as publications about Rust.
 * [Ownership is Theft: Experiences Building an Embedded OS in Rust - Amit Levy, et. al.](http://amitlevy.com/papers/tock-plos2015.pdf)
 * [You can't spell trust without Rust](https://raw.githubusercontent.com/Gankro/thesis/master/thesis.pdf). Alexis Beingessner's master's thesis.
 * [Rust-Bio: a fast and safe bioinformatics library](http://bioinformatics.oxfordjournals.org/content/early/2015/10/06/bioinformatics.btv573). Johannes Köster
-* [Safe, Correct, and Fast Low-Level Networking](https://octarineparrot.com/assets/msci_paper.pdf). Robert Clipsham's master's thesis.
+* [Safe, Correct, and Fast Low-Level Networking](http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.704.1768). Robert Clipsham's master's thesis.
 * [Formalizing Rust traits](http://hdl.handle.net/2429/55609). Jonatan Milewski's master's thesis.
 * [Rust as a Language for High Performance GC Implementation](http://users.cecs.anu.edu.au/~steveb/downloads/pdf/rust-ismm-2016.pdf)
 * [Simple Verification of Rust Programs via Functional Purification](https://github.com/Kha/electrolysis). Sebastian Ullrich's master's thesis.


src/stability.md
```diff
@@ -0,0 +1,35 @@
...
+
+This section is about the stability attributes and schemes that allow stable APIs to use unstable
+APIs internally in the rustc standard library.
+
+For instructions on stabilizing a feature see [Stabilizing Features](./stabilization_guide.md).
+
+# unstable
+
+The `#[unstable(feature = "foo", issue = "1234", reason = "lorem ipsum")]` attribute explicitly
+marks an item as unstable. This infects all sub-items, where the attribute doesn't have to be
```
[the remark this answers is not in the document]
oli-obk: should I be explaining that? I assumed that to be obvious, but since you're asking it seems that it is not. It's essentially the same rules that `pub` has

src/stability.md
```diff
@@ -0,0 +1,35 @@
+# Stability
```
Centril:
```suggestion
# Stability attributes
```

src/stability.md
```diff
@@ -0,0 +1,35 @@
...
+
+This section is about the stability attributes and schemes that allow stable APIs to use unstable
+APIs internally in the rustc standard library.
+
+For instructions on stabilizing a feature see [Stabilizing Features](./stabilization_guide.md).
```
Centril:
```suggestion
For instructions on stabilizing a language feature see [Stabilizing Features](./stabilization_guide.md).
```

src/stability.md
```diff
@@ -0,0 +1,35 @@
...
+# unstable
+
+The `#[unstable(feature = "foo", issue = "1234", reason = "lorem ipsum")]` attribute explicitly
+marks an item as unstable. This infects all sub-items, where the attribute doesn't have to be
+reapplied. So if you apply this to a module, all items in the module will be unstable.
+
+# stable
+
+The `#[stable(feature = "foo", "since = "1.420.69")]` attribute explicitly marks an item as
+stabilized. In order to do this follow the instructions in
```
Centril:
```suggestion
stabilized. To do this, follow the instructions in
```

src/stability.md
```diff
@@ -0,0 +1,35 @@
...
+[Stabilizing Features](./stabilization_guide.md).
+
+Note that stable functions may use unstable things in their body.
+
+# allow_internal_unstable
+
+Macros, compiler desugarings and `const fn`s expose their bodies to the call site. In order to
```
Centril:
```suggestion
Macros, compiler desugarings and `const fn`s expose their bodies to the call site. To
```

src/stability.md
```diff
@@ -0,0 +1,35 @@
...
+`#[allow_internal_unstable(feature1, feature2)]` attribute that whitelists the given features for
+usage in stable macros or `const fn`s.
+
+Note that `const fn`s are even more special in this regard. You can't just whitelist any feature,
+the features need an implementation in `qualify_min_const_fn.rs`. For example the `const_fn_union`
+feature gate allows accessing fields of unions inside stable `const fn`s. The rules for when it's
+ok to use such a feature gate are that behavior matches the runtime behavior of the same code
+(see also https://www.ralfj.de/blog/2018/07/19/const.html). This means that you may not create a
+`const fn` that e.g. transmutes a memory address to an integer, because the addresses of things
+are nondeterministic and often unknown at compile-time.
```
Centril: A note re. FCP & lang team would be good also.

src/stability.md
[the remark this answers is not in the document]
Centril: Probably a good idea to add a note re. "essentially `pub`".

src/stability.md
oli-obk: This is a libs thing, why does it need lang team FCP? It seems like a waste of resources to me to require lang team FCP. We don't do lang team FCPs on arbitrary feature gate usage in macros or desugarings.

Centril: At minimum I want to be personally notified time one of these attributes are used to ensure that the policy re. determinism is actually upheld and if it isn't then I don't want any libs member objecting when I eventually do `@bors r-`.

Centril: Saying something akin to *"Always ping @oli-obk, @RalfJung, and @Centril"* would do it I suppose.

oli-obk: Noone objects to things getting `r-` with a good reason.

Adding the suggestion to ping us makes sense though.

pietroalbini:
> (I don't have r+ here -- @pietroalbini can that be arranged for?)

I guess ask Niko?

Centril:
cc @nikomatsakis ^---

src/stability.md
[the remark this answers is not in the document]
shepmaster: The first example makes some amount of sense to me intuitively, but the second doesn't. How do I use a stable function inside of an unstable module? Wouldn't I get a warning just by accessing the module to import the function?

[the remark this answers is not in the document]
oli-obk: similar to `pub`, this pattern is often used when the function is reexported elsewhere, but the module is still unstable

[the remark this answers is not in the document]
RalfJung: I think this is worth explaining more, actually. In particular, there is a limitation here in our stability check which *does* make the location of the stable item in the unstable module available to stable code, so the module cannot be renamed or removed any more. The document should note that problem.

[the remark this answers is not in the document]
RalfJung: The relevant Rust bug is https://github.com/rust-lang/rust/issues/15702.

spastorino:
Looks good to me, unsure if you wanted to add more information or not, ended not being very clear to me from the conversations.

src/stability.md
```diff
@@ -0,0 +1,42 @@
...
+Macros, compiler desugarings and `const fn`s expose their bodies to the call site. To
+work around not being able to use unstable things in the standard library's macros, there's the
+`#[allow_internal_unstable(feature1, feature2)]` attribute that whitelists the given features for
+usage in stable macros or `const fn`s.
+
+Note that `const fn`s are even more special in this regard. You can't just whitelist any feature,
+the features need an implementation in `qualify_min_const_fn.rs`. For example the `const_fn_union`
+feature gate allows accessing fields of unions inside stable `const fn`s. The rules for when it's
+ok to use such a feature gate are that behavior matches the runtime behavior of the same code
+(see also https://www.ralfj.de/blog/2018/07/19/const.html). This means that you may not create a
```
mark-i-m: Could this be made into an actual link please? e.g. `See this [blog post](https://www.ralfj.de/blog/2018/07/19/const.html)`.

mark-i-m:
Hmm... I was trying to update this. It seems I broke it... I will create a new PR and merge it...

RalfJung:
I also left some comments in the text that seem to be gone now...?!?

PR #432, opened 2019-08-26
Updated the mdbook-linkcheck version

Michael-F-Bryan:
I recently rewrote large chunks of `mdbook-linkcheck` to take advantage of caching and parallelism, and thought you might want to take advantage of it for CI.

I'm not sure how this would interact with [these lines](https://github.com/rust-lang/rustc-guide/blob/78b63d19e57cfe3b309d6bf394640dfc193f403e/ci/build-ignore-timeouts.sh#L14-L15) in the `build-ignore-timeouts.sh` script, though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ rust:
 - nightly
 cache:
 - cargo
+- directories:
+  - book/linkcheck/
 before_install:
 - shopt -s globstar
 - MAX_LINE_LENGTH=100 bash ci/check_line_lengths.sh src/**/*.md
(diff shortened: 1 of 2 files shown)


Michael-F-Bryan:
@mark-i-m, done.

Michael-F-Bryan:
Do you know why certain links time out? There's an option for increasing the timeout period, but that won't necessarily fix the problem.

mark-i-m:
> Do you know why certain links time out? There's an option for increasing the timeout period, but that won't necessarily fix the problem.

Unfortunately, no. It seems to be Travis throttling us, since I don't see this failures locally.

Issue #431, opened 2019-08-25
libtest build step has been merged with that of libstd


tshepang:
That means How to Build and Run the Compiler needs updating.

For reference, this change was done in https://github.com/rust-lang/rust/pull/63637.

mark-i-m:
@tshepang Would you be able to submit a PR?

tshepang:
I will

tshepang:
status: diagram still needs updating

PR #434, opened 2019-08-26
merge libtest build step with libstd

tshepang:
Partially handles #431


tshepang:
The diagram needs updating also, something I will have to figure out later (unless someone that's done it before can do it in a separate commit)

tshepang:
Rather not wait

PR #437, opened 2019-09-02
Fix a link in the query docs.

vext01:
I'm not sure why the link checker didn't catch this...

There are a couple of other broken links I've not touched too:
```
error: Unable to retrieve "https://public.etherpad-mozilla.org/p/rust-compiler-meeting": https://public.e
therpad-mozilla.org/p/rust-compiler-meeting: error trying to connect: Connection refused (os error 111)
- compiler-team.md:32:1                                                                                 
   |                                                                                                    
32 | [the rust-compiler-meeting etherpad][etherpad]. It works roughly as                                
   |                                                                                                    
```

^ Genuinely a dead link.

```
error: Unable to retrieve "http://www.cs.bgu.ac.il/%7Ehendlerd/papers/p280-hendler.pdf": https://www.cs.b
gu.ac.il/%7Ehendlerd/papers/p280-hendler.pdf: error trying to connect: error:141A318A:SSL routines:tls_pr
ocess_ske_dhe:dh key too small:../ssl/statem/statem_clnt.c:2156:         
- appendix/bibliography.md:33:3                                          
   |                                                                       
33 | * [Non-blocking steal-half work queues](http://www.cs.bgu.ac.il/%7Ehendlerd/papers/p280-hendler.pdf)
   |                                                                             
```

^ Link works for me.

### DIFF
--- a/src/query.md
+++ b/src/query.md
@@ -35,9 +35,9 @@ will in turn demand information about that crate, starting from the
 However, that vision is not fully realized. Still, big chunks of the
 compiler (for example, generating MIR) work exactly like this.
 
-### The Query Evaluation Model in Detail
+### Incremental Compilation in Detail
 
-The [Query Evaluation Model in Detail][query-model] chapter gives a more
+The [Incremental Compilation in Detail][query-model] chapter gives a more
 in-depth description of what queries are and how they work.
 If you intend to write a query of your own, this is a good read.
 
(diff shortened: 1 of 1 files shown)


mark-i-m:
cc @Michael-F-Bryan

Michael-F-Bryan:
That second error seems funny. The original link uses http, but then it upgrades to https and the TLS key exchange fails with a "key too small" error. 

I'm guessing it was rejected because the version of TLS used secure enough... I'll have a look and see if reqwest let's you tweak how TLS is handled, then we can add an option for relaxing TLS to the config.

PR #446, opened 2019-09-17
Add some instructions to fix toolstate

mark-i-m:
r? @spastorino

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,3 +41,31 @@ be found.
 
 You will need `mdbook` version `>= 0.2`. `linkcheck` will be run automatically
 when you run `mdbook build`.
+
+## How to fix toolstate failures
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,3 +41,31 @@ be found.
...
+
+7. Click on the log and Ctrl-f to get a search box in the log
+
+8. Search for rustc-guide. This gets you to the place where the links are checked. It is usually ~11K lines into the log
+
+9. Look at the links in the log near that point in the log
+
+10. Fix those links in the rustc-guide (by making a PR in the rustc-guide repo)
+
+11. Make a PR on the rust-lang/rust repo to update the rustc-guide git submodule in src/docs/rustc-guide
```
JohnTitor: IMO: It'd be nice that we have the guidance to make a PR something like this here: https://github.com/rust-lang/rust-clippy/blob/master/CONTRIBUTING.md#fixing-build-failures-caused-by-rust

spastorino: @JohnTitor :+1:, I've just merged this but we could work on something like what you're suggesting on a different PR. Feel free to provide a PR for that or open an issue.

PR #450, opened 2019-09-20
I prefers long options in documentation

tshepang:


spastorino:
IMO short options are more common but let's wait what others say.

tshepang:
With long options, there's better chances that user that does not know (or forgot) will make a good guess what the option means, instead of having to look it up.

mark-i-m:
I prefer short options because they're shorter :P

But I think it's a good idea to have the long options for intuition. Perhaps we can have both?

tshepang:
Since this is documentation, one could simply copy-paste (and perhaps modify) stuff, so no extra typing needed.

mark-i-m:
True, but it's also a teaching aid, and it's useful to know the shortcuts. I think it's ok to just mention both...

tshepang:
Descending order of what I prefer...

- go for `--long-options`
- go for `-s` options
- go for `-s | --long-options`

The last one means I lose copy-paste convenience, unless you mean duplicating instructions?

spastorino:
I'm not convinced but I may agree with changing `-i` with `--incremental` because this is a rustc guide. About git, I'm less convinced because it's more common to use `-m` than `--message`, the same for `-u` and this is even not a git guide.

spastorino:
BTW about `-i` and `--incremental` somewhere in the guide it should be clear that `-i` is a shortcut for `--incremental` unsure if that's explained somewhere but it should. After we explain all the flags and their shortcuts, I think it's fine to just use shortcuts everywhere.

tshepang:
There's lots of CLI tools out there, and seeing yet-another `-o` to remember in another such tool adds to cognitive load. To the regular user of said tool who is tired of typing `--option`, let it be on them to look for `--help` so as to see if there is a short option they can use. I rather we optimize for the new user... I want them to have a good guess as to what the `--option` means, instead of them having to reach out to `--help` to find out.

mark-i-m:
I was thinking more of having a note after the command that says something like "-i is the short version of --incremental"

PR #457, opened 2019-10-01
fix chalk links

tshepang:


JohnTitor:
According to the CI, we should update more links.

JohnTitor:
You should update the links in this file (https://github.com/rust-lang/rustc-guide/blob/master/src/macro-expansion.md) from `ext/tt` to `ext/mbe`.

tshepang:
Ok, was only focused on chalk links... will fix those also.

JohnTitor:
Thanks! Looks good to me.

PR #458, opened 2019-10-02
Removes timeout script so CI does not have false positives

amanjeev:


amanjeev:
@mark-i-m Seems like the build is ok. Please check.

ci/build-ignore-timeouts.sh
```diff
@@ -1,29 +0,0 @@
...
-output=$(mktemp)
-
-mkdir -p book/
-cp -r $HOME/linkcheck/ book/
-RUST_LOG=mdbook_linkcheck=debug mdbook-linkcheck -s 2>&1 | tee -a $output
-cp -r book/linkcheck $HOME/
-
-mdbook build
-
-result=${PIPESTATUS[0]}
```
mark-i-m: We need to keep at least this part because this is what invokes the linkchecker.

ci/build-ignore-timeouts.sh
```diff
@@ -1,29 +0,0 @@
...
-timeouts=$(cat $output | grep "error while fetching" | wc -l)
-
-# if all errors are timeouts, ignore them...
-if [ "$errors" -eq "$timeouts" ] ; then
-    echo "Ignoring $timeouts timeouts";
-    exit 0;
-else
-    echo "Non-timeout errors found";
-    exit 1;
-fi
```
mark-i-m: This is the part we want to remove.

amanjeev: :rofl: my bad. the script is called `build-ignore-timeouts.sh` so threw it out of the window!

ci/build.sh
```diff
@@ -0,0 +1,17 @@
...
+cp -r book/linkcheck $HOME/
+
+mdbook build
+
+result=${PIPESTATUS[0]}
+
+# if passed, great!
+if [ "$result" -eq "0" ] ; then
+    exit 0 ;
+fi
```
mark-i-m: Do we need to `exit 1` or something in the `else` case?

amanjeev: I agree. Better be explicit.

amanjeev:
> Thanks @amanjeev!
> 
> Did you get a chance to test it out? e.g. could you intentionally change one of the links to `foo.com/asdf` and make sure that `bash -x ci/build.sh` returns a non-zero exit code?

You what, I ran it and assumed it was working. However, no matter if you have broken links, the exit status is still `0` as `PIPESTATUS[0]` is `0`.

amanjeev:
@mark-i-m  I made some changes. My bash is not that strong so please feel free to revoke the approval and comment.

mark-i-m:
@amanjeev Thanks!

As I was looking at it, I realized that we don't need to do all that stuff to capture the output. I ended up with the following simplified version, which seems to work locally for me:

```bash
#!/bin/bash -x

set -e

mkdir -p book/
cp -r $HOME/linkcheck/ book/
RUST_LOG=mdbook_linkcheck=debug mdbook-linkcheck -s
cp -r book/linkcheck $HOME/

mdbook build
```

amanjeev:
Thank you @mark-i-m  I have changed the script and pushed.

Is it odd that only some of the links fail the linkcheck? For example changing this link does not fail the linkcheck https://github.com/rust-lang/rustc-guide/blame/master/src/debugging-support-in-rustc.md#L302

amanjeev:
Can someone summon @bors  for merging if all ok? :smile:

PR #460, opened 2019-10-07
Don't include ">" in bash scripts

hbina:
This makes copying inconvenient as `> ./x.py build -i --stage 1 src/libstd` will actually render your `x.py` file empty...

### DIFF
--- a/src/profiling/with_perf.md
+++ b/src/profiling/with_perf.md
@@ -28,7 +28,7 @@ of events, though, like cache misses and so forth.
 The basic `perf` command is this:
 
 ```bash
-> perf record -F99 --call-graph dwarf XXX
+perf record -F99 --call-graph dwarf XXX
 ```
 
 The `-F99` tells perf to sample at 99 Hz, which avoids generating too
@@ -39,7 +39,7 @@ information from debuginfo, which is accurate. The `XXX` is the
 command you want to profile. So, for example, you might do:
 
 ```bash
-> perf record -F99 --call-graph dwarf cargo +<toolchain> rustc
+perf record -F99 --call-graph dwarf cargo +<toolchain> rustc
 ```
 
 to run `cargo` -- here `<toolchain>` should be the name of the toolchain
@@ -59,7 +59,7 @@ do that, the first step is to clone
 [the rustc-perf repository][rustc-perf-gh]:
 
 ```bash
-> git clone https://github.com/rust-lang-nursery/rustc-perf
+git clone https://github.com/rust-lang-nursery/rustc-perf
 ```
 
 [rustc-perf-gh]: https://github.com/rust-lang-nursery/rustc-perf
@@ -75,13 +75,13 @@ do profiling for you! You can find
 For example, to measure the clap-rs test, you might do:
 
 ```bash
-> ./target/release/collector
-    --output-repo /path/to/place/output
-    profile perf-record
-    --rustc /path/to/rustc/executable/from/your/build/directory
-    --cargo `which cargo`
-    --filter clap-rs
-    --builds Check
+./target/release/collector                                      \
+  --output-repo /path/to/place/output                           \
+  profile perf-record                                           \
+  --rustc /path/to/rustc/executable/from/your/build/directory   \
+  --cargo `which cargo`                                         \
+  --filter clap-rs                                              \
+  --builds Check                                                \
 ```
 
 You can also use that same command to use cachegrind or other profiling tools.
@@ -97,7 +97,7 @@ example:
 [dir]: https://github.com/rust-lang-nursery/rustc-perf/tree/master/collector/benchmarks
 
 ```bash
-> cd collector/benchmarks/clap-rs
+cd collector/benchmarks/clap-rs
 ```
 
 In this case, let's say we want to profile the `cargo check`
@@ -106,8 +106,8 @@ build the dependencies:
 
 ```bash
 # Setup: first clean out any old results and build the dependencies:
-> cargo +<toolchain> clean
-> CARGO_INCREMENTAL=0 cargo +<toolchain> check
+cargo +<toolchain> clean
+CARGO_INCREMENTAL=0 cargo +<toolchain> check
 ```
 
 (Again, `<toolchain>` should be replaced with the name of the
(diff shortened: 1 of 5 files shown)


spastorino:
We could use a $ sign but please check out through the whole guide because there are tons of "```bash" pieces and we would need to update them all.

hbina:
>We could use a `$` sign
I don't quite understand. Even if I replace `>` it will simply copy the entire text and `$` is not a valid bash command...

> please check out through the whole guide because there are tons of "```bash" pieces and we would need to update them all.
Okay, I am going to apply the changes to all of them now....

spastorino:
> I don't quite understand. Even if I replace `>` it will simply copy the entire text and `$` is not a valid bash command...

Yeah, I just meant that would be a bit more obvious that you don't need to copy that and more obvious visually that you're on a shell. IMO the best solution is to have the $ sign but once you try to highlight all the text the dollar sign is not highlighted.

hbina:
But the "copy to clipboard* button will copy the entire text, including the `$`...I guess there are no good solution to this?

shepmaster:
> ```bash" pieces

FWIW, it's generally incorrect to use ` ```bash ` for code blocks that represent user interaction, especially multiple interactions or with output. That should be reserved for actual bash shell script. As a concrete example, `$` denotes a variable usage in bash, `#` is a comment, and `>` is a redirect, yet all of those are used as prompt indicators.

spastorino:
Sorry because I may be having hard time to express my idea :). I feel like in these blocks we should have just the thing that needs to be run, in this case `./x.py build -i --stage 1 src/libstd` but I guess we may want to apply some css style to this and maybe prepend a prompt indicator.

Also what @shepmaster says is true. I'm not sure what would be the best solution /cc @mark-i-m

tshepang:
I myself use prompt characters only when including command output (for easy visual separation), otherwise I do not add it for copy-paste convenience.

mark-i-m:
@tshepang That seems like a reasonable rule to me

spastorino:
So in order to make progress with this my suggestion is ...

1. Let's get rid of all of the prompt symbols in the guide
2. Open an issue to handle script blocks with css/js styling. I think we should add a prompt but as part of presentation and once you select text that part is not selected.

mark-i-m:
@spastorino I don't know how to do #2 in mdbook, but it sounds good to me.

mark-i-m:
@hbina Thanks for the PR! Are you still interested in working on this? If so, could you remove all the `>` thoughout the book and rebase on top of master?

hbina:
Uh oh, I guess I better start finish reading the book. Sure, I can work on them. I only know how to do the first one tho, I am not entirely knowledgeable in CSS styling stuff

spastorino:
@hbina going over the first one is good enough for now :)

mark-i-m:
I think we can leave CSS for a follow-up pr. For now, perhaps we can start by just removing the characters?

hbina:
My current plan is to remove `>` from code blocks that are meant to be executed, I left and sometimes added `$` and `/` where necessary. Plus, those code are not meant to be copied anyway.

src/profiling/with_perf.md
```diff
@@ -260,7 +260,7 @@ usually also want to give `--tree-min-percent` or
 `--tree-max-depth`. The result looks like this:
 
 ```bash
-> perf focus '{do_mir_borrowck}' --tree-callees --tree-min-percent 3
+$ perf focus '{do_mir_borrowck}' --tree-callees --tree-min-percent 3
```
spastorino: Same

src/profiling/with_perf.md
```diff
@@ -232,7 +232,7 @@ Perhaps we'd like to know how much time MIR borrowck spends in the
 trait checker. We can ask this using a more complex regex:
 
 ```bash
-> perf focus '{do_mir_borrowck}..{^rustc::traits}'
+$ perf focus '{do_mir_borrowck}..{^rustc::traits}'
```
spastorino: Same

src/profiling/with_perf.md
```diff
@@ -191,7 +191,7 @@ function of the MIR borrowck is called `do_mir_borrowck`, so we can do
 this command:
 
 ```bash
-> perf focus '{do_mir_borrowck}'
+$ perf focus '{do_mir_borrowck}'
```
spastorino: Same here

src/how-to-build-and-run.md
```diff
@@ -385,7 +385,7 @@ should see a version number ending in `-dev`, indicating a build from
 your local environment:
 
 ```bash
-> rustc +stage1 -vV
+$ rustc +stage1 -vV
```
spastorino: Hmm I see why you added $ there but I'm not sure what to think about this.
cc @mark-i-m

src/profiling/with_perf.md
```diff
@@ -311,7 +311,7 @@ could get our percentages relative to the borrowck itself
 like so:
 
 ```bash
-> perf focus '{do_mir_borrowck}' --tree-callees --relative --tree-max-depth 1 --tree-min-percent 5
+$ perf focus '{do_mir_borrowck}' --tree-callees --relative --tree-max-depth 1 --tree-min-percent 5
```
spastorino: Same

src/profiling/with_perf.md
hbina: I thought the text below this are meant to show an example of what will be displayed? In this case, the "copy to clipboard" button is meaningless regardless. So thats why I thought adding `$` is okay to convey what would appear in a console. My original PR was really about "texts that are meant to be executed, should be executable from what is copied to the clipboard".

mark-i-m:
@spastorino Personally, I think this is ok. One would not copy/paste the whole block anyway in these cases, and I think having the prompt will improve the syntax highlighting of mdbook...

IMHO, we should just merge...

spastorino:
@hbina I've just approved the changes and I was going to merge them but there are some conflicts now, do you mind rebasing?. Sorry for not merging it before and hitting this new conflicts.

mark-i-m:
Umm... looks like rebase went wrong?

hbina:
Oh...that's really bad...
I did `git rebase --onto master patch-1` but I guess that's wrong?

hbina:
Should be okay now?

mark-i-m:
Yeah, I'm not really sure why that happens. Occasionally I do that too.

For whatever reason, `git checkout your-branch; git rebase master` seems to fix it. I think that is equivalent to what you ran but for some reason running it again seems to fix things...

mark-i-m:
Yep, that looks much better :)

qlcom:
This code blocks should not be marked as `bash`, there is nothing `bash` specific in them. They should be marked as `sh`.

shepmaster:
> They should be marked as `sh`.

[I disagree](https://github.com/rust-lang/rustc-dev-guide/pull/460#issuecomment-539146105)

PR #462, opened 2019-10-09
Expand on stability documentation.

ehuss:
@Centril requested some extra documentation on stability (https://github.com/rust-lang/rust/pull/64882#discussion_r332688034), so here it is.  Along the way I added some related things that seemed like it would be good to document. I am not an expert on these things, but they seem mostly correct.

### DIFF
--- a/src/stability.md
+++ b/src/stability.md
@@ -3,13 +3,18 @@
 This section is about the stability attributes and schemes that allow stable APIs to use unstable
 APIs internally in the rustc standard library.
 
-For instructions on stabilizing a language feature see 
+For instructions on stabilizing a language feature see
 [Stabilizing Features](./stabilization_guide.md).
 
-# unstable
+## unstable
 
 The `#[unstable(feature = "foo", issue = "1234", reason = "lorem ipsum")]` attribute explicitly
-marks an item as unstable. This infects all sub-items, where the attribute doesn't have to be
+marks an item as unstable. Items that are marked as "unstable" cannot be used
+without a corresponding `#![feature]` attribute on the crate, even on a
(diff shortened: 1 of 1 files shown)


src/stability.md
```diff
@@ -49,4 +59,67 @@ are nondeterministic and often unknown at compile-time.
...
+future version, then the `deprecated_in_future` lint is triggered which is
+default `allow`, but most of the standard library raises it to a warning with
+`#![warn(deprecated_in_future)]`.
+
+[`deprecated` attribute]: https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute
+
+## -Zforce-unstable-if-unmarked
+
+The `-Zforce-unstable-if-unmarked` flag has a variety of purposes to help
+enforce that the correct crates are marked as unstable, but can still use
```
mark-i-m: I don't understand this sentence. What do you mean by "can still use private crates"

ehuss: Maybe an example would help:

`chalk-engine` gets marked as unstable `rustc_private`.  `chalk-engine` depends on `chalk-macros` which also gets marked as unstable `rustc_private`.  Normally it would not be able to access the `chalk-macros` crate without the `#![feature(rustc_private)]` flag, but that is not something most crates.io crates have. So [this code](https://github.com/rust-lang/rust/blob/58b54911fabcd1b328fab78014b6833a4e8cfaa3/src/librustc/middle/stability.rs#L746-L759) ensures that once something has been "forced" unstable, it can also access other forced unstable crates.

I pushed a change that tries to clarify it, but it still seems a little awkward.

mark-i-m:
Thanks!

PR #473, opened 2019-10-18
Trim

tshepang:


src/compiler-documenting.md
```diff
@@ -35,28 +35,18 @@ Much like individual tests or building certain components you can build only
...
-But, when enabled, compiler documentation does include internal items.
+Compiler documentation is not built by default. To enable it:
 
-Next open up config.toml and make sure these two lines are set to true:
-
-```bash
-docs = true
+```toml
+[build]
 compiler-docs = true
```
JohnTitor: It is useful to know that this flag is in `config.toml`.

tshepang: oh, indeed

src/compiler-documenting.md
```diff
@@ -35,28 +35,18 @@ Much like individual tests or building certain components you can build only
...
 ```
 
-When you want to build the compiler docs as well run this command:
-
-```bash
-./x.py doc
```
mark-i-m: Do we have this command documented somewhere?

tshepang: it is, beginning of the document

mark-i-m:
It looks like there is a merge conflict

PR #476, opened 2019-10-22
Add some docs around the lint store

Mark-Simulacrum:
r? (?) @nikomatsakis 

This accompanies https://github.com/rust-lang/rust/pull/65193. I would appreciate thoughts as to _what_ we should be documenting (and how) -- once we figure things out here I'll also delete https://forge.rust-lang.org/compiler/bug-fix-procedure.html#issuing-future-compatibility-warnings as I believe it's subsumed by the documentation in rustc-guide.

### DIFF
--- a/src/diagnostics.md
+++ b/src/diagnostics.md
@@ -171,11 +171,17 @@ crate.
 
 [builtin]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lint/index.html
 
-Each lint is defined as a `struct` that implements the `LintPass` `trait`. The
-trait implementation allows you to check certain syntactic constructs the
-linter walks the source code. You can then choose to emit lints in a very
-similar way to compile errors. Finally, you register the lint to actually get
-it to be run by the compiler by using the `declare_lint!` macro.
+Every lint is implemented via a `struct` that implements the `LintPass` `trait`
+(you also implement one of the more specific lint pass traits, either
+`EarlyLintPass` or `LateLintPass`).  The trait implementation allows you to
+check certain syntactic constructs the linter walks the source code. You can
+then choose to emit lints in a very similar way to compile errors.
+
+You also declare the metadata of a particular lint via the `declare_lint!`
+macro. This includes the name, the default level, a short description, and some
+more details.
+
+Note that the lint and the lint pass must be registered with the compiler.
 
 For example, the following lint checks for uses
 of `while true { ... }` and suggests using `loop { ... }` instead.
@@ -196,11 +202,15 @@ declare_lint! {
 #[derive(Copy, Clone)]
 pub struct WhileTrue;
 
-impl LintPass for WhileTrue {
-    fn get_lints(&self) -> LintArray {
-        lint_array!(WHILE_TRUE)
-    }
-}
+// This declares a lint pass, providing a list of associated lints.  The
+// compiler currently doesn't use the associated lints directly (e.g., to not
+// run the pass or otherwise check that the pass emits the appropriate set of
+// lints). However, it's good to be accurate here as it's possible that we're
+// going to register the lints via the get_lints method on our lint pass (that
+// this macro generates).
+impl_lint_pass!(
+    WhileTrue => [WHILE_TRUE],
+);
 
 // LateLintPass has lots of methods. We only override the definition of
 // `check_expr` for this lint because that's all we need, but you could
@@ -242,9 +252,24 @@ declare_lint! {
 This makes the `ANONYMOUS_PARAMETERS` lint allow-by-default in the 2015 edition
 but warn-by-default in the 2018 edition.
 
-Lints that represent an incompatibility (i.e. error) in the upcoming edition
-should also be registered as `FutureIncompatibilityLint`s in
-[`register_builtins`][rbuiltins] function in [`rustc_lint::lib`][builtin].
+A future-incompatible lint should be declared with the `@future_incompatible`
+additional "field":
+
+```rust,ignore
+declare_lint! {
+    pub ANONYMOUS_PARAMETERS,
+    Allow,
+    "detects anonymous parameters",
+    @future_incompatible = FutureIncompatibleInfo {
+        reference: "issue #41686 <https://github.com/rust-lang/rust/issues/41686>",
+        edition: Some(Edition::Edition2018),
+    };
+}
+```
+
+If you need a combination of options that's not supported by the
+`declare_lint!` macro, you can always define your own static with a type of
+`&Lint` but this is currently linted against in the compiler tree.
 
 ### Lint Groups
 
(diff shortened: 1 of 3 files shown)


src/diagnostics.md
```diff
@@ -171,11 +171,17 @@ crate.
...
 
-Each lint is defined as a `struct` that implements the `LintPass` `trait`. The
-trait implementation allows you to check certain syntactic constructs the
-linter walks the source code. You can then choose to emit lints in a very
-similar way to compile errors. Finally, you register the lint to actually get
-it to be run by the compiler by using the `declare_lint!` macro.
+Every lint is implemented via a `struct` that implements the `LintPass` `trait`
+(you also implement one of the more specific lint pass traits, either
+`EarlyLintPass` or `LateLintPass`).  The trait implementation allows you to
+check certain syntactic constructs the linter walks the source code. You can
```
nikomatsakis:
```suggestion
check certain syntactic constructs as the linter walks the source code. You can
```

weird grammar, maybe this is what you meant?

src/diagnostics.md
```diff
@@ -171,11 +171,17 @@ crate.
...
+then choose to emit lints in a very similar way to compile errors.
+
+You also declare the metadata of a particular lint via the `declare_lint!`
+macro. This includes the name, the default level, a short description, and some
+more details.
+
+Note that the lint and the lint pass must be registered with the compiler.
```
nikomatsakis: can we maybe link to a method in the rustc API docs here? e.g., [`register_lints`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/lint/struct.LintStore.html#method.register_lints)?

src/diagnostics/lintstore.md
```diff
@@ -0,0 +1,40 @@
+# LintStore
+
+This page documents some of the machinery around lint registration and how we
+run lints in the compiler.
+
+The `LintStore` is the central piece of infrastructure, around which everything
+rotates. It's not available during the early parts of compilation (i.e., before
+TyCtxt) in most of the code, as we need to fill it in with all of the lints,
+which can only happen after plugin registration.
```
nikomatsakis: Can you describe where the `LintStore` is created and how it eventually gets frozen? IIRC, it is created by the plugin registrar (where it is mutable), but then it is "frozen" (put into a `Rc`) and given to the tcx?

src/diagnostics/lintstore.md
```diff
@@ -0,0 +1,40 @@
...
+
+We store essentially two separate things in the `LintStore`: a list of all known
```
nikomatsakis: I think I would move the `## Registration` section up here -- basically say:

> During the mutable period, we can register lints by providing a "constructor function", like so:

This existing paragraph says "We store constructors for the lint passes rather than the passes
themselves" but this feels (to me) like something that makes more sense if you happen to be familiar with the old code than from first principles. i.e., there is nothing all that strange about the fact that we register a function that can produce a lint (also, it's not always a *function* right? it's basically any closure?)

So I would just describe it as is: to register the lint, you give metadata + a constructor. When it is time for us to run the lints, we invoke the closure and get back a `Box<dyn LintPass>` value. We can then invoke the visitor methods on this lint pass. Note that the lint pass methods get `&mut self` so the lint can keep internal, mutable state that it updates as we traverse the input program.

src/diagnostics/lintstore.md
```diff
@@ -0,0 +1,40 @@
...
+constructed per-module if the module changes in an incremental session.
+
+## Registration
+
+We must register both every lint and every lint pass into the `LintStore` for
+the code to work properly. This often is as simple as calling
+`register_lints(PassType::get_lints())` and then `register_late_pass(||
+Box::new(PassType))`.
+
+Within the compiler, for performance reasons, we usually do not register dozens
```
nikomatsakis: This is a good detail to include. I would add a section header before it like

\#\# Compiler lint passes are combined into one pass

Mark-Simulacrum:
Okay, I've pretty much restructured the text completely, let me know what you think. I tried to address the review comments, though some of them are sort of indirectly addressed.

Mark-Simulacrum:
I plan to do some additional work around lint buffering -- but I think having that happen in a separate PR to have at least some docs sooner makes sense?

src/diagnostics/lintstore.md
```diff
@@ -0,0 +1,105 @@
...
+just one -- the compiler doesn't track whether a pass is in any way associated with a particular
+lint, and frequently lints are emitted as part of other work (e.g., type checking, etc.).
+
+## Registration
+
+### High-level overview
+
+The lint store is created and all lints are registered during plugin registration, in
+[`rustc_interface::register_plugins`]. There are three 'sources' of lint: the internal lints, plugin
+lints, and `rustc_interface::Config` `register_lints`. All are registered here, in
```
lzutao: `[`register_lints`]`

src/diagnostics/lintstore.md
```diff
@@ -0,0 +1,105 @@
...
+Ideally, we'd not have to do this, since it certainly adds to the complexity of
+understanding the code. However, with the current type-erased lint store
+approach, it is beneficial to do so for performance reasons.
+
+New lints being added likely want to join one of the existing declarations like
+`late_lint_mod_passes` in `librustc_lint/lib.rs`, which would then
+auto-propagate into the other.
+
+[`LintStore::register_lint`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc/lint/struct.LintStore.html#method.register_lints
+[`rustc_interface::register_plugins`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_interface/passes/fn.register_plugins.html)
```
lzutao: Remove )

src/diagnostics/lintstore.md
```diff
@@ -0,0 +1,105 @@
...
+[`rustc_lint::register_builtins`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lint/fn.register_builtins.html
+[`rustc_lint::register_internals`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lint/fn.register_internals.html
+[`rustc_lint::new_lint_store`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lint/fn.new_lint_store.html
+[`rustc::declare_lint!`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc/macro.declare_lint.html
```
lzutao: ``[`declare_lint!`]:``

src/diagnostics/lintstore.md
```diff
@@ -0,0 +1,105 @@
...
+[`rustc::declare_tool_lint!`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc/macro.declare_tool_lint.html
```
lzutao: ``[`declare_tool_lint!`]:``

Issue #478, opened 2019-10-23
Missing link in ice-breaker/about.md


CreepySkeleton:
I think `[claim the issue]` is supposed to lead somewhere, but it doesn't

amanjeev:
Can you please elaborate? Are you a member you wants to have the issue assigned yourself?

mark-i-m:
@amanjeev I think they mean that the `[claim this issue]` text in https://github.com/rust-lang/rustc-guide/blob/master/src/ice-breaker/about.md is likely supposed to be a link but was not checked.

@Michael-F-Bryan is there any way to make the link checker check these sorts of links?

@spastorino @nikomatsakis What was this supposed to link to?

CreepySkeleton:
@mark-i-m 

> I think they mean that the `[claim this issue]` text in https://github.com/rust-lang/rustc-guide/blob/master/src/ice-breaker/about.md is likely supposed to be a link but was not checked.

Yes, exactly.

> @Michael-F-Bryan is there any way to make the link checker check these sorts of links?

Yes, I'm about to send such a PR.

spastorino:
I guess that should have been https://github.com/rust-lang/triagebot/wiki/Assignment or just saying claim this issue by commenting with `@rustbot claim`. Unsure what exactly the intention if @nikomatsakis was though.

CreepySkeleton:
Hm, it looks like there's checking via `mdbook-linkcheck` but it doesn't check for situations like this.

ehuss:
>  is there any way to make the link checker check these sorts of links?

It is possible, but difficult.  pulldown-cmark has a broken link callback, but it doesn't work very well (it is global, and does not tell you *where* the link is).  I've experimented with this in mdBook, and found a ton of false positives, because `[foo]` is valid markdown, and is intentionally allowed.  It might have to be a pedantic-style check, due to the high false positive risk.

Fixing the callback is tracked in https://github.com/raphlinus/pulldown-cmark/issues/165 and https://github.com/raphlinus/pulldown-cmark/issues/373.  It is also a blocker for updating pulldown-cmark in rustdoc, so it would be nice to get some movement on it.  Someone just needs to propose a good API that could work with pulldown-cmark's architecture.

There's a similar issue of broken references are not validated (https://github.com/raphlinus/pulldown-cmark/issues/217).

CreepySkeleton:
OK, then we might set up a convention not to use any `[link]` links and replace them with `[link][]`. What do you think?

[...]

PR #479, opened 2019-10-23
Check links in src/*.md files

CreepySkeleton:
This should help to avoid problems like #478


CreepySkeleton:
Oops, it looks like this is already covered by `mdbook-linkcheck`

PR #481, opened 2019-10-24
Move readme to guide

mark-i-m:
Currently, this is in the rust tree. I'm moving it here for consistency. It's pretty messy, but better than nothing, and we can clean it up as needed.

This is pretty much a straight copy/paste, as I don't have time to clean it up right now. We can do cleanup in followup PRs.


mark-i-m:
Removed from src/librustc_codegen_ssa/README.md

spastorino:
I like the better than nothing approach a lot, you may have noticed that already. People also start reading it and contributing back so we can keep improving things.

PR #484, opened 2019-10-27
Explain `rustc_on_unimplemented`

Centril:
Extracted from the unstable book since it needs to be removed from there.
Doing this as part of https://github.com/rust-lang/rust/pull/65794.


src/diagnostics.md
```diff
@@ -342,3 +342,150 @@ The JSON emitter defines [its own `Diagnostic`
...
+  = help: the trait `std::iter::Iterator` is not implemented for `Foo`
+  = note: required by `std::iter::IntoIterator::into_iter`
+
+error[E0277]: `&str` is not an iterator
+ --> src/main.rs:5:16
+  |
+5 |     for foo in "" {}
+  |                ^^ `&str` is not an iterator
+  |
+  = note: call `.chars()` or `.bytes() on `&str`
```
jens1o: missing ` after `.bytes()` :)

PR #490, opened 2019-11-03
expand and update Miri description

RalfJung:
r? @oli-obk @eddyb

Cc @iwikal whose questions triggered this PR: feedback would be welcome! As someone who is still new to rustc and Miri, how comprehensible do you think this all is?


iwikal:
That was a good read! It seems reasonably comprehensible. I'm sure it's only as complicated as it needs to be. I'm a bit confused by the MIR example just above your edits, as I've never seen a text representation of MIR before. The stuff about provenance also went above my head.

mark-i-m:
CI is failing

RalfJung:
CI wants me to make the diff unreadable by breaking the lines. Well, I guess I'll do that then.

PR #495, opened 2019-11-05
Fix some broken fragment links.

ehuss:


mark-i-m:
Thanks!

(Btw also https://github.com/rust-lang/rustc-guide/pull/496, so thanks @ehuss and @Michael-F-Bryan! Very helpful :) )

PR #502, opened 2019-11-07
rustc_back has long since been renamed to rustc_target

tshepang:


JohnTitor:
Ci is red because arXiv is temporarily unavailable, the change looks good.

PR #506, opened 2019-11-08
Clarify function name and add links

JOE1994:
rustc_interface::interface::run_compiler 
(https://doc.rust-lang.org/nightly/nightly-rustc/rustc_interface/interface/fn.run_compiler.html)


JOE1994:
Build is currently failing not because of my commit, but because of a web link in another file currently not working.

PR #507, opened 2019-11-08
Define the acronym FCP where it is first used.

mikehcox:
See section "How do I introduce an abbreviation in the text?" in https://blog.apastyle.org/apastyle/abbreviations/#Q2


JohnTitor:
Could you fix the line length error? Otherwise, the change looks good.

PR #510, opened 2019-11-12
Fix: backtraces do work on Mac

ksqsf:
This description

    backtraces **don't work** on Mac and on MinGW

seems outdated. Just checked and it works.

### DIFF
--- a/src/compiler-debugging.md
+++ b/src/compiler-debugging.md
@@ -22,9 +22,9 @@ could be useful for debugging.
 
 When you have an ICE (panic in the compiler), you can set
 `RUST_BACKTRACE=1` to get the stack trace of the `panic!` like in
-normal Rust programs.  IIRC backtraces **don't work** on Mac and on MinGW,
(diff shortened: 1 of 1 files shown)


src/compiler-debugging.md
```diff
@@ -22,7 +22,7 @@ could be useful for debugging.
 
 When you have an ICE (panic in the compiler), you can set
 `RUST_BACKTRACE=1` to get the stack trace of the `panic!` like in
-normal Rust programs.  IIRC backtraces **don't work** on Mac and on MinGW,
+normal Rust programs.  IIRC backtraces **don't work** on MinGW,
 sorry. If you have trouble or the backtraces are full of `unknown`,
 you might want to find some way to use Linux or MSVC on Windows.
```
JohnTitor: Also, could you update this line to include Mac?

JohnTitor:
Ah, we support macOS now, according to https://github.com/rust-lang/rust/issues/24346

PR #523, opened 2019-11-25
Website has disappeared

mark-i-m:
cc @QuietMisdreavus


JohnTitor:
We can remove this link for now and re-visit there later (if they re-post it), I think.

mark-i-m:
Hmm... actually I just realized that the latest PR passed without this, so I am reverting...

PR #530, opened 2019-11-27
Summarize the lecture of ty into a chapter

### LINKED ISSUE #375: Transcribe the Compiler Lecture Representing types in rustc
Video: https://www.youtube.com/watch?v=c01TsOsr3-c

This issue has been assigned to @igaray via [this comment](https://github.com/rust-lang/rustc-guide/issues/375#issuecomment-506992148).

spastorino:
This is coming out as part of the effort from the @rust-lang/wg-learning working group :tada: :tada: :tada:.

As currently is I have two comments that may worth fixing:

1. This probably needs a very deep review in the way things are located and explained. I may have messed some stuff after adding a lot of content and after merging with what the guide already had. Right now I need a little break from this, maybe when I come back I will realize all my mistakes :smile: 
2. There are probably a lot of style mistakes, this was just exported from Paper.

This may be better fixed using this [link](https://hackmd.io/gwqG8yppTjWPd_uteyWLjw), then I can place the fixes into the PR.

Closes #375

### DIFF
--- a/src/ty.md
+++ b/src/ty.md
@@ -1,3 +1,7 @@
+---
+tags: rustc, ty
+---
+
 # The `ty` module: representing types
 
 The `ty` module defines how the Rust compiler represents types internally. It also defines the
@@ -6,18 +10,21 @@ The `ty` module defines how the Rust compiler represents types internally. It al
 ## `ty::Ty`
 
 When we talk about how rustc represents types,  we usually refer to a type called `Ty` . There are
-quite a few modules and types for `Ty` in the compiler ([Ty
-documentation](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/ty/index.html)).
+quite a few modules and types for `Ty` in the compiler ([Ty documentation][ty]).
 
-The specific `Ty` we are referring to is
-[`rustc::ty::Ty`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/ty/type.Ty.html) (and not
-[`rustc::hir::Ty`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/hir/struct.Ty.html)). The
-distinction is important, so we will discuss it first before going into the details of `ty::Ty`.
+[ty]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc/ty/index.html
+
+The specific `Ty` we are referring to is [`rustc::ty::Ty`][ty_ty] (and not
+[`rustc::hir::Ty`][hir_ty]). The distinction is important, so we will discuss it first before going
+into the details of `ty::Ty`.
+
+[ty_ty]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc/ty/type.Ty.html
+[hir_ty]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc/hir/struct.Ty.html
 
 ## `hir::Ty` vs `ty::Ty`
 
 The HIR in rustc can be thought of as the high-level intermediate representation. It is more or less
-the AST (see [this chapter](https://rust-lang.github.io/rustc-guide/hir.html)) as it represents the
+the AST (see [this chapter](hir.md)) as it represents the
 syntax that the user wrote, and is obtained after parsing and some *desugaring*. It has a
 representation of types, but in reality it reflects more of what the user wrote, that is, what they
 wrote so as to represent that type.
@@ -44,29 +51,31 @@ whole program, not a specific usage, unlike `hir::Ty`.
 
 Here is a summary:
 
-| [`hir::Ty`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/hir/struct.Ty.html) | [`ty::Ty`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/ty/type.Ty.html) |
+| [`hir::Ty`][hir_ty] | [`ty::Ty`][ty_ty] |
 | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | Describe the *syntax* of a type: what the user wrote (with some desugaring).  | Describe the *semantics* of a type: the meaning of what the user wrote. |
 | Each `hir::Ty` has its own spans corresponding to the appropriate place in the program. | Doesn’t correspond to a single place in the user’s program. |
-| `hir::Ty` has generics and lifetimes, even those that were elided; however, some of those lifetimes are special markers like "elided" | `ty::Ty` has the full type, including generics and lifetimes, even if the user left them out |
+| `hir::Ty` has generics and lifetimes; however, some of those lifetimes are special markers like [`LifetimeName::Implicit`][implicit]. | `ty::Ty` has the full type, including generics and lifetimes, even if the user left them out |
 | `fn foo(x: u32) → u32 { }` - Two `hir::Ty` representing each usage of `u32`. Each has its own `Span`s, etc.- `hir::Ty` doesn’t tell us that both are the same type | `fn foo(x: u32) → u32 { }` - One `ty::Ty` for all instances of `u32` throughout the program.- `ty::Ty` tells us that both usages of `u32` mean the same type. |
-| `fn foo(x: &u32) -> &u32)`- Two `hir::Ty` again.- Lifetimes for the references show up in the `hir::Ty`s using a special marker, [LifetimeName::Implicit](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/hir/enum.LifetimeName.html#variant.Implicit) | `fn foo(x: &u32) -> &u32)`- A single `ty::Ty`.- The `ty::Ty` has the hidden lifetime param |
+| `fn foo(x: &u32) -> &u32)`- Two `hir::Ty` again.- Lifetimes for the references show up in the `hir::Ty`s using a special marker, [`LifetimeName::Implicit`][implicit]. | `fn foo(x: &u32) -> &u32)`- A single `ty::Ty`.- The `ty::Ty` has the hidden lifetime param |
+
+[implicit]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc/hir/enum.LifetimeName.html#variant.Implicit
 
 **Order** HIR is built directly from the AST, so it happens before any `ty::Ty` is produced. After
 HIR is built, some basic type inference and type checking is done. During the type inference, we
 figure out what the `ty::Ty` of everything is and we also check if the type of something is
 ambiguous. The `ty::Ty` then, is used for type checking while making sure everything has the
-expected type. The [astconv
-module](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_typeck/astconv/index.html),
-is where the code responsible for converting a `hir::Ty` into a `ty::Ty`
-is located. This occurs during the type-checking phase, but also in
-other parts of the compiler that want to ask questions like "what
-argument types does this function expect"?.
+expected type. The [`astconv` module][astconv], is where the code responsible for converting a
+`hir::Ty` into a `ty::Ty` is located. This occurs during the type-checking phase, but also in other
+parts of the compiler that want to ask questions like "what argument types does this function
+expect"?.
 
+[astconv]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_typeck/astconv/index.html
 
-**How semantics drive the two instances of `Ty`** You can think of HIR as the “default” perspective
-of the type information. We assume two things are distinct until they are proven to be the same
-thing. In other words, we know less about them, so we should assume less about them.
+**How semantics drive the two instances of `Ty`** You can think of HIR as the perspective
+of the type information that assumes the least. We assume two things are distinct until they are
+proven to be the same thing. In other words, we know less about them, so we should assume less about
+them.
 
 They are syntactically two strings: `"u32"` at line N column 20 and `"u32"` at line N column 35. We
 don’t know that they are the same yet. So, in the HIR we treat them as if they are different. Later,
@@ -74,22 +83,45 @@ we determine that they semantically are the same type and that’s the `ty::Ty` 
 
 Consider another example: `fn foo<T>(x: T) -> u32` and suppose that someone invokes `foo::<u32>(0)`.
 This means that `T` and `u32` (in this invocation) actually turns out to be the same type, so we
-would end up with the same `ty::Ty` in the end, but we have distinct `hir::Ty`.
+would eventually end up with the same `ty::Ty` in the end, but we have distinct `hir::Ty`. (This is
+a bit over-simplified, though, since during type checking, we would check the function generically
+and would still have a `T` distinct from `u32`. Later, when doing code generation, we would always
+be handling "monomorphized" (fully substituted) versions of each function, and hence we would know
+what `T` represents (and specifically that it is `u32`).
+
+Here is one more example:
+
+```rust
+mod a {
+    type X = u32;
+    pub fn foo(x: X) -> i32 { 22 }
+}
+mod b {
+    type X = i32;
+    pub fn foo(x: X) -> i32 { x }
+}
+```
+
+Here the type `X` will vary depending on context, clearly. If you look at the `hir::Ty`, you will
+get back that `X` is an alias in both cases (though it will be mapped via name resolution to
+distinct aliases). But if you look at the `ty::Ty` signature, it will be either `fn(u32) -> u32` or
+`fn(i32) -> i32` (with type aliases fully expanded).
 
 ## `ty::Ty` implementation
 
-[`rustc::ty::Ty`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/ty/type.Ty.html) is actually
-a type alias to [`&TyS`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/ty/struct.TyS.html)
-(more about that later). `TyS` (Type Structure) is where the main functionality is located. You can
-ignore `TyS` struct in general - you will basically never access it explicitly. We always pass it by
-reference using the `Ty` alias - the only exception is to define inherent methods on types. In
-particular, `TyS` has a [`kind`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/ty/struct.TyS.html#structfield.kind) field of type
-[`TyKind`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/ty/enum.TyKind.html), which
-represents the key type information. `TyKind` is a big enum which represents different kinds of
-types (e.g. primitives, references, abstract data types, generics, lifetimes, etc). `TyS` also has 2
-more fields, `flags` and `outer_exclusive_binder`. They are convenient hacks for efficiency and
-summarize information about the type that we may want to know, but they don’t come into the picture
-as much here.
+[`rustc::ty::Ty`][ty_ty] is actually a type alias to [`&TyS`][tys] (more about that later). `TyS`
+(Type Structure) is where the main functionality is located. You can ignore `TyS` struct in general;
+you will basically never access it explicitly. We always pass it by reference using the `Ty` alias.
+The only exception is to define inherent methods on types. In particular, `TyS` has a [`kind`][kind]
+field of type [`TyKind`][tykind], which represents the key type information. `TyKind` is a big enum
+which represents different kinds of types (e.g. primitives, references, abstract data types,
+generics, lifetimes, etc). `TyS` also has 2 more fields, `flags` and `outer_exclusive_binder`. They
+are convenient hacks for efficiency and summarize information about the type that we may want to
+know, but they don’t come into the picture as much here.
+
+[tys]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc/ty/struct.TyS.html
+[kind]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc/ty/struct.TyS.html#structfield.kind
+[tykind]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc/ty/enum.TyKind.html
 
 Note: `TyKind` is **NOT** the functional programming concept of *Kind*.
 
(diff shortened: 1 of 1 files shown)


mark-i-m:
cc @nikomatsakis who gave the lecture...

amanjeev:
Who reviews this? Niko?

src/ty.md
```diff
@@ -4,55 +4,60 @@ The `ty` module defines how the Rust compiler represents types
...
+
+## `ty::Ty` implementation
+
+`[rustc::ty::Ty](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/ty/type.Ty.html)` is actually a type alias to `&TyS` ([rustdoc](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/ty/struct.TyS.html)) (more about that later). `TyS` (Type Structure) is where the main functionality is located. You can ignore `TyS` struct in general - you will basically never access it explicitly. We always pass it by reference using the `Ty` alias - the only exception I think is to define inherent methods on types. In particular, `TyS` has a `sty` field of type `TyKind` ([rustdoc](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/ty/enum.TyKind.html)), which represents the key type information. `TyKind` is a big enum which represents different kinds of types (e.g. primitives, references, abstract data types, generics, lifetimes, etc). `TyS` also has 2 more fields, `flags` and `outer_exclusive_binder` they are convenient hacks for efficiency and summarize information about the type that we may want to know, but they don’t come into the picture as much here.
```
matthewjasper: The field is now called `kind`.

src/ty.md
```diff
@@ -4,55 +4,60 @@ The `ty` module defines how the Rust compiler represents types
...
+
+    fn foo(x: Ty<'tcx>) {
+      match x.sty {
         ...
+      }
     }
-}
-```
 
 The `sty` field (the origin of this name is unclear to me; perhaps
```
matthewjasper:
```suggestion
The `kind` field (the origin of this name is unclear to me; perhaps
```
and so on...

src/ty.md
```diff
@@ -63,6 +68,67 @@ defining all of the different kinds of types in the compiler.
...
+
+At the beginning of the compilation we make a buffer and each time we need to allocate a type we use some of this memory buffer and if we run out of space we get another one. The lifetime of that buffer is `'tcx`. Our types are tied to that lifetime, so when compilation finishes all the memory related to that buffer is freed and our `'tcx` references would be invalid. This is called arena allocation.
+
+However, there are a lot of types that we don’t want to keep around for the entire compilation. For example, when we are type-checking a specific function,  we will have a bunch of local types (for things in the body of the function) that are irrelevant outside the function. We don’t need them around after we finish type checking the function, and keeping them around would just waste memory. For that reason, we actually have two types of arenas: the *global arena* is for types that are relevant everywhere (e.g. `u32` or a `struct` definition) and the fresh *local arena* whenever we start to type check specific function. Local types of the function are allocated in the local arena,  and when we are done, we can discard the entire arena, freeing up the space.
```
matthewjasper: There's not a local arena any more.

src/ty.md
spastorino: Fixed on the HackMD doc.

src/ty.md
spastorino: @matthewjasper ideas on how can we adapt the text?.

matthewjasper: I think that this paragraph can be removed.

spastorino: Done!

src/ty.md
```diff
@@ -4,65 +4,130 @@ The `ty` module defines how the Rust compiler represents types
...
-queries. The struct `TyCtxt` defines a reference to this shared context:
+When we talk about how rustc represents types,  we usually refer to a type called `Ty` . There are quite a few modules and types for `Ty` in the compiler ([Ty documentation](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/ty/index.html)). 
 
-```rust,ignore
-tcx: TyCtxt<'tcx>
-//          ----
-//          |
-//          arena lifetime
-```
+The specific `Ty` we are referring to is `rustc::ty::Ty` (and not `rustc::hir::Ty`). The distinction is important, so we will discuss it first before going into the details of `ty::Ty`.
```
nikomatsakis: we should link to the API docs when mentioning types. e.g.

* https://doc.rust-lang.org/nightly/nightly-rustc/rustc/ty/type.Ty.html
* https://doc.rust-lang.org/nightly/nightly-rustc/rustc/hir/struct.Ty.html

this can also help us detect bitrot

src/ty.md
```diff
@@ -4,65 +4,130 @@ The `ty` module defines how the Rust compiler represents types
...
-Rust types are represented using the `Ty<'tcx>` defined in the `ty`
-module (not to be confused with the `Ty` struct from [the HIR]). This
-is in fact a simple type alias for a reference with `'tcx` lifetime:
+In contrast, `ty::Ty` represents the semantics of a type, that is, the *meaning* of what the user wrote. For example, `hir::Ty` would record the fact that a user used the name `u32` twice in their program, but the `ty::Ty` would record the fact that both usages refer to the same type.
 
-```rust,ignore
-pub type Ty<'tcx> = &'tcx TyS<'tcx>;
-```
+**Example: `fn foo(x: u32) → u32 { }`**
+In this function we see that `u32` appears twice. We know that that is the same type, i.e. the function takes an argument and returns an argument of the same type, but from the point of view of the HIR there would be two distinct type instances because these are occurring in two different places in the program. That is, they have two different *spans* (locations).
```
nikomatsakis: span could link too?

src/ty.md
spastorino: Fixed.

src/ty.md
spastorino: Fixed.

spastorino:
This is now ready for review. CI failures are unrelated. cc @nikomatsakis

Mark-Simulacrum:
I will be fixing the CI failure due to changes of how early lints are defined today, which should hopefully fix CI on this PR.

mark-i-m:
@Mark-Simulacrum : @JohnTitor made a PR to fix the links, but we would still really appreciate a PR to update the section on early lints, if you have time :)

mark-i-m:
Ping @nikomatsakis :)

We would love to get this merged

src/ty.md
```diff
@@ -17,110 +179,378 @@ tcx: TyCtxt<'tcx>
...
+let array_ty = tcx.mk_array(elem_ty, len * 2);
+```
+
+These methods all return a `Ty<'tcx>` – note that the lifetime you get back is the lifetime of the
+innermost arena that this `tcx` has access to. In fact, types are always canonicalized and interned
+(so we never allocate exactly the same type twice) and are always allocated in the outermost arena
+where they can be (so, if they do not contain any inference variables or other "temporary" types,
+they will be allocated in the global arena). However, the lifetime `'tcx` is always a safe
+approximation, so that is what you get back.
+
```
gilescope: Would be interested to know examples of when an arena would start and stop - is global=crate level?

src/ty.md
```diff
@@ -17,110 +179,378 @@ tcx: TyCtxt<'tcx>
...
+## Beyond types: other kinds of arena-allocated data structures
+
+In addition to types, there are a number of other arena-allocated data structures that you can
+allocate, and which are found in this module. Here are a few examples:
+
+- [`Substs`][subst], allocated with `mk_substs` – this will intern a slice of types, often used to
+  specify the values to be substituted for generics (e.g. `HashMap<i32, u32>` would be represented
+  as a slice `&'tcx [tcx.types.i32, tcx.types.u32]`).
+- `TraitRef`, typically passed by value – a **trait reference** consists of a reference to a trait
+  along with its various type parameters (including `Self`), like `i32: Display` (here, the def-id
```
gilescope: First usage of def id. Good to point out it’s defined and discussed in depth later

src/ty.md
```diff
@@ -17,110 +179,378 @@ tcx: TyCtxt<'tcx>
...
+MyStruct<A> ==> MyStruct<B>
+```
+
+in an example like this, we can subst from `MyStruct<A>` to `MyStruct<B>` (and so on) very cheaply,
+by just replacing the one reference to `A` with `B`. But if we eagerly substituted all the fields,
+that could be a lot more work.
+
+A bit more deeply, this corresponds to structs in Rust being *nominal* types — which means that they
+are defined by their *name* (and that their contents are then indexed from the definition of that
+name, and not carried along “within” the type itself).
```
gilescope: Would it be possible having set out the above to explain how to query if a ty:Ty implements a specific Trait? 
(Thank you so much for this chapter - it is like the rosetta stone)

src/ty.md
mark-i-m: I believe global means the whole compilation (ie the whole crate graph)

src/ty.md
mark-i-m: @gilescope Perhaps take a look at the trait-solving chapters (already exist). Does that answer your question?

mark-i-m:
Ping @spastorino Could you address the review comments above? Also, is the hackmd still the canonical version of the chapter or this PR?

Ping @nikomatsakis `feedback.await` :)

spastorino:
> Ping @spastorino Could you address the review comments above? Also, is the hackmd still the canonical version of the chapter or this PR?

The Hackmd document is the canonical version of the chapter yeah. When some review is made I sync both things.
About the reviews above I'm going to check them in a bit.

src/ty.md
```diff
@@ -17,110 +179,378 @@ tcx: TyCtxt<'tcx>
...
+`List`).
+
+So why do we use this `List` type instead of making it really a slice? The reason is that we may
+want to compare subslices of the list of generics for equality. For example, imagine we had a `List`
+`a = [u32, f32]` and  a `List` `b = [f32]`. Now suppose that we have a subslice of `a` : `a2 =
+[f32]`. Recall that lists are interned, so if we want to compare `a` with `b`, we can just compare
+their pointers and see whether they are equivalent or not. But we may *also* want to check the
+equality of `b` and `a2` (i.e. compare subslices of lists). Since the subslice `a2`  would also be
+interned, it would actually end up being the same as `b`, so we would be able to efficiently compare
+them by just comparing the pointers (just like with the full lists). You wouldn’t be able to do that
```
nikomatsakis: I don't understand this paragraph actually. I *think* that it may be a bit misleading, in fact. I believe a `List` is distinct from a slice in two ways:

* it has the `len` "inline", so `&List` is only 32 bits
* as a consequence, it cannot be "subsliced" (that only works if the len is out of line)

This also implies that yes you can check two Lists for equality via `==` (which would be not be possible for ordinary slices). This is precisely because they never represent a "sub-list", only the complete list, which has thus been hashed and interned.

src/ty.md
```diff
@@ -17,110 +179,378 @@ tcx: TyCtxt<'tcx>
...
+with normal slices and would have to iterate over the contents to check for equality.
+
+So pulling it all together, let’s go back to our example above:
```
nikomatsakis: as above, I think the example is really a type like `MyStruct<u32>`, not the definition itself

src/ty.md
```diff
@@ -17,110 +179,378 @@ tcx: TyCtxt<'tcx>
...
+query](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/ty/struct.TyCtxt.html#method.adt_def).
+The `AdtDef`s are all interned (as you can see `'tcx` lifetime on it).
+
+We also have an internal map to go from `def_id` to what’s called "Def path". "Def path" is like a
+module path but a bit more rich. For our example, it may be `crate::foo::MyStruct` that identifies
+this definition uniquely. It’s a bit different than a module path because it might include the type
+parameter `T`, which you can't write in normal rust, like `crate::foo::MyStruct::T`. These are used
+in incremental compilation.
+
+### Generics and substitutions
```
nikomatsakis: I wonder if it makes sense to break up this single file into multiple subsections. I tend to think yes (can be a follow-up PR, obviously)

src/ty.md
```diff
@@ -17,110 +179,378 @@ tcx: TyCtxt<'tcx>
...
+So above, recall that in our example the `MyStruct` struct had a generic type `T`. When we are (for
+example) type checking functions that use `MyStruct`, we will need to be able to refer to this type
+`T` without actually knowing what it is. In general, this is true inside all generic definitions: we
+need to be able to work with unknown types. This is done via `TyKind::Param` (which we mentioned in
+the example above).
+
+Each `TyKind::Param` contains two things: the index of the type parameter and the name. The more
+important one is the index, since the name can change across scopes (e.g. we can define struct
+`MyStruct<A>` by then implement `impl<B> MyStruct<B>` ; the names are different, but they mean the
+same type parameter).
```
nikomatsakis: This is actually rather incorrect -- they are not the same type parameter, but they happen to play the same role in this instance. Consider `impl<B> MyStruct<Vec<B>>` -- here, the index of `B` is zero, but it is not "the same" as `A`. In fact, `A` (in the definition) will be replaced with `Vec<B>`.

More generally, we should never mix code with type parameters from distinct scopes -- so you should never simultaneously have a a type referencing `B` (defined on the impl) with a type referencing `A` (defined on the struct). Instead, you would always substitute `A` for some other type.

I would say it more like this:

* It contains the name and the index. In general, the index fully defines the parameter and is used by most of the code. The name is included for debug print-outs. There are two reasons for this. First, the index is convenient, it allows you to include into the list of generic arguments when substituting. Second, the index is more robust. For example, you could in principle have two distinct type parameters that use the same name, e.g. `impl<A> Foo<A> { fn bar<A>() { .. } }`, although the rules against shadowing make this difficult (but those language rules could change in the future).

src/ty.md
```diff
@@ -17,110 +179,378 @@ tcx: TyCtxt<'tcx>
...
+
+You may have a couple of followup questions…
+
+ **`type_of`** How do we get the “generic type of `x`"? You can get the type of pretty much anything
+ with the   `tcx.type_of(def_id)` query. In this case, we would pass the `DefId` of the field `x`.
+ The `type_of` query always returns the definition with the generics that are in scope of the
+ definition. For example, `tcx.type_of(def_id_of_my_struct)` would return the “self-view” of
+ `MyStruct`: `Adt(Foo, &[Param(0), Param(1)])`.
+
+**`subst`** How do we actually do the substitutions? There is a query for that too! You use
```
nikomatsakis: This is not, strictly speaking, a "query" -- it's just a function. Queries are memoized and integrated with incremental compilation.

src/ty.md
```diff
@@ -17,110 +179,378 @@ tcx: TyCtxt<'tcx>
...
 
-> N.B. inspecting the `sty` field on types during type inference can be
-> risky, as there may be inference variables and other things to
-> consider, or sometimes types are not yet known and will become
-> known later.
+[Here is an example of actually using `subst` in the
+compiler](https://github.com/rust-lang/rust/blob/597f432489f12a3f33419daa039ccef11a12c4fd/src/librustc_typeck/astconv.rs#L942-L953).
+The exact details are not too important, but in this piece of code, we happen to be converting from
+the `hir::Ty` to a real `ty::Ty`. You can see that we first get some substitutions (`substs`) and
+then we call `type_of` to get a type and call `subst(substs)` to get a new type with the
```
nikomatsakis: I would say: ...and call `ty.subst(substs)` to get a new version of `ty` with the substitutions made.

src/ty.md
```diff
@@ -17,110 +179,378 @@ tcx: TyCtxt<'tcx>
...
+substitutions made.
 
-To allocate a new type, you can use the various `mk_` methods defined
-on the `tcx`. These have names that correspond mostly to the various kinds
-of type variants. For example:
+### `TypeFoldable` and `TypeFolder`
```
nikomatsakis: this would definitely be good to move to a "subsection"

src/ty.md
```diff
@@ -17,110 +179,378 @@ tcx: TyCtxt<'tcx>
...
+and all that does is index into the list of substitutions with the index of the `Param`.
+
+## Errors
+
+It is possible for the indices in `Param` to not match with what we expect. For example, the index
+could be out of bounds or it could be the index of a lifetime when we were expecting a type. These
+sorts of errors would be caught earlier in the compiler when translating from a `hir::Ty` to a
+`ty::Ty`. If they occur later, that is a compiler bug.
+
+There is a `TyKind::Error` that is produced when the user makes this sort of error. The idea is that
```
nikomatsakis: This is sort of distinct from the previous paragraph. That is, the `Error` type is used as the "result type" for expressions or other things that turn out to be internally inconsistent -- i.e., where the user had a bug. In contrast, the indices for a `Param` should *always* be in bounds for a substitution or else our compiler code is wrong.

One thing worth mentioning about `TyKind::Error` -- there is an important invariant. You should **never** return the 'error type' unless you **know** that an error has already been reported to the user. This is usually because (a) you just reported it right there or (b) you are propagating an existing `Error` type (in which case the error should've been reported when that error type was produced).

It's important  to maintain this invariant because the whole point of the `Error` type is to suppress other errors -- i.e., we don't report them. If we were to produce an `Error` type without actually reporting an error, then this could cause later errors to be suppressed, and the compilation might inadvertently succeed!

Sometimes there is a third case. You believe that an error has been reported, but you believe it would've been reported earlier in the compilation, not locally. In that case, you can invoke `tcx.delay_span_bug`. This will create a file away that you expect compilation to yield an error -- if however compilation should succeed, then it will trigger a bug report.

src/ty.md
```diff
@@ -17,110 +179,378 @@ tcx: TyCtxt<'tcx>
...
+we would propagate this type and suppress other errors that come up due to it so as not to overwhelm
+the user with cascading compiler error messages.
+
+## Question: Why not substitute “inside” the adt-def?
+
+The adt-def logically represents just the *name* of the struct (e.g., `Vec`). Its contents are
+always assumed to be expressed in terms of the “internal” or “self-view” — i.e., expressed in terms
+of the generics on the struct. If nothing else, this has an efficiency win:
```
nikomatsakis: Even though this is reflecting what I said, I found it kind of confusing. =) Basically, we're arguing for why we represent `MyStruct<u32>` as the pair of an "unsubstituted" definition (`MyStruct`, the `AdtDef`) combined with a list of substitutions (the `[u32]`). This requires us then to apply those substitutions whenever we extract a type from the struct definition, such as the type of a field.

The alternative would be to do the substitution "eagerly" -- i.e., create a new, fully substituted form of the `AdtDef`, where the types of all fields are already substituted. I think it's worth spelling out this alternative in the text to help explain what you are arguing *against*.

src/ty.md
```diff
@@ -1,14 +1,176 @@
...
+Consider another example: `fn foo<T>(x: T) -> u32` and suppose that someone invokes `foo::<u32>(0)`.
+This means that `T` and `u32` (in this invocation) actually turns out to be the same type, so we
+would end up with the same `ty::Ty` in the end, but we have distinct `hir::Ty`.
+
+## `ty::Ty` implementation
+
+[`rustc::ty::Ty`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/ty/type.Ty.html) is actually
+a type alias to [`&TyS`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/ty/struct.TyS.html)
+(more about that later). `TyS` (Type Structure) is where the main functionality is located. You can
+ignore `TyS` struct in general - you will basically never access it explicitly. We always pass it by
```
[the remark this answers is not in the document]
mark-i-m: We do link to the `TyS` struct in the previous sentence.

src/ty.md
spastorino: @gilescope thanks for noting that. Can you open an issue that ask for that information?. I think would be better if we get this PR in and we continue with follow ups.

src/ty.md
spastorino: Done

src/ty.md
```diff
@@ -1,14 +1,176 @@
...
+reference using the `Ty` alias - the only exception is to define inherent methods on types. In
+particular, `TyS` has a `kind` field of type
+[`TyKind`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/ty/enum.TyKind.html), which
+represents the key type information. `TyKind` is a big enum which represents different kinds of
+types (e.g. primitives, references, abstract data types, generics, lifetimes, etc). `TyS` also has 2
+more fields, `flags` and `outer_exclusive_binder`. They are convenient hacks for efficiency and
+summarize information about the type that we may want to know, but they don’t come into the picture
+as much here.
+
+Note: `TyKind` is **NOT** the functional programming concept of *Kind*.
```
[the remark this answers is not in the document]
mark-i-m: We link to `TyKind` docs when we first mention it.

src/ty.md
```diff
@@ -1,14 +1,176 @@
...
+the same, all we need to do is compare the pointers which is efficient. `TyS` which represents types
+is carefully setup so you never construct them on the stack. You always allocate them from this
+arena and you always intern them so they are unique.
+
+At the beginning of the compilation we make a buffer and each time we need to allocate a type we use
+some of this memory buffer. If we run out of space we get another one. The lifetime of that buffer
+is `'tcx`. Our types are tied to that lifetime, so when compilation finishes all the memory related
+to that buffer is freed and our `'tcx` references would be invalid.
+
+Finally, interned types are stored in canonical form. You can read more about [canonicalization in
```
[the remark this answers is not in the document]
mark-i-m: I deleted the sentence.

src/ty.md
```diff
@@ -17,110 +179,378 @@ tcx: TyCtxt<'tcx>
...
+that correspond mostly to the various kinds of type variants. For example:
+
+```rust,ignore
+let array_ty = tcx.mk_array(elem_ty, len * 2);
+```
+
+These methods all return a `Ty<'tcx>` – note that the lifetime you get back is the lifetime of the
+innermost arena that this `tcx` has access to. In fact, types are always canonicalized and interned
+(so we never allocate exactly the same type twice) and are always allocated in the outermost arena
+where they can be (so, if they do not contain any inference variables or other "temporary" types,
```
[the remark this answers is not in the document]
mark-i-m: I updated the paragraph

src/ty.md
```diff
@@ -17,110 +179,378 @@ tcx: TyCtxt<'tcx>
...
+
+> NB. Because types are interned, it is possible to compare them for equality efficiently using `==`
+> – however, this is almost never what you want to do unless you happen to be hashing and looking
+> for duplicates. This is because often in Rust there are multiple ways to represent the same type,
+> particularly once inference is involved. If you are going to be testing for type equality, you
+> probably need to start looking into the inference code to do it right.
 
-### Allocating and working with types
+You can also find various common types in the `tcx` itself by accessing `tcx.types.bool`,
+`tcx.types.char`, etc (see `CommonTypes` for more).
```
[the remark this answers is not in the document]
mark-i-m: I can't seem to find the rustdocs for it?

src/ty.md
```diff
@@ -17,110 +179,378 @@ tcx: TyCtxt<'tcx>
...
+Notice that the code above does not generate a new `DefId` for `u32` because it is not defined in
+that code (it is only referenced).
+
+`AdtDef` is more or less a wrapper around `DefId` with lots of useful helper methods. There is
+essentially a one-to-one relationship between `AdtDef` and `DefId`. You can get the `AdtDef` for a
+`DefId` with the [`tcx.adt_def(def_id)`
+query](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/ty/struct.TyCtxt.html#method.adt_def).
+The `AdtDef`s are all interned (as you can see `'tcx` lifetime on it).
+
+We also have an internal map to go from `def_id` to what’s called "Def path". "Def path" is like a
```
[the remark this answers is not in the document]
mark-i-m: opened https://github.com/rust-lang/rustc-guide/pull/548

mark-i-m:
@spastorino I addressed some of the low-hanging fruit in the hackmd. Also, it looks like the followup PR should create the following (sub)chapters:

- Allocation, Interning
- `TypeFolder`, `TypeFoldable`
- Generics and Substitutions

src/ty.md
```diff
@@ -1,14 +1,176 @@
...
+whole program, not a specific usage, unlike `hir::Ty`.
+
+Here is a summary:
+
+| [`hir::Ty`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/hir/struct.Ty.html) | [`ty::Ty`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/ty/type.Ty.html) |
+| ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Describe the *syntax* of a type: what the user wrote (with some desugaring).  | Describe the *semantics* of a type: the meaning of what the user wrote. |
+| Each `hir::Ty` has its own spans corresponding to the appropriate place in the program. | Doesn’t correspond to a single place in the user’s program. |
+| `hir::Ty` only has generics and lifetimes the user wrote (since elided params don’t show up in the syntax) | `ty::Ty` has the full type, including generics and lifetimes, even if the user left them out |
```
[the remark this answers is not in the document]
spastorino: Edited. @nikomatsakis can you check it out again? and also, how would you write the example below? the one that talks about lifetimes.

src/ty.md
```diff
@@ -1,14 +1,176 @@
...
+| `fn foo(x: u32) → u32 { }` - Two `hir::Ty` representing each usage of `u32`. Each has its own `Span`s, etc.- `hir::Ty` doesn’t tell us that both are the same type | `fn foo(x: u32) → u32 { }` - One `ty::Ty` for all instances of `u32` throughout the program.- `ty::Ty` tells us that both usages of `u32` mean the same type. |
+| `fn foo(x: &u32) -> &u32)`- Two `hir::Ty` again.- No lifetimes in the `hir::Ty`s, because the user elided them | `fn foo(x: &u32) -> &u32)`- A single `ty::Ty` again.- The `ty::Ty` has the hidden lifetime param |
```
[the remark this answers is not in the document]
spastorino: :+1:, fixed.

src/ty.md
```diff
@@ -1,14 +1,176 @@
...
+
+**Order** HIR is built directly from the AST, so it happens before any `ty::Ty` is produced. After
+HIR is built, some basic type inference and type checking is done. During the type inference, we
+figure out what the `ty::Ty` of everything is and we also check if the type of something is
+ambiguous. The `ty::Ty` then, is used for type checking while making sure everything has the
+expected type.
```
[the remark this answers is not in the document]
spastorino: Addressed.

src/ty.md
spastorino: Addressed.

src/ty.md
spastorino: :+1:

src/ty.md
spastorino: Addressed.

src/ty.md
spastorino: :+1:

src/ty.md
spastorino: Done.

spastorino:
Just pushed with @mark-i-m fixes and my fixes.
Most of @nikomatsakis comments are addressed but:

- https://github.com/rust-lang/rustc-guide/pull/530#discussion_r362002707
- https://github.com/rust-lang/rustc-guide/pull/530#discussion_r362030160
- https://github.com/rust-lang/rustc-guide/pull/530#discussion_r362030906

src/ty.md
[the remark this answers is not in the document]
mark-i-m: Nevermind. Found it.

mark-i-m:
@rust-lang/wg-learning I've resolved the remaining comments and rebased over master. The hackmd and this PR should be consistent.

If it looks good, then let's go ahead and merge, and subsequent comments can go in a followup.

Both @spastorino and I are going to be unavailable next week, btw...

PR #538, opened 2019-12-16
Old information where JSON error emitter can be found

### LINKED ISSUE #537: Wrong information where "json" error emitter can be found
I am reading the rustc documentation right now, and I stumbled upon a misinformation as to where the error emitter for the json format lies in the rustc.
It looks like it was moved at some point to the same crate as the regular emitter `librustc_errors`.

I've creating a pull request which does change the paragraph.

TheSamsa:
Right now the documentation states it lives under "libsyntax" but this is not true anymore, it can be found with the human readable emitter under "librustc_errors"

Closes #537

### DIFF
--- a/src/diagnostics.md
+++ b/src/diagnostics.md
@@ -357,14 +357,12 @@ string; this was introduced so that UI tests could both make use of
 the structured JSON and see the "human" output (well, _sans_ colors)
 without having to compile everything twice.
 
-The JSON emitter currently lives in libsyntax/json.rs. (But arguably
-it should live in librustc_errors along with the "human" emitter? It's
-not obvious to the present author why it wasn't moved from libsyntax
-to librustc_errors at the same [time the "human" emitter was
-moved](https://github.com/rust-lang/rust/commit/6ae3502134).)
+The "human" readable and the json format emitter can be found under
+librustc_errors, both were moved from the libsyntax crate to the
+[librustc_errors crate](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_errors/index.html).
 
 The JSON emitter defines [its own `Diagnostic`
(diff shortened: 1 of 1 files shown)


src/diagnostics.md
```diff
@@ -357,14 +357,11 @@ string; this was introduced so that UI tests could both make use of
 the structured JSON and see the "human" output (well, _sans_ colors)
 without having to compile everything twice.
 
-The JSON emitter currently lives in libsyntax/json.rs. (But arguably
-it should live in librustc_errors along with the "human" emitter? It's
-not obvious to the present author why it wasn't moved from libsyntax
-to librustc_errors at the same [time the "human" emitter was
-moved](https://github.com/rust-lang/rust/commit/6ae3502134).)
+The "human" readable and the json format emitter can be found under
+librustc_errors, both where moved from the libsyntax crate.
```
mark-i-m: where -> were

Also would be nice if you could make it a link to the rustdocs: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_errors/index.html

src/diagnostics.md
```diff
@@ -357,14 +357,11 @@ string; this was introduced so that UI tests could both make use of
...
 
 The JSON emitter defines [its own `Diagnostic`
-struct](https://github.com/rust-lang/rust/blob/b2c6b8c29f13f8d1f242da89e587960b95337819/src/libsyntax/json.rs#L85-L99)
+struct](https://github.com/rust-lang/rust/blob/master/src/librustc_errors/json.rs#L133-L145)
```
mark-i-m: Could you instead link to the rustdocs here: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_errors/json/struct.Diagnostic.html

Sorry, I know you were just doing what was previously done.

TheSamsa: No problem, will do.

src/diagnostics.md
TheSamsa: Thanks, I'll change that

PR #542, opened 2019-12-23
Introduce `rustc_const_stable` and explain `rustc_const_unstable`

oli-obk:
r? @Centril

### DIFF
--- a/src/stability.md
+++ b/src/stability.md
@@ -49,6 +49,28 @@ marks an item as stabilized. To do this, follow the instructions in
 
 Note that stable functions may use unstable things in their body.
 
+## rustc_const_unstable
+
+The `#[rustc_const_unstable(feature = "foo", issue = "1234", reason = "lorem ipsum")]`
+has the same interface as the `unstable` attribute. It is used to mark
+`const fn` as having their constness be unstable. This allows you to make a
+function stable without stabilizing its constness or even just marking an existing
+stable function as `const fn` without instantly stabilizing the `const fn`ness.
+
+Furthermore this attribute is needed to mark an intrinsic as `const fn`, because
+there's no way to add `const` to functions in `extern` blocks for now.
+
+## rustc_const_stable
+
+The `#[stable(feature = "foo", "since = "1.420.69")]` attribute explicitly marks
+a `const fn` as having its constness be `stable`. This attribute can make sense
+even on an `unstable` function, if that function is called from another
+`rustc_const_stable` function.
+
+Furthermore this attribute is needed to mark an intrinsic as callable from
+`rustc_const_stable` functions.
+
+
 ## allow_internal_unstable
 
 Macros, compiler desugarings and `const fn`s expose their bodies to the call


src/stability.md
```diff
@@ -49,6 +49,28 @@ marks an item as stabilized. To do this, follow the instructions in
...
+## rustc_const_unstable
+
+The `#[rustc_const_unstable(feature = "foo", issue = "1234", reason = "lorem ipsum")]`
+has the same interface as the `unstable` attribute. It is used to mark
+`const fn` as having their constness be unstable. This allows you to make a
+function stable without stabilizing its constness or even just marking an existing
+stable function as `const fn` without instantly stabilizing the `const fn`ness.
+
+Furthermore this attribute is needed to mark an intrinsic as `const fn`, because
+there's no way to add `const` to functions in `extern` blocks for now.
```
Centril: argh... this is a syntactic problem? (I should fix this as part of my item-parsing unification work...)

oli-obk: well... it's more of a `const` is not a part of https://doc.rust-lang.org/nightly/nightly-rustc/syntax/ast/struct.FnDecl.html

But since we have *safe* intrinsics nowadays, and `const` intrinsics, maybe we really should move to using a `FnHeader`. Though we don't want this as a general thing for `extern` items. The `rust-intrinsic` ABI really should go away and we should have a better scheme for declaring intrinsics. I think @eddyb had some thoughts on this, not sure if those were ever written down, but maybe we have an issue about it?

Centril: > well... it's more of a `const` is not a part of https://doc.rust-lang.org/nightly/nightly-rustc/syntax/ast/struct.FnDecl.html

It should be an [`FnSig`](https://doc.rust-lang.org/nightly/nightly-rustc/syntax/ast/struct.FnSig.html) syntactically in [`ForeignItemKind`](https://doc.rust-lang.org/nightly/nightly-rustc/syntax/ast/enum.ForeignItemKind.html#variant.Fn). There should just be one function item grammar (aside from the snafu of requiring parameter names sometimes and sometimes not but that's a minor issue controlled by a flag basically).

Though I think you're about the semantic issue... we wanted `#[rustc_intrinsic]` iirc and there's an issue somewhere about that.

Centril: Found it, https://github.com/rust-lang/rust/issues/63585.

oli-obk: > It should be an FnSig syntactically in ForeignItemKind. 

That won't work out of the box because the default for extern functions is `unsafe`, so you'd still need parser hacks.

Centril: You don't need parser hacks. If we want to prevent anything, e.g. `async fn` in `extern`, then we can do it in `ast_validation`.

oli-obk: It's not about denying, it's that extern functions are unsafe by default, so the normal parser will mark them safe

Centril: That's something easily fixed in lowering which is the right place to draw such semantic distinctions.

PR #556, opened 2020-01-09
add cleanup crew

nikomatsakis:
First draft of describing the "cleanup crew" ICE-breakers team.

### DIFF
--- a/src/ice-breaker/cleanup-crew.md
+++ b/src/ice-breaker/cleanup-crew.md
@@ -80,10 +80,11 @@ PR; for older regressions, it will simply identify a nightly.
 To learn to use [cargo-bisect-rustc], check out [this blog
 post][learn], which gives a quick introduction to how it works. You
 can also ask questions at the Zulip stream
-`#t-compiler/cargo-bisect-rustc`, or help in improving the tool.
+[`#t-compiler/cargo-bisect-rustc`][zcbr], or help in improving the tool.
 
 [cargo-bisect-rustc]: https://github.com/rust-lang/cargo-bisect-rustc/
 [learn]: https://blog.rust-lang.org/inside-rust/2019/12/18/bisecting-rust-compiler.html
+[zcbr]: https://rust-lang.zulipchat.com/#narrow/stream/217417-t-compiler.2Fcargo-bisect-rustc
 
 ### identifying the range of PRs in a nightly
 
--- a/src/ice-breaker/cleanup-crew.md
+++ b/src/ice-breaker/cleanup-crew.md
@@ -87,13 +87,16 @@ can also ask questions at the Zulip stream
 
 ### identifying the range of PRs in a nightly
 
-If you've managed to narrow things down to a particular nightly build,
-it is then helpful if we can identify the set of PRs that this
-corresponds to. One helpful command in that regard is `rustc +nightly
--vV`, which will cause it to output a number of useful bits of version
-info, including the `commit-hash`. Given the commit-hash of two
-nightly versions, you can find all of PRs that have landed in between
-by taking the following steps:
+If the regression occurred more than 90 days ago, then
+cargo-bisect-rustc will not able to identify the particular PR that
+caused the regression, just the nightly build. In that case, we can
+identify the set of PRs that this corresponds to by using the git
+history. 
+
+The command `rustc +nightly -vV` will cause rustc to output a number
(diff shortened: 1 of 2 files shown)


src/ice-breaker/cleanup-crew.md
```diff
@@ -0,0 +1,66 @@
+# Cleanup Crew
+
+**Github Label:** [ICEBreaker-Cleanup-Crew]
+
+[ICEBreaker-LLVM]: https://github.com/rust-lang/rust/labels/ICEBreaker-Cleanup-Crew
```
lqd:
```suggestion
[ICEBreaker-Cleanup-Crew]: https://github.com/rust-lang/rust/labels/ICEBreaker-Cleanup-Crew
```

mark-i-m: Ping @nikomatsakis ^^^

nikomatsakis:
OK, I pushed some basic advice about cargo-bisect-rustc. I'd like to push some advice about finding the set of PRs in a nightly, but I'm actually not sure the best way to do that.

pnkfelix:
> OK, I pushed some basic advice about cargo-bisect-rustc. I'd like to push some advice about finding the set of PRs in a nightly, but I'm actually not sure the best way to do that.

The biggest problem here is that if you want to do something like `git log --author=bors --format=oneline COMMIT1..COMMIT2`, you need a local checkout, right? Because I would agree with that, and is a further argument for investing effort in resolving https://github.com/rust-lang/cargo-bisect-rustc/issues/40

(I'm not sure if you find the above `git log` invocation awkward; it seems pretty stream-lined to me.)

nikomatsakis:
Yes, the biggest problem is probably needing a local checkout -- but I also think we could get more streadmlined. My ideal would be something that extracts and prints markdown with a list of PR numbers

```
* "blah blah blah", PR #123 by @so-and-so
```

src/ice-breaker/cleanup-crew.md
```diff
@@ -0,0 +1,110 @@
...
+To help in figuring out the cause of a regression we have a tool
+called [cargo-bisect-rustc]. It will automatically download and test
+various builds of rustc. For recent regressions, it is even able to
+use the builds from our CI to track down the regression to a specific
+PR; for older regressions, it will simply identify a nightly.
+
+To learn to use [cargo-bisect-rustc], check out [this blog
+post][learn], which gives a quick introduction to how it works. You
+can also ask questions at the Zulip stream
+`#t-compiler/cargo-bisect-rustc`, or help in improving the tool.
```
spastorino: Started reading this and I thought, I'm going to comment that we should mention the stream :smile:. This is brilliant.

src/ice-breaker/cleanup-crew.md
```diff
@@ -0,0 +1,110 @@
...
+2. Execute the command `git log --author=bors --format=oneline SHA1..SHA2`
+  * This will list out all of the commits by bors, which is our merge bot
+  * Each commit corresponds to one PR, and information about the PR should be in the description
+3. Copy and paste that information into the bug report
+
+Often, just eye-balling the PR descriptions (which are included in the
+commit messages) will give you a good idea which one likely caused the
+problem. But if you're unsure feel free to just ping the compiler team
+(`@rust-lang/compiler`) or else to ping the authors of the PR
+themselves.
```
spastorino: I'm not sure about this section, or at least is not clear to me why you're not relying on `cargo-bisect-rustc` for this case too. Is this for the case where the artifacts related to those PRs are not around anymore?.

spastorino:
@nikomatsakis this is great, left a comment about the last section.

src/ice-breaker/cleanup-crew.md
nikomatsakis: Yes, just for that case... I'll make that clearer.

nikomatsakis: done

src/ice-breaker/cleanup-crew.md
lqd: this here could even link to the mentioned [Zulip stream](https://rust-lang.zulipchat.com/#narrow/stream/217417-t-compiler.2Fcargo-bisect-rustc)

spastorino:
Looks good to me now, maybe as @lqd mentioned in one case, we should link all the streams that show up there.

PR #560, opened 2020-01-14
Update .gitattributes

jethrogb:
See https://github.com/rust-lang/rust/pull/57858

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,7 @@
 # Set the default behavior, in case people don't have core.autocrlf set.
 * text=auto eol=lf
+
+# Older git versions try to fix line endings on images, this prevents it.
+*.png binary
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -1,2 +1,7 @@
 # Set the default behavior, in case people don't have core.autocrlf set.
 * text=auto eol=lf
+
+# Older git versions try to fix line endings on images, this prevents it.
+*.png binary
+*.jpg binary
+*.ico binary
```
JohnTitor: At a glance, it seems this repository doesn't contain any JPEG or ICO files. Where do we need them or am I missing something?

jethrogb: Yeah they're not needed. I was just being exhaustive

PR #571, opened 2020-02-03
mir: begin documenting user variable debuginfo.

eddyb:
This is the companion to https://github.com/rust-lang/rust/pull/56231, which added these separate debuginfo annotations.
I tried to stick in as much information as I could, but the example does not help me - compare its single `scope 1` (with only a `debug` inside, no `let`s), with the example in https://github.com/rust-lang/rust/pull/56231.

r? @nikomatsakis cc @rust-lang/wg-mir-opt

### DIFF
--- a/src/mir/index.md
+++ b/src/mir/index.md
@@ -84,11 +84,7 @@ with a bunch of variable declarations. They look like this:
 
 ```mir
 let mut _0: ();                      // return place
-scope 1 {
-    let mut _1: std::vec::Vec<i32>;  // "vec" in scope 1 at src/main.rs:2:9: 2:16
-}
-scope 2 {
-}
+let mut _1: std::vec::Vec<i32>;      // in scope 0 at src/main.rs:2:9: 2:16
 let mut _2: ();
 let mut _3: &mut std::vec::Vec<i32>;
 let mut _4: ();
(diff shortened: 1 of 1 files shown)


spastorino:
@eddyb cool, I was wondering if something could also be documented on rustc-guide :).

nikomatsakis:
@eddyb you are my hero

src/mir/index.md
```diff
@@ -97,11 +93,27 @@ let mut _5: &mut std::vec::Vec<i32>;
...
+scope 1 {
+    debug vec => _1;                 // in scope 1 at src/main.rs:2:9: 2:16
+}
+```
+Each `debug <Name> => <Place>;` annotation describes a named user variable,
+and where (i.e. the place) a debugger can find the data of that variable.
+Here the mapping is trivial, but optimizations may complicate the place,
+or lead to multiple user variables sharing the same place.
+Additionally, closure captures are described using the same system, and so
+they're complicated even without optimizations, e.g.: `debug x => (*((*_1).0: &T));`.
```
nikomatsakis: it would be good, I think, to have some links to the data structures that describe this in the MIR. I dont' thnk we have to go into details, but just a few links would help people know where to get started.

eddyb: I agree, but it feels like this introduction to MIR is trying to stay light, and that makes it awkward to go into details.

nikomatsakis: yeah, really I think what we need is to add a subsection with more details. For now, I'm going to merge this PR as is.

eddyb: MIR feels like it should be a chapter with sections for every part, perhaps including thing like semantic constraints. I don't know why I expected that to already be a thing at first, I guess I haven't kept up with `rustc-guide`.

PR #578, opened 2020-02-14
Add links to the rustc docs

loris-intergalactique:
Hi,

I'm currently learning Rust and want to contribute to Rustc, so I think that adding links to the docs could be useful. I think it gives quick examples to newcomers like me. What do you think about it ? (There's also a small typo fix line 115)

The thing I'm not sure about : I kept a ~70 characters limit per line so I had to update some lines (like from lines 154 to 159) to keep it consistent. 

Thanks !


JohnTitor:
Overall looks good. Could you move all links to the bottom of the file? So that we can find them easily.

mark-i-m:
> Overall looks good. Could you move all links to the bottom of the file? So that we can find them easily.

Is that something we want to enforce? I personally tend to put the link next to the thing I'm linking, but I'm fine if we want to change it. But if we are going to put all the links at the bottom, we should go through the whole book and do it everywhere, IMHO.

JohnTitor:
> But if we are going to put all the links at the bottom, we should go through the whole book and do it everywhere, IMHO.

Ah, so I'm fine as it is!

loris-intergalactique:
I understand that making a bloc of links at the end of the file could be interesting, a little like the references section at the end of papers. But if we take an example like this one : https://raw.githubusercontent.com/rust-lang/rustc-guide/d6a6122b951c475fb12f8c407016efa397460cfa/src/borrow_check/region_inference.md

The file has a lot of references, and if we had a big list of links at the end of the file, I think that it would not be that user-friendly at the end, and even maybe prone to errors when maintaining it. What do you think ?

JohnTitor:
That's why I've approved as above now :)

loris-intergalactique:
Thanks hahaha

PR #580, opened 2020-02-14
Fix link

JohnTitor:
Also `The macro parser` section is outdated. But I'm really not familiar with macro parser. cc @spastorino


mark-i-m:
On the whole, the macro parser still works as described (I wrote that section)...

JohnTitor:
Oh, but we should update the signature at least. I'll submit a follow-up PR.

PR #582, opened 2020-02-16
Implicit Caller Location section.

anp:
Documenting the overall feature here as the current implementation is spread across several PRs and has deviated a good bit from the RFC.

recommended in https://github.com/rust-lang/rust/issues/47809#issuecomment-572497766

### DIFF
--- a/src/codegen/implicit-caller-location.md
+++ b/src/codegen/implicit-caller-location.md
@@ -1,0 +1,280 @@
+# Implicit Caller Location
+
+Approved in [RFC 2091], this feature enables the accurate reporting of caller location during panics
+initiated from functions like `Option::unwrap`, `Result::expect`, and `Index::index`. This feature 
+adds the [`#[track_caller]`][attr-reference] attribute for functions, the 
+[`caller_location`][intrinsic] intrinsic, and the stabilization-friendly 
+[`core::panic::Location::caller`][wrapper] wrapper.
+
+## Motivating Example
+
+Take this example program:
+
+```rust
+fn main() {
+    let foo: Option<()> = None;
+    foo.unwrap(); // this should produce a useful panic message!
+}
+```
+
+Prior to Rust 1.42, panics like this `unwrap()` printed a location in libcore:
+
+```
+$ rustc +1.41.0 example.rs; example.exe
+thread 'main' panicked at 'called `Option::unwrap()` on a `None` value',...core\macros\mod.rs:15:40
+note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
+```
+
+As of 1.42, we get a much more helpful message:
+
+```
+$ rustc +1.42.0 example.rs; example.exe 
+thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', example.rs:3:5
+note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
+```
+
+These error messages are achieved through a combination of changes to `panic!` internals to make use
+of `core::panic::Location::caller` and a number of `#[track_caller]` annotations in the standard 
+library which propagate caller information.
+
+## Reading Caller Location
+
+Previously, `panic!` made use of the `file!()`, `line!()`, and `column!()` macros to construct a
+[`Location`] pointing to where the panic occurred. These macros couldn't be given an overridden
+location, so functions which intentionally invoked `panic!` couldn't provide their own location, 
+hiding the actual source of error.
+
+Internally, `panic!()` now calls [`core::panic::Location::caller()`][wrapper] to find out where it 
+was expanded. This function is itself annotated with `#[track_caller]` and wraps the 
+[`caller_location`][intrinsic] compiler intrinsic implemented by rustc. This intrinsic is easiest 
+explained in terms of how it works in a `const` context.
+
+## Caller Location in `const`
+
+There are two main phases to returning the caller location in a const context: walking up the stack
+to find the right location and allocating a const value to return.
+
+### Finding the right `Location`
+
+In a const context we "walk up the stack" from where the intrinsic is invoked, stopping when we 
+reach the first function call in the stack which does *not* have the attribute. This walk is in
+[`InterpCx::find_closest_untracked_caller_location()`][const-find-closest].
+
+Starting at the bottom, we iterate up over stack [`Frame`][const-frame]s in the 
+[`InterpCx::stack`][const-stack], calling
+[`InstanceDef::requires_caller_location`][requires-location] on the 
+[`Instance`s from each `Frame`][frame-instance]. We stop once we find one that returns `false` and
+return the span of the *previous* frame which was the "topmost" tracked function.
+
+### Allocating a static `Location`
+
+Once we have a `Span`, we need to allocate static memory for the `Location`, which is performed by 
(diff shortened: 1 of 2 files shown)


eddyb:
Probably an easy way to explain it is to show the definition and what it codegens as, both in regular Rust code (even if nothing before codegen, not even MIR, encodes the caller location passing).

src/codegen/implicit-caller-location.md
```diff
@@ -0,0 +1,154 @@
...
+up on the tracking issue. During the course of implementing that, it was also discovered that an
+implementation was possible without modifying the number of arguments in a function's MIR, which
+would simplify later stages and unlock use in traits.
+
+## The Attribute
+
+The `#[track_caller]` attribute is checked alongside other codegen attrs to ensure the function:
+
+* has `"Rust"` ABI (as opposed to `"C"`, etc...)
+* is not in an extern block
```
eddyb: We probably use "extern block" elsewhere, but I wish we could say "is not a foreign import". or "FFI import'

Eh, at the very least, maybe write `extern {...}` instead of "extern".

anp:
cc @nikomatsakis you had specifically asked for this to be able to learn more about the feature. Would love to hear if it achieves that goal!

src/codegen/implicit-caller-location.md
```diff
@@ -0,0 +1,278 @@
...
+
+## Generating code for `#[track_caller]` callees
+
+To generate efficient code for a tracked function and its callers we need to provide the same 
+behavior from the intrinsic's point of view without having a stack to walk up at runtime. We invert
+the approach: as we grow the stack down we pass an additional argument to calls of tracked functions
+rather than walking up the stack when the intrinsic is called. That additional argument can be
+returned wherever the caller location is queried.
+
+The argument we append is of type `&'static core::panic::Location<'staic>`. A reference was chosen
```
yaahc:
```suggestion
The argument we append is of type `&'static core::panic::Location<'static>`. A reference was chosen
```

src/codegen/implicit-caller-location.md
```diff
@@ -0,0 +1,278 @@
...
+[`Location`] pointing to where the panic occurred. These macros couldn't be given an overridden
+location, so functions which intentionally invoked `panic!` couldn't provide their own location, 
+hiding the actual source of error.
+
+Internally, `panic!()` now calls [`core::panid::Location::caller()`][wrapper] to find out where it 
+was expanded. This function is itself annotated with `#[track_caller]` and wraps the 
+[`caller_location`][intrinsic] compiler intrinsic implemented by rustc. This intrinsic is easiest 
+explained in terms of how it works in a `const` context.
+
+## Caller Location in `const`
```
yaahc: This is information dense and I'm not sure its actually clearer than the next section, to me adding an extra arg where we forward information down every time its a `#[track_caller]` is much more intuitive

src/codegen/implicit-caller-location.md
```diff
@@ -0,0 +1,278 @@
...
+the return value of [`InstanceDef::requires_caller_location`][requires-location] which is in turn
+used in both const and codegen contexts to ensure correct propagation.
+
+### Traits
+
+When applied to a trait method prototype, the attribute takes effect on all implementations of the
+trait method. When applied to a default trait method implementation, the attribute takes effect on
+that implementation *and* any overrides. It is valid to apply the attribute to a regular
+implementation of a trait method, regardless of whether the defining trait does. It is a no-op to
+apply the attribute to trait methods with the attribute in the trait definition, but a valid one.
```
yaahc: the last two sentences here were a little difficult to follow

src/codegen/implicit-caller-location.md
anp: I think I've addressed this, let me know if you still find it dense!

src/codegen/implicit-caller-location.md
```diff
@@ -0,0 +1,278 @@
...
+library which propagate caller information.
+
+## Reading Caller Location
+
+Previously, `panic!` made use of the `file!()`, `line!()`, and `column!()` macros to construct a
+[`Location`] pointing to where the panic occurred. These macros couldn't be given an overridden
+location, so functions which intentionally invoked `panic!` couldn't provide their own location, 
+hiding the actual source of error.
+
+Internally, `panic!()` now calls [`core::panid::Location::caller()`][wrapper] to find out where it
```
nikomatsakis:
```suggestion
Internally, `panic!()` now calls [`core::panic::Location::caller()`][wrapper] to find out where it 
```

src/codegen/implicit-caller-location.md
```diff
@@ -0,0 +1,278 @@
...
+
+If the calling function is tracked, `get_caller_location` returns the local in
+[`FunctionCx::caller_location`][fcx-location] which was populated by the current caller's caller.
+In these cases the intrinsic "returns" a reference which was actually provided in an argument to its
+caller.
+
+If the calling function is not tracked, `get_caller_location` allocates a `Location` static from
+the current `Span` and returns a reference to that.
+
+By chaining together the `caller_location` fields of multiple `FunctionCx`s as we grow the bottom of
```
nikomatsakis: Nit: the term "chaining together" is confusing to me. Are we building a linked list in memory?

nikomatsakis: Oh, I guess we are. OK, great.

src/codegen/implicit-caller-location.md
```diff
@@ -0,0 +1,278 @@
...
+There are two main phases to returning the caller location in a const context: walking up the stack
+to find the right location and allocating a const value to return.
+
+### Finding the right `Location`
+
+In a const context we "walk up the stack" from where the intrinsic is invoked, stopping when we 
+reach the first function call in the stack which does *not* have the attribute. This walk is in
+[`InterpCx::find_closest_untracked_caller_location()`][const-find-closest].
+
+Starting at the bottom, we iterate up over stack [`Frame`][const-frame]s in the [InterpCx::stack],
```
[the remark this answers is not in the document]
anp: Good catch! The anchor reference needs to be updated too, which reflows the paragraph, which...

src/codegen/implicit-caller-location.md
eddyb: Not AFAIK, it's more of a conceptual chaining, every `#[track_caller]` function using their location parameter, to which a `#[track_caller]` caller would've passed in their own location parameter etc.

anp: `We achieve the same behavior as a loop starting from the bottom by passing a single `&Location` value through the `caller_location` fields of multiple `FunctionCx`s as we grow the stack downward."

?

anp: Actually just going to commit that and push, it's the last bit of feedback.

anp:
OK! Thank you for the reviews and the kind words <3. I think I've addressed all the feedback.

Once this is re-approved, it's gated on https://github.com/rust-lang/rustc-dev-guide/pull/641 and https://github.com/rust-lang/rustc-dev-guide/pull/646 to get CI green. I'll add a link from the code to this once it lands?

mark-i-m:
@anp Could you rebase on top of the most recent master? It has a bunch of link fixes.

anp:
Actually just finished doing that, those breakages were from my own links 😅 -- fixed up.

PR #585, opened 2020-02-18
Interning chapter

### LINKED ISSUE #419: Explain what interning means
At the bottom of https://rust-lang.github.io/rustc-guide/the-parser.html, the phrase "interns identifiers" is used. But to me, its meaning wasn't clear on first read-through. I checked in Zulip::/t-compiler/help and got back : "It means that you are not creating a heap allocation for an identifier every time, but instead have a global table where they are stored and deduplicated. You then just need to store the index into the global table whenever you have an identifier" from @oli (oliobk?). It would be helpful to have this explained early on (and maybe a link back to the definition in later uses?)

mark-i-m:
as discussed in #530 and elsewhere

resolves #419

### DIFF
--- a/src/type-inference.md
+++ b/src/type-inference.md
@@ -43,13 +43,6 @@ tcx.infer_ctxt().enter(|infcx| {
 })
 ```
 
-Each inference context creates a short-lived type arena to store the
-fresh types and things that it will create, as described in the
-[chapter on the `ty` module][ty-ch]. This arena is created by the `enter`
-function and disposed of after it returns.
-
-[ty-ch]: ty.html
-
 Within the closure, `infcx` has the type `InferCtxt<'cx, 'tcx>` for some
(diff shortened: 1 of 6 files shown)


src/type-inference.md
```diff
@@ -43,13 +43,6 @@ tcx.infer_ctxt().enter(|infcx| {
 })
 ```
 
-Each inference context creates a short-lived type arena to store the
-fresh types and things that it will create, as described in the
-[chapter on the `ty` module][ty-ch]. This arena is created by the `enter`
-function and disposed of after it returns.
-
-[ty-ch]: ty.html
```
LeSeulArtichaut: *Psst!* this is used later, at line 48

mark-i-m: Thanks! Fixed.

PR #589, opened 2020-02-19
Minor clarification and fix

JOE1994:


JohnTitor:
@JOE1994 Do you have any context of the first commit? I feel the current explanation is fine :)

JOE1994:
> Although you can obtain a valid `ParamEnv` for any item via
`tcx.param_env(def_id)`, this `ParamEnv` can be too generic for your use case.
Using the `ParamEnv` from the surrounding context can allow you to evaluate more things.

@JohnTitor My understanding was that, the above sentence is intending to say:

'*using the **immediate (innermost or tightest)** surrounding context allows to evaluate more things* (compared to using an **outer (looser or more general)** surrounding context).'

I was initially a bit confused with the meaning of the sentence, so I thought adding a clarifying term could help..

Perhaps, I could either
1. Undo the first commit, *OR*
2. Change the wording of **innermost** to **immediate**,  *OR*
3. Keep it as it is

Either way is fine for me. Thanks for reviewing! 🦀

JohnTitor:
Ah, your explanation seems reasonable to me! But actually I'm not English-native and might be missing the nuance so would like to up to other members :)

mark-i-m:
I actually understood the sentence to mean the opposite: "The context on the particular def_id may be too generic (i.e. doesn't have enough information), so you need to expand the scope of what context to gather more info", but I agree the sentence is a bit vague, and I'm not at all confident that I'm correct.

@oli-obk Could you please clarify where this additional context comes from and why it contains more information than `tcx.param_env(def_id)`? Perhaps you could give an example?

oli-obk:
Basically if you start out with processing a generic function `foo`, you have the param env of that function via `tcx.param_env(def_id_of_function_foo)`. Now if that function calls generic function `bar`, we may know some things inside `bar` that we wouldn't know if we just fetched `bar`'s param env. Example:

```
fn foo<T: SomeTrait>() where <T as SomeTrait>::SomeAssocTy: OtherTrait {
   bar::<T::SomeAssocTy>()
}
fn bar<T: ThirdTrait>() {
    if T::stuff() { mep() } else { mop() }
}
```

Due to knowing the OtherTrait bound we may be able to do some analysis.

This is a very contrived example that makes no sense in our existing analyses, but I can't come up with one where it's relevant without being on a PC. I can try creating one next week. Likely something with associated constants on generic traits or traits with assoc types.

mark-i-m:
Ping @JOE1994 Would you be able to update the PR?

JOE1994:
> Ping @JOE1994 Would you be able to update the PR?

I apologize for the late update.. :crying_cat_face:  I removed the wording 'innermost' that I had inserted in the previous commit. CI is currently failing due to a failing web link irrelevant to this PR.

mark-i-m:
Could you rebase on the latest master? (that will fix the ci issue)

JOE1994:
> Could you rebase on the latest master? (that will fix the ci issue)

I rebased/squashed to the latest master. CI check is passing now :smiley_cat:

PR #590, opened 2020-02-25
Add introductory section for name resolution chapter

phansch:
I've been looking into `librustc_resolve` the past couple of days and saw that the rustc-guide could use some updates. This adds a small intro section to the name resolution chapter.

### DIFF
--- a/src/name-resolution.md
+++ b/src/name-resolution.md
@@ -12,10 +12,10 @@ let x: x = 1;
 let y: x = 2;
 ```
 
-How do we know on line 3 whether `x` is a type (u32) or a value (1)?
-These conflicts are resolved in `librustc_resolve`.
-In this specific case, name resolution defines that type names and variable
-names live in separate namespaces and therefore can co-exist.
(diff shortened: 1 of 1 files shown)


src/name-resolution.md
```diff
@@ -1,6 +1,23 @@
...
+program:
+
+```rust
+type x = u32;
+let x: x = 1;
+let y: x = 2;
+```
+
+How do we know on line 3 whether `x` is a type (u32) or a value (1)?
+These conflicts are resolved in `librustc_resolve`.
```
JohnTitor: Now, the below part (L35-L36) seems redundant:

```
The name resolution lives in the librustc_resolve crate, with the meat in lib.rs and some helpers or symbol-type specific logic in the other modules.
```

I think it can be squashed with this line?

phansch: Oh, good catch. https://github.com/rust-lang/rustc-guide/pull/590/commits/ee9a681de2e3e3bc595b87e3772d9d631c5d5e78 changes `resolved in 'librustc_resolve'` to `resolved during name resolution` (and reformats the paragraph).

PR #594, opened 2020-02-27
Fix broken link

### LINKED ISSUE #591: broken link
In the query.md file:
https://github.com/rust-lang/rustc-guide/blob/master/src/query.md
the first link (which should point to the high level overview of rustc page) is broken. Easy to fix but i don't know which link  exactly it is supposed so i will let someone do a PR to fix/fix it myself if someone tells me which link it's supposed to be.

shika-blyat:
Fix #591

### DIFF
--- a/src/query.md
+++ b/src/query.md
@@ -9,7 +9,7 @@ simple. You have various queries that compute things about the input
 the def-id of some item, will compute the type of that item and return
 it to you.
 
-[hl]: high-level-overview.html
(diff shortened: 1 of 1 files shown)


JohnTitor:
Actually, we expect the link should work on https://rust-lang.github.io/rustc-guide/, not on https://github.com. So this fix is unnecessary.

shika-blyat:
Oh okay 😅, my bad.

PR #621, opened 2020-03-18
Example of rustc interface 

georgewfraser:
It's pretty hard to figure out how to use rustc_interface. This adds a simple "hello world" example of using rustc_interface to parse and typecheck a single file.

See discussion in https://internals.rust-lang.org/t/using-the-rustc-interface-crate/11982


JohnTitor:
I'd like to create another directory and put `.rs` files there since I'd prefer `src/` directory should contain markdown files or assets only, otherwise, it may make it hard that we find examples.

eddyb:
cc @Zoxc

georgewfraser:
@JohnTitor if you tell me what you want that directory to be named, I will set up 2 PRs, one with the example, one with the link.

src/rustc-driver-example.rs
```diff
@@ -0,0 +1,103 @@
...
+        // Registry of diagnostics codes.
+        registry: errors,
+    };
+    interface::run_compiler(config, |compiler| {
+        compiler.enter(|queries| {
+            // Parse the program and print the syntax tree.
+            let parse = queries.parse().unwrap().take();
+            println!("{:#?}", parse);
+            // Analyze the program and inspect the types of definitions.
+            queries.global_ctxt().unwrap().take().enter(|ctx| {
```
Zoxc: `ctx` should be named `tcx` here to match the compiler convention.

src/rustc-driver-example.rs
```diff
@@ -0,0 +1,103 @@
...
+
+        input: config::Input::Str {
+            name: source_map::FileName::Custom(String::from(filename)),
+            input: String::from(contents),
+        },
+        // Option<PathBuf>
+        input_path: None,
+        // Option<PathBuf>
+        output_dir: None,
+        // Option<PathBuf>
```
Zoxc: Not sure listing the types is very helpful.

JohnTitor:
@georgewfraser `examples` or something else would be fine. Anyway we can rename it later.

src/rustc-driver-example.rs
georgewfraser: If you know nothing about the compiler internals (like me!) it's quite helpful for understanding what each of these options actually does.

georgewfraser:
OK @mark-i-m this is just the .rs file now, once this is merged I will make a new PR with the changes to the .md file.

PR #623, opened 2020-03-19
Add section: Source file checksums in debug info

arlosi:
Implementation PR in Rust: https://github.com/rust-lang/rust/pull/69718

### DIFF
--- a/src/debugging-support-in-rustc.md
+++ b/src/debugging-support-in-rustc.md
@@ -268,6 +268,45 @@ Focus is to let macros decide what to do. This can be achieved by having some ki
 that lets the macro tell the compiler where the line marker should be. This affects where you
 set the breakpoints and what happens when you step it.
 
+## Source file checksums in debug info
+
+Both DWARF and CodeView (PDB) support embedding a cryptographic hash of each source file that
+contributed to the associated binary.
+
+The cryptographic hash can be used by a debugger to verify that the source file matches the
+executable. If the source file does not match, the debugger can provide a warning to the user.
+
+The hash can also be used to prove that a given source file has not been modified since it was
+used to compile an executable. Because MD5 and SHA1 both have demonstrated vulnerabilities,
+using SHA256 is recommended for this application.
+
+The Rust compiler stores the hash for each source file in the corresponding `SourceFile` in
+the `SourceMap`. The hashes of input files to external crates are stored in `rlib` metadata.
+
+A default hashing algorithm is set in the target specification. This allows the target to
+specify the best hash available, since not all targets support all hash algorithms.
+
+The hashing algorithm for a target can also be overridden with the `-Z source-file-checksum=`
+command-line option.
+
+#### DWARF 5
+DWARF version 5 supports embedding an MD5 hash to validate the source file version in use.
+DWARF 5 - Section 6.2.4.1 opcode DW_LNCT_MD5
+
+#### LLVM
+LLVM IR supports MD5 and SHA1 (and SHA256 in LLVM 11+) source file checksums in the DIFile node.
+
+[LLVM DIFile documentation](https://llvm.org/docs/LangRef.html#difile)
+
+#### Microsoft Visual C++ Compiler /ZH option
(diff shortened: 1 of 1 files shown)


eddyb:
(this repo doesn't have any bots AFAIK)

JohnTitor:
I think this should be blocked until that PR is merged?

src/debugging-support-in-rustc.md
```diff
@@ -268,6 +268,36 @@ Focus is to let macros decide what to do. This can be achieved by having some ki
...
+#### LLVM
+LLVM IR supports MD5, SHA1, and SHA256 source file checksums in the DIFile node.
+
+LLVM DIFile documentation: [https://llvm.org/docs/LangRef.html#difile]
+
+#### Microsoft Visual C++ Compiler /ZH option
+The MSVC compiler supports embedding MD5, SHA1, or SHA256 hashes in the PDB using the `/ZH`
+compiler option.
+
+MSVC /ZH documentation: [https://docs.microsoft.com/en-us/cpp/build/reference/zh]
```
nikomatsakis: Nit: Maybe make this proper links, like so?

[MSVC /ZH documentation](https://docs.microsoft.com/en-us/cpp/build/reference/zh)

src/debugging-support-in-rustc.md
```diff
@@ -268,6 +268,36 @@ Focus is to let macros decide what to do. This can be achieved by having some ki
...
+
+Both DWARF and CodeView (PDB) support embedding a cryptographic hash of each source file that
+contributed the associated binary.
+
+The cryptographic hash can be used by a debugger to verify that the source file matches the
+executable. If the source file does not match, the debugger can provide a warning to the user.
+
+The hash can also be used to prove that a given source file has not been modified since it was
+used to compile an executable. Because MD5 and SHA1 both have demonstrated vulnerabilities,
+using SHA256 is recommended for this application.
```
nikomatsakis: It might be worth noting roughly where in the code this information is attached -- e.g., the name of a function or module. Is this something that is done in the LLVM-specific code, or is it portable across backends by default?

arlosi: I added some additional detail. It is portable across backends.

src/debugging-support-in-rustc.md
```diff
@@ -268,6 +268,45 @@ Focus is to let macros decide what to do. This can be achieved by having some ki
...
+
+The Rust compiler stores the hash for each source file in the corresponding `SourceFile` in
+the `SourceMap`. The hashes of input files to external crates are stored in `rlib` metadata.
+
+A default hashing algorithm is set in the target specification. This allows the target to
+specify the best hash available, since not all targets support all hash algorithms.
+
+The hashing algorithm for a target can also be overridden with the `-Z source-file-checksum=`
```
arlosi: Questions for reviewers: Do we even want a command line option to override this? Or is rare enough that people wanting to override this could create a custom target specification?

src/debugging-support-in-rustc.md
```diff
@@ -268,6 +268,36 @@ Focus is to let macros decide what to do. This can be achieved by having some ki
 that lets the macro tell the compiler where the line marker should be. This affects where you
 set the breakpoints and what happens when you step it.
 
+## Source file checksums in debug info
+
+Both DWARF and CodeView (PDB) support embedding a cryptographic hash of each source file that
+contributed the associated binary.
```
sivadeilra: nit: "contributed the" --> "contributed to the"

src/debugging-support-in-rustc.md
sivadeilra: Would it be a good idea to spec this as a list of checksums, e.g. `-Z sha256,md5`, in case a target can support more than one?

arlosi: I would prefer to keep it as a single checksum for now. Both LLVM and CodeView (PDB) only support a single type of checksum. If a backend starts supporting multiple checksum algorithms, we can extend it then.

mark-i-m:
rust-lang/rust#69718 has merged

PR #624, opened 2020-03-20
Move GenericArgs chapter under ty

mark-i-m:


LeSeulArtichaut:
By the way, is there any reason this cannot be included in the "Generics and substitutions" chapter?

mark-i-m:
@LeSeulArtichaut It probably could be. I just noticed that it should be a subchapter and made a quick PR. But if you are interested in doing that, it would be great :)

LeSeulArtichaut:
Let's merge this one?

PR #631, opened 2020-03-24
Change ICE-breaker ping to use the new aliases

LeSeulArtichaut:
I'm not sure if this is really useful, but I think the "full" invocation is really long and not convenient to learn.

Related to rust-lang/rust#69949

### DIFF
--- a/src/ice-breaker/about.md
+++ b/src/ice-breaker/about.md
@@ -58,13 +58,21 @@ To tag an issue as appropriate for an ICE-breaker group, you give
 team. For example:
 
 ```text
+@rustbot ping icebreakers-llvm
+@rustbot ping icebreakers-cleanup-crew
+```
+
+To make these commands shorter and easier to remember, there are aliases,
+defined in the [`triagebot.toml`] file. For example:
+
+```text
 @rustbot ping llvm
 @rustbot ping cleanup
 ```
 
-These are aliases that are easier to remember than the full invocation.
-They are declared in the [`triagebot.toml`] file and might be subject to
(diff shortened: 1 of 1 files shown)


JohnTitor:
I believe it's reasonable to tell that we have aliases for simplicity. We have some aliases and they might be changed/added so I think it's better to add a note and point to the triagebot.toml on rust-lang/rust.

LeSeulArtichaut:
@JohnTitor I wrote a note in 6e11e5b, but I'm not fully satisfied with the wording. Would you mind reviewing? Thanks 😄

pnkfelix:
I think the thrust of @JohnTitor 's comment was that it would be best to provide the non-aliased versions first, since I suspect those will never change (it seems more likely we'd retire a given ICE-breaker group than rename it).

And then, after that, state that some aliases exist (with the link to the triagebot toml), that they are subject to change, and *then* give a second code block showing the invocations with the aliases.

Does that make sense?

pnkfelix:
(One reason it could be good to include the non-aliased versions in these docs is that if someone writes some tools, e.g. something to assist with triage, then it would be good for those tools to probably use the non-aliased versions...)

JohnTitor:
> I think the thrust of @JohnTitor 's comment was that it would be best to provide the non-aliased versions first, since I suspect those will never change (it seems more likely we'd retire a given ICE-breaker group than rename it).

Yeah, that's exactly what I mean :)

mark-i-m:
Ping @LeSeulArtichaut  :)

PR #633, opened 2020-03-26
Overview

### LINKED ISSUE #320: Write an overview chapter
cc @lokalmatador

Copying from the dropbox doc:

More or less an expansion of the "High-level overview of the compiler source" section)
Content (or links to):

- compiler high level structure
- system requirements & dependencies
- compiler module structure
- crate dependencies
- compiler configuration
- compiler inputs
- compiler outputs
- compiler state
- Walk-through of compiling a meaningful example

Current plans are to walk through a simple example program and show how it moves through the compiler.

This issue has been assigned to @lokalmatador via [this comment](https://github.com/rust-lang/rustc-guide/issues/320#issuecomment-506993854).

mark-i-m:
This is our draft so far. Feedback welcome! And thanks to all who contributed :tada: 

cc @chrissimpkins @rust-lang/wg-learning 

resolves #320


src/overview.md
```diff
@@ -0,0 +1,327 @@
...
+challenges with lock contention, maintaining query-system invariants under
+concurrency, and the complexity of the code base. One can try out the current
+work by enabling parallel compilation in `config.toml`. It's still early days,
+but there are already some promising performance improvements.
+
+### Bootstrapping
+
+**TODO (or do we want such a section)?**
+
+## A flow chart or walkthrough diagram
```
chrissimpkins: Is this a duplicate of the diagram above?

crlf0710:
There's a link to `syntax`, which is now `rustc_ast`, i guess.

src/overview.md
```diff
@@ -0,0 +1,327 @@
...
+    - The different libraries/binaries are linked together to produce the final
+      binary. **TODO: reference for this section?**
+
+[`librustc_lexer`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lexer/index.html
+[`librustc_driver`]: https://rust-lang.github.io/rustc-guide/rustc-driver.html
+[lex]: https://rust-lang.github.io/rustc-guide/the-parser.html
+[`StringReader`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_parse/lexer/struct.StringReader.html
+[`librustc_parse`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_parse/index.html
+[parser]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lexer/index.html
+[hir]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lexer/index.html
```
ebroto: This should point to https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/index.html instead

src/overview.md
```diff
@@ -0,0 +1,327 @@
...
+      **TODO: reference for this section?**
+    - The different libraries/binaries are linked together to produce the final
+      binary. **TODO: reference for this section?**
+
+[`librustc_lexer`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lexer/index.html
+[`librustc_driver`]: https://rust-lang.github.io/rustc-guide/rustc-driver.html
+[lex]: https://rust-lang.github.io/rustc-guide/the-parser.html
+[`StringReader`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_parse/lexer/struct.StringReader.html
+[`librustc_parse`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_parse/index.html
+[parser]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lexer/index.html
```
ebroto: Not sure if we wanted to link to https://doc.rust-lang.org/nightly/nightly-rustc/rustc_parse/index.html here?

src/overview.md
```diff
@@ -0,0 +1,323 @@
...
+guide completely up to date on everything!
+
+The compiler definitely has rough edges, but because of its design it is able
+to keep up with the requirements above.
+
+### Intermediate representations
+
+As with most compilers, `rustc` uses some intermediate representations (IRs) to
+facilitate computations. In general, working directly with the source code is
+extremely inconvenient. Source code is designed to be human-friendly while at
```
Centril, line 142:
```suggestion
extremely inconvenient and error-prone. Source code is designed to be human-friendly while at
```

src/overview.md
```diff
@@ -0,0 +1,323 @@
...
+the same time being unambiguous, but it's less convenient for doing something
+like, say, type checking.
+
+Instead most compilers, including `rustc`, build some sort of IR out of the
+source code which is easier to analyze. `rustc` has a few IRs, each optimized
+for different things:
```
Centril, line 148:
```suggestion
for different purposes:
```

src/overview.md
```diff
@@ -0,0 +1,323 @@
...
+
+- Abstract Syntax Tree (AST): the abstract syntax tree is built from the stream
```
Centril, line 150: We can consider the stream of tokens as an IR in its own right.

src/overview.md
```diff
@@ -0,0 +1,323 @@
...
+  of tokens produced by the lexer directly from the source code. It represents
+  pretty much exactly what the user wrote. It helps to do some syntactic sanity
+  checking (e.g. checking that a type is expected where the user wrote one).
+- High-level IR (HIR): This is a sort of very desugared AST. It's still close
```
Centril, line 154:
```suggestion
- High-level IR (HIR): This is a sort of desugared AST. It's still close
```
It's not that desugared tbh.

src/overview.md
```diff
@@ -0,0 +1,323 @@
...
+  to what the user wrote syntactically, but it includes some implicit things
+  such as some elided lifetimes, etc. This IR is amenable to type checking.
+- HAIR: This is an intermediate between HIR and MIR. This only exists to make
+  it easier to lower HIR to MIR.
```
Centril, line 158: It also exists to do exhaustiveness checking for the pattern fragment. Worthy of mention here is that this AST is explicitly typed and e.g., method calls and implicit derefs are made explicit.

src/overview.md
```diff
@@ -0,0 +1,323 @@
...
+- Middle-level IR (MIR): This IR is basically a Control-Flow Graph (CFG). A CFG
+  is a type of diagram that shows the basic blocks of a program and how control
+  flow can go between them. Likewise, MIR also has a bunch of basic blocks with
+  simple typed statements inside them (e.g. assignment, simple computations,
+  dropping values, etc). MIR is used for borrow checking and a bunch of other
```
Centril, line 163: Drop is a terminator though.

Centril, line 163:
```suggestion
  dropping values, etc). MIR is used for borrow checking and other
```
(general note: try reducing the wordiness of "a bunch" et. al in various places)

src/overview.md
```diff
@@ -0,0 +1,323 @@
...
+  important dataflow based checks, such as checking for uninitialized values.
+  It is also used for a bunch of optimizations and for constant evaluation (via
```
Centril, line 165: "a bunch" again

src/overview.md
```diff
@@ -0,0 +1,323 @@
...
+  MIRI). Because MIR is still generic, we can do a lot of analyses here more
```
Centril, line 166:
```suggestion
  Miri). Because MIR is still generic, we can do a lot of analyses here more
```

src/overview.md
```diff
@@ -0,0 +1,323 @@
...
+  efficiently than after monomorphization.
+- LLVM IR: This is the standard form of all input to the LLVM compiler. LLVM IR
+  is basically a sort of typed assembly language with lots of annotations. It's
```
Centril, line 169:
```suggestion
  is a sort of typed assembly language with lots of annotations. It's
```

src/overview.md
```diff
@@ -0,0 +1,323 @@
...
+### Queries
+
+The first big implementation choice is the _query_ system. The rust compiler
+uses a query system which is unlike most textbook compilers, which are
+organized as a series of passes over the code that execute sequentially. The
+compiler does this to make incremental compilation possible -- that is, if the
+user makes a change to their program and recompiles, we want to do as little
+redundant work as possible to produce the new binary.
+
+In rustc, all the major steps above are organized as a bunch of queries that
```
Centril, line 184:
```suggestion
In `rustc`, all the major steps above are organized as a bunch of queries that
```

src/overview.md
```diff
@@ -0,0 +1,323 @@
...
+call each other. For example, there is a query to ask for the type of something
+and another to ask for the optimized MIR of a function, and so on. These
```
Centril, line 186:
```suggestion
and another to ask for the optimized MIR of a function. These
```

src/overview.md
```diff
@@ -0,0 +1,323 @@
...
+queries can call each other and are all tracked through the query system, and
+the results of the queries are cached on disk so that we can tell which
```
Centril, line 188: Long sentence.

src/overview.md
```diff
@@ -0,0 +1,323 @@
...
+`collect_and_partition_mono_items` query, which first recursively requests the
+`optimized_mir` for all reachable functions, which in turn runs `mir_borrowck`
+for that function and then creates codegen units. This kind of split will need
+to remain to ensure that unreachable functions still have their errors emitted.
+
+[passes]: https://github.com/rust-lang/rust/blob/45ebd5808afd3df7ba842797c0fcd4447ddf30fb/src/librustc_interface/passes.rs#L824
+
+Moreover, the compiler wasn't originally built to use a query system; the query
+system has been retrofitted into the compiler, so parts of it are not
+query-fied yet. Also, LLVM isn't our code, so obviously that isn't querified
```
Centril, line 212:
```suggestion
query-fied yet. Also, LLVM isn't our code, so that isn't querified
```
Nothing is ever obvious. :)

src/overview.md
```diff
@@ -0,0 +1,323 @@
...
+either. The plan is to eventually query-fy all of the steps listed in the
+previous section, but as of this writing, only the steps between HIR and
+LLVM-IR are query-fied. That is, lexing and parsing are done all at once for
+the whole program.
+
+One other thing to mention here is the all-important "typing context",
```
Centril, line 218: It's massively confusing to call this the "typing context" -- It doesn't actually correspond to some `Γ` or `Δ` in the sense of typing rules.

src/overview.md
```diff
@@ -0,0 +1,323 @@
...
+[`TyCtxt`], which is a giant struct that is at the center of all things. All
+queries are defined as methods on the [`TyCtxt`] type, and the in-memory query
+cache is stored there too. In the code, there is usually a variable called
+`tcx` which is a handle on the typing context. You will also see lifetimes with
+the name `'tcx`, which means that something is tied to the lifetime of the
+`TyCtxt` (usually it is stored or _interned_ there).
```
Centril, line 224: maybe mention "arena" here :)

src/overview.md
```diff
@@ -0,0 +1,323 @@
...
+on [`ty::Ty`][ty], but for now, we just want to mention that it exists and is the way
+`rustc` represents types!
+
+Oh, and also the `rustc::ty` module defines the `TyCtxt` struct we mentioned before.
+
+[ty]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc/ty/type.Ty.html
+
+### Parallelism
+
+Compiler performance is a problem that we would very much like to improve on
```
Centril, line 242:
```suggestion
Compiler performance is a problem that we would like to improve on
```

src/overview.md
```diff
@@ -0,0 +1,323 @@
...
+(and are always working on). One aspect of that is attempting to parallelize
```
Centril, line 243:
```suggestion
(and are always working on). One aspect of that is parallelizing
```

src/overview.md
```diff
@@ -0,0 +1,323 @@
...
+`rustc` itself.
+
+Currently, there is only one part of rustc that is already parallel: codegen.
```
Centril, line 246: cc @Zoxc

src/overview.md
```diff
@@ -0,0 +1,323 @@
...
+approach is (**TODO: verify**) to turn `RefCell`s into `Mutex`s -- that is, we
+switch to thread-safe internal mutability. However, there are ongoing
+challenges with lock contention, maintaining query-system invariants under
+concurrency, and the complexity of the code base. One can try out the current
+work by enabling parallel compilation in `config.toml`. It's still early days,
+but there are already some promising performance improvements.
+
+### Bootstrapping
+
+**TODO (or do we want such a section)?**
```
Centril, line 264: Yes. :)

src/overview.md
```diff
@@ -0,0 +1,323 @@
...
+    - Main entry point: **TODO**
+- Lexical Analysis: Lex the user program to a stream of tokens
+    - Guide: [Lexing and Parsing](https://rust-lang.github.io/rustc-guide/the-parser.html)
+    - Lexer definition: [`librustc_lexer`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lexer/index.html)
+    - Main entry point: **TODO**
+- Parsing: Parse the stream of tokens to an Abstract Syntax Tree (AST)
+    - Guide: [Lexing and Parsing](https://rust-lang.github.io/rustc-guide/the-parser.html)
+    - Parser definition: [`librustc_parse`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_parse/index.html)
+    - Main entry point: **TODO**
+    - AST definition: [`librustc_ast`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_ast/ast/index.html)
```
Centril, line 292: Expansion, nameres, ast validation, and feature gating, and early linting should probably be mentioned here.

src/overview.md
```diff
@@ -0,0 +1,323 @@
...
+- The High Level Intermediate Representation (HIR)
```
Centril, line 293: Something something late linting...

chrissimpkins:
@mark-i-am OK to open a PR against this branch for work on the Overview?

mark-i-m:
@chrissimpkins Yes, PRs are welcome :) Thanks in advance!

chrissimpkins:
Great! I will start working on this tonight.

src/overview.md
mark-i-m: Hmm... I guess I have always had this mental model of "Lexing and Parsing" being one large black box. To what extent is there a clean separation in the code?

src/overview.md
mark-i-m: Yeah, this sentence is a bit imprecise. I'm wondering how much to say about the structure of MIR here...

src/overview.md
mark-i-m: "many"?

src/overview.md
mark-i-m: I thought it was an acronym "MIR Interpreter"? cc @oli-obk

src/overview.md
mark-i-m: Yep, I very much agree, but that's what it's called in the source (I think for historical reasons). We should add a note about this here.

src/overview.md
mark-i-m: Good to know! Do you have specific things you think we should mention?

mark-i-m:
Oh :sob: I rebased over master and force-pushed over centril suggestions which I just applied here... NOTE TO SELF: manually reapply them all.

src/overview.md
```diff
@@ -0,0 +1,323 @@
...
+  We now know what the exact work is we need to do (e.g. which nightly features
+  are enabled, whether we are doing a `check`-only build or emiting LLVM-IR or
+  a full compilation).
+- Then, we start to do compilation...
+- We first [_lex_ the user program][lex]. This turns the program into a stream
+  of _tokens_ (yes, the same sort of tokens as `proc_macros` (sort of)).
+  [`StringReader`] from [`librustc_parse`] integrates [`librustc_lexer`] with
+  `rustc` data structures.
+- We then [_parse_ the stream of tokens][parser] to build an Abstract Syntax
+  Tree (AST).
```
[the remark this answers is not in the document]
chrissimpkins: Added bullet points for these topics in https://github.com/rust-lang/rustc-dev-guide/pull/659.

src/overview.md
```diff
@@ -0,0 +1,322 @@
...
+# References
+
+- Command line parsing
+    - Guide: [The Rustc Driver and Interface](https://rust-lang.github.io/rustc-guide/rustc-driver.html)
+    - Driver definition: [`rustc_driver`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_driver/)
+    - Main entry point: **TODO**
+- Lexical Analysis: Lex the user program to a stream of tokens
+    - Guide: [Lexing and Parsing](https://rust-lang.github.io/rustc-guide/the-parser.html)
+    - Lexer definition: [`librustc_lexer`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lexer/index.html)
+    - Main entry point: **TODO**
```
LeSeulArtichaut, line 286: I'm really unsure but it probably is [`rustc_lexer::tokenize`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lexer/fn.tokenize.html).
```suggestion
    - Main entry point: [`rustc_lexer::tokenize`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lexer/fn.tokenize.html)
```

src/overview.md
```diff
@@ -0,0 +1,322 @@
...
+- Parsing: Parse the stream of tokens to an Abstract Syntax Tree (AST)
+    - Guide: [Lexing and Parsing](https://rust-lang.github.io/rustc-guide/the-parser.html)
+    - Parser definition: [`librustc_parse`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_parse/index.html)
+    - Main entry point: **TODO**
```
LeSeulArtichaut, line 290: I believe this is [`rustc_parse::parse_in`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_parse/fn.parse_in.html).
```suggestion
    - Main entry point: [`rustc_parse::parse_in`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_parse/fn.parse_in.html)
```

src/overview.md
```diff
@@ -0,0 +1,322 @@
...
+    - Guide: [Identifiers in the HIR](https://rust-lang.github.io/rustc-guide/hir.html#identifiers-in-the-hir)
+    - Guide: [The HIR Map](https://rust-lang.github.io/rustc-guide/hir.html#the-hir-map)
+    - Guide: [Lowering AST to HIR](https://rust-lang.github.io/rustc-guide/lowering.html)
+    - How to view HIR representation for your code `cargo rustc -- -Zunpretty=hir-tree`
+    - Rustc HIR definition: [`rustc_hir`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/index.html)
+    - Main entry point: **TODO**
+- Type Inference
+    - Guide: [Type Inference](https://rust-lang.github.io/rustc-guide/type-inference.html)
+    - Guide: [The ty Module: Representing Types](https://rust-lang.github.io/rustc-guide/ty.html) (semantics)
+    - Main entry point: **TODO**
```
LeSeulArtichaut, line 303: Reading the Guide, I think this is [`rustc_infer::infer::InferCtxtBuilder::enter`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_infer/infer/struct.InferCtxtBuilder.html#method.enter).
```suggestion
    - Main entry point: [`rustc_infer::infer::InferCtxtBuilder::enter`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_infer/infer/struct.InferCtxtBuilder.html#method.enter)
```

src/overview.md
```diff
@@ -0,0 +1,322 @@
...
+- The Mid Level Intermediate Representation (MIR)
+    - Guide: [The MIR (Mid level IR)](https://rust-lang.github.io/rustc-guide/mir/index.html)
+    - Definition: [`librustc/mir`](https://github.com/rust-lang/rust/tree/master/src/librustc/mir)
+    - Definition of source that manipulates the MIR: [`librustc_mir`](https://github.com/rust-lang/rust/tree/master/src/librustc_mir)
+    - Main entry point: **TODO**
+- The Borrow Checker
+    - Guide: [MIR Borrow Check](https://rust-lang.github.io/rustc-guide/borrow_check.html)
+    - Definition: [`rustc_mir/borrow_check`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_mir/borrow_check/index.html)
+    - Main entry point: **TODO**
```
LeSeulArtichaut, line 312: This is the [`mir_borrowck`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_mir/borrow_check/fn.mir_borrowck.html) query.

src/overview.md
```diff
@@ -0,0 +1,322 @@
...
+- MIR Optimizations
+    - Guide: [MIR Optimizations](https://rust-lang.github.io/rustc-guide/mir/optimizations.html)
+    - Definition: [`rustc_mir/transform`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_mir/transform/index.html) **TODO: is this correct?**
+    - Main entry point: **TODO**
```
LeSeulArtichaut, line 316: I think this is the [`optimized_mir`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_mir/transform/fn.optimized_mir.html) query.

src/overview.md
```diff
@@ -0,0 +1,322 @@
...
+- Code Generation
+    - Guide: [Code Generation](https://rust-lang.github.io/rustc-guide/codegen.html)
+    - Guide: [Generating LLVM IR](https://rust-lang.github.io/rustc-guide/codegen.html#generating-llvm-ir) - **TODO: this is not available yet**
+    - Generating Machine Code from LLVM IR with LLVM - **TODO: reference?**
+    - Main entry point MIR -> LLVM IR: **TODO**
+    - Main entry point LLVM IR -> Machine Code **TODO**
```
LeSeulArtichaut, line 322: Is it [`rustc_codegen_ssa::base::codegen_crate`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_codegen_ssa/base/fn.codegen_crate.html)?
```suggestion
    - Main entry point LLVM IR -> Machine Code: [`rustc_codegen_ssa::base::codegen_crate`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_codegen_ssa/base/fn.codegen_crate.html)
```

LeSeulArtichaut:
@mark-i-m You left many links to the old domain https://rust-lang.github.io/rustc-guide. Is it intentional?

src/overview.md
Centril: No, that's a function that's primarily used for parsing the `TokenStream` inside an `Attribute`, and not for general parsing.

mark-i-m:
@LeSeulArtichaut Good catch. I will fix that.

I'll try to work on this a bit more tomorrow...

src/overview.md
```diff
@@ -0,0 +1,322 @@
...
+
+So first, let's look at what the compiler does to your code. For now, we will
+avoid mentioning how the compiler implements these steps except as needed;
+we'll talk about that later.
+
+**TODO: Would be great to have a diagram of this once we nail down the details...**
+
+**TODO: someone else should confirm this vvv**
+
+- The compile process begins when a user writes a Rust source program in text and invokes the `rustc` compiler on it. The work that the compiler needs to perform is defined by command-line options. For example, it is possible to enable nightly features (`-Z` flags), perform `check`-only builds, or emit LLVM-IR rather than executable machine code. The `rustc` executable call may be indirect through the use of `cargo`.
```
mark-i-m, line 22: https://github.com/rust-lang/rustc-dev-guide/pull/659#discussion_r402821994

> Should we have a soft wrap here?

> Open question how we will format the entire document. Maybe this will become standard body copy with paragraphs rather than a bullet list?

src/overview.md
```diff
@@ -0,0 +1,322 @@
...
+- Command line argument parsing occurs in the [`librustc_driver`]. This crate defines the compile configuration that is requested by the user and passes it to the rest of the compilation process as a [`rustc_interface::Config`].
+- The raw Rust source text is analyzed by a low-level lexer located in [`librustc_lexer`]. At this stage, the source text is turned into a stream of atomic source code units known as _tokens_. (**TODO**: chrissimpkins - Maybe discuss Unicode handling during this stage?)
+- The token stream passes through a higher-level lexer located in [`librustc_parse`] to prepare for the next stage of the compile process. The [`StringReader`] struct is used at this stage to perform a set of validations and turn strings into interned symbols.
```
mark-i-m, line 25: https://github.com/rust-lang/rustc-dev-guide/pull/659#discussion_r402848391

> Did we define interning before? If not, maybe include a link to the chapter about interning?

src/overview.md
mark-i-m: I think if we are going to do paragraphs, we should make them actually paragraphs, rather than bullets. But if we are going to do bullets, we should keep the, short.

Personally I'm ok with doing paragraphs.

src/overview.md
```diff
@@ -19,16 +19,16 @@ we'll talk about that later.
...
-- We first [_lex_ the user program][lex]. This turns the program into a stream
-  of _tokens_ (yes, the same sort of tokens as `proc_macros` (sort of)).
-  [`StringReader`] from [`librustc_parse`] integrates [`librustc_lexer`] with
-  `rustc` data structures.
+- The compile process begins when a user writes a Rust source program in text and invokes the `rustc` compiler on it. The work that the compiler needs to perform is defined by command-line options. For example, it is possible to enable nightly features (`-Z` flags), perform `check`-only builds, or emit LLVM-IR rather than executable machine code. The `rustc` executable call may be indirect through the use of `cargo`.
+- Command line argument parsing occurs in the [`librustc_driver`]. This crate defines the compile configuration that is requested by the user and passes it to the rest of the compilation process as a [`rustc_interface::Config`].
+- The raw Rust source text is analyzed by a low-level lexer located in [`librustc_lexer`]. At this stage, the source text is turned into a stream of atomic source code units known as _tokens_. (**TODO**: chrissimpkins - Maybe discuss Unicode handling during this stage?)
```
mark-i-m, line 24: @chrissimpkins is there anything specific you had in mind, or would saying "we handle unicode codepoint here" be sufficient?

src/overview.md
mark-i-m: I don't think we have.

src/overview.md
chrissimpkins: > Personally I'm ok with doing paragraphs.

I like that idea too.  I think that we may be hitting lengths that are a bit too much for lists.

mark-i-m: One thing, though is that we should resolve all of the above comments first. Otherwise, github will mark them all as outdated.

src/overview.md
chrissimpkins: This is an area of interest for me and maybe not necessary in the Overview.  I am trying to understand the Unicode handling in Rust.  I came across `librustc_parse::lexer::unicode_chars` and thought that it might be interesting to add a comment about Unicode support.  If it doesn't work here I will see if I can work it into the Guide chapter once I understand it.

src/overview.md
mark-i-m: @chrissimpkins ?

src/overview.md
mark-i-m: @Centril Can you point to the correct entry point?

src/overview.md
mark-i-m: @eddyb confirm?

src/overview.md
mark-i-m: @bjorn3 can you confirm?

src/overview.md
```diff
@@ -0,0 +1,322 @@
...
+So first, let's look at what the compiler does to your code. For now, we will
+avoid mentioning how the compiler implements these steps except as needed;
+we'll talk about that later.
+
+**TODO: Would be great to have a diagram of this once we nail down the details...**
+
+**TODO: someone else should confirm this vvv**
+
+- The compile process begins when a user writes a Rust source program in text and invokes the `rustc` compiler on it. The work that the compiler needs to perform is defined by command-line options. For example, it is possible to enable nightly features (`-Z` flags), perform `check`-only builds, or emit LLVM-IR rather than executable machine code. The `rustc` executable call may be indirect through the use of `cargo`.
+- Command line argument parsing occurs in the [`librustc_driver`]. This crate defines the compile configuration that is requested by the user and passes it to the rest of the compilation process as a [`rustc_interface::Config`].
```
bjorn3, line 23: ``[`rustc_interface::Config`]`` isn't rendered as link.

src/overview.md
```diff
@@ -0,0 +1,322 @@
...
+- We then take the AST and [convert it to High-Level Intermediate
+  Representation (HIR)][hir]. This is a compiler-friendly representation of the
+  AST.  This involves a lot of desugaring of things like loops and `async fn`.
+- We use the HIR to do [type inference]. This is the process of automatic
+  detection of the type of an expression. **TODO: how `ty` module fits in
+  here**
+- **TODO: Maybe some other things are done here? I think initial type checking
+  happens here? And trait solving?**
+- The HIR is then [lowered to Mid-Level Intermediate Representation (MIR)][mir].
+- The MIR is used for [borrow checking].
```
bjorn3, line 43: This is not rendered as link.

src/overview.md
```diff
@@ -0,0 +1,322 @@
...
+  - Definition: [`librustc/mir`](https://github.com/rust-lang/rust/tree/master/src/librustc/mir)
+  - Definition of source that manipulates the MIR: [`librustc_mir`](https://github.com/rust-lang/rust/tree/master/src/librustc_mir)
+  - Main entry point: **TODO**
+- The Borrow Checker
+  - Guide: [MIR Borrow Check](https://rust-lang.github.io/rustc-guide/borrow_check.html)
+  - Definition: [`rustc_mir/borrow_check`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_mir/borrow_check/index.html)
+  - Main entry point: **TODO**
+- MIR Optimizations
+  - Guide: [MIR Optimizations](https://rust-lang.github.io/rustc-guide/mir/optimizations.html)
+  - Definition: [`rustc_mir/transform`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_mir/transform/index.html) **TODO: is this correct?**
```
bjorn3, line 315: Yes this is correct.

src/overview.md
mark-i-m: Also, as I understand it, starting codegen also transitively causes monomorphization for the CGU and then generates LLVM IR from the collector's results.

src/overview.md
```diff
@@ -0,0 +1,322 @@
...
+  - Main entry point: **TODO**
+- Code Generation
+  - Guide: [Code Generation](https://rust-lang.github.io/rustc-guide/codegen.html)
+  - Guide: [Generating LLVM IR](https://rust-lang.github.io/rustc-guide/codegen.html#generating-llvm-ir) - **TODO: this is not available yet**
+  - Generating Machine Code from LLVM IR with LLVM - **TODO: reference?**
+  - Main entry point MIR -> LLVM IR: **TODO**
```
bjorn3, line 321: https://github.com/rust-lang/rust/blob/8926bb497d9b127eb318aea5aed0e745d8381591/src/librustc_codegen_llvm/base.rs#L129

src/overview.md
mark-i-m: Actually, is the "entry point" just the `type_of` query?

src/overview.md
chrissimpkins: I *think* so.  I am trying to track down the path from `rustc_interface` to the lexer calls.  https://doc.rust-lang.org/nightly/nightly-rustc/src/rustc_interface/interface.rs.html#207-215 ...

src/overview.md
mark-i-m: Hmmm... this is another place where we talk about interning with defining it.

src/overview.md
eddyb: `type_of` is usually the declared type, `typeck_tables_of` is the entry-point for type-checking *bodies* (e.g. those of functions and consts), and the `tcx.infer_ctxt().enter(...)` thing (you can't decouple those methods btw, although I wonder if they still need to be separate? we removed the `'tcx`/`'gcx` distinction so maybe the bug is fixed) is the entry-point for doing *inference* (which is something that type-checking does but it's also used elsewhere).

src/overview.md
mark-i-m: I added a bit. Let me know what you think.

src/overview.md
mark-i-m: @bjorn3 Is that this: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_codegen_llvm/mono_item/enum.MonoItem.html#method.define ?

src/overview.md
Centril: Well there are a few entry points depending on the situation. For example, the lib.rs file has a different entry point (called by a query in the rustc_interface) than parsing an outline module (called by expansion), and then there are black box parsers for macro fragments. Which do you want?

mark-i-m: Preferably all of them if possible.

chrissimpkins: Does `rustc_interface` call the parser which calls the lexer?  This seems to (maybe) be the case and is what I am expecting to find but I am having a tough time locating the connections beginning at `rustc_interface`.

Centril: > Preferably all of them if possible.

- Entry point for first file in crate is: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_interface/passes/fn.parse.html

- Entry point for outline module parsing is: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_expand/module/fn.parse_external_mod.html

- Entry point for macro fragments is: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_expand/mbe/macro_parser/fn.parse_nt.html (I might move this into the parser to reduce visibilities there.)

- There are more, but these are the most notable ones.

These lead into the parser crate itself, but all of them are outside.

> Does `rustc_interface` call the parser which calls the lexer? This seems to (maybe) be the case and is what I am expecting to find but I am having a tough time locating the connections beginning at `rustc_interface`.

The interface calls a glue function between the lexer and the parser that composes them together for convenience.

src/overview.md
chrissimpkins: Note to self. See https://github.com/rust-lang/rustc-dev-guide/pull/633#discussion_r403515467

chrissimpkins: @petrochenkov Vadim, any guidance on the best lexer entry point for our `rustc` Overview docs?

src/overview.md
bjorn3: > @bjorn3 can you confirm?

`rustc_codegen_ssa::base::codegen_crate` is responsible for calling the right functions to translate everything in a codegen unit to LLVM ir. It then starts a background thread for that codegen unit to optimize it and write the resulting object file. After all threads are started it returns. You then have to call [`rustc_codegen_ssa::back::write::OngoingCodegen::join`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_codegen_ssa/back/write/struct.OngoingCodegen.html#method.join) from [`rustc_codegen_ssa::traits::CodegenBackend::join_codegen](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_codegen_ssa/traits/trait.CodegenBackend.html#tymethod.join_codegen) to join all threads. That function then returns all compiled modules. This split exists to be able to free the `TyCtxt` before all codegen units are optimized and written, potentially reducing max RSS.

> Also, as I understand it, starting codegen also transitively causes monomorphization for the CGU and then generates LLVM IR from the collector's results.

Monomorphization happens lazily. Every time something part of the MIR (eg a statement, or a part of a statement) is codegened, `FunctionCx::monomorphize` is called on that part.

src/overview.md
bjorn3: Yes, but maybe it should refer to [`rustc_codegen_ssa::base::codegen_instance`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_codegen_ssa/base/fn.codegen_instance.html) instead. That function is called by `MonoItem::define`, when the `MonoItem` is a function.

src/overview.md
chrissimpkins: I added Mazdak's links in 5243c7a3f7bd1d0a7b85ee4247219f5581ce2e30 (#669)

src/overview.md
chrissimpkins: Agree with you.  I added a simple statement about Unicode support in 55e61ed8408d41302368683868bc47883e520b4f (#669)

src/overview.md
chrissimpkins: I think that I have the first file in crate full call path worked out from `rustc_interface` -> `parser` entry point, including how the lexer is instantiated and what is called to return a `TokenStream` (it appears to be part of the `Parser` instantiation).  

I'll summarize this for the Overview in #669 and will add more detailed documentation in the Guide chapter.

chrissimpkins: @Centril I added an initial draft of Overview parser documentation in  https://github.com/rust-lang/rustc-dev-guide/pull/669/commits/79c19ab99818a9184e037dad0bb496eac4c0292f

This does not include information about macro expansion, AST validation, nameres, or early linting yet.  How detailed do you feel that we should dive into these topics in the overview?  Mention that they occur and then expand on them in the Guide chapter?

Centril: @chrissimpkins I'd just include the entry point in the overview and then call it a day; most of the additional stuff belongs in a page about the parser imo.

chrissimpkins: > @chrissimpkins I'd just include the entry point in the overview and then call it a day; most of the additional stuff belongs in a page about the parser imo.

Thank you! I will update it today and begin to move the other information over to the Guide chapter.

src/overview.md
chrissimpkins: Aleksey indicated that the lexer entry point is https://github.com/rust-lang/rust/blob/39b62533c7f9d0581a6ea9b9fc2cc51f21c3b5b0/src/librustc_lexer/src/lib.rs#L246

src/overview.md
chrissimpkins: I think that I addressed this as you recommended in https://github.com/rust-lang/rustc-dev-guide/pull/669/commits/b3e75e8b7b32c71e7d97f2e98a262ea279cfed27

src/overview.md
```diff
@@ -0,0 +1,368 @@
...
+  - macro expansion (**TODO** chrissimpkins)
+  - ast validation (**TODO** chrissimpkins)
+  - nameres (**TODO** chrissimpkins)
+  - early linting (**TODO** chrissimpkins)
+
+- We then take the AST and [convert it to High-Level Intermediate
+  Representation (HIR)][hir]. This is a compiler-friendly representation of the
+  AST.  This involves a lot of desugaring of things like loops and `async fn`.
+- We use the HIR to do [type inference]. This is the process of automatic
+  detection of the type of an expression. **TODO: how `ty` module fits in
```
mark-i-m, line 50: I think we can drop this TODO, as we discuss `ty` down below.

src/overview.md
```diff
@@ -0,0 +1,368 @@
...
+  here**
+- **TODO: Maybe some other things are done here? I think initial type checking
+  happens here? And trait solving?**
+- The HIR is then [lowered to Mid-Level Intermediate Representation (MIR)][mir].
+  - Along the way, we construct the HAIR, which is an even more desugared HIR.
+    HAIR is used for pattern and exhaustiveness checking. It is also more
+    convenient to convert into MIR than HIR is.
```
spastorino, line 57: I liked how @eddyb has explained HAIR here https://rust-lang.zulipchat.com/#narrow/stream/182449-t-compiler.2Fhelp/topic/history.20of.20typeck.20and.20borrowck. Maybe @eddyb has more thoughts about the Overview :).

spastorino: Saw that down here https://github.com/rust-lang/rustc-dev-guide/pull/633/files#diff-c872a24fac64fbf63e412eb49ccf468aR175 some of the things said in that Zulip topic were included.

src/overview.md
```diff
@@ -0,0 +1,368 @@
...
+  of tokens produced by the lexer. It represents
+  pretty much exactly what the user wrote. It helps to do some syntactic sanity
+  checking (e.g. checking that a type is expected where the user wrote one).
+- High-level IR (HIR): This is a sort of desugared AST. It's still close
+  to what the user wrote syntactically, but it includes some implicit things
+  such as some elided lifetimes, etc. This IR is amenable to type checking.
+- HAIR: This is an intermediate between HIR and MIR. It is like the HIR but it
+  is fully typed and a bit more desugared (e.g. method calls and implicit
+  dereferences are made fully explicit). Moreover, it is easier to lower to MIR
+  than HIR.
```
spastorino, line 178:
```suggestion
  from HAIR than from HIR.
```

src/overview.md
```diff
@@ -0,0 +1,368 @@
...
+  C compiler also outputs LLVM IR). LLVM IR is designed to be easy for other
+  compilers to emit and also rich enough for LLVM to run a bunch of
+  optimizations on it.
+
+### Queries
+
+The first big implementation choice is the _query_ system. The rust compiler
+uses a query system which is unlike most textbook compilers, which are
+organized as a series of passes over the code that execute sequentially. The
+compiler does this to make incremental compilation possible -- that is, if the
```
spastorino, line 201: Is this the only reason to use a query system? I think it's also a better design.

src/overview.md
```diff
@@ -0,0 +1,368 @@
...
+### Bootstrapping
+
+`rustc` itself is written in Rust. So how do we compile the compiler? We use an
+older compiler to compile the newer compiler. This is called _bootstrapping_.
+
+Bootstrapping has a lot of interesting implications. For example, it means that one
+of the major users of Rust is Rust, so we are constantly testing our own
+software ("eating our own dogfood"). Also, it means building the compiler can
+take a long time because one must first build the compiler and then use that to
+build the new compiler (sometimes you can get away without the full 2-stage
```
spastorino, line 295: I guess people may be confused if they've never thought about this on the why this needs to be built twice. What if without a lot of detail we explain that in the first build you're compiling your program and second build is like you're executing your program.

src/overview.md
mark-i-m: It's a fair question. Incremental was the reason I kept hearing.

src/overview.md
```diff
@@ -0,0 +1,368 @@
...
+build, but for release artifacts you need the 2-stage build).
+
+Bootstrapping also has implications for when features are usable in the
+compiler itself. The build system uses the current beta compiler to build the
+stage-1 bootstrapping compiler. This means that the compiler source code can't
+use some features until they reach beta (because otherwise the beta compiler
+doesn't support them). On the other hand, for compiler intrinsics and internal
```
spastorino, line 302: Hmm, wouldn't people read this and don't understand what intrinsics are?.

src/overview.md
```diff
@@ -0,0 +1,368 @@
...
+features, we may be able to use them immediately because the stage-1
+bootstrapping compiler will support them.
```
spastorino, line 304: Also unsure if this has the right context to be understood.

src/overview.md
mark-i-m: Will do this in a followup ( #320)

src/overview.md
mark-i-m: The explanation in the other chapter makes sense, but it's a bit long. Basically, the point is that we want to end up with a new compiler built with the new compiler. So the first stage is to compile the new compiler with the old compiler, and then we use stage 1 to build the new compiler again.

src/overview.md
mark-i-m: Possibly, but it's a standard compilers term, I think, so we can include it in the glossary perhaps?

src/overview.md
mark-i-m: @chrissimpkins I cherry-picked the commit with the entry points from your branch

src/overview.md
mark-i-m: Added a para about interning and arenas.

src/overview.md
chrissimpkins: > @chrissimpkins I cherry-picked the commit with the entry points from your branch

Ty!

src/overview.md
mark-i-m: cherry-picked this commit.

mark-i-m:
Ok, so as per our discussion in the meeting, we now have a final-ish draft to attempt to merge. Followup work and improvements should be noted on #674. For now, comments and commits on the PR should be limited to fixing incorrect statements.

src/overview.md
spastorino: @mark-i-m yeah, I know but the thing is how to briefly explain this without saying things that readers may not comprehend, hmmm. I kind of like the idea of stage 1 compile your "program", stage 2 execute your "program", the thing is that your "program" is a compiler which is what entangles all this. :/

src/overview.md
spastorino: :+1:, glossary and link from here to the glossary?.

spastorino:
In my opinion this looks already good enough.

src/overview.md
```diff
@@ -1,3 +1,400 @@
...
+in turn in this chapter, and in the rest of the guide, we will look at all the
+individual pieces in more detail.
+
+## What the compiler does to your code
+
+So first, let's look at what the compiler does to your code. For now, we will
+avoid mentioning how the compiler implements these steps except as needed;
+we'll talk about that later.
+
+**TODO: Would be great to have a diagram of this once we nail down the details...**
```
chrissimpkins, line 18: This line should be removed before the merge
```suggestion

```

src/overview.md
```diff
@@ -1,3 +1,400 @@
...
+  during this stage of the compile process.
+- The parser uses the standard `DiagnosticBuilder` API for error handling, but we
+  try to recover, parsing a superset of Rust's grammar, while also emitting an error.
+- `rustc_ast::ast::{Crate, Mod, Expr, Pat, ...}` AST nodes are returned from the parser.
+- We then take the AST and [convert it to High-Level Intermediate
+  Representation (HIR)][hir]. This is a compiler-friendly representation of the
+  AST.  This involves a lot of desugaring of things like loops and `async fn`.
+- We use the HIR to do [type inference]. This is the process of automatic
+  detection of the type of an expression.
+- **TODO: Maybe some other things are done here? I think initial type checking
```
chrissimpkins, line 70: Should be confirmed and removed before merge
```suggestion

```

src/overview.md
```diff
@@ -1,3 +1,400 @@
...
+  happens here? And trait solving?**
+- The HIR is then [lowered to Mid-Level Intermediate Representation (MIR)][mir].
+  - Along the way, we construct the HAIR, which is an even more desugared HIR.
+    HAIR is used for pattern and exhaustiveness checking. It is also more
+    convenient to convert into MIR than HIR is.
+- The MIR is used for [borrow checking].
+- **TODO: const eval fits in somewhere here I think**
```
chrissimpkins, line 77: Should be confirmed and removed before merge
```suggestion

```

src/overview.md
```diff
@@ -1,3 +1,400 @@
...
+- We (want to) do [many optimizations on the MIR][mir-opt] because it is still
+  generic and that improves the code we generate later, improving compilation
+  speed too. (**TODO: size optimizations too?**)
```
chrissimpkins, line 80: Should be removed before merge
```suggestion
  speed too.
```

src/overview.md
```diff
@@ -1,3 +1,400 @@
...
+  - The [code generation stage (codegen)][codegen] is when higher level
+    representations of source are turned into an executable binary. `rustc`
+      uses LLVM for code generation.  The first step is the MIR is then
+    converted to LLVM Intermediate Representation (LLVM IR). This is where
+    the MIR is actually monomorphized, according to the list we created in
+    the previous step.
+  - The LLVM IR is passed to LLVM, which does a lot more optimizations on it.
+    It then emits machine code. It is basically assembly code with additional
+    low-level types and annotations added. (e.g. an ELF object or wasm).
+    **TODO: reference for this section?**
```
chrissimpkins, line 99: Remove before merge
```suggestion

```

src/overview.md
```diff
@@ -1,3 +1,400 @@
...
+  - The different libraries/binaries are linked together to produce the final
+    binary. **TODO: reference for this section?**
```
chrissimpkins, line 101: Remove before merge
```suggestion
    binary.
```

src/overview.md
```diff
@@ -1,3 +1,400 @@
...
+[codegen]: https://rustc-dev-guide.rust-lang.org/backend/codegen.html
+
+## How it does it
+
+Ok, so now that we have a high-level view of what the compiler does to your
+code, let's take a high-level view of _how_ it does all that stuff. There are a
+lot of constraints and conflicting goals that the compiler needs to
+satisfy/optimize for. For example,
+
+- Compilation speed: how fast is it to compile a program. More/better
```
chrissimpkins, line 125: Maybe we can remove "Compilation", "Compiler" from these bullet points?  There is a lot of repetition and I don't know that it needs to be stated in each case?
```suggestion
- Speed: how fast is it to compile a program. More/better
```

src/overview.md
```diff
@@ -1,3 +1,400 @@
...
+  compile-time analyses often means compilation is slower.
+  - Also, we want to support incremental compilation, so we need to take that
+    into account. How can we keep track of what work needs to be redone and
+    what can be reused if the user modifies their program?
+    - Also we can't store too much stuff in the incremental cache because
+      it would take a long time to load from disk and it could take a lot
+      of space on the user's system...
+- Compiler memory usage: while compiling a program, we don't want to use more
```
chrissimpkins, line 133:
```suggestion
- Memory usage: while compiling a program, we don't want to use more
```

src/overview.md
```diff
@@ -1,3 +1,400 @@
...
+  memory than we need.
+- Program speed: how fast is your compiled program. More/better compile-time
+  analyses often means the compiler can do better optimizations.
+- Program size: how large is the compiled binary? Similar to the previous
+  point.
+- Compiler compilation speed: how long does it take to compile the compiler?
```
chrissimpkins, line 139:
```suggestion
- Compilation speed: how long does it take to compile the compiler?
```

src/overview.md
```diff
@@ -1,3 +1,400 @@
...
+  This impacts contributors and compiler maintenance.
+- Compiler implementation complexity: building a compiler is one of the hardest
```
chrissimpkins, line 141:
```suggestion
- Implementation complexity: building a compiler is one of the hardest
```

src/overview.md
```diff
@@ -1,3 +1,400 @@
...
+  things a person/group can do, and Rust is not a very simple language, so how
+  do we make the compiler's code base manageable?
+- Compiler correctness: the binaries produced by the compiler should do what
```
chrissimpkins, line 144:
```suggestion
- Correctness: the binaries produced by the compiler should do what
```

src/overview.md
```diff
@@ -1,3 +1,400 @@
...
+  the input programs says they do, and should continue to do so despite the
+  tremendous amount of change constantly going on.
+- Compiler integration: a number of other tools need to use the compiler in
```
chrissimpkins, line 147:
```suggestion
- Integration: a number of other tools need to use the compiler in
```

src/overview.md
```diff
@@ -1,3 +1,400 @@
...
+  various ways (e.g. cargo, clippy, miri, RLS) that must be supported.
+- Compiler stability: the compiler should not crash or fail ungracefully on the
```
chrissimpkins, line 149:
```suggestion
- Stability: the compiler should not crash or fail ungracefully on the
```

src/overview.md
```diff
@@ -1,3 +1,400 @@
...
+  stable channel.
+- Rust stability: the compiler must respect rust's stability guarantees by not
```
chrissimpkins, line 151:
```suggestion
- Rust stability: the compiler must respect Rust's stability guarantees by not
```

src/overview.md
```diff
@@ -1,3 +1,400 @@
...
+
+### `ty::Ty`
+
+Types are really important in Rust, and they form the core of a lot of compiler
+analyses. The main type (in the compiler) that represents types (in the user's
+program) is [`rustc::ty::Ty`][ty]. This is so important that we have a whole chapter
+on [`ty::Ty`][ty], but for now, we just want to mention that it exists and is the way
+`rustc` represents types!
+
+Oh, and also the `rustc::ty` module defines the `TyCtxt` struct we mentioned before.
```
chrissimpkins, line 286:
```suggestion
Also note that the `rustc::ty` module defines the `TyCtxt` struct we mentioned before.
```

src/overview.md
```diff
@@ -1,3 +1,400 @@
...
+Currently, there is only one part of rustc that is already parallel: codegen.
+During monomorphization, the compiler will split up all the code to be
+generated into smaller chunks called _codegen units_. These are then generated
+by independent instances of LLVM. Since they are independent, we can run them
+in parallel. At the end, the linker is run to combine all the codegen units
+together into one binary.
+
+However, the rest of the compiler is still not yet parallel. There have been
+lots of efforts spent on this, but it is generally a hard problem. The current
+approach is (**TODO: verify**) to turn `RefCell`s into `Mutex`s -- that is, we
```
chrissimpkins, line 305: Verify and remove TODO
```suggestion
approach is to turn `RefCell`s into `Mutex`s -- that is, we
```

src/overview.md
```diff
@@ -1,3 +1,400 @@
...
+compiler itself. The build system uses the current beta compiler to build the
+stage-1 bootstrapping compiler. This means that the compiler source code can't
+use some features until they reach beta (because otherwise the beta compiler
+doesn't support them). On the other hand, for compiler intrinsics and internal
+features, we may be able to use them immediately because the stage-1
+bootstrapping compiler will support them.
+
+# Unresolved Questions
+
+**TODO: find answers to these**
```
chrissimpkins, line 335:
```suggestion

```

src/overview.md
mark-i-m: @spastorino Hmm... I guess, I don't quite see what you are getting at. stage 2 is also compiling no?

src/overview.md
mark-i-m: added to the followup issue

src/overview.md
mark-i-m: added to the followup issue

src/overview.md
mark-i-m: added to the followup issue

src/overview.md
mark-i-m: I think it is important to say that we are talking about the compiler here: do you mean the speed of the compiler or the speed of the compiled program or the speed of compiling the compiler?

Likewise for the others...

src/overview.md
mark-i-m: added to followup issue

src/overview.md
mark-i-m: added to followup issue

src/overview.md
mark-i-m: added to followup issue

src/overview.md
mark-i-m: added to followup issue

src/overview.md
```diff
@@ -1,3 +1,398 @@
...
+  atomic source code units known as _tokens_.  The lexer supports the
+  Unicode character encoding.
+- The token stream passes through a higher-level lexer located in
+  [`librustc_parse`] to prepare for the next stage of the compile process. The
+  [`StringReader`] struct is used at this stage to perform a set of validations
+  and turn strings into interned symbols (_interning_ is discussed later).
+- The lexer has a small interface and doesn't depend directly on the
+  diagnostic infrastructure in `rustc`. Instead it provides diagnostics as plain
+  data which are emitted in `librustc_parse::lexer::mod` as real diagnostics.
+- The lexer preseves full fidelity information for both IDEs and proc macros.
```
JohnTitor, line 39:
```suggestion
- The lexer preserves full fidelity information for both IDEs and proc macros.
```

src/overview.md
```diff
@@ -1,3 +1,398 @@
...
+
+# Unresolved Questions
+
+
+- Does LLVM ever do optimizations in debug builds?
+- How do I explore phases of the compile process in my own sources (lexer,
+  parser, HIR, etc)? - e.g., `cargo rustc -- -Zunpretty=hir-tree` allows you to
+  view HIR representation
+- What is the main source entry point for `X`?
+- Where do phases diverge for cross-compilation to machine  code across
```
LeSeulArtichaut, line 340: There is a trailing space here :eyes:
```suggestion
- Where do phases diverge for cross-compilation to machine code across
```

src/overview.md
```diff
@@ -1,3 +1,398 @@
...
+use some features until they reach beta (because otherwise the beta compiler
+doesn't support them). On the other hand, for [compiler intrinsics][intrinsics]
+and internal features, we may be able to use them immediately because the
+stage-1 bootstrapping compiler will support them.
+
+[intrinsics]: ./appendix/glossary.md#intrinsic
+
+# Unresolved Questions
+
+
```
LeSeulArtichaut, line 334: There are two lines here, unlike the rest of the document.
```suggestion
```

src/overview.md
mark-i-m: I'll make a new PR for these

mark-i-m:
:tada: :tada: :tada:

mark-i-m:
Thanks to everyone who was involved!

Followups should go in #674

PR #659, opened 2020-04-03
Overview: Command line argument parsing, lexer

chrissimpkins:
Work in progress on the Overview document.

Based on [@mark-i-m branch](https://github.com/mark-i-m/rustc-guide/tree/overview) in #633

### DIFF
--- a/src/overview.md
+++ b/src/overview.md
@@ -20,7 +20,7 @@ we'll talk about that later.
 **TODO: someone else should confirm this vvv**
 
 - The compile process begins when a user writes a Rust source program in text and invokes the `rustc` compiler on it. The work that the compiler needs to perform is defined by command-line options. For example, it is possible to enable nightly features (`-Z` flags), perform `check`-only builds, or emit LLVM-IR rather than executable machine code. The `rustc` executable call may be indirect through the use of `cargo`.
-- Command line argument parsing occurs in the [`librustc_driver`]. This crate defines the compile configuration that is requested by the user.
+- Command line argument parsing occurs in the [`librustc_driver`]. This crate defines the compile configuration that is requested by the user and passes it to the rest of the compilation process as a [`rustc_interface::Config`].
 - The raw Rust source text is analyzed by a low-level lexer located in [`librustc_lexer`]. At this stage, the source text is turned into a stream of atomic source code units known as _tokens_. (**TODO**: chrissimpkins - Maybe discuss Unicode handling during this stage?)
 - The token stream passes through a higher-level lexer located in [`librustc_parse`] to prepare for the next stage of the compile process. The [`StringReader`] struct is used at this stage to perform a set of validations and turn strings into interned symbols.
 - (**TODO**: chrissimpkins - Expand info on parser) We then [_parse_ the stream of tokens][parser] to build an Abstract Syntax Tree (AST).
--- a/src/overview.md
+++ b/src/overview.md
@@ -50,7 +50,7 @@ we'll talk about that later.
       doesn't seem to be able to optimize the pattern the [`simplify_try`] mir
       opt looks for.
 - Rust code is _monomorphized_, which means making copies of all the generic
-  code with the type parameters replaced by concrete types. In order to do
+  code with the type parameters replaced by concrete types. To do
   this, we need to collect a list of what concrete types to generate code for.
   This is called _monomorphization collection_.
 - We then begin what is vaguely called _code generation_ or _codegen_.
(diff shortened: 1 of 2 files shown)


chrissimpkins:
@matklad @petrochenkov Mind having a look at the details on the lexers added here for our `rustc-dev-guide` Overview chapter?  Do you have any suggestions about how we can expand upon/clarify the high-level details of the two stages of the lexical analysis?

src/overview.md
```diff
@@ -19,16 +19,16 @@ we'll talk about that later.
...
-- First, we parse command line flags, etc. This is done in [`librustc_driver`].
-  We now know what the exact work is we need to do (e.g. which nightly features
-  are enabled, whether we are doing a `check`-only build or emiting LLVM-IR or
-  a full compilation).
-- Then, we start to do compilation...
-- We first [_lex_ the user program][lex]. This turns the program into a stream
-  of _tokens_ (yes, the same sort of tokens as `proc_macros` (sort of)).
-  [`StringReader`] from [`librustc_parse`] integrates [`librustc_lexer`] with
-  `rustc` data structures.
+- The compile process begins when a user writes a Rust source program in text and invokes the `rustc` compiler on it. The work that the compiler needs to perform is defined with command line options. For example, it is possible to optionally enable nightly features, perform `check`-only builds, or emit LLVM-IR rather than complete the entire compile process defined here. The `rustc` executable call may be indirect through the use of `cargo`.
```
LeSeulArtichaut, line 22: Should we have a soft wrap here?

LeSeulArtichaut, line 22: Few suggestions:
```suggestion
- The compile process begins when a user writes a Rust source program in text and invokes the `rustc` compiler on it. The work that the compiler needs to perform is defined by command-line options. For example, it is possible to enable nightly features (`-Z` flags), perform `check`-only builds, or emit LLVM-IR rather than executable machine code. The `rustc` executable call may be indirect through the use of `cargo`.
```

src/overview.md
```diff
@@ -19,16 +19,16 @@ we'll talk about that later.
...
+- Command line argument parsing occurs in the [`librustc_driver`]. This crate defines the compile configuration that is requested by the user.
```
LeSeulArtichaut, line 23: I'm tempted to add what the configuration is turned into:
```suggestion
- Command line argument parsing occurs in the [`librustc_driver`]. This crate defines the compile configuration that is requested by the user and passes it to the rest of the compilation process as a [`rustc_interface::Config`].
```
(`rustc_interface::Config` is https://doc.rust-lang.org/nightly/nightly-rustc/rustc_interface/interface/struct.Config.html)

But I don't know if this too much of a detail to be included in the Overview...

src/overview.md
```diff
@@ -19,16 +19,16 @@ we'll talk about that later.
...
+- The raw Rust source text is analyzed by a low-level lexer located in [`librustc_lexer`]. At this stage, the source text is turned into a stream of atomic source code units known as _tokens_. (**TODO**: chrissimpkins - Maybe discuss Unicode handling during this stage?)
+- The token stream passes through a higher-level lexer located in [`librustc_parse`] to prepare for the next stage of the compile process. The [`StringReader`] struct is used at this stage to perform a set of validations and turn strings into interned symbols.
```
LeSeulArtichaut, line 25: Did we define interning before? If not, maybe include a link to the chapter about interning?

src/overview.md
tshepang: am happy with the change... it's not too much

src/overview.md
chrissimpkins: Ty!

src/overview.md
chrissimpkins: Agree but don't know this area of the code.  Can you confirm that this is the correct reference?  If not, I'll take a look this weekend and commit your suggestion here.

src/overview.md
chrissimpkins: Open question how we will format the entire document.  I am thinking that maybe this will become paragraph-based body copy rather than a bullet list?

mark-i-m:
Hi everyone: in an effort to avoid merge conflicts, I'm proposing that we merge this into #633 and continue working there. Does anyone have objections to this?

cc @Centril @petrochenkov @matklad  @rust-lang/wg-rustc-dev-guide

src/overview.md
chrissimpkins: @mark-i-m Open issue here if we merge

src/overview.md
chrissimpkins: @mark-i-m Open issue here if we merge

mark-i-m:
Merged in #633 

Thanks all!

chrissimpkins:
Ty!

mark-i-m:
I've copied over the comments from this thread.

PR #665, opened 2020-04-04
remove stupid-stats, and some references to removed API

tshepang:


mark-i-m:
For posterity: this was removed because the API has changed sufficiently that stupid stats is no longer that helpful. Moreover, recently we added some more info about how to use the `rustc_interface`, which is the current API.

PR #673, opened 2020-04-11
codegen tests: add link to LLVM FileCheck docs

RalfJung:


JohnTitor:
Hm, I feel it'd be better a bit to add a sentence like "See LLVM doc" since we might expect that link to point to _the direct Rust content_.

(Btw, Travis is under maintainance and CI won't run: https://www.traviscistatus.com/)

RalfJung:
So, more like this?

JohnTitor:
Yeah, looks good to me!

PR #675, opened 2020-04-17
Start with the basics

badboy:


src/building/how-to-build-and-run.md
```diff
@@ -6,15 +6,23 @@ be hacking on `rustc`, you'll want to tweak the configuration of the compiler.
 The default configuration is oriented towards running the compiler as a user,
 not a developer.
 
+## Get the source code
+
+The very first step to work on `rustc` is to clone the repository:
+
+```text
+git clone https://github.com/rust-lang/rust.git
+cd rust
```
badboy, line 15: Am I right that initializing the submodules is taking care of by `./x.py`, so no manual step required for that?

JohnTitor: Yeah, that's right.

src/building/how-to-build-and-run.md
```diff
@@ -6,15 +6,23 @@ be hacking on `rustc`, you'll want to tweak the configuration of the compiler.
 The default configuration is oriented towards running the compiler as a user,
 not a developer.
 
+## Get the source code
+
+The very first step to work on `rustc` is to clone the repository:
+
+```text
```
JohnTitor, line 13: It'd be nice that we had the same annotation as well as the below.

```suggestion
```bash
```

badboy: Ah right, wanted to comment on that.
I did that first but it gave some weird looking coloring on `clone` (but the rest stayed uncolored, so I opted for `text` instead.

Obviously semantically that's not the right thing, but UI wise it felt better.
But happy to change that back to `bash`

JohnTitor: It's the syntax highlighting for bash, like: https://rustc-dev-guide.rust-lang.org/profiling/with_perf#gathering-a-perf-profile-from-a-perfrust-langorg-test
I think it's fine here since we use it in other places.

PR #682, opened 2020-04-25
Remove info about diag ice breakers; doesn't exist yet

mark-i-m:
cc @spastorino @estebank


LeSeulArtichaut:
There exists a WG-Diagnostics in the rust-lang org, right?

mark-i-m:
@LeSeulArtichaut WG-diagnostics is not the same as the ICE-breakers group, which does not exist

spastorino:
Exactly, anyway it would be great to add a Diagnostics ICE Breaker group. ICE Breakers groups are working great in my opinion and diagnostics is an area that may make great use of it. cc @estebank

PR #689, opened 2020-05-01
Reduce cache timeout to catch broken links faster.

mark-i-m:
It would be nice to catch broken links a bit faster. The tradeoff is that we might end up with more spurious failures.

I propose that we shorten the cache timeout from 2 days to 1 day as an experiment.

Thoughts?


mark-i-m:
cc @rust-lang/wg-rustc-dev-guide ?

mark-i-m:
Ok, let's give this a try, then

PR #693, opened 2020-05-02
Example of getting diagnostics

georgewfraser:
Follow-up of [forum chat](https://internals.rust-lang.org/t/using-the-rustc-interface-crate/11982/20), https://github.com/rust-lang/rust/issues/71770, and [zulip chat](https://rust-lang.zulipchat.com/#narrow/stream/182449-t-compiler.2Fhelp/topic/Is.20rustc_interface.20a.20rustc.20interface.3F)

Using `rustc_session::DiagnosticOutput::Raw` and parsing the resulting JSON seems to be the recommended way to get diagnostics from rustc, so this is a minimal example of how to do that.

The code example is long, but there are a bunch of top-level subtleties, and so I think it is necessary to give a complete working program. If we just give the central snippet, I think future readers will have difficulty figuring out the surrounding context they need to get the program to work.

### DIFF
--- a/src/rustc-driver-getting-diagnostics.md
+++ b/src/rustc-driver-getting-diagnostics.md
@@ -4,100 +4,38 @@
 
 ## Getting diagnostics
 
-NOTE: For the example to compile, you will need to first run the following:
-
-    rustup component add rustc-dev
-
 To get diagnostics from the compiler, 
 configure `rustc_interface::Config` to output diagnostic to a buffer, 
 and run `TyCtxt.analysis`:
 
 ```rust
-#![feature(rustc_private)]
-
-extern crate rustc;
-extern crate rustc_error_codes;
-extern crate rustc_errors;
-extern crate rustc_hash;
-extern crate rustc_hir;
-extern crate rustc_interface;
-extern crate rustc_session;
-extern crate rustc_span;
-
-use rustc_errors::registry;
-use rustc_session::config;
-use rustc_span::source_map;
-use std::io;
-use std::path;
-use std::process;
-use std::str;
-use std::sync;
-
-// Buffer diagnostics in a Vec<u8>.
-#[derive(Clone)]
-pub struct DiagnosticSink(sync::Arc<sync::Mutex<Vec<u8>>>);
-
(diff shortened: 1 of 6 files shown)


georgewfraser:
@mark-i-m @Mark-Simulacrum @bjorn3 FYI

src/rustc-driver-getting-diagnostics.md
```diff
@@ -0,0 +1,103 @@
...
+NOTE: For the example to compile, you will need to first run the following:
+
+    rustup component add rustc-dev
+
+To get diagnostics from the compiler, 
+configure `rustc_interface::Config` to output diagnostic to a buffer, 
+and run `TyCtxt.analysis`:
+
+```rust
+#![feature(rustc_private)]
```
mark-i-m, line 16: I mildly prefer to put the whole file in `src/examples/` as an actual rust file and only quote the relevant parts here? I'm not actually sure what the best way to handle this is...

cc @rust-lang/wg-rustc-dev-guide perhaps we should establish some sort of convention?

mark-i-m: It might also be worthwhile to run CI on the examples to make sure they continue to compile...

georgewfraser:
Let me see if I can come up with a sensible snippet that has enough detail but not too much.

src/rustc-driver-getting-diagnostics.md
spastorino: Is it possible that we place files in `src/examples` and then display them in some way in the guide?. I imagine reading this in a tablet or something and being able to see the example right there in the section of the guide would be nice. @mark-i-m is that what you were suggesting? if so :+1:, I think that could be a nice convention.

chrissimpkins: > It might also be worthwhile to run CI on the examples to make sure they continue to compile...

Good idea!

georgewfraser:
@mark-i-m OK I've refactored the examples a little so there's always a complete example in `examples/` and then a snippet in the docs.

georgewfraser:
Yeah I agree, one thing at a time 😄

PR #697, opened 2020-05-07
Add a bit about various type system concepts

### LINKED ISSUE #177: region name resolution, early vs late bound lifetimes
It'd be nice to make this concept obsolete someday. *Until then*, here is some discussion from Discord on the topic:

[region name resolution](https://discordapp.com/channels/442252698964721669/443113264751181824/476838477770784768), [early vs late](https://discordapp.com/channels/442252698964721669/443113264751181824/476839507983859727)

mark-i-m:
A lot of this is taken straight from Niko's writings here:
- https://gist.github.com/nikomatsakis/fe440aec437f7e8ad4cce8319954a973
- https://rust-lang.zulipchat.com/#narrow/stream/144729-wg-traits/topic/reintegrating.20chalk.20into.20rustc/near/190518417

r? @nikomatsakis 

resolves #177 I think

### DIFF
--- a/src/appendix/background.md
+++ b/src/appendix/background.md
@@ -151,10 +151,10 @@ initialized in (C)). But this is not the case; in particular, `init = false` in
 (A), and `x` is not initialized in (C).  Thus, `init = false` in (C); we can
 report an error that "`x` may not be initialized before use".
 
-There is definitely a lot more that can be said about dataflow analyses, and
+There is definitely a lot more that can be said about dataflow analyses. There is an
 extensive body of research literature on the topic, including a lot of theory.
 We only discussed a forwards analysis, but backwards dataflow analysis is also
-useful. For example, rather than starting from block (A) and moving backwards,
+useful. For example, rather than starting from block (A) and moving forwards,
 we might have started with the usage of `x` and moved backwards to try to find
 its initialization.
 
--- a/src/appendix/background.md
+++ b/src/appendix/background.md
@@ -136,6 +136,8 @@ so that `x` is actually always initialized. It could also be the case that
 initialized. In general, we cannot know statically (due to [Rice's
 Theorem][rice]).  So what should the value of `init` be in block (C)?
 
+[rice]: https://en.wikipedia.org/wiki/Rice%27s_theorem
+
 In this case, we want to be able to prove definitively that `x` must be
 initialized before use. This forces us to be conservative and assume that
 `some_cond` might be false sometimes. Thus, we can only make `init = true` in
--- a/src/appendix/background.md
+++ b/src/appendix/background.md
@@ -120,7 +120,7 @@ A CFG for this code might look like this:
  +---------+
 ```
 
-We can do the dataflow analysis as follows: we will start of with a flag `init`
+We can do the dataflow analysis as follows: we will start off with a flag `init`
 which indicates if we know `x` is initialized. As we walk the CFG, we will
 update the flag. At the end, we can check its value.
 
@@ -129,7 +129,7 @@ So first, in block (A), the variable `x` is declared but not initialized, so
 initialized. So at the end of (B), `init = true`.
 
 Block (C) is where things get interesting. Notice that there are two incoming
-edges from (A) and (B), corresponding to whether `some_cond` is true or not.
+edges, one from (A) and one from (B), corresponding to whether `some_cond` is true or not.
 But we cannot know that! It could be the case the `some_cond` is always true,
 so that `x` is actually always initialized. It could also be the case that
 `some_cond` depends on something random (e.g. the time), so `x` may not be
@@ -188,7 +188,7 @@ bound, it is well-typed: `∀ F, 'a: (F: Fn(&'a u8)) => well_typed(foo)`.
 
 ## What is a DeBruijn Index?
 
-DeBruijn indices are a way of representing the which variables are bound in
+DeBruijn indices are a way of representing which variables are bound in
 which binders using only integers. They were [originally invented][wikideb] for
 use in lambda calculus evaluation. In `rustc`, we use a similar idea for the
 [representation of generic types][sub].
--- a/src/appendix/background.md
+++ b/src/appendix/background.md
@@ -138,12 +138,25 @@ Theorem][rice]).  So what should the value of `init` be in block (C)?
 
 [rice]: https://en.wikipedia.org/wiki/Rice%27s_theorem
 
+Generally, in dataflow analyses, if a block has multiple parents (like (C) in
+our example), its dataflow value will be some function of all it's parents (and
+of course, what happens in (C)).  Which function we use depends on the analysis
+we are doing.
+
 In this case, we want to be able to prove definitively that `x` must be
 initialized before use. This forces us to be conservative and assume that
-`some_cond` might be false sometimes. Thus, we can only make `init = true` in
-(C) if it is true in both (A) and (B), i.e., both parents of (C). Since, `init
-= false` in (A), `init = false` in (C), and we can report to the user that "`x`
-may not be initialized before use".
+`some_cond` might be false sometimes. So our "merging function" is "and". That
+is, `init = true` in (C) if `init = true` in (A) _and_ in (B) (or if `x` is
+initialized in (C)). But this is not the case; in particular, `init = false` in
+(A), and `x` is not initialized in (C).  Thus, `init = false` in (C); we can
+report an error that "`x` may not be initialized before use".
+
+There is definitely a lot more that can be said about dataflow analyses, and
+extensive body of research literature on the topic, including a lot of theory.
+We only discussed a forwards analysis, but backwards dataflow analysis is also
+useful. For example, rather than starting from block (A) and moving backwards,
+we might have started with the usage of `x` and moved backwards to try to find
+its initialization.
 
 <a name="quantified"></a>
 
--- a/src/appendix/background.md
+++ b/src/appendix/background.md
@@ -185,17 +185,25 @@ Another example:
 ```rust,ignore
 fn foo<'a>(_: &'a usize)
 ```
-This function claims that there is some lifetime `'a` (determined by the
-caller) such that it is well-typed: `∃ 'a: well_typed(foo)`.
+This function claims that for any lifetime `'a` (determined by the
+caller), it is well-typed: `∀ 'a: well_typed(foo)`.
+
+Another example:
+
+```rust,ignore
+fn foo<F>()
+where for<'a> F: Fn(&'a u8)
+```
+This function claims that it is well-typed for all types `F` such that for all
+lifetimes `'a`, `F: Fn(&'a u8)`: `∀ F: ∀ 'a: (F: Fn(&'a u8)) => well_typed(foo)`.
 
 One more example:
 
 ```rust,ignore
-fn foo<F>()
-where for<'a> F: Fn(&'a u8),
+fn foo(_: dyn Debug)
 ```
-This function claims that for all lifetimes `'a` and types `F` satisfying the
-bound, it is well-typed: `∀ F, 'a: (F: Fn(&'a u8)) => well_typed(foo)`.
+This function claims that there exists some type `T` that implements `Debug`
+such that the function is well-typed: `∃ T:  (T: Debug) and well_typed(foo)`.
 
 <a name="variance"></a>
 
--- a/src/appendix/background.md
+++ b/src/appendix/background.md
@@ -147,7 +147,40 @@ may not be initialized before use".
 
 ## What is "universally quantified"? What about "existentially quantified"?
 
-*to be written*
+In math, a predicate may be _universally quantified_ or _existentially
+quantified_:
+
+- _Universal_ quantification:
+  - the predicate holds if it is true for all possible inputs.
+  - Traditional notation: ∀x: P(x). Read as "for all x, P(x) holds".
+- _Existential_ quantification:
+  - the predicate holds if there is any input where it is true, i.e., there
+    only has to be a single input.
+  - Traditional notation: ∃x: P(x). Read as "there exists x such that P(x) holds".
+
+In Rust, they come up in type checking and trait solving. For example,
+
+```rust
+fn foo<T>()
+```
+This function claims that the function is well-typed for all types `T`: `∀ T: well_typed(foo)`.
+
+Another example:
+
+```rust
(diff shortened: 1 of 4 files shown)


spastorino:
This looks good!

src/appendix/background.md
```diff
@@ -79,7 +79,69 @@ cycle.
...
+  else    \ +-------+
+    |      \| x = 1 | (B)
+    |       +-------+
+    |      /
+ +---------+
+ | dbg!(x) | (C)
+ +---------+
+```
+
+We can do the dataflow analysis as follows: we will start of with a flag `init`
```
tshepang, line 123:
```suggestion
We can do the dataflow analysis as follows: we will start off with a flag `init`
```

src/appendix/background.md
```diff
@@ -79,7 +79,69 @@ cycle.
...
+which indicates if we know `x` is initialized. As we walk the CFG, we will
+update the flag. At the end, we can check its value.
+
+So first, in block (A), the variable `x` is declared but not initialized, so
+`init = false`. In block (B), we initialize the value, so we know that `x` is
+initialized. So at the end of (B), `init = true`.
+
+Block (C) is where things get interesting. Notice that there are two incoming
+edges from (A) and (B), corresponding to whether `some_cond` is true or not.
```
tshepang, line 132:
```suggestion
edges, one from (A) and one from (B), corresponding to whether `some_cond` is true or not.
```
that looks more clear

src/appendix/background.md
```diff
@@ -79,7 +79,69 @@ cycle.
...
+`some_cond` depends on something random (e.g. the time), so `x` may not be
+initialized. In general, we cannot know statically (due to [Rice's
+Theorem][rice]).  So what should the value of `init` be in block (C)?
+
+In this case, we want to be able to prove definitively that `x` must be
+initialized before use. This forces us to be conservative and assume that
+`some_cond` might be false sometimes. Thus, we can only make `init = true` in
+(C) if it is true in both (A) and (B), i.e., both parents of (C). Since, `init
+= false` in (A), `init = false` in (C), and we can report to the user that "`x`
+may not be initialized before use".
```
tshepang, line 147: I don't understand this last sentence... can you rephrase.

src/appendix/background.md
```diff
@@ -184,6 +184,30 @@ bound, it is well-typed: `∀ F, 'a: (F: Fn(&'a u8)) => well_typed(foo)`.
 
 <a name="variance"></a>
 
+## What is a DeBruijn Index?
+
+DeBruijn indices are a way of representing the which variables are bound in
+which binders using only integers. They were [originally invented][wikideb] for
```
tshepang, line 192:
```suggestion
DeBruijn indices are a way of representing which variables are bound in
which binders using only integers. They were [originally invented][wikideb] for
```

src/early-late-bound.md
```diff
@@ -0,0 +1,122 @@
...
+let x = foo; // T, 'b have to be substituted here
+x(...);      // 'a substituted here, at the point of call
+x(...);      // 'a substituted here with a different value
+```
+
+## Early-bound parameters
+
+Early-bound parameters in Rustc are identified by an index, stored in the
+[`ParamTy`] struct for types or the [`EarlyBoundRegion`] struct for lifetimes.
+The index counts from the outermost declaration in scope. This meansthat as you
```
tshepang, line 68:
```suggestion
The index counts from the outermost declaration in scope. This means that as you
```

src/early-late-bound.md
```diff
@@ -0,0 +1,122 @@
...
+supported, though with GATs that has to change). We indicate their potential
+presence by a [`Binder`] type. The [`Binder`] doesn't know how many variables
+there are at that binding level. This can only be determined by walking the
+type itself and collecting them. So a type like `for<'a, 'b> ('a, 'b)` would be
+`for (^0.a, ^0.b)`. Here, we just write `for` because we don't know the names
+of the things bound within.
+
+Moreover, a reference to a late-bound lifetime is written `^0.a`:
+
+    - The `0` is the index; it identifies that this lifetime is bound in the
```
tshepang, line 104: this renders as a code block... is it intended

src/appendix/background.md
mark-i-m: I rephrased it and expanded a bit. Let me know what you think.

src/appendix/background.md
```diff
@@ -79,16 +79,150 @@ cycle.
...
+
+```rust,ignore
+fn foo<T>()
+```
+This function claims that the function is well-typed for all types `T`: `∀ T: well_typed(foo)`.
+
+Another example:
+
+```rust,ignore
+fn foo<'a>(_: &'a usize)
```
nikomatsakis, line 186: This is also universal quantification -- for any lifetime `'a` that the caller chooses, if you supply a `&'a usize`, `foo` will succeed.

src/appendix/background.md
```diff
@@ -79,16 +79,150 @@ cycle.
...
+```
+This function claims that there is some lifetime `'a` (determined by the
+caller) such that it is well-typed: `∃ 'a: well_typed(foo)`.
```
nikomatsakis, line 189: Existential would be like a `dyn Debug` type, which is kind of `exists<T> { T: Debug }`. i.e., you have a value of some type `T` that is `Debug` (but there is some concrete type `T`).

src/appendix/background.md
```diff
@@ -79,16 +79,150 @@ cycle.
...
+
+One more example:
+
+```rust,ignore
+fn foo<F>()
+where for<'a> F: Fn(&'a u8),
+```
+This function claims that for all lifetimes `'a` and types `F` satisfying the
+bound, it is well-typed: `∀ F, 'a: (F: Fn(&'a u8)) => well_typed(foo)`.
```
nikomatsakis, line 198: Not quite. This function claims that it is well-typed for al types `F` where:

* for all lifetimes `'a`, `F` implements `Fn(&'a u8)`

nikomatsakis:
```
forall<F> {
    if (forall<'a> { F: Fn(&'a u8) }) {
        /* foo::<F> is well typed */
    }
}
```

src/appendix/background.md
```diff
@@ -79,16 +79,150 @@ cycle.
...
+```rust,ignore
+|x| {
+    f(x) // de Bruijn index of `x` is 1 because `x` is bound 1 level up
+
+    |y| {
+        g(x, y) // index of `x` is 2 because it is bound 2 levels up
+                // index of `y` is 1 because it is bound 1 level up
+    }
+}
+```
```
nikomatsakis, line 224: yep

src/early-late-bound.md
```diff
@@ -0,0 +1,122 @@
+# Early and Late Bound Variables
+
+A generic type may be [_universally_ or _existentially_ quantified][quant]. For
```
nikomatsakis, line 3: I'm not sure this is true. All the generics in Rust correspond to universal quantification.

src/early-late-bound.md
```diff
@@ -0,0 +1,122 @@
...
+example,
+
+[quant]: ./appendix/background.md#quantified
+
+```rust,ignore
+fn foo<T>()
+```
+This function claims that the function is well-typed for all types `T`. To use
+the chalk notation: `forall<T> foo<T>`.
```
nikomatsakis, line 12: This seems to duplicate the material above.

src/early-late-bound.md
```diff
@@ -0,0 +1,122 @@
...
+Consider the following example:
+
+```rust,ignore
+fn foo<'a, 'b, T>(x: &'a u32, y: &'b T) where T: 'b { ... }
+```
+
+We cannot treat `'a`, `'b`, and `T` in the same way.  Types in Rust can't have
+`for<T> { .. }`, only `for<'a> {...}`, so whenever you reference `foo` the type
+you get back can't be `for<'a, 'b, T> fn(&'a u32, y: &'b T)`. Instead, the `T`
+must be substituted early. In particular, you have:
```
nikomatsakis, line 56: This seems correct, yes.

src/early-late-bound.md
```diff
@@ -0,0 +1,122 @@
...
+  latter arise from types like `fn(&u32, &u32)`, which are equivalent to
+  something like `for<'a, 'b> fn(&'a u32, &'b u32)`, but the names of those
+  lifetimes must be generated.
+
+This setup of not knowing the full set of variables at a binding level has some
+advantages and some disadvantages. The disadvantage is that you must walk the
+type to find out what is bound at the given level and so forth. The advantage
+is primarily that, when constructing types from Rust syntax, if we encounter
+anonymous regions like in `fn(&u32)`, we just create a fresh index and don't have
+to update the binder.
```
nikomatsakis, line 118: correct

src/early-late-bound.md
nikomatsakis: I'm also not sure how relevant it is here

src/appendix/background.md
```diff
@@ -79,16 +79,150 @@ cycle.
...
+initialized in (C)). But this is not the case; in particular, `init = false` in
+(A), and `x` is not initialized in (C).  Thus, `init = false` in (C); we can
+report an error that "`x` may not be initialized before use".
+
+There is definitely a lot more that can be said about dataflow analyses, and
+extensive body of research literature on the topic, including a lot of theory.
+We only discussed a forwards analysis, but backwards dataflow analysis is also
+useful. For example, rather than starting from block (A) and moving backwards,
+we might have started with the usage of `x` and moved backwards to try to find
+its initialization.
```
nikomatsakis, line 159: seems good!

src/early-late-bound.md
```diff
@@ -0,0 +1,122 @@
...
+
+```rust,ignore
+let x = foo; // T, 'b have to be substituted here
+x(...);      // 'a substituted here, at the point of call
+x(...);      // 'a substituted here with a different value
+```
+
+## Early-bound parameters
+
+Early-bound parameters in Rustc are identified by an index, stored in the
```
tshepang, line 66:
```suggestion
Early-bound parameters in rustc are identified by an index, stored in the
```

src/early-late-bound.md
```diff
@@ -0,0 +1,122 @@
...
+in [this chapter](./generics.md).
+
+[`ParamTy`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.ParamTy.html
+[`EarlyBoundRegion`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.EarlyBoundRegion.html
+[`Generics`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.Generics.html
+
+## Late-bound parameters
+
+Late-bound parameters in `rustc` are handled quite differently (they are also
+specialized to lifetimes, since right now only late-bound lifetimes are
```
tshepang, line 94:
```suggestion
specialized to lifetimes since, right now, only late-bound lifetimes are
```

src/appendix/background.md
```diff
@@ -79,16 +79,150 @@ cycle.
...
+But we cannot know that! It could be the case the `some_cond` is always true,
+so that `x` is actually always initialized. It could also be the case that
+`some_cond` depends on something random (e.g. the time), so `x` may not be
+initialized. In general, we cannot know statically (due to [Rice's
+Theorem][rice]).  So what should the value of `init` be in block (C)?
+
+[rice]: https://en.wikipedia.org/wiki/Rice%27s_theorem
+
+Generally, in dataflow analyses, if a block has multiple parents (like (C) in
+our example), its dataflow value will be some function of all it's parents (and
```
tshepang, line 142:
```suggestion
our example), its dataflow value will be some function of all its parents (and
```

mark-i-m:
@nikomatsakis I think I addressed your comments. Could you take a look at the beginning of the early/late bounds chapter again? (basically the last commit)...

src/early-late-bound.md
```diff
@@ -1,46 +1,13 @@
...
-```txt
-forall<'a, F> {
-    if F: Fn(&'a u8) {
-        foo<'a, f>
-    }
-}
-```
-
+In Rust, generic types are generally [_universally_ quantified][quant].
 Notice, however, that in Rust, we don't have (at the language level)
```
nikomatsakis, line 4: Oops, sorry @mark-i-m for the delay -- but this doesn't seem right to me. The type `for<'a> fn(&'a u32)`, for example, is a universally quantified type, as is `dyn for<'a> Foo<&'a u32>` or something like that. It is true that the quantifiers are attached to specific kinds of types (`fn` types, `dyn` types), and we don't have some general purpose "type quantification" mechanism. 

(As far as early- vs late- splits go, any region that is bound in a type is late-bound.)

mark-i-m: Hi @nikomatsakis, I adjusted this paragraph again in the latest commit, but I'm not sure if I'm just misunderstanding you...

src/early-late-bound.md
```diff
@@ -0,0 +1,90 @@
+# Early and Late Bound Variables
+
+In Rust, generic types are generally [_universally_ quantified][quant].
```
nikomatsakis, line 3: well, I think I would phrase this like so

In Rust, item definitions (like `fn`) can often have generic parameters, which are always [_universally_ quantified][quant]. That is, if you have a function like

```rust
fn foo<T>(x: T) { }
``` 

this function is defined "for all T" (not "for some specific T", which would be existentially quantified).

While Rust *items* can be quantified over types, lifetimes, and constants, the types of values in Rust are only ever quantified over lifetimes. So you can have a type like `for<'a> fn(&'a u32)`, which represents a function pointer that takes a reference with any lifetime, or `for<'a> dyn Trait<'a>`, which is a `dyn` trait for a trait implemented for any lifetime; but we have no type like `for<T> fn(T)`, which would be a function that takes a value of *any type* as a parameter. This is a consequence of monomorphization -- to support a value of type `for<T> fn(T)`, we would need a single function pointer that can be used for a parameter of any type, but in Rust we generate customized code for each parameter type. 

One consequence of this asymmetry is a weird split in how we represesent some generic types: _early-_ and _late-_ bound parameters.

mark-i-m: I replaced the intro with this (let me know if I didn't replace enough). Sorry, I'm only slowly starting to understand this distinction; you're intro helps a lot.

mark-i-m:
Oh, huh? I thought I had already merged it! Sorry!

PR #700, opened 2020-05-10
glossary: make soundness definition more readable

tshepang:


src/appendix/glossary.md
```diff
@@ -67,7 +67,7 @@ rib <div id="rib"/>                      |  A data structure in the name resolve
 sess <div id="sess"/>                    |  The compiler session, which stores global data used throughout compilation
 side tables <div id="side-tables"/>      |  Because the AST and HIR are immutable once created, we often carry extra information about them in the form of hashtables, indexed by the id of a particular node.
 sigil <div id="sigil"/>                  |  Like a keyword but composed entirely of non-alphanumeric tokens. For example, `&` is a sigil for references.
-soundness <div id="soundness"/>          |  A technical term in type theory. Roughly, if a type system is sound, then if a program type-checks, it is type-safe; i.e. I can never (in safe rust) force a value into a variable of the wrong type. (see "completeness").
+soundness <div id="soundness"/>          |  A technical term in type theory. Roughly, if a type system is sound and a program type-checks, it is type-safe. That is, one can never (in safe rust) force a value into a variable of the wrong type. (see "completeness").
```
mark-i-m, line 70:
```suggestion
soundness <div id="soundness"/>          |  A technical term in type theory. Roughly, if a type system is sound, then a program that type-checks is type-safe. That is, one can never (in safe rust) force a value into a variable of the wrong type. (see "completeness").
```

PR #709, opened 2020-05-23
Fix BodyId link typo in glossary

Nashenas88:


chrissimpkins:
Hmmm... CI failing on something unrelated to your changes here Paul.  

https://travis-ci.com/github/rust-lang/rustc-dev-guide/builds/167907317#L338

@mark-i-m OK to merge with this fail?

Nashenas88:
@chrissimpkins I fixed the broken link in this case as well. Seems the enum was split into an enum kind and a wrapping struct. I pointed to the struct for now since it seems to be the main point of use.

LeSeulArtichaut:
I opened #711 to fix the broken link (I didn't see it was fixed here).

LeSeulArtichaut:
> @mark-i-m OK to merge with this fail?

Btw @chrissimpkins I think we can't merge when CI fails because it is marked as "required".

PR #721, opened 2020-05-27
Broken Link to the high-level overview

Michael-F-Bryan:
I'm about to release version `0.6.0` of `mdbook-linkcheck` and found this broken link while testing it on the rustc dev guide. 

Hopefully I'm correct in guessing `high-level-overview.md` was renamed to `overview.md`. If not, let me know and I'll update it accordingly.

### DIFF
--- a/src/query.md
+++ b/src/query.md
@@ -10,7 +10,7 @@ the [def-id] of some item, will compute the type of that item and return
 it to you.
 
 [def-id]: appendix/glossary.md#def-id
-[hl]: high-level-overview.html
+[hl]: compiler-src.html
 
 Query execution is **memoized** – so the first time you invoke a
 query, it will go do the computation, but the next time, the result is


Michael-F-Bryan:
This version has switched to a different backend for linkchecking (the [linkcheck](https://crates.io/crates/linkcheck) crate) which uses async IO to check links concurrently.

While testing it locally I got a lot of *429 Too Many Requests* errors from GitHub. The rustc guide has a lot more GitHub links than your average book, have you seen this before?

src/query.md
```diff
@@ -10,7 +10,7 @@ the [def-id] of some item, will compute the type of that item and return
 it to you.
 
 [def-id]: appendix/glossary.md#def-id
-[hl]: high-level-overview.html
+[hl]: overview.html
```
JohnTitor, line 13: Given from git blame, I think it should be https://github.com/rust-lang/rustc-dev-guide/blob/master/src/compiler-src.md.

JohnTitor:
> While testing it locally I got a lot of 429 Too Many Requests errors from GitHub. The rustc guide has a lot more GitHub links than your average book, have you seen this before?

Yeah, we've seen that failures on toolstate check as well and currently we don't have any workaround iirc :/

Michael-F-Bryan:
My approach was to run the linkcheck multiple times. That way it'll cache the first X requests and after two or three runs it'll work its way through all GitHub links.

I'm guessing you could generate an API token and use mdbook-linkcheck's variable interpolation to attach a `Authorization: Basic $API_TOKEN` header to hopefully get a higher rate limit. But then because `mdbook-linkcheck` doesn't have a way of saying *"only set this header when interpolation succeeds"*, *everyone* wanting to develop the guide would need to have a token and set the `$API_TOKEN` variable.

JohnTitor:
Sounds good! Let's continue the discussion about the 429 failures on Zulip: https://rust-lang.zulipchat.com/#narrow/stream/196385-t-compiler.2Fwg-rustc-dev-guide/topic/toolstate

PR #729, opened 2020-05-29
[WIP] Document remote testing

tblah:
Document how to run tests on remote machines (which may be emulated).

This provides accompanying documentation for https://github.com/rust-lang/rust/pull/72704/, and should remain WIP until that is merged. 

r? @ehuss


LeSeulArtichaut:
@mark-i-m You merged this before rust-lang/rust#72704 landed, is that intentional?

mark-i-m:
Ah, I wasn't paying attention to that! Sorry!

In any case, it looks like that PR will be merged soon too, so I think we can just leave it.

PR #730, opened 2020-06-02
Playing with travis images

mark-i-m:
I want to see if we can update to ubuntu 20.04 (the latest LTS). Perhaps that will solve some of the cert issues?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: rust
+dist: focal
 rust:
 - nightly
 cache:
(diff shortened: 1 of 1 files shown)


JohnTitor:
IIRC we cannot use 20.04 on travis with `dist: focal`.

mark-i-m:
Hmm... that might be because they fixed their certs though... Let me try rerunning the other CI...

mark-i-m:
Nope! Indeed, it seems to be this PR that fixes it. I'm going to go ahead and merge.

PR #737, opened 2020-06-05
Do a rebase before link checking

mark-i-m:
The goal here is that people won't have to rebase over the latest master to get link fixes; instead, travis will automatically rebase on master before testing so that the latest fixes are automatically pulled in.

cc @rust-lang/wg-rustc-dev-guide This is unconventional to do in a CI tmk, so let me know if anyone sees problems here.

One disadvantage is that if there are conflicts, the build will fail, but I think that is acceptable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ install:
 - cargo install mdbook --version 0.3.7
 - cargo install mdbook-linkcheck --version 0.5.0
 script:
+- git rebase master # this will exit with an error on conflict
 - mdbook build
 notifications:
   email:


LeSeulArtichaut:
I'm in favor of this change. It has happened quite often that PRs had to be rebased on master, which isn't necessary otherwise. Allowing CI to rebase things itself could spare us some time trying to navigate through CI logs to find out that the error is unrelated to the PR and has been fixed already 😄

> This is unconventional to do in a CI tmk

IIUC @bors merges PRs before testing them? So it doesn't seem so unconventional to me. Of course, the CI testing is much more important on `rust-lang/rust`, but rebasing should be pretty cheap, since the commits usually aren't very large.

chrissimpkins:
Workflow will be that, if an automated rebase leads to a conflict, it fails and author rebases locally to resolve it?

mark-i-m:
@chrissimpkins Yes, that's already the case; if there is a conflict, someone needs to manually resolve it. AFAICT, this change shouldn't affect anyone's workflow, but it allows us to just restart a build rather than wait for the author to rebase.

chrissimpkins:
LGTM! 👍

PR #739, opened 2020-06-09
Rebrand ICE-breakers, introduce Windows notification group

nikomatsakis:
Part of https://github.com/rust-lang/compiler-team/issues/282

r? @spastorino


src/notification-groups/llvm.md
```diff
@@ -1,10 +1,10 @@
-# LLVM ICE-breakers
+# LLVM Notification group
 
-**Github Label:** [ICEBreaker-LLVM]
+**Github Label:** [A-LLVM]
```
LeSeulArtichaut, line 3: What about the `ICEBreaker-Cleanup-Crew` label? Will it stay the same? If not we should edit the reference to it in the `cleanup-crew.md`.

nikomatsakis: I guess we should just change it to Cleanup-Crew

spastorino: I've merged this with the idea that the labels can be fixed in a follow-up PR.

PR #741, opened 2020-06-09
complete the ICE-breaker => notification group rename

nikomatsakis:


nikomatsakis:
@mark-i-m I was hoping to create an ICE-breaker group after some time has passed, that is truly just focused on "ICEs"

PR #731, opened 2020-06-02
Getting started guide

mark-i-m:
cc @rust-lang/wg-rustc-dev-guide @rust-lang/compiler @rust-lang/libs (not sure if I missed any other teams)

The eventual goal is to replace CONTRIBUTING.md and the README.md in rust-lang/rust with pointers to this document.

This is still very WIP, but it aims to address some of the concerns that WG-rustc-dev-guide was feeling with respect to organization, as well as some of the concerns that have come up in the contributor survey.

Feedback is welcome, especially on the structure and content of the document (e.g. if you feel something else should be added, or something should be removed)! It still pretty early, so I expect that this will change a bit before things settle.

Also, please excuse me if there is some delay in responding.

PR #753, opened 2020-06-17
Move CONTRIBUTING.md to rustc-dev-guide

mark-i-m:
cc https://github.com/rust-lang/compiler-team/issues/296 @ecstatic-morse @rust-lang/compiler 

This PR moves CONTRIBUTING.md to the rust-dev-guide. It should not merge before #731. The objective is to point all new contributors in one place. In the contributor survey, a surprisingly high number of people (more than half) pointed to CONTRIBUTING.md as a major resource they use to contribute to the compiler, whereas the new "Getting Started" guide is likely a better place to start.

That said, the content in CONTRIBUTING.md is still useful, so we move it here and take the opportunity to also move some other contributing-related chapters together in the guide.

Feedback is welcome.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -28,6 +28,7 @@
 
 # Contributing to Rust
 
+- [Introduction](./contributing.md)
 - [About the compiler team](./compiler-team.md)
 - [Walkthrough: a typical contribution](./walkthrough.md)
 - [Bug Fix Procedure](./bug-fix-procedure.md)
--- a/src/contributing.md
+++ b/src/contributing.md
@@ -306,7 +306,7 @@ Caused by:
 failed to run: ~/rust/build/x86_64-unknown-linux-gnu/stage0/bin/cargo build --manifest-path ~/rust/src/bootstrap/Cargo.toml
 ```
 
-The [`[patch]`][patchsec] section of `Cargo.toml` can be very useful for
+The [`\[patch\]`][patchsec] section of `Cargo.toml` can be very useful for
 testing. In addition to that, you should read the [Overriding
 dependencies][overriding] section of the documentation.
 
--- a/src/contributing.md
+++ b/src/contributing.md
@@ -374,11 +374,14 @@ You can also use `rustdoc` directly to check small fixes. For example,
 `rustdoc src/doc/reference.md` will render reference to `doc/reference.html`.
 The CSS might be messed up, but you can verify that the HTML is right.
 
-Additionally, contributions to the [rustc-dev-guide] are always welcome. Contributions
-can be made directly at [the
-rust-lang/rustc-dev-guide](https://github.com/rust-lang/rustc-dev-guide) repo. The issue
-tracker in that repo is also a great way to find things that need doing. There
-are issues for beginners and advanced compiler devs alike!
+Additionally, contributions to the [rustc-dev-guide][rdg] are always welcome.
+Contributions can be made directly at [the rust-lang/rustc-dev-guide
+repo][rdgrepo]. The issue tracker in that repo is also a great way to find
+things that need doing. There are issues for beginners and advanced compiler
+devs alike!
+
+[rdg]: https://rustc-dev-guide.rust-lang.org/
+[rdgrepo]: https://github.com/rust-lang/rustc-dev-guide
 
 ## Issue Triage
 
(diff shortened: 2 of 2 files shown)


src/SUMMARY.md
```diff
@@ -16,24 +15,26 @@
...
-    - [Stabilizing Features](./stabilization_guide.md)
     - [Debugging the Compiler](./compiler-debugging.md)
     - [Profiling the compiler](./profiling.md)
         - [with the linux perf tool](./profiling/with_perf.md)
-    - [Coding conventions](./conventions.md)
     - [crates.io Dependencies](./crates-io.md)
     - [Errors and Lints](diagnostics.md)
         - [`LintStore`](./diagnostics/lintstore.md)
         - [Diagnostic Codes](./diagnostics/diagnostic-codes.md)
+- [Part 1.5: Contributing to Rust](./contributing.md)
```
nikomatsakis, line 25: I think this shows that incorporating "part numbers" into the URL was probably a mistake... not sure of the best way to fix. Maybe just let the part numbers diverge from the URLs? Maybe bite the bullet and break? Anyway, I think that this naming scheme is good as a start.

mark-i-m: > Maybe bite the bullet and break?

I'm tempted to just do this (but perhaps in another PR). I don't think people tend to link to the part intro chapters, though I might be wrong.

Also, mdbook 0.4 will have support for [breaking the book into parts](https://github.com/rust-lang/mdBook/pull/1171) and doing automatic numbering, so this will not be as much of a problem...

src/contributing.md
```diff
@@ -0,0 +1,499 @@
...
+Contributors with sufficient permissions on the Rust repo can help by adding
+labels to triage issues:
+
+* Yellow, **A**-prefixed labels state which **area** of the project an issue
+  relates to.
+
+* Magenta, **B**-prefixed labels identify bugs which are **blockers**.
+
+* Dark blue, **beta-** labels track changes which need to be backported into
+  the beta branches.
```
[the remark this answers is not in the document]
mark-i-m: I just copy/pasted the document with very minor edits. I don't know much about triage/tagging, so it might be out of date. Feel free to propose additions.

src/contributing.md
```diff
@@ -0,0 +1,506 @@
...
+*(This error may change in the future to include more information.)*
+```
+error: failed to resolve patches for `https://github.com/rust-lang/rustfmt`
+
+Caused by:
+  patch for `rustfmt-nightly` in `https://github.com/rust-lang/rustfmt` did not resolve to any crates
+failed to run: ~/rust/build/x86_64-unknown-linux-gnu/stage0/bin/cargo build --manifest-path ~/rust/src/bootstrap/Cargo.toml
+```
+
+The [`[patch]`][patchsec] section of `Cargo.toml` can be very useful for
```
jyn514, line 309: CI is failing:
```suggestion
The [`\[patch\]`][patchsec] section of `Cargo.toml` can be very useful for
```
```
error: Potential incomplete link

     ┌── contributing.md:309:7 ───

     │

 309 │ The [`[patch]`][patchsec] section of `Cargo.toml` can be very useful for

     │       ^^^^^^^ Did you forget to define a URL for `patch`?

     │

     = hint: declare the link's URL. For example: `[patch]: http://example.com/`
```

mark-i-m: Thanks! I was wondering how to fix that...

src/contributing.md
```diff
@@ -0,0 +1,506 @@
...
+[overriding]: http://doc.crates.io/specifying-dependencies.html#overriding-dependencies
+
+Specifically, the following [section in Overriding dependencies][testingbugfix]
+reveals what the problem is:
+
+[testingbugfix]: http://doc.crates.io/specifying-dependencies.html#testing-a-bugfix
+
+> Next up we need to ensure that our lock file is updated to use this new
+> version of uuid so our project uses the locally checked out copy instead of
+> one from crates.io. The way [patch] works is that it'll load the dependency
```
jyn514, line 323:
```suggestion
> one from crates.io. The way `[patch]` works is that it'll load the dependency
```

src/contributing.md
```diff
@@ -320,7 +320,7 @@ reveals what the problem is:
 
 > Next up we need to ensure that our lock file is updated to use this new
 > version of uuid so our project uses the locally checked out copy instead of
-> one from crates.io. The way [patch] works is that it'll load the dependency
+> one from crates.io. The way `\[patch\]` works is that it'll load the dependency
```
jyn514, line 323: You only need one or the other, this will render like this:

`\[patch\]`

mark-i-m: Hmm... I remember when I tried this locally it broke :/ I'll have to take a look again later...

mark-i-m: I am now unable to reproduce the original error... looks like it was fixed in the latest version of mdbook?

PR #757, opened 2020-06-18
Fix master branch links

mark-i-m:


mark-i-m:
Ok, so it looks like the problem is #737 ... something about travis's setup is weird.

mark-i-m:
ping @rust-lang/wg-rustc-dev-guide It would be great to get this fixed, as it is block master from being rebuilt...

src/building/prerequisites.md
```diff
@@ -0,0 +1,3 @@
+# Prerequesites
```
tshepang, line 1:
```suggestion
# Prerequisites
```

Issue #748, opened 2020-06-15
Add info about IDs in the compiler


mark-i-m:
We should add the following content about IDs:

ID memo:
- `LocalDefId` - the best variant, if you can use it for something do it. Cannot address entities in other crates and fine-grained entities like individual expressions in the current crate. Can address any items.
- `DefId` - the second choice. Can address entities in other crates, but cannot address fine-grained entities like individual expressions.
- `HirId` - Can address fine-grained entities like individual expressions in the current crate. Fine-grained entities from other crates can never be addressed in rustc.
- `NodeId` - The last choice. Can address fine-grained entities like individual expressions when `HirId`s are not yet created.

_Originally posted by @petrochenkov in https://github.com/rust-lang/rust/pull/73291#issuecomment-643829018_

PR #759, opened 2020-06-21
Update HIR chapter to reflect the removal of `NodeId`s

LeSeulArtichaut:
See rust-lang/rust#50928.

cc @marmeladema @zackmdavis can you confirm that everything is correct? I haven't really followed the `HirId`-ification initiative, and I just wrote what I could find in the docs and what I could guess.

### DIFF
--- a/src/hir.md
+++ b/src/hir.md
@@ -88,7 +88,7 @@ sorts of identifiers in active use:
     expected, because it prevents [`DefId`] from upstream crates from
     being passed instead and causing bugs at compile time.
   - They can still be transformed back into `DefId`s as needed by using
-    the [`LocalDefId::to_def_id`] method.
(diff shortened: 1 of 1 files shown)


marmeladema:
Documentation seems correct to me. One note though, `NodeId` are not entirely removed, they remain available / accessible before HIR lowering because that's how each AST nodes are identified. Most notably, they are heavily relied upon during name resolution passes.

mark-i-m:
@LeSeulArtichaut Would you be willing to add a new chapter about identifiers throughout the compiler and move generic info there (and maybe also stuff like #748)?

LeSeulArtichaut:
Sure, will take a look :)

mark-i-m:
Friendly ping @LeSeulArtichaut

LeSeulArtichaut:
Closing in favor of #872

PR #767, opened 2020-06-24
Move Errors and Lints to Analysis

mark-i-m:
This seems to fit better with analysis IMHO.

Thoughts?


mark-i-m:
We haven't received any new comments here, so I will go ahead and merge.

Thanks @JohnTitor !

PR #768, opened 2020-06-24
Move Getting Started to a "prefix chapter"

mark-i-m:
Getting started feels more general than just "Building and debugging" ...

Thoughts?


mark-i-m:
We haven't received any new comments here, so I will go ahead and merge.

Thanks @JohnTitor !

PR #773, opened 2020-06-27
add stub inline asm chapter

mark-i-m:


JohnTitor:
This is intend to merge as TBD, isn't it? If so r=me, it's a good entry point.

mark-i-m:
@JohnTitor Yep, that was the intent.

PR #780, opened 2020-07-02
Suggest selectively disabling debug and optimizations

Nadrieril:
As discussed on [Zulip](https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/incremental.20compilation.20question/near/202742715). Those suggestions are not mine and I can't personally judge if they are sensible. I linked to their sources.

cc @mark-i-m

### DIFF
--- a/src/building/suggested.md
+++ b/src/building/suggested.md
@@ -87,6 +87,22 @@ You can also use `--keep-stage 1` when running tests. Something like this:
 - Initial test run: `./x.py test -i --stage 1 src/test/ui`
 - Subsequent test run: `./x.py test -i --stage 1 src/test/ui --keep-stage 1`
 
+## Fine-tuning optimizations
+
+Setting `optimize = false` makes the compiler too slow for tests. However, to
+improve the test cycle, you can disable optimizations selectively only for the
+crates you'll have to rebuild
+([source](https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/incremental.20compilation.20question/near/202712165)).
+For example, when working on `rustc_mir_build`, the `rustc_mir_build` and
+`rustc_driver` crates take the most time to incrementally rebuild. You could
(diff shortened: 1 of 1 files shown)


src/building/suggested.md
```diff
@@ -87,6 +87,32 @@ You can also use `--keep-stage 1` when running tests. Something like this:
 - Initial test run: `./x.py test -i --stage 1 src/test/ui`
 - Subsequent test run: `./x.py test -i --stage 1 src/test/ui --keep-stage 1`
 
+## Fine-tuning debug and optimizations
+
+By default, `config.toml` sets `debug = true`. Unless you plan to run a
```
mark-i-m, line 92: I think the default is `debug = false`, right?

Also, it might make sense to update the following references with the new advice, if you don't mind:
```
src/compiler-debugging.md
48:If you want line numbers for the stack trace, you can enable `debug = true` in
50:line numbers, but `debug = true` gives full debuginfo). Then the backtrace will

src/getting-started.md
129:- `debug = true`: enables debug symbols and `debug!` logging, takes a bit longer to compile.

src/building/how-to-build-and-run.md
46:# `debug-assertions = true`. Note that `debug = true` will make your build
49:debug = true
```

Nadrieril: Hmm, that would be updating a significant chunk of advice, but I have no clue if what I suggest is correct. I would need approval by someone who knows that stuff before merging that

Nadrieril: At this point I'm inclined to remove the first paragraph of advice and leave only the thing about crate-wise optimizations

mark-i-m: @Nadrieril Sure, that makes sense. There will be more updates to the recommendations and x.py defaults soon, I think...

Nadrieril: I did that

PR #781, opened 2020-07-03
Minor updates to stage docs.

### LINKED ISSUE #431: libtest build step has been merged with that of libstd

ehuss:
The libtest and libstd steps were merged a long time ago, so update the image for that.

Also, try to clarify some differences between stage1/2.

Fixes #431


JohnTitor:
The change itself looks good to me. #434 updates the wording but it didn't do the diagram. As a bonus-point, it'd be great if the table on the bootstrapping chapter mentions `libtest` as well as `libstd`. Then I think this should close #431.

ehuss:
Ok, I added `libtest` to the table.  (Also fixed the `?` about libstd in stage2.)

PR #793, opened 2020-07-13
Fix the difference of title in appendix A

giraffate:
Left side is `Background material` but title is `Background topics`, so I fix it to `Background topics`. ( If `Background material` is preferrable, I fix it like that.)

スクリーンショット 2020-07-13 12 09 43


giraffate:
Thanks!

PR #795, opened 2020-07-14
Resync config.toml and suggest better settings

spastorino:


src/building/how-to-build-and-run.md
```diff
@@ -30,23 +30,57 @@ settings (and possibly others, such as `llvm.ccache`):
 
 ```toml
 [llvm]
-# Enables LLVM assertions, which will check that the LLVM bitcode generated
-# by the compiler is internally consistent. These are particularly helpful
-# if you edit `codegen`.
+# Indicates whether the LLVM assertions are enabled or not
 assertions = true
```
jyn514, line 34: Why do we recommend this? I don't expect most new contributors to be editing codegen and it will slow down the compiler.

src/building/how-to-build-and-run.md
```diff
@@ -30,23 +30,57 @@ settings (and possibly others, such as `llvm.ccache`):
...
+#
+# Defaults to 2 if debug is true
+debuginfo-level = 1
+
+# Whether to always use incremental compilation when building rustc
+incremental = true
+
+# Emits extraneous output from tests to ensure that failures of the test
+# harness are debuggable just from logfiles.
+verbose-tests = true
```
jyn514, line 83:
```suggestion
# Emits extra output from tests so test failures are debuggable just from logfiles.
verbose-tests = true
```
Extraneous means it's not useful or irrelevant.

spastorino: @jyn514 :+1: but consider changing this on rust-lang/rust, I've just copied what it's there already.

src/building/how-to-build-and-run.md
spastorino: :+1: I think you're right here, I've just left it there because it was there already but given that we are fixing suggestions it may be a good idea to remove it.

What others think about this? why was this originally added?.

spastorino: Discussed here https://rust-lang.zulipchat.com/#narrow/stream/196385-t-compiler.2Fwg-rustc-dev-guide/topic/meeting.2007.2E14.2E2020

src/building/how-to-build-and-run.md
```diff
@@ -30,23 +30,56 @@ settings (and possibly others, such as `llvm.ccache`):
...
 [llvm]
-# Enables LLVM assertions, which will check that the LLVM bitcode generated
-# by the compiler is internally consistent. These are particularly helpful
-# if you edit `codegen`.
+# Indicates whether the LLVM assertions are enabled or not
 assertions = true
 
 [rust]
-# This will make your build more parallel; it costs a bit of runtime
-# performance perhaps (less inlining) but it's worth it.
```
eddyb, line 40: Was this from `config.toml.example`? It looks like these two lines explained why the `= 0` value was set, and this was lost.

src/building/how-to-build-and-run.md
```diff
@@ -30,23 +30,56 @@ settings (and possibly others, such as `llvm.ccache`):
...
+# and needs to be enabled separately with `debuginfo-level-tests`.
+#
+# Defaults to 2 if debug is true
+debuginfo-level = 1
+
+# Whether to always use incremental compilation when building rustc
+incremental = true
+
+# Emits extra output from tests so test failures are debuggable just from logfiles.
+verbose-tests = true
```
eddyb, line 82: Why was this added? The default is that failures are listed at the end, without noisy output while running the tests.

src/building/how-to-build-and-run.md
```diff
@@ -30,23 +30,56 @@ settings (and possibly others, such as `llvm.ccache`):
...
+# `2` - full debug info with variable and type information
+# Can be overridden for specific subsets of Rust code (rustc, std or tools).
+# Debuginfo for tests run with compiletest is not controlled by this option
+# and needs to be enabled separately with `debuginfo-level-tests`.
+#
+# Defaults to 2 if debug is true
+debuginfo-level = 1
+
+# Whether to always use incremental compilation when building rustc
+incremental = true
```
eddyb, line 79: According [to Cargo documentation](https://doc.rust-lang.org/nightly/cargo/reference/profiles.html#codegen-units), the incremental default for `codegen-units` is 256, which makes `codegen-units = 0` unnecessary/undesirable (unless you're compiling Rust on a machine with more than 256 hardware cores/threads, heh).

So my understanding is that when adding `incremental = true`, we also need to remove `codegen-units = 0`. cc @Mark-Simulacrum

src/building/how-to-build-and-run.md
eddyb: This change doesn't seem right, specifically "test harness" refers to *the code running the tests*, rather than the tests themselves. And it is extraneous for the common case of e.g. `libtest` not crashing.

src/building/how-to-build-and-run.md
jyn514: Yes, this is from `config.toml.example`. I'll add it back in https://github.com/rust-lang/rust/pull/76141.

PR #797, opened 2020-07-14
Suggest system llvm or assertions

mark-i-m:
cc @spastorino @jyn514

### DIFF
--- a/src/building/how-to-build-and-run.md
+++ b/src/building/how-to-build-and-run.md
@@ -94,7 +94,7 @@ assertions = true
 If you are _not_ planning to work on LLVM-related things, you can avoid
 building LLVM altogether by using a system LLVM:
 
-```
+```toml
 [rust]
 # Path to the `llvm-config` binary of the installation of a custom LLVM to link
 # against. Note that if this is specified we don't compile LLVM at all for this
--- a/src/building/how-to-build-and-run.md
+++ b/src/building/how-to-build-and-run.md
@@ -26,13 +26,10 @@ cp config.toml.example config.toml
 ```
 
 Then you will want to open up the file and change the following
-settings (and possibly others, such as `llvm.ccache`):
+settings (and possibly others, such as `llvm.ccache`). Also, see the notes
+below about LLVM-related settings.
 
 ```toml
-[llvm]
-# Indicates whether the LLVM assertions are enabled or not
-assertions = true
-
 [rust]
 # Indicates that the build should be configured for debugging Rust. A
 # `debug`-enabled compiler and standard library will be somewhat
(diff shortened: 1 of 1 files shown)


src/building/how-to-build-and-run.md
```diff
@@ -82,6 +79,33 @@ incremental = true
 verbose-tests = true
 ```
 
+Moreover, there are a couple of other configs to consider, depending on whether
+you are planning to work on something LLVM-related.
+
+If you are planning to work on LLVM-related things, you may want to enable
+assertions for improved debugging. This comes at a cost in compile time:
```
jyn514, line 86: It would be great to know how _much_ of a cost :/ but we've already discussed that we don't know. Other than that this looks great.

src/building/how-to-build-and-run.md
```diff
@@ -82,6 +79,33 @@ incremental = true
...
+```toml
+[llvm]
+# Indicates whether the LLVM assertions are enabled or not.
+assertions = true
+```
+
+If you are _not_ planning to work on LLVM-related things, you can avoid
+building LLVM altogether by using a system LLVM:
+
+```
```
JohnTitor, line 97:
```suggestion
```toml
```

src/building/how-to-build-and-run.md
```diff
@@ -26,13 +26,10 @@ cp config.toml.example config.toml
...
 
 Then you will want to open up the file and change the following
-settings (and possibly others, such as `llvm.ccache`):
+settings (and possibly others, such as `llvm.ccache`). Also, see the notes
+below about LLVM-related settings.
 
 ```toml
-[llvm]
-# Indicates whether the LLVM assertions are enabled or not
-assertions = true
```
spastorino, line 34: I remember talking with @nikomatsakis about this and him suggesting to leave this in. I'm also not sure if removing this is a better suggestion.

src/building/how-to-build-and-run.md
```diff
@@ -82,6 +79,33 @@ incremental = true
...
+assertions for improved debugging. This comes at a cost in compile time:
+
+```toml
+[llvm]
+# Indicates whether the LLVM assertions are enabled or not.
+assertions = true
+```
+
+If you are _not_ planning to work on LLVM-related things, you can avoid
+building LLVM altogether by using a system LLVM:
```
spastorino, line 95: I also wonder how good this suggestion is. If I do some changes let say on MIR and then try to run ui tests which compile and compare the output, couldn't using a different and unpatched LLVM version give problems?.

jyn514: Isn't MIR above the LLVM level? Why would it be affected by the llvm version?

spastorino: Maybe my example wasn't clear, I've just used MIR as an example my question is more general. By using some random LLVM system version that may not match the LLVM rustc version and do not have the rustc patches, are we sure rustc tests are going to pass? in particular ui tests.

src/building/how-to-build-and-run.md
nikomatsakis: I personally find this useful. It does make builds and bootstraps take longer, though, and it's often not relevant -- but when you're debugging changes to codegen or other LLVM-related problems, it's invaluable (and people often forget to enable it at that time).

src/building/how-to-build-and-run.md
mark-i-m: I think the UI tests should always pass, right? Most of them are long before LLVM runs. I would expect it to only matter for codegen. Is there a particular example you are thinking of?

src/building/how-to-build-and-run.md
mark-i-m: Sorry, looking over this PR again after a while. We add the suggestion to turn this on below. Are you suggesting that we suggest always turning it on (and if they use system LLVM it doesn't matter)?

mark-i-m:
Does anyone remember what the status of this PR is?

jyn514:
It's waiting on a response to https://github.com/rust-lang/rustc-dev-guide/pull/797#discussion_r464139356 from @nikomatsakis I think. Although in general it might not be necessary after https://github.com/rust-lang/rust/pull/76349.

spastorino:
Maybe even after https://github.com/rust-lang/rust/pull/76349 we would need to check and unsuggest system llvm?.

mark-i-m:
Hmm... yeah it probably makes sense to close this PR then.

PR #798, opened 2020-07-15
Change `rustc::*` to `rustc_middle::*`

giraffate:
Now it looks like `rustc_middle::*` is used not `rustc::*`.
I didn't change everything. For example, I didn't know if I should change `src/compiler-debugging.md` because it includes the backtrace message.

### DIFF
--- a/src/diagnostics/lintstore.md
+++ b/src/diagnostics/lintstore.md
@@ -15,7 +15,7 @@ Unfortunately, a lot of the documentation we have refers to both of these as jus
 
 First, we have the lint declarations themselves: this is where the name and default lint level and
 other metadata come from. These are normally defined by way of the [`declare_lint!`] macro, which
-boils down to a static with type `&rustc_middle::lint::Lint`. We lint against direct declarations
+boils down to a static with type `&rustc_session::lint::Lint`. We lint against direct declarations
 without the use of the macro today (though this may change in the future, as the macro is somewhat
 unwieldy to add new fields to, like all macros by example).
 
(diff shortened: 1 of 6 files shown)


src/diagnostics/lintstore.md
```diff
@@ -55,9 +55,9 @@ internally.
 Note, these include both rustc-internal lints, and the traditional lints, like, for example the dead
 code lint.
 
-These are primarily described in two places: `rustc::lint::builtin` and `rustc_lint::builtin`. The
-first provides the definitions for the lints themselves, and the latter provides the lint pass
-definitions (and implementations).
+These are primarily described in two places: `rustc_middle::lint::builtin` and
```
JohnTitor, line 58:
```suggestion
These are primarily described in two places: `rustc_session::lint::builtin` and
```

It isn't placed on `rustc_middle` but `rustc_session`.

src/diagnostics/lintstore.md
```diff
@@ -15,9 +15,9 @@ Unfortunately, a lot of the documentation we have refers to both of these as jus
 
 First, we have the lint declarations themselves: this is where the name and default lint level and
 other metadata come from. These are normally defined by way of the [`declare_lint!`] macro, which
-boils down to a static with type `&rustc::lint::Lint`. We lint against direct declarations without
-the use of the macro today (though this may change in the future, as the macro is somewhat unwieldy
-to add new fields to, like all macros by example).
+boils down to a static with type `&rustc_middle::lint::Lint`. We lint against direct declarations
```
JohnTitor, line 18:
```suggestion
boils down to a static with type `&rustc_session::lint::Lint`. We lint against direct declarations
```

src/diagnostics/lintstore.md
giraffate: I fixed it. Thanks.

src/diagnostics/lintstore.md
giraffate: I fixed it. Thanks.

giraffate:
> We can also update them following the comments:
> https://github.com/rust-lang/rustc-dev-guide/blob/master/src/compiler-debugging.md#L177-L183

Thanks for your information! It seems good.

> It seems the profiling perf chapter is also slightly outdated but I think it's better to go in another PR.

Yes. I try another PR.

giraffate:
The test seems to fail for another cause.

PR #804, opened 2020-07-17
Fixes menu for rustdoc page

FedericoPonzi:


src/SUMMARY.md
```diff
@@ -14,7 +14,7 @@
     - [Bootstrapping](./building/bootstrapping.md)
     - [Distribution artifacts](./building/build-install-distribution-artifacts.md)
     - [Documenting Compiler](./building/compiler-documenting.md)
-    - [Rustdoc](./rustdoc.md)
+    - [Rustdoc](./building/rustdoc.md)
```
JohnTitor, line 17: So, the problem here is that "The Rustc Driver and Interface" has also the rustdoc chapter, not the path.
It's introduced by 093b9032de1821f8a9f507f80f985aca1920858b but I assume it isn't intentional (cc @mark-i-m). I think it's a better way just to remove that entry, not to change the path.

FedericoPonzi:
`mdbook build` working locally, CI is failed because of lots of 429:
```
    │                     ^ Server returned 429 Too Many Requests for https://github.com/servo/homu
```
Can someone plesae restart travis? Thanks!

JohnTitor:
Nah, I mean, we should remove this:
https://github.com/FedericoPonzi/rustc-dev-guide/blob/summary-fix/src/SUMMARY.md#L66

FedericoPonzi:
ugh, sorry for the misunderstanding! I've just pushed the required changes, let me know if it's ok now!

PR #805, opened 2020-07-19
Add basic steps for a new target

shepmaster:


src/building/new-target.md
```diff
@@ -0,0 +1,105 @@
...
+
+git remote add my-target-llvm some-llvm-repository
+
+git checkout my-target-llvm/my-branch
+
+cd ..
+
+git add llvm_target
+
+git commit -m 'Use my custom LLVM'
```
jyn514, line 28:
```suggestion
cd src/llvm_project
git remote add my-target-llvm some-llvm-repository
git checkout my-target-llvm/my-branch
cd ..
git add llvm_target
git commit -m 'Use my custom LLVM'
```

src/building/new-target.md
```diff
@@ -0,0 +1,105 @@
...
+
+These are a set of steps to add support for a new target. There are
+numerous end states and paths to get there, so not all sections may be
+relevant to your desired goal.
+
+## Specifying a new LLVM
+
+For very new targets, you may need to use a different fork of LLVM
+than what is currently shipped with Rust. In that case, navigate to
+the `src/llvm_project` git submodule, check out the appropriate commit
```
jyn514, line 11: You might need to run `x.py check` at least once so the submodule is updated.

src/building/new-target.md
```diff
@@ -0,0 +1,105 @@
...
+[patch]: https://doc.rust-lang.org/stable/cargo/reference/overriding-dependencies.html#the-patch-section
+
+## Cross-compiling
+
+Once you have a target specification in JSON and in the code, you can
+cross-compile `rustc`:
+
+```
+DESTDIR=/path/to/install/in \
+x.py install -i --stage 1 --host aarch64-apple-darwin.json --target aarch64-apple-darwin \
```
jyn514, line 100:
```suggestion
./x.py install -i --stage 1 --host aarch64-apple-darwin.json --target aarch64-apple-darwin \
```

src/building/new-target.md
shepmaster: Sure. I find this harder to read, but if that's the convention.

src/building/new-target.md
```diff
@@ -0,0 +1,101 @@
...
+index be15e50e2bc..4fb1248ba99 100644
+--- a/Cargo.toml
++++ b/Cargo.toml
+@@ -66,10 +66,11 @@ cargo = { path = "src/tools/cargo" }
+ [patch.crates-io]
+ # Similar to Cargo above we want the RLS to use a vendored version of `rustfmt`
+ # that we're shipping as well (to ensure that the rustfmt in RLS and the
+ # `rustfmt` executable are the same exact version).
+ rustfmt-nightly = { path = "src/tools/rustfmt" }
++libc = { git = "https://github.com/rust-lang/libc", rev = "0bf7ce340699dcbacabdf5f16a242d2219a49ee0" }
```
androm3da, line 76: Is it possible to describe when or why you might want to patch the libc used by the top-level `Cargo.toml` versus situations where you'd patch the libc dependency of `rustc-std-workspace-{std,core,alloc}` (assuming there are some)?

src/building/new-target.md
```diff
@@ -0,0 +1,101 @@
...
+cross-compile `rustc`:
+
+```
+DESTDIR=/path/to/install/in \
+./x.py install -i --stage 1 --host aarch64-apple-darwin.json --target aarch64-apple-darwin \
+src/librustc src/libstd
+```
+
+If your target specification is already available in the bootstrap
+compiler, you can use it instead of the JSON file for both arguments.
```
androm3da, line 101: I think this is really great info.  Other content that might be good to include:
* reference to config.toml.example, sample changes that show how it's used for another target
* reference to a sample commit that adds support for a target to libc or librustc_target (these unfortunately age very quickly)
* do's/don'ts:
* do (or do-not?) use `rustup toolchain link` to get a limited target integrated with a more complete toolchain
* do (or do-not?) use cargo xbuild to experiment with local builds of libc/libcore/liballoc/libstd
* do not add target-specific behavior to librustc_foo,  librustc_bar - instead add to existing members of librustc_target or customize the `TargetResult` with new args...

These might be good for follow-ups to this doc (I've at least worked on target support for the new inline asm and could probably add some info about it):
* inline asm
* intrinsics
* do some/most targets require customization of lang items?
* reference to the `target_tier_policy` draft RFC

src/building/new-target.md
shepmaster: It may be possible, but **I** don't know the reason. I only needed to patch the listed Cargo.toml.

src/building/new-target.md
shepmaster: > * reference to config.toml.example, sample changes that show how it's used for another target

In my case, there were none, so I've nothing useful to add.

> * reference to a sample commit that adds support for a target to libc or librustc_target (these unfortunately age very quickly)

We could link to a specific commit (like https://github.com/rust-lang/rust/pull/74541) *and* to https://github.com/rust-lang/rust/commits/master/src/librustc_target/spec and say "look for newer examples". WDYT?

> * do (or do-not?) use `rustup toolchain link` to get a limited target integrated with a more complete toolchain

I don't have a good sense of exactly what to encourage here. *I* used rustup, but is that relevant to this section? That seems more general to a "so you built the compiler" section.

> * do (or do-not?) use cargo xbuild to experiment with local builds of libc/libcore/liballoc/libstd

I did not do this, so I don't have anything useful to add about it.

> * do not add target-specific behavior to librustc_foo,  librustc_bar - instead add to existing members of librustc_target or customize the `TargetResult` with new args...

I don't know what this means, really. The AVR port [requires changes to a number of files outside of target](https://github.com/rust-lang/rust/pull/73270/files); should that PR be disallowed?

androm3da: > WDYT?

SGTM

> I don't know what this means, really. The AVR port requires changes to a number of files outside of target; should that PR be disallowed?

Yeah, sorry I guess I was just spitballing and was wrong here. Please disregard.

mark-i-m:
Yeah, this is a weird failure. I restarted CI...

src/building/new-target.md
```diff
@@ -0,0 +1,101 @@
...
+ # Similar to Cargo above we want the RLS to use a vendored version of `rustfmt`
+ # that we're shipping as well (to ensure that the rustfmt in RLS and the
+ # `rustfmt` executable are the same exact version).
+ rustfmt-nightly = { path = "src/tools/rustfmt" }
++libc = { git = "https://github.com/rust-lang/libc", rev = "0bf7ce340699dcbacabdf5f16a242d2219a49ee0" }
+
+ # See comments in `src/tools/rustc-workspace-hack/README.md` for what's going on
+ # here
+ rustc-workspace-hack = { path = 'src/tools/rustc-workspace-hack' }
+ ```
```
JohnTitor, line 81: This one white-space before the fence confuses our line length checker :P

PR #809, opened 2020-07-22
Fix links for cranelift

JohnTitor:
The wasmtime renamed their default branch to `main` thus its link returns 404 now.
This replaces `master` with `HEAD` which is an alias for the default branch on GitHub, not `main`, just in case.


JohnTitor:
> I re-ran the checks.

Thanks! It's now green.

PR #815, opened 2020-07-29
Rename `src/libstd` to `library/std` etc.

JohnTitor:
Follow-up of rust-lang/rust#73265
I renamed them mechanically so the text may be wrong in places.
cc @mark-i-m


tshepang:
Failure is due to TLS cert that expired yesterday

JohnTitor:
> Failure is due to TLS cert that expired yesterday

I contacted the owner and they fixed the issue: https://github.com/Munksgaard/munksgaard.me/issues/1

goosnarrggh:
Was it perhaps premature to land this change publicly, since the current "stable" release 1.45.2 is still using the old directory layout?

For someone who is brand new to attempting to build rustc (such as myself), and who wants to maximize their chances of success by starting by attempting to rebuild the stable release before trying their luck with something more bleeding-edge, the content of https://rustc-dev-guide.rust-lang.org/building/how-to-build-and-run.html led to quite a bit of consternation when I was doing everything exactly as the documentation suggested (shame on me for checking out a tagged release, I suppose) and having the bootstrap bail out almost as quickly as it began.

jyn514:
@goosnarrggh I don't building from a tagged release is really supported. Is there some documentation that pointed you to do that?

goosnarrggh:
@jyn514 I suppose it came from my (perhaps flawed) presumption that the head of the default branch might be occasionally broken by work-in-progress, and I took the initiative to check out a tagged branch in the hopes that it might actually increase my chances of success. Like I said, shame on me.

In any event, I suppose I shall see in a few hours whether or not building from a tagged release can actually yield successful results.

jyn514:
> the head of the default branch might be occasionally broken by work-in-progress

Well, sort of. Occasionally tools (like clippy and rustfmt) do get broken. But the master branch of the compiler is guaranteed to always build, there's quite a lot of CI making sure that PRs never break the build. So you don't have to worry about work in progress when getting started :)

Note also that the compiler is very unstable internally, so working on a codebase from 16 weeks ago means a lot of your changes might be out of date when you try to transfer them to the master branch.

goosnarrggh:
Thanks for the explanation, that seems completely reasonable.

PR #816, opened 2020-07-31
Adjust mentions about CI to GitHub Actions

JohnTitor:
So, we've migrated our CI to GitHub Actions almost, except for macOS.
There're some mentions to Pipelines and even Travis in the guide, let's replace them with GHA. I don't think we have to mention macOS build but any suggestions are welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,8 +103,9 @@ There is not a workaround for this error at the moment.  Comment out the `[outpu
 
 ## How to fix toolstate failures
 
-> **NOTE**: Currently, we do not track the toolstate due to the spurious failure,
-but we leave this for when we do it again in the future.
+> **NOTE**: Currently, we do not track the rustc-dev-guide toolstate due to
+[the spurious failure](https://github.com/rust-lang/rust/pull/71731),
+but we leave this instructions for when we do it again in the future.
 
 1. You will get a ping from the toolstate commit. e.g. https://github.com/rust-lang-nursery/rust-toolstate/commit/8ffa0e4c30ac9ba8546b7046e5c4ccc2b96ebdd4
 
--- a/README.md
+++ b/README.md
@@ -103,13 +103,16 @@ There is not a workaround for this error at the moment.  Comment out the `[outpu
 
 ## How to fix toolstate failures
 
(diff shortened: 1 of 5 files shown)


src/tests/intro.md
```diff
@@ -116,27 +116,29 @@ including:
...
 branch. When a PR is approved, it goes into a [queue] where merges are tested
-one at a time on a wide set of platforms using Azure Pipelines (currently over
-50 different configurations). Most platforms only run the build steps, some run
-a restricted set of tests, only a subset run the full suite of tests (see
-Rust's [platform tiers]).
-
-[Azure Pipelines]: https://dev.azure.com/rust-lang/rust/
+one at a time on a wide set of platforms using GitHub Actions (currently over
+50 different configurations). Due to the limit on the number of parallel jobs,
+we run CI under the [rust-lang-ci] organization except for PRs.
```
JohnTitor, line 128: > Due to the limit on the number of parallel jobs

Is this correct? @Mark-Simulacrum

README.md
```diff
@@ -103,13 +103,16 @@ There is not a workaround for this error at the moment.  Comment out the `[outpu
 
 ## How to fix toolstate failures
 
+> **NOTE**: Currently, we do not track the toolstate due to the spurious failure,
+but we leave this for when we do it again in the future.
```
jyn514, line 107: What spurious failure is this referring to?

JohnTitor: I think this PR has detail explanation: https://github.com/rust-lang/rust/pull/71731

JohnTitor: So, it makes more sense to add this link here?

jyn514: Yes, I think so. Otherwise the first question is 'why is it turned off?' and there's not enough info here to figure it out.

jyn514: Also the wording is ambiguous and makes it look like _all_ toolstate is untracked, not just for rustc-dev-guide.

JohnTitor: Addressed via 7e24fae. Does it make you happy? :)

jyn514: Looks great :)

PR #818, opened 2020-08-01
Rename `HAIR` to `THIR`

### LINKED ISSUE #806: Rename `HAIR` (High-level Abstruct IR) to `THIR` (Typed HIR) in this guide
Follow-up of https://github.com/rust-lang/rust/pull/74582
This proposes to rename HAIR to THIR and it'll be likely accepted.
We have some mention of HAIR here, so the follow-up work will be needed as Niko said.
I'll work on this but in case let's create an issue just as a reminder.

JohnTitor:
rust-lang/rust#74582 is merged, let's cut our hair here as well :)
The nightly rustc doc isn't updated so the links are currently broken.
Closes #806

### DIFF
--- a/src/mir/construction.md
+++ b/src/mir/construction.md
@@ -14,6 +14,7 @@ list of items:
 The lowering is triggered by calling the [`mir_built`] query.
 There is an intermediate representation
 between [HIR] and [MIR] called the [THIR] that is only used during the lowering.
+[THIR] means "Typed HIR" and used to be called "HAIR (High-level Abstract IR)".
 The [THIR]'s most important feature is that the various adjustments (which happen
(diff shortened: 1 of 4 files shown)


src/mir/construction.md
```diff
@@ -1,4 +1,4 @@
-# HAIR and MIR construction
+# THIR and MIR construction
```
mark-i-m, line 2: Could you add a note that THIR used to be called HAIR? That way people that come across comments or old discussions have some context.

JohnTitor: Good point! Addressed via 19832d56b01e76a40f7c7aabf973483b1bb4a656.

PR #819, opened 2020-08-03
Fix link/wording for `parse_nt`

JohnTitor:
Follow-up of https://github.com/rust-lang/rust/pull/74826

### DIFF
--- a/src/overview.md
+++ b/src/overview.md
@@ -38,11 +38,11 @@ we'll talk about that later.
   data which are emitted in `librustc_parse::lexer::mod` as real diagnostics.
 - The lexer preserves full fidelity information for both IDEs and proc macros.
 - The parser [translates the token stream from the lexer into an Abstract Syntax
-  Tree (AST)][parser].  It uses a recursive descent (top-down) approach to syntax
-  analysis. The crate entry points for the parser are the `Parser.parse_crate_mod()` and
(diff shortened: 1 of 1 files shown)


src/overview.md
```diff
@@ -42,7 +42,7 @@ we'll talk about that later.
   analysis. The crate entry points for the parser are the `Parser.parse_crate_mod()` and
   `Parser.parse_mod()` methods found in `librustc_parse::parser::item`. The external
   module parsing entry point is `librustc_expand::module::parse_external_mod`. And
-  the macro parser entry point is `rustc_expand::mbe::macro_parser::parse_nt`.
+  the macro parser entry point is [`Parser.parse_nonterminal()`][parse_nonterminal].
```
mark-i-m, line 45:
```suggestion
  the macro parser entry point is [`Parser::parse_nonterminal()`][parse_nonterminal].
```

JohnTitor: I use `.` because we have `Parser.parse_crate_mod()` above. Should we replace them too?

mark-i-m: Oh, I didn't notice that. Yeah, would you mind changing the other ones to :: too?

JohnTitor: Sure! Addressed via bbb27ac.

PR #820, opened 2020-08-03
Fix examples not working

giraffate:


giraffate:
I just confirmed that `cargo run` was successful. But as you said, `cargo run` succeeded even if I changed rust to rustc_middle. Well, I don't know why this is.

I used this version.
```
$ rustc -V
rustc 1.47.0-nightly (05762e3d6 2020-08-01)
`

giraffate:
I'm sorry if I made a mistake, but isn't `rustc_middle` simply unused?

PR #821, opened 2020-08-03
Avoid "whitelist" according to rust-lang/rust changes

giraffate:
Fixed according to https://github.com/rust-lang/rust/commit/62cf767a4a39b47677d18110359d9e7152dc9d1c.

### DIFF
--- a/src/crates-io.md
+++ b/src/crates-io.md
@@ -14,7 +14,7 @@ reasons:
 
 TODO: what is the vetting process?
 
-## Pemitted dependencies
+## Permitted dependencies
 
 The `tidy` tool has [a list of crates that are allowed]. To add a
 dependency that is not already in the compiler, you will need to add it to these
--- a/src/crates-io.md
+++ b/src/crates-io.md
@@ -14,10 +14,10 @@ reasons:
 
 TODO: what is the vetting process?
 
-## Whitelist
+## Pemitted dependencies
 
-The `tidy` tool has a [whitelist] of crates that are allowed. To add a
-dependency that is not already in the compiler, you will need to add it to this
-whitelist.
+The `tidy` tool has [crates that are allowed]. To add a
+dependency that is not already in the compiler, you will need to add it to these
+crates.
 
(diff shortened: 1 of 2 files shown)


src/crates-io.md
```diff
@@ -14,10 +14,10 @@ reasons:
 
 TODO: what is the vetting process?
 
-## Whitelist
+## Pemitted dependencies
 
-The `tidy` tool has a [whitelist] of crates that are allowed. To add a
-dependency that is not already in the compiler, you will need to add it to this
-whitelist.
+The `tidy` tool has [crates that are allowed]. To add a
```
mark-i-m, line 19: perhaps "a list of crates..."?

giraffate: You're right, thanks!

src/crates-io.md
```diff
@@ -14,10 +14,10 @@ reasons:
 
 TODO: what is the vetting process?
 
-## Whitelist
+## Pemitted dependencies
```
jyn514, line 17: Pemitted -> Permitted

giraffate: Thanks!

src/crates-io.md
```diff
@@ -14,10 +14,10 @@ reasons:
...
 
-## Whitelist
+## Permitted dependencies
 
-The `tidy` tool has a [whitelist] of crates that are allowed. To add a
-dependency that is not already in the compiler, you will need to add it to this
-whitelist.
+The `tidy` tool has [a list of crates that are allowed]. To add a
+dependency that is not already in the compiler, you will need to add it to these
+crates.
```
jyn514, line 21:
```suggestion
dependency that is not already in the compiler, you will need to add it to the list.
```

src/stability.md
```diff
@@ -76,11 +76,11 @@ Furthermore this attribute is needed to mark an intrinsic as callable from
 Macros, compiler desugarings and `const fn`s expose their bodies to the call
 site. To work around not being able to use unstable things in the standard
 library's macros, there's the `#[allow_internal_unstable(feature1, feature2)]`
-attribute that whitelists the given features for usage in stable macros or
+attribute that allows the given features for usage in stable macros or
```
jyn514, line 79:
```suggestion
attribute that allows the given features to be used in stable macros or
```

PR #822, opened 2020-08-04
Some small fixes

giraffate:
- Enable a link `https://github.com/rust-lang/rust/issues/47389`
- Fix a broken link `https://crates.io/rayon`
- Fix typo

### DIFF
--- a/src/queries/incremental-compilation-in-detail.md
+++ b/src/queries/incremental-compilation-in-detail.md
@@ -507,9 +507,10 @@ session. The overhead of doing so is a few percent of total compilation time.
 Data structures used as query results could be factored in a way that removes
 edges from the dependency graph. Especially "span" information is very volatile,
 so including it in query result will increase the chance that that result won't
-be reusable. See https://github.com/rust-lang/rust/issues/47389 for more
+be reusable. See [https://github.com/rust-lang/rust/issues/47389] for more
 information.
 
 
 
 [query-model]: ./query-evaluation-model-in-detail.html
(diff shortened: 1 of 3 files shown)


src/queries/incremental-compilation-in-detail.md
```diff
@@ -507,9 +507,10 @@ session. The overhead of doing so is a few percent of total compilation time.
 Data structures used as query results could be factored in a way that removes
 edges from the dependency graph. Especially "span" information is very volatile,
 so including it in query result will increase the chance that that result won't
-be reusable. See https://github.com/rust-lang/rust/issues/47389 for more
+be reusable. See [https://github.com/rust-lang/rust/issues/47389] for more
```
jyn514, line 510: You can use a shortcut link here I think so you don't duplicate the url below.
```suggestion
be reusable. See <https://github.com/rust-lang/rust/issues/47389> for more
```

jyn514:
I was wondering how we didn't catch the broken link then I realized crates.io returns 200 even if the page isn't found 🤦

src/queries/incremental-compilation-in-detail.md
giraffate: I didn't know angle brackets to turn URL into a link. Thanks!

jyn514:
> I was wondering how we didn't catch the broken link then I realized crates.io returns 200 even if the page isn't found 🤦

This is https://github.com/rust-lang/crates.io/issues/788#issuecomment-312495830.

PR #823, opened 2020-08-11
Some small fixes

giraffate:
- Fix some typos
- Update from `typeck_tables_of` to `typeck`

### DIFF
--- a/src/diagnostics/diagnostic-codes.md
+++ b/src/diagnostics/diagnostic-codes.md
@@ -1,5 +1,5 @@
 # Diagnostic Codes
-We generally try to assign each error messages a unique code like `E0123`. These
+We generally try to assign each error message a unique code like `E0123`. These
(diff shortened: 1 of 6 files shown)


src/diagnostics/diagnostic-codes.md
```diff
@@ -1,5 +1,5 @@
 # Diagnostic Codes
-We generally try assign each error message a unique code like `E0123`. These
+We generally try to assign each error messages a unique code like `E0123`. These
```
jyn514, line 2: This should still be 'message': 'each message' refers to a single one.

jyn514:
Actually CI just looks like a network error, so hopefully will be fixed on the next push.

src/diagnostics/diagnostic-codes.md
giraffate: Thanks for your review. I fixed it.

PR #826, opened 2020-08-16
Remove unneeded hashmarks

mark-i-m:
r? @JohnTitor


src/serialization.md
```diff
@@ -36,9 +36,9 @@ deserialization to the fields of the struct or enum. For a struct those impls
 look something like this:
 
 ```rust,ingore
```
jyn514, line 38:
```suggestion
```rust,ignore
```
Although maybe if it worked before we shouldn't ignore it?

PR #828, opened 2020-08-17
Document how to use build task for x.py check

matklad:
That's the workflow I personally use (as I am using auto-save), so I guess others might find it useful as well (where "others" includes "me on the other laptop").


JohnTitor:
Thanks!

PR #847, opened 2020-09-04
Document CGU partioning in case of generic and inline functions

arora-aman:
The `<br>` make sure that the first two columns don't get completely wrapped.

Looks something like: 
image

r? @wesleywiser 

Closes https://github.com/rust-lang/wg-incr-comp/issues/2

### DIFF
--- a/src/backend/monomorph.md
+++ b/src/backend/monomorph.md
@@ -55,8 +55,28 @@ The monomorphization collector is run just before MIR lowering and codegen.
 collection and then partitions them into [codegen
 units](../appendix/glossary.md#codegen-unit).
 
+## Codegen Unit (CGU) partitioning
+
+For better incremental build times, the CGU partitioner creates two CGU for each source level 
+modules. One is for "stable" i.e. non-generic code and the other is more volatile code i.e. 
+monoporphized/specialized instances.
+
+For depenencies, consider Crate A and Crate B, such that Crate B depends on Crate A.
+The following table lists different scenarios for a function in Crate A that might be used by one 
+or more modules in Crate B.
+
+| Crate A function | Behavior |
+| - | - |
+| Non-generic function | Crate A function doesn't appear in any codegen units of Crate B |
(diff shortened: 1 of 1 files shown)


jyn514:
It looks like 'uses once in a module' and 'uses in two different modules' is always the same? Is there a reason for including both?

arora-aman:
> It looks like 'uses once in a module' and 'uses in two different modules' is always the same? Is there a reason for including both?

My understanding of the issue is to verify the possibility of there being a difference in the two cases. I decided to leave them separate to make it explicit that both cases behave similarly. 

I can unify them as "Uses in any number of different modules"

jyn514:
> I can unify them as "Uses in any number of different modules

I think that would be helpful, yeah. That way you don't have to double check the graph to make sure you're not misreading it.

arora-aman:
@jyn514 updated :) 

image

src/backend/monomorph.md
```diff
@@ -55,6 +55,17 @@ The monomorphization collector is run just before MIR lowering and codegen.
 collection and then partitions them into [codegen
 units](../appendix/glossary.md#codegen-unit).
 
+Consider Crate A and Crate B, such that Crate B depends on Crate A.
```
wesleywiser, line 58: It might be good to add a header here for "Codegen Unit Partitioning" or similar. This isn't really part of the "Collection" step anymore (in my mind anyway). 

If you want to write a sentence or two introducing that topic, that would would be awesome, otherwise we can add more content here later.

arora-aman:
@wesleywiser updated

arora-aman:
The build seems to failing because of too many calls to the Github api (assuming during linkcheck), is there a fix for that?

jyn514 (approved):

jyn514:
> The build seems to failing because of too many calls to the Github api (assuming during linkcheck), is there a fix for that?

Try repushing the same change? `git commit --amend && git push -f`? I don't think we have @bors retry in this repo.

mark-i-m:
I restarted it

PR #856, opened 2020-09-08
Minor cleanup for "How to build and run"

camelid:


camelid:
Is this section in the first paragraph still true? Cc @jyn514 re defaults changes in `x.py`.

> The default configuration is oriented towards running the compiler as a user,
not a developer.

jyn514:
> Is this section in the first paragraph still true? Cc @jyn514 re defaults changes in x.py.

>    The default configuration is oriented towards running the compiler as a user,
    not a developer.

It's a mess and there's no consensus.

src/building/how-to-build-and-run.md
```diff
@@ -215,13 +213,13 @@ your local environment:
...
-rustc 1.25.0-dev
+rustc 1.48.0-dev
 binary: rustc
 commit-hash: unknown
 commit-date: unknown
 host: x86_64-unknown-linux-gnu
-release: 1.25.0-dev
-LLVM version: 4.0
+release: 1.48.0-dev
+LLVM version: 11.0
```
jyn514, line 222: It's not a big deal since you're updating other things, but there's much point in updating this version - it gets out of date again every 6 weeks and it still gets the point across.

src/building/how-to-build-and-run.md
```diff
@@ -252,4 +250,5 @@ everything up then you only need to run one command!
 ./x.py clean
 ```
 
-`rm -rf build` works too, but then you have to rebuild LLVM.
+`rm -rf build` works too, but then you have to rebuild LLVM, which can take
+a long time (on the order of a half-hour).
```
jyn514, line 254: I don't like saying half hour because it can be a lot more or a lot less depending on your computer. "Long time" seems fine, you'll know after the first time you build it.

src/building/how-to-build-and-run.md
camelid: I know, but I figure I might as well update it so it's not *so* out-of-date :)

src/building/how-to-build-and-run.md
camelid: I just wanted to provide a *very* rough timescale. Is there a way to convey that without providing a specific time?

jyn514: 'a little longer than a stage 1 build'?

jyn514: The reason I don't think this makes sense is some people (like `@Mark-Simulacrum`) can build from scratch in 5 minutes, while others like `@mark-i-m` need over 2 hours. So there's no time we can give that would be anywhere _close_ for both of them.

camelid: Wow, I want `@Mark-Simulacrum`'s computer ;)

camelid: What about this?

```suggestion
a long time even on fast computers.
```

My computer is quadcore and it still takes 30 minutes.

jyn514: Yeah, that sounds reasonable.

jyn514 (approved):

PR #869, opened 2020-09-10
Update "Helpful Links" section

camelid:


src/contributing.md
```diff
@@ -476,27 +476,26 @@ For people new to Rust, and just starting to contribute, or even for
...
+* The [generated documentation for Rust's compiler][gdfrustc]
+* The [Rust reference][rr], even though it doesn't specifically talk about
+  Rust's internals is a great resource nonetheless
 * Although out of date, [Tom Lee's great blog article][tlgba] is very helpful
 * [rustaceans.org][ro] is helpful, but mostly dedicated to IRC
 * The [Rust Compiler Testing Docs][rctd]
-* For [@bors][bors], [this cheat sheet][cheatsheet] is helpful
-(though you'll need to replace `@homu` with `@bors` in any commands)
-* **Google!** ([search only in Rust Documentation][gsearchdocs] to find types,
-  traits, etc. quickly)
```
camelid, line 494: This was added in the original rust-lang/rust `CONTRIBUTING.md` file [5 years ago](https://github.com/rust-lang/rust/commit/aa9666e5c1cbf0e1295e266f738d764f911f48f3). I'm guessing that was before the Rustdoc search feature?

src/contributing.md
```diff
@@ -476,27 +476,29 @@ For people new to Rust, and just starting to contribute, or even for
...
 * [Rust Forge][rustforge] contains additional documentation, including
   write-ups of how to achieve common tasks
 * The [Rust Internals forum][rif], a place to ask questions and
   discuss Rust's internals
-* The [generated documentation for rust's compiler][gdfrustc]
-* The [rust reference][rr], even though it doesn't specifically talk about
-  Rust's internals, it's a great resource nonetheless
+* The [generated documentation for Rust's compiler][gdfrustc]
+* The [Rust reference][rr], even though it doesn't specifically talk about
+  Rust's internals is a great resource nonetheless
```
jyn514, line 487:
```suggestion
  Rust's internals, is a great resource nonetheless
```

src/contributing.md
jyn514: Rustdoc only searches in the crates you're currently looking at; search results in https://doc.rust-lang.org/nightly/std/primitive.bool.html won't show results for the nomicon. So I think recommending google search is still useful.

src/contributing.md
camelid: I wasn't sure about whether there should be a comma, but I guess it's probably better to err on the side of having one :)

src/contributing.md
camelid: Okay, I'll add it back and then people will have two options.

jyn514 (approved):

Issue #850, opened 2020-09-06
Add a section on how to use git


jyn514:
Git is a major part of contributing to rust-lang/rust, but it sees almost no discussion in the guide. There should be instructions on:
- How to pull, branch, push, etc. This can just be a link to a git tutorial, but it should be there. It should also have basic instructions on GitHub (even if it's just 'the create pull request button is here').
- How to rebase. A lot of people are afraid of rebasing if they haven't used it a lot: the commands are not intuitive, it's rewriting history which already sounds scary, and you run into merge conflicts as you go which make the whole thing even more painful. There should be a detailed walkthrough on this.
- The no-merges policy. This is really discouraging to newcomers: they _have a change that works_, but it's not accepted because of what seems like paperwork. It should say why we have this policy and how to fix your PR if you made a merge by accident.

Writing this requires no knowledge of rust or the compiler (although having made PRs to rust-lang/rust is a bonus). It should go in the 'Contributing to Rust' section, possibly in its own page.

jyn514:
Possibly there should be instructions on how to do this in a GUI too, but I've never used a git gui for rebasing and I wouldn't be able to help with that bit.

workingjubilee:
People should be reminded that making a branch is effectively free in git, so it's fine to just pop over to another branch to save their workstate and then pop back.

jyn514:
'Git tips' would also be a useful section (maybe near `git worktree` in the suggested workflows page?) but I see it as something separate from this issue, which is 'how do I contribute at _all_'.

We should mention how to fix issues from adding a submodule by accident though: https://github.com/rust-lang/rust/pull/75949#issuecomment-687631020
Maybe we should also suggest `git add <file>` instead of `git add .`? That would help prevent committing submodules in the first place.

workingjubilee:
Right, I mean that as a heads-up before rebasing. "rebasing sounds intimidating but you can make an additional branch, push it to your fork on GH or whatever, and guaranteed you have your work saved."

Yeah, people should be adding by-file, or at least by-specific-directory, in a deliberate fashion, not `git add .`.

workingjubilee:
Graphical GUIs are always going to be a case-by-case for each, so covering rebasing in any of them will require someone who actually has used that tool, or finding other people's tutorials. However, many people lean on VS Code as a dev tool, and since VS Code is semi-officially-supported by rust-analyzer, so we should probably lean on assisting people with VS Code's support for git, as a start.

[...]

PR #870, opened 2020-09-10
Clean up 'Contributing to Rust - Pull Requests'

### LINKED ISSUE #866: Document better that it's OK to use CI for testing
Currently this is documented as
> If your machine is not very powerful, a common strategy is to only use ./x.py check on your local machine and let the CI build test your changes when you push to a PR branch.

But it's hidden all the way down in https://rustc-dev-guide.rust-lang.org/building/prerequisites.html#hardware, which is a long way away from https://rustc-dev-guide.rust-lang.org/contributing.html#pull-requests (which is itself not easy to find).

We should
- [x] Add a note near the 'pull request' section that Rust has plenty of CI capacity for PRs and it's fine to use it
- [ ] Make the pull request section easier to find
- [ ] Link to it from more places

For context, this is meant to be a replacement for https://github.com/rust-lang/rust/pull/76272.

LeSeulArtichaut:
Also helps with #866.

### DIFF
--- a/src/contributing.md
+++ b/src/contributing.md
@@ -92,7 +92,7 @@ to the end of the pull request description, and [@rust-highfive][rust-highfive] 
 
 In addition to being reviewed by a human, pull requests are automatically tested
 thanks to continuous integration (CI). Basically, every time you open and update
-a pull request, the CI builds the compiler and tests it against the
+a pull request, CI builds the compiler and tests it against the
 [compiler test suite][rctd], and also performs other tests such as checking that
 your pull request is in compliance with Rust's style guidelines.
 
@@ -103,7 +103,8 @@ of the status of a particular pull request.
 Rust has plenty of CI capacity, and you should never have to worry about wasting
 computational resources each time you push a change. It is also perfectly fine
 (and even encouraged!) to use the CI to test your changes if it can help your
-productivity, e.g. if your machine is not very powerful.
+productivity. In particular, we don't recommend running the full `x.py test` suite locally,
+since it takes a very long time to execute.
 
 After someone has reviewed your pull request, they will leave an annotation
 on the pull request with an `r+`. It will look something like this:
@@ -136,11 +137,11 @@ should be aware of.
 
 All pull requests should be filed against the `master` branch, except in very
 particular scenarios. Unless you know for sure that you should target another
-branch, `master` will be the right choice.
+branch, `master` will be the right choice (it's also the default).
 
 Make sure your pull request is in compliance with Rust's style guidelines by running
 
-    $ ./x.py test tidy
+    $ ./x.py test tidy --bless
 
 We recommand to make this check before every pull request (and every new commit
 in a pull request); you can add [git hooks](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks)
--- a/src/contributing.md
+++ b/src/contributing.md
@@ -143,7 +143,7 @@ Make sure your pull request is in compliance with Rust's style guidelines by run
 
     $ ./x.py test tidy --bless
 
-We recommand to make this check before every pull request (and every new commit
+We recommend to make this check before every pull request (and every new commit
 in a pull request); you can add [git hooks](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks)
 before every push to make sure you never forget to make this check. The
 CI will also run tidy and will fail if tidy fails.
--- a/src/contributing.md
+++ b/src/contributing.md
@@ -67,40 +67,14 @@ in the appropriate provided template.
 
 ## Pull Requests
 
-Pull requests are the primary mechanism we use to change Rust. GitHub itself
-has some [great documentation][about-pull-requests] on using the Pull Request feature.
-We use the "fork and pull" model [described here][development-models], where
-contributors push changes to their personal fork and create pull requests to
+Pull requests (or PRs for short) are the primary mechanism we use to change Rust.
+GitHub itself has some [great documentation][about-pull-requests] on using the
+Pull Request feature. We use the "fork and pull" model [described here][development-models],
+where contributors push changes to their personal fork and create pull requests to
 bring those changes into the source repository.
 
 [about-pull-requests]: https://help.github.com/articles/about-pull-requests/
 [development-models]: https://help.github.com/articles/about-collaborative-development-models/
-
-Please make pull requests against the `master` branch.
-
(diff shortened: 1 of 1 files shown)


LeSeulArtichaut:
@jyn514 Do you think this is enough to address #866 btw?

src/contributing.md
```diff
@@ -116,32 +90,76 @@ make a documentation change, add
 to the end of the pull request description, and [@rust-highfive][rust-highfive] will assign
 [@steveklabnik][steveklabnik] instead of a random person. This is entirely optional.
 
+In addition to being reviewed by a human, pull requests are automatically tested
+thanks to continuous integration (CI). Basically, every time you open and update
+a pull request, the CI builds the compiler and tests it against the
```
jyn514, line 95: nit: CI is a process, not a noun
```suggestion
a pull request, CI builds the compiler and tests it against the
```
or
```suggestion
a pull request, Github Actions builds the compiler and tests it against the
```

src/contributing.md
```diff
@@ -116,32 +90,76 @@ make a documentation change, add
...
+your pull request is in compliance with Rust's style guidelines.
+
+Running continuous integration tests allows PR authors to catch mistakes early
+without going through a first review cycle, and also helps reviewers stay aware
+of the status of a particular pull request.
+
+Rust has plenty of CI capacity, and you should never have to worry about wasting
+computational resources each time you push a change. It is also perfectly fine
+(and even encouraged!) to use the CI to test your changes if it can help your
+productivity, e.g. if your machine is not very powerful.
```
jyn514, line 106: This is true even for people with powerful machines ;) rustc is a bit of a hog.
```suggestion
productivity. In particular, we don't recommend running the full `x.py test` suite locally,
since it takes a very long time to execute.
```

src/contributing.md
```diff
@@ -116,32 +90,76 @@ make a documentation change, add
...
+with one another are marked as "always roll up".
 
 [rust-highfive]: https://github.com/rust-highfive
 [steveklabnik]: https://github.com/steveklabnik
 [@bors]: https://github.com/bors
 [merge-queue]: https://buildbot2.rust-lang.org/homu/queue/rust
 
-Speaking of tests, Rust has a comprehensive test suite. More information about
-it can be found [here][rctd].
+### Opening a PR
```
jyn514, line 132: Hmm, I think I'd prefer for this to be a separate page to avoid making this one too long. But there's already a lot of sections, so maybe splitting them all up should be a separate PR?

src/contributing.md
```diff
@@ -116,32 +90,76 @@ make a documentation change, add
...
+
+You are now ready to file a pull request? Great! Here are a few points you
+should be aware of.
+
+All pull requests should be filed against the `master` branch, except in very
+particular scenarios. Unless you know for sure that you should target another
+branch, `master` will be the right choice.
```
jyn514, line 139:
```suggestion
branch, `master` will be the right choice (it's also the default).
```

src/contributing.md
```diff
@@ -116,32 +90,76 @@ make a documentation change, add
...
+
+Make sure your pull request is in compliance with Rust's style guidelines by running
+
+    $ ./x.py test tidy
```
jyn514, line 143: No need to make them run it twice.
```suggestion
    $ ./x.py test tidy --bless
```

src/contributing.md
```diff
@@ -116,32 +90,76 @@ make a documentation change, add
...
+We recommand to make this check before every pull request (and every new commit
+in a pull request); you can add [git hooks](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks)
+before every push to make sure you never forget to make this check. The
+CI will also run tidy and will fail if tidy fails.
+
+Rust follows a _no merge-commit policy_, meaning, when you encounter merge
+conflicts you are expected to always rebase instead of merging.  E.g. always use
+rebase when bringing the latest changes from the master branch to your feature
+branch.  Also, please make sure that fixup commits are squashed into other
+related commits with meaningful commit messages.
```
jyn514, line 154: Once someone writes up #850, this should link there. But no need to solve two issues at once ;)

jyn514:
> @jyn514 Do you think this is enough to address #866 btw?

I think it's not discoverable enough still ... I think to mark that as closed we should split up the contributing page into multiple smaller ones, then I'd be happy enough. As it is there's a lot of text to wade through to find what you're looking for.

jyn514:
This is definitely much better than before though!

LeSeulArtichaut:
Changed it to not close #866.

src/contributing.md
LeSeulArtichaut: I think I've seen occurences of "the CI" in other parts of the guide... Not sure that's worth fixing, I'll just change that for my PR.

src/contributing.md
LeSeulArtichaut: I think I could do that in a separate PR

jyn514 (approved):
This is somewhat of a big change - maybe someone else wants to take a look? but LGTM

mark-i-m (approved):
Thanks! LGTM :)

src/contributing.md
```diff
@@ -116,32 +90,77 @@ make a documentation change, add
 to the end of the pull request description, and [@rust-highfive][rust-highfive] will assign
 [@steveklabnik][steveklabnik] instead of a random person. This is entirely optional.
 
+In addition to being reviewed by a human, pull requests are automatically tested
+thanks to continuous integration (CI). Basically, every time you open and update
+a pull request, CI builds the compiler and tests it against the
```
tshepang, line 95:
```suggestion
a PR, CI builds the compiler and tests it against the
```

tshepang: This avoids having to decide between *pull request* and *Pull Request*

LeSeulArtichaut: I don't think Pull Request vs pull request is a problem however

tshepang: The inconsistency bothered me

src/contributing.md
```diff
@@ -116,32 +90,77 @@ make a documentation change, add
...
+
+All pull requests should be filed against the `master` branch, except in very
+particular scenarios. Unless you know for sure that you should target another
+branch, `master` will be the right choice (it's also the default).
+
+Make sure your pull request is in compliance with Rust's style guidelines by running
+
+    $ ./x.py test tidy --bless
+
+We recommand to make this check before every pull request (and every new commit
```
tshepang, line 146:
```suggestion
We recommend to make this check before every pull request (and every new commit
```

PR #872, opened 2020-09-10
Add a chapter on all the identifiers used through `rustc`

### LINKED ISSUE #748: Add info about IDs in the compiler

LeSeulArtichaut:
@rust-lang/wg-rustc-dev-guide I don't really know what's the best way organizing all of this. I'm not satisfied with the HIR IDs being described in detail in both the HIR chapter and this new one, and at the same time I'm not sure we should remove them from the HIR chapter because they are needed to understand the rest of the chapter.

Replaces #759, closes #748.

### DIFF
--- a/src/identifiers.md
+++ b/src/identifiers.md
@@ -27,7 +27,7 @@ like macro expansion and name resolution.
 The HIR uses a bunch of different identifiers that coexist and serve different purposes.
 
 - A [`DefId`], as the name suggests, identifies a particular definition, or top-level
-  item, in a given grate. It is composed of two parts: a [`CrateNum`] which identifies
+  item, in a given crate. It is composed of two parts: a [`CrateNum`] which identifies
   the crate the definition comes from, and a [`DefIndex`] which identifies the definition
   within the crate. Unlike [`NodeId`]s, there isn't a [`DefId`] for every expression, which
   makes them more stable across compilations.
--- a/src/identifiers.md
+++ b/src/identifiers.md
@@ -1,7 +1,7 @@
 # Identifiers in the Compiler
 
-If you have read the few previous chapters, you now know that the `rustc` uses
-many different intermediate representations to perform different kinds of analysis.
+If you have read the few previous chapters, you now know that `rustc` uses
+many different intermediate representations to perform different kinds of analyses.
 However, like in every data structure, you need a way to traverse the structure
 and refer to other elements. In this chapter, you will find information on the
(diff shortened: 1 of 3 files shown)


src/identifiers.md
```diff
@@ -0,0 +1,63 @@
+# Identifiers in the Compiler
+
+If you have read the few previous chapters, you now know that the `rustc` uses
```
tshepang, line 3:
```suggestion
If you have read the few previous chapters, you now know that `rustc` uses
```

src/identifiers.md
```diff
@@ -0,0 +1,63 @@
...
+many different intermediate representations to perform different kinds of analysis.
```
tshepang, line 4:
```suggestion
many different intermediate representations to perform different kinds of analyses.
```

jyn514 (approved):

mark-i-m:
You might want to update the HIR chapter to point at this one. I think there is some content about identifiers there too.

LeSeulArtichaut:
@mark-i-m Could you check out d777b6d and tell me if you like it or not? :D

src/identifiers.md
```diff
@@ -0,0 +1,63 @@
...
+like macro expansion and name resolution.
+
+[`NodeId`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_ast/node_id/struct.NodeId.html
+
+## In the HIR
+
+The HIR uses a bunch of different identifiers that coexist and serve different purposes.
+
+- A [`DefId`], as the name suggests, identifies a particular definition, or top-level
+  item, in a given grate. It is composed of two parts: a [`CrateNum`] which identifies
```
mark-i-m, line 30:
```suggestion
  item, in a given crate. It is composed of two parts: a [`CrateNum`] which identifies
```

mark-i-m (approved):
Thanks 💯

PR #875, opened 2020-09-11
Bump mdbook version

camelid:
0.4.1 -> 0.4.3

Looking at the changelog, it looks like only minor changes and improvements.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ before_install:
 - MAX_LINE_LENGTH=100 bash ci/check_line_lengths.sh src/**/*.md
 install:
 - source ~/.cargo/env || true
-- cargo install mdbook --version 0.4.1
+- cargo install mdbook --version 0.4.3
 - cargo install mdbook-linkcheck --version 0.7.0
 script:
 - git checkout -b ci
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,8 @@ before_install:
 - MAX_LINE_LENGTH=100 bash ci/check_line_lengths.sh src/**/*.md
 install:
 - source ~/.cargo/env || true
-- cargo install mdbook --version 0.4.3
-- cargo install mdbook-linkcheck --version 0.7.0
+- cargo install mdbook --version '^0.4.3'
+- cargo install mdbook-linkcheck --version '^0.7.0'
 script:
 - git checkout -b ci
 - git rebase origin/master


camelid:
Perhaps we should change the version constraint to download the latest patch version? Is that possible with `cargo install`?

camelid:
It looks like it is possible with Cargo.

.travis.yml
```diff
@@ -11,7 +11,7 @@ before_install:
 - MAX_LINE_LENGTH=100 bash ci/check_line_lengths.sh src/**/*.md
 install:
 - source ~/.cargo/env || true
-- cargo install mdbook --version 0.4.1
+- cargo install mdbook --version 0.4.3
```
camelid, line 14: I don't know if it makes sense, but if you want to get the latest semver-compatible version, I *think* it should be:

```suggestion
- cargo install mdbook --version '^0.4.3'
```

JohnTitor: My concern is that it may lose the integration with `mdbook-linkcheck` if we don't check it explicitly. But I think it should be quite an minor issue, I'd be happy to accept that change :)

camelid: Well, patch releases shouldn't break things, right? But it might be a good idea to have the latest patch release of `mdbook-linkcheck` too.

.travis.yml
```diff
@@ -11,7 +11,7 @@ before_install:
...
 - cargo install mdbook-linkcheck --version 0.7.0
```
camelid, line 15:
```suggestion
- cargo install mdbook-linkcheck --version '^0.7.0'
```

camelid:
Okay, I changed it to use the latest patch release of both `mdbook` and `mdbook-linkcheck`! Let me know if I made a mistake with the semver constraints - I'm still somewhat inexperienced with those!

.travis.yml
```diff
@@ -11,8 +11,8 @@ before_install:
 - MAX_LINE_LENGTH=100 bash ci/check_line_lengths.sh src/**/*.md
 install:
 - source ~/.cargo/env || true
-- cargo install mdbook --version 0.4.1
-- cargo install mdbook-linkcheck --version 0.7.0
+- cargo install mdbook --version '^0.4.3'
+- cargo install mdbook-linkcheck --version '^0.7.0'
```
camelid, line 15: I can skip having a caret requirement for `mdbook-linkcheck` if you want to manually update it. I don't think it's updated that often anyway.

```suggestion
- cargo install mdbook-linkcheck --version 0.7.0
```

jyn514 (approved):
Changelog: https://github.com/rust-lang/mdBook/blob/master/CHANGELOG.md

camelid:
Yay, now syntax highlighting is working again!

image

PR #876, opened 2020-09-11
Improve link and description of entry points in parser section

guswynn:
This seemed to be out of date!


JohnTitor (approved):
Good point, thanks!

PR #877, opened 2020-09-12
Correct stability attribute example

Monadic-Cat:
It looks like there was a copy and paste mistake
on the `#[rustc_const_stable(...)]` attribute section.


jyn514 (approved):
Good catch, thanks!

PR #878, opened 2020-09-13
Document new way to not build LLVM

matklad:
Implemented in https://github.com/rust-lang/rust/pull/76349

r? @Mark-Simulacrum

### DIFF
--- a/src/building/suggested.md
+++ b/src/building/suggested.md
@@ -161,12 +161,21 @@ git worktree add -b my-feature ../rust2 master
 You can then use that rust2 folder as a separate workspace for modifying
 and building `rustc`!
 
-## Building with system LLVM
+## Skipping LLVM Build
 
-By default, LLVM is built from source, and that can take significant amount of
-time.  An alternative is to use LLVM already installed on your computer.
+By default, LLVM is built from source, and that takes significant amount of
+time. One way to avoid that is to add this to `config.toml`:
 
-This is specified in the `target` section of `config.toml`:
+```toml
+[llvm]
+download-ci-llvm = true
+```
+
+Downloading LLVM from CI is still experimental though, and might not be
+available on all platforms. Otherwise, we'd make it a default!
+
+Another alternative is to use LLVM already installed on your computer. This is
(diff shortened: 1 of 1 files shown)


src/building/suggested.md
```diff
@@ -161,12 +161,21 @@ git worktree add -b my-feature ../rust2 master
...
+time. One way to avoid that is to add this to `config.toml`:
 
-This is specified in the `target` section of `config.toml`:
+```toml
+[llvm]
+download-ci-llvm = true
+```
+
+Downloading LLVM from CI is still experimental though, and might not be
+available on all platforms. Otherwise, we'd make it a default!
```
jyn514, line 175:
```suggestion
Downloading LLVM from CI is still experimental though, and is currently only
available on Linux. Otherwise, we'd make it a default!
```

matklad: I somewhat intentionally phrased this vaguely, so that it does not bitrot, but I am fine with either wording.

tshepang, line 175: This does not feel like the right place to indicate planned policy/default, here referring to "we'd make it a default".

matklad: This also was done in the interest of future compatibility. The thing is in flux, so I find it useful:

* to document it early, even if incomplete
* to be clear that there might be changes, and that the docs might be out of date
* to not be super strict with keeping the docs up to date.

tshepang: I was only referring to "Other, we'd make it a default". First sentence is all good.

LeSeulArtichaut:
Since the previous discussion stalled out, I'd be in favour of landing this to get it in the guide now, and eventually rework the wording in a follow-up PR

src/building/suggested.md
mark-i-m: I think we can keep it as is personally.

mark-i-m (approved):
I agree with @LeSeulArtichaut. I think we should go ahead and merge this and iterate later if it causes confusion.

jyn514:
Sounds good to me.

PR #880, opened 2020-09-14
Improve MIR sections in appendix

camelid:
Adds a new graphical representation of one of the examples of MIR and
basic blocks!

And makes some other minor improvements.

r? @oli-obk

### DIFF
--- a/src/mir/index.md
+++ b/src/mir/index.md
@@ -44,7 +44,8 @@ This section introduces the key concepts of MIR, summarized here:
   leading underscore, like `_1`. There is also a special "local"
   (`_0`) allocated to store the return value.
 - **Places:** expressions that identify a location in memory, like `_1` or
-  `_1.f`.
+  `_1.f`. (Some people call these "lvalues"; the "l" meaning that they occur on
+  the left-hand side of an assignment.)
 - **Rvalues:** expressions that produce a value. The "R" stands for
   the fact that these are the "right-hand side" of an assignment.
   - **Operands:** the arguments to an rvalue, which can either be a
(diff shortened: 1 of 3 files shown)


camelid:
For some reason it's very satisfying to look at control-flow graphs.

src/mir/index.md
```diff
@@ -44,7 +44,8 @@ This section introduces the key concepts of MIR, summarized here:
   leading underscore, like `_1`. There is also a special "local"
   (`_0`) allocated to store the return value.
 - **Places:** expressions that identify a location in memory, like `_1` or
-  `_1.f`.
+  `_1.f`. (Some people call these "lvalues"; the "l" meaning that they occur on
+  the left-hand side of an assignment.)
```
jyn514, line 48: This is traditional in C (especially in compiler textbooks) but it's not correct in Rust, because they can be on the right hand side. For example `&mut MyTemporary{}` is a place expression that can be used in an expression context.

src/mir/index.md
```diff
@@ -44,7 +44,8 @@ This section introduces the key concepts of MIR, summarized here:
...
 - **Rvalues:** expressions that produce a value. The "R" stands for
```
jyn514, line 49: Similarly this should be 'value expressions'.
https://doc.rust-lang.org/reference/expressions.html#place-expressions-and-value-expressions

src/mir/index.md
camelid: Hmm, okay.

src/mir/index.md
camelid: Actually it seems they're called ["rvalues" in MIR](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/enum.Rvalue.html).

src/mir/index.md
jyn514: The only place in the whole compiler that uses this is https://github.com/rust-lang/rust/blob/9e5f7d5631b8f4009ac1c693e585d4b7108d4275/compiler/rustc_mir/src/transform/simplify.rs#L463
Maybe it should be changed?

oli-obk (approved):

PR #881, opened 2020-09-15
Improve instructions for adding a new test

jyn514:
- Add a section for library tests
- Expand section on rustdoc tests

From https://github.com/rust-lang/rust/issues/76268#issuecomment-692867148

### DIFF
--- a/src/tests/adding.md
+++ b/src/tests/adding.md
@@ -29,15 +29,24 @@ rough heuristics:
   - need to run gdb or lldb? use the `debuginfo` test suite
   - need to inspect LLVM IR or MIR IR? use the `codegen` or `mir-opt` test
     suites
-  - need to run rustdoc? Prefer a `rustdoc` or `rustdoc-ui` test. Occasionally you'll need `rustdoc-js` as well.
+  - need to run rustdoc? Prefer a `rustdoc` or `rustdoc-ui` test.
+    Occasionally you'll need `rustdoc-js` as well.
   - need to inspect the resulting binary in some way? Then use `run-make`
-- Library tests should go in `library/$crate/tests` (where `$crate` is usually `core`, `alloc`, or `std`).
-- For most other things, [a `ui` (or `ui-fulldeps`) test](#ui) is to be
-  preferred:
-  - `ui` tests subsume both `run-pass`, `compile-fail`, and `parse-fail` tests
+- Library tests should go in `library/${crate}/tests` (where `${crate}` is
+  usually `core`, `alloc`, or `std`). Library tests include:
+  - tests that an API behaves properly, including accepting various types or
+    having some runtime behavior
+  - tests where any compiler warnings are not relevant to the test
+  - tests that a use of an API gives a compile error, where the exact error
+    message is not relevant to the test. These should have an
+    [error number](`E0XXX`) in the code block to make sure it's the correct error.
+- For most other things, [a `ui` (or `ui-fulldeps`) test](#ui) is to be preferred:
+  - [`ui`](#ui) tests subsume both `run-pass`, `compile-fail`, and `parse-fail` tests
   - in the case of warnings or errors, `ui` tests capture the full output,
     which makes it easier to review but also helps prevent "hidden" regressions
     in the output
+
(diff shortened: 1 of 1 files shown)


src/tests/adding.md
```diff
@@ -29,11 +29,12 @@ rough heuristics:
   - need to run gdb or lldb? use the `debuginfo` test suite
   - need to inspect LLVM IR or MIR IR? use the `codegen` or `mir-opt` test
     suites
-  - need to run rustdoc? Prefer a `rustdoc` test
+  - need to run rustdoc? Prefer a `rustdoc` or `rustdoc-ui` test. Occasionally you'll need `rustdoc-js` as well.
   - need to inspect the resulting binary in some way? Then use `run-make`
+- Library tests should go in `library/$crate/tests` (where `$crate` is usually `core`, `alloc`, or `std`).
```
workingjubilee, line 34: I usually prefer to add `${brackets}` in a case like this, rather than hoping the shell parses it correctly. :^) Actually, it might even be better to strictly match Rust's `fmt` syntax, if we're going to use shorthand...

Would be good to mention that a "library test" is one that tests an API's functionality directly, merely verifying it succeeds or has some simple behavior... basically, to specify that `run-pass` tests should usually be library tests, without directly saying "don't add run-pass tests".

src/tests/adding.md
```diff
@@ -29,11 +29,12 @@ rough heuristics:
...
 - For most other things, [a `ui` (or `ui-fulldeps`) test](#ui) is to be
   preferred:
-  - `ui` tests subsume both run-pass, compile-fail, and parse-fail tests
+  - `ui` tests subsume both `run-pass`, `compile-fail`, and `parse-fail` tests
```
workingjubilee, line 37: Might be good to link directly to https://rustc-dev-guide.rust-lang.org/tests/adding.html#ui here, and inline the comment from there that UI tests are usually for detecting errors in parsing or compilation. That would add a little more backpressure, nudging people backwards to "library test" if they are testing for success.

workingjubilee:
>`The command "MAX_LINE_LENGTH=100 bash ci/check_line_lengths.sh src/**/*.md" failed and exited with 1 during .`

src/tests/adding.md
jyn514: I linked to `#ui`, good idea. I'm not sure if more info needs to be added - it already says

> in the case of warnings or errors, `ui` tests capture the full output,
    which makes it easier to review but also helps prevent "hidden" regressions
    in the output

which makes it pretty clear to me.

workingjubilee: That's fair~
...wait, there's already a link to #ui just above it???
I hadn't actually processed it was in the same file... 
えーと —... It looks good to me now. :sweat_drops:

workingjubilee:
This looks good. I think I may have also wanted to point at https://rustc-dev-guide.rust-lang.org/tests/intro.html#compiletest-test-suites but that's nonessential.

workingjubilee (approved):
This looks good (though obviously I don't have the relevant privileges to make much of it). I think I may have also wanted to point at https://rustc-dev-guide.rust-lang.org/tests/intro.html#compiletest-test-suites but that's nonessential.

jyn514:
> I think I may have also wanted to point at https://rustc-dev-guide.rust-lang.org/tests/intro.html#compiletest-test-suites but that's nonessential.

That's already linked on the sidebar, I'm not sure where it would go in the text itself.

image

src/tests/adding.md
```diff
@@ -29,15 +29,25 @@ rough heuristics:
...
-  preferred:
-  - `ui` tests subsume both run-pass, compile-fail, and parse-fail tests
+- Library tests should go in `library/${crate}/tests` (where `${crate}` is
+  usually `core`, `alloc`, or `std`). Library tests include:
+  - tests that an API behaves properly, including accepting various types or
+    having some runtime behavior
+  - tests where any compiler warnings are not relevant to the test
+  - tests that a use of an API gives a compile error, where the exact error
+    message is not relevant to the test. These should have an
+    [error number](`E0XXX`) in the code block to make sure it's the correct error.
```
JohnTitor, line 42: Whitespace is needed.

```suggestion
    [error number] (`E0XXX`) in the code block to make sure it's the correct error.
```

JohnTitor (approved):

PR #884, opened 2020-09-17
debug-assertions=true instead of debug-assertions=yes

macdonaldo:


src/compiler-debugging.md
```diff
@@ -207,10 +207,10 @@ $ RUSTDOC_LOG=rustdoc rustdoc +local my-file.rs
 
 While calls to `error!`, `warn!` and `info!` are included in every build of the compiler,
 calls to `debug!` and `trace!` are only included in the program if
-`debug-assertions=yes` is turned on in config.toml (it is
+`debug-assertions=true` is turned on in config.toml (it is
```
jyn514, line 210: Good catch, thanks! Although this actually changed recently to be more specific: https://github.com/rust-lang/rust/pull/76588
```suggestion
`debug-logging=true` is turned on in config.toml (it is
```

src/compiler-debugging.md
```diff
@@ -207,10 +207,10 @@ $ RUSTDOC_LOG=rustdoc rustdoc +local my-file.rs
...
 turned off by default), so if you don't see `DEBUG` logs, especially
 if you run the compiler with `RUSTC_LOG=rustc rustc some.rs` and only see
-`INFO` logs, make sure that `debug-assertions=yes` is turned on in your
+`INFO` logs, make sure that `debug-assertions=true` is turned on in your
```
jyn514, line 213:
```suggestion
`INFO` logs, make sure that `debug-logging=true` is turned on in your
```

jyn514 (approved):

PR #885, opened 2020-09-22
`ConstEvalResult` is now `EvalToConstValueResult`

camelid:
I believe this was changed in rust-lang/rust#74949, so r? @oli-obk

### DIFF
--- a/src/const-eval.md
+++ b/src/const-eval.md
@@ -28,7 +28,7 @@ in which the constant is evaluated (e.g. the function within which the constant 
 and a [`GlobalId`]. The `GlobalId` is made up of an `Instance` referring to a constant
 or static or of an `Instance` of a function and an index into the function's `Promoted` table.
 
-Constant evaluation returns a [`ConstEvalResult`] with either the error, or the a
+Constant evaluation returns a [`EvalToConstValueResult`] with either the error, or a
 representation of the constant. `static` initializers are always represented as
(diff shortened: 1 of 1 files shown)


oli-obk (approved):

LeSeulArtichaut:
CI failure seems spurious, I just re-ran the checks

camelid:
Yeah, it was due to `429 Too Many Requests` to some sites. Is there any way to ignore 429 and related errors, and if so should we?

LeSeulArtichaut (approved):

LeSeulArtichaut:
> Yeah, it was due to `429 Too Many Requests` to some sites. Is there any way to ignore 429 and related errors, and if so should we?

I know there's been discussions about this in the past, not sure if we ever took a decision or not. I also know that we removed the `rustc-dev-guide` from toolstate because of these too

camelid:
The README says:

>Please note: You may receive errors like the following when link checks are active on local mdbook builds:
>
> ```
> error: The server responded with 429 Too Many Requests for "https://github.com/rust-lang/rust/tree/master/src/tools/compiletest"
> [...]
> ```
>
> There is not a workaround for this error at the moment. Comment out the [output.linkcheck] field in the book.toml using the build instructions above to complete a local site build without link validations.

jyn514:
More discussion here: https://rust-lang.zulipchat.com/#narrow/stream/196385-t-compiler.2Fwg-rustc-dev-guide/topic/move.20to.20gha

PR #888, opened 2020-09-23
Link to the -Zmir-opt-level proposal

bugadani:
I was confused about what the different levels mean. This PR adds a paragraph that shortly explains the rationale behind different mir-opt-levels and links to the current proposal where there's some information about them.

Let me know if I wrote something unreasonable or incorrect.

### DIFF
--- a/src/mir/optimizations.md
+++ b/src/mir/optimizations.md
@@ -36,7 +36,7 @@ implemented in its own submodule of the [`rustc_mir::transform`][trans] module.
 
 [rop]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_mir/transform/fn.run_optimization_passes.html
 [`MirPass`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_mir/transform/trait.MirPass.html
-[trans]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_mir/transform/index.html 
+[trans]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_mir/transform/index.html
 
 Some examples of passes are:
 - `CleanupNonCodegenStatements`: remove some of the info that is only needed for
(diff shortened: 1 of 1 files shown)


src/mir/passes.md
```diff
@@ -53,6 +53,18 @@ want to use a [MIR visitor]. MIR visitors are a handy way to walk all
 the parts of the MIR, either to search for something or to make small
 edits.
 
+### MIR optimization levels
+
+MIR optimizations can come in various levels of readiness. Experimental
+optimizations may cause miscompilations, or slow down compile times.
+However, these passes are complex enough to be included in nightly builds
+even in these states, to enable validating them or otherwise gather feedback.
```
jyn514, line 61: I don't think it's because they're _complex_, it's so they're in-tree and multiple people can work on them at once.
```suggestion
These passes are still included in nightly builds to gather feedback and make it easier to modify the pass.
```

src/mir/passes.md
```diff
@@ -53,6 +53,18 @@ want to use a [MIR visitor]. MIR visitors are a handy way to walk all
...
+To enable working with slow or otherwise experimental optimization passes,
+you can specify the `-Z mir-opt-level` debug flag. You can find the
+definitions of the levels [here][1]. If you are developing a MIR pass and
```
jyn514, line 64:
```suggestion
definitions of the levels in the [compiler MCP]. If you are developing a MIR pass and
```
and the link below updated to use `[compiler MCP]:` instead of [1]:

src/mir/passes.md
bugadani: I don't think I wanted to use complex there but I can't remember. Thanks for the suggestions!

bugadani:
I moved the section from passes to optimizations, maybe it has a better place there

jyn514 (approved):

PR #891, opened 2020-09-24
Remove reference to copyright notices.

JakobDegen:
Cleans up a reference to copyright notices that are not supposed to be there in the first place.


LeSeulArtichaut (approved):
Good catch, thanks!

PR #901, opened 2020-09-29
--verbose is useful when not compiling and when compiling

spastorino:
For instance https://github.com/rust-lang/rust/blob/master/compiler/rustc_mir/src/util/pretty.rs#L333


LeSeulArtichaut (approved):
Agreed, I think the previous wording didn't make much sense either.

PR #904, opened 2020-10-02
Fix instructions for squashing

camelid:
`git merge-base master` does not work, you need
`git merge-base master HEAD` instead.

r? @jyn514


camelid:
We were giving people Git instructions that don't work!

jyn514 (approved):
Good catch, thanks!

PR #905, opened 2020-10-03
Document the usage of cargo-llvm-lines and -Ztimings.

Julian-Wollersberger:
Here I document what I learned while making rust-lang/rust#76680:
* How to use cargo-llvm-lines with x.py. 
* You can do `CARGOFLAGS="-Ztimings" ./x.py whatever` to get a nice timing chart.
* Unrelated, I played around with [tracy](https://github.com/wolfpld/tracy). Would need more work though to integrate with measureme.

~~This is blocked on dtolnay/cargo-llvm-lines#32.~~

### DIFF
--- a/src/profiling.md
+++ b/src/profiling.md
@@ -12,5 +12,84 @@ Depending on what you're trying to measure, there are several different approach
     See [their docs](https://github.com/rust-lang/measureme/blob/master/summarize/Readme.md) for more information.
     
 - If you want function level performance data or even just more details than the above approaches:
-  - Consider using a native code profiler such as [perf](profiling/with_perf.html).
+  - Consider using a native code profiler such as [perf](profiling/with_perf.html) 
+  - or [tracy](https://github.com/nagisa/rust_tracy_client) for a nanosecond-precision, 
+    full-featured graphical interface.
 
+- If you want a nice visual representation of the compile times of your crate graph, 
+  you can use [cargo's `-Ztimings` flag](https://doc.rust-lang.org/cargo/reference/unstable.html#timings), 
+  eg. `cargo -Ztimings build`.
+  You can use this flag on the compiler itself with `CARGOFLAGS="-Ztimings" ./x.py build`
+  
+## Optimizing rustc's self-compile-times with cargo-llvm-lines
+
+Using [cargo-llvm-lines](https://github.com/dtolnay/cargo-llvm-lines) you can count the 
+number of lines of LLVM IR across all instantiations of a generic function.
+Since most of the time compiling rustc is spent in LLVM, the idea is that by
+reducing the amount of code passed to LLVM, compiling rustc gets faster.
+
+Example usage:
+```
+cargo install cargo-llvm-lines
+# On a normal crate you could now run `cargo llvm-lines`, but x.py isn't normal :P
+
+# Do a clean before every run, to not mix in the results from previous runs.
+./x.py clean
+RUSTFLAGS="--emit=llvm-ir" ./x.py build --stage 0 compiler/rustc
+
+# Single crate, eg. rustc_middle
+cargo llvm-lines --files ./build/x86_64-unknown-linux-gnu/stage0-rustc/x86_64-unknown-linux-gnu/debug/deps/rustc_middle* > llvm-lines-middle.txt
+# Whole compiler at once
+cargo llvm-lines --files ./build/x86_64-unknown-linux-gnu/stage0-rustc/x86_64-unknown-linux-gnu/debug/deps/*.ll > llvm-lines.txt
+```
+
+Example output:
+```
+  Lines            Copies        Function name
+  -----            ------        -------------
+  11802479 (100%)  52848 (100%)  (TOTAL)
+   1663902 (14.1%)   400 (0.8%)  rustc_query_system::query::plumbing::get_query_impl::{{closure}}
+    683526 (5.8%)  10579 (20.0%) core::ptr::drop_in_place
+    568523 (4.8%)    528 (1.0%)  rustc_query_system::query::plumbing::get_query_impl
+    472715 (4.0%)   1134 (2.1%)  hashbrown::raw::RawTable<T>::reserve_rehash
+    306782 (2.6%)   1320 (2.5%)  rustc_middle::ty::query::plumbing::<impl rustc_query_system::query::QueryContext for rustc_middle::ty::context::TyCtxt>::start_query::{{closure}}::{{closure}}::{{closure}}
+    212800 (1.8%)    514 (1.0%)  rustc_query_system::dep_graph::graph::DepGraph<K>::with_task_impl
+    194813 (1.7%)    124 (0.2%)  rustc_query_system::query::plumbing::force_query_impl
+    158488 (1.3%)      1 (0.0%)  rustc_middle::ty::query::<impl rustc_middle::ty::context::TyCtxt>::alloc_self_profile_query_strings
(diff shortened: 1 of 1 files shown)


src/profiling.md
```diff
@@ -12,5 +12,84 @@ Depending on what you're trying to measure, there are several different approach
...
+  - Consider using a native code profiler such as [perf](profiling/with_perf.html) 
+  - or [tracy](https://github.com/nagisa/rust_tracy_client) for a nanosecond-precision, 
+    full-featured graphical interface.
 
+- If you want a nice visual representation of the compile times of your crate graph, 
+  you can use [cargo's `-Ztimings` flag](https://doc.rust-lang.org/cargo/reference/unstable.html#timings), 
+  eg. `cargo -Ztimings build`.
+  You can use this flag on the compiler itself with `CARGOFLAGS="-Ztimings" ./x.py build`
+  
+## Optimizing rustc's self-compile-times with cargo-llvm-lines
```
jyn514, line 24:
```suggestion
## Optimizing rustc's bootstrap times with `cargo-llvm-lines`
```

src/profiling.md
```diff
@@ -12,5 +12,84 @@ Depending on what you're trying to measure, there are several different approach
...
+reducing the amount of code passed to LLVM, compiling rustc gets faster.
+
+Example usage:
+```
+cargo install cargo-llvm-lines
+# On a normal crate you could now run `cargo llvm-lines`, but x.py isn't normal :P
+
+# Do a clean before every run, to not mix in the results from previous runs.
+./x.py clean
+RUSTFLAGS="--emit=llvm-ir" ./x.py build --stage 0 compiler/rustc
```
jyn514, line 38: Hmm, does this work? I thought you needed to `export RUSTFLAGS` for it to be passed to child processes, but maybe bootstrap handles it specially.

src/profiling.md
```diff
@@ -12,5 +12,84 @@ Depending on what you're trying to measure, there are several different approach
...
+
+# Caution: This changes the output of llvm-lines.
+# Using a single codegen unit gives more accurate output, but is slower to compile.
+# Changing it to the number of cores on my machine increased the output 
+# from 3.5GB to 4.1GB and decreased compile times from 5½ min to 4 min.
+codegen-units = 1
+#codegen-units = 0 # num_cpus
+```
+
+What I'm still not sure about is if inlining in MIR optimizations affect llvm-lines.
```
jyn514, line 92:
```suggestion
TODO: does inlining in MIR optimizations affect llvm-lines?
```

src/profiling.md
```diff
@@ -12,5 +12,84 @@ Depending on what you're trying to measure, there are several different approach
...
+The output with `-Zmir-opt-level=0` and `-Zmir-opt-level=1` is the same,
```
jyn514, line 93: Did you try `-Zmir-opt-level=2`? This might also be related to `optimize=false`.

src/profiling.md
```diff
@@ -12,5 +12,84 @@ Depending on what you're trying to measure, there are several different approach
...
+but it feels like that some functions that show up at the top should be to small
```
jyn514, line 94:
```suggestion
but it feels like that some functions that show up at the top should be too small
```

src/profiling.md
```diff
@@ -12,5 +12,84 @@ Depending on what you're trying to measure, there are several different approach
...
+to have such a high impact. Inlining should only happens in LLVM though.
```
jyn514, line 95: There are plans to eventually do inlining in mir-opts, but I don't think they're implemented yet.
```suggestion
to have such a high impact. Inlining only currently happens in LLVM though.
```

src/profiling.md
```diff
@@ -12,5 +12,84 @@ Depending on what you're trying to measure, there are several different approach
...
+```
+
+Since this doesn't seem to work with incremental compilation or `x.py check`, 
+you will be compiling rustc _a lot_.
+I recommend changing a few settings in `config.toml` to make it bearable:
+```
+[rust]
+# A debug build takes _a fourth_ as long on my machine, 
+# but compiling more than stage0 rustc becomes unbearably slow.
+optimize = false
```
jyn514, line 77: Nice trick. Does this affect llvm-lines at all? Maybe it's a good thing if it does since it reflects how much LLVM IR is getting sent in the first place, not how small LLVM can make it with optimizations.

src/profiling.md
```diff
@@ -12,5 +12,84 @@ Depending on what you're trying to measure, there are several different approach
...
+# Whole compiler at once
+cargo llvm-lines --files ./build/x86_64-unknown-linux-gnu/stage0-rustc/x86_64-unknown-linux-gnu/debug/deps/*.ll > llvm-lines.txt
+```
+
+Example output:
+```
+  Lines            Copies        Function name
+  -----            ------        -------------
+  11802479 (100%)  52848 (100%)  (TOTAL)
+   1663902 (14.1%)   400 (0.8%)  rustc_query_system::query::plumbing::get_query_impl::{{closure}}
```
jyn514, line 51: This seems like another good place to use `dyn` or just separate out most of the function. That's a _lot_ of LLVM lines for only 400 instantiations.

camelid:
> This is blocked on [dtolnay/cargo-llvm-lines#32](https://github.com/dtolnay/cargo-llvm-lines/pull/32).

That PR was merged!

src/profiling.md
camelid: I thought `ENV_VAR='abc' foo` was the same as

```sh
$ export ENV_VAR='abc'
$ foo
```

except that it was only for that command. But maybe I don't understand my shell ;)

src/profiling.md
camelid: Hmm, I think it might be worded better as:

```suggestion
to have such a high impact. Inlining currently only happens in LLVM though.
```

src/profiling.md
jyn514: Export passes it to processed that `foo` spawns, using it inline does not. Your example is only the same if foo does not spawn child processes, but bootstrap launches lots of them.

jyn514: Or at least, that seems to have been the behavior when I tried it out. Maybe I'm just misunderstanding how bootstrap works.

Julian-Wollersberger: I always do it this way, so it works.  I guess bootstrap handles it, like it does with `CARGOFLAGS`.

src/profiling.md
Julian-Wollersberger: Apparently yes.
I did some more experimenting and the interactions between the different flags are more complicated than I thought. I'll rewrite the last paragraph.

src/profiling.md
Julian-Wollersberger: Probably. I haven't had much time yet to look into it, but it's tricky because the relevant traits aren't dyn safe.

src/profiling.md
Julian-Wollersberger: Tried it now and the difference is -0.2GB, but it doesn't seem to inline anything. [This discussion](https://rust-lang.zulipchat.com/#narrow/stream/189540-t-compiler.2Fwg-mir-opt/topic/MIR-level.20inlining) says MIR inlining doesn't work yet.

src/profiling.md
Julian-Wollersberger: Removed this paragraph.

Julian-Wollersberger:
@jyn514 I've replaced the paragraph about mir-opt. 

@rustbot modify labels +waiting-on-review -waiting-on-author

jyn514 (approved):
This looks great. Thank you so much for working on this!

PR #910, opened 2020-10-04
Allow anyone to change the status of a PR

LeSeulArtichaut:
Allows anyone to change the status of PRs by using the `@rustbot modify labels` command, so [this](https://github.com/rust-lang/rustc-dev-guide/pull/905#issuecomment-703255140) is allowed now.

### DIFF
--- a/triagebot.toml
+++ b/triagebot.toml
@@ -1,1 +1,8 @@
 [assign]
+
+[relabel]
+allow-unauthenticated = [
+    "waiting-on-review",
+    "waiting-on-author",
+    "blocked",
+]


jyn514:
Is there a way to limit this only to the author of the PR? I don't want random people coming along and changing the status.

LeSeulArtichaut:
Pretty sure that's not possible, but the same thing is allowed on `rust-lang/rust` and I have never seen any problem (and I think doing that there would bring much more chaos!)

jyn514 (approved):
Fair enough, I definitely want people to be able to add 'waiting-on-review'.

LeSeulArtichaut:
@jyn514 Should we also add an "auto-labeler" to add `waiting-on-review` to all PRs when opened?

jyn514:
Sounds good to me!

PR #913, opened 2020-10-05
Linkcheck only changed files (except for cron jobs)

### LINKED ISSUE #886: Only run linkcheck on changed files
From @mark-i-m in [this Zulip thread](https://rust-lang.zulipchat.com/#narrow/stream/196385-t-compiler.2Fwg-rustc-dev-guide/topic/move.20to.20gha/near/209911752):

> It doesn't sound too hard to add a way for mdbook-linkcheck to only check a list of given files. Then, we could have the CI script list the changed files from git as those arguments and only check the changed files. And we could have a cron job that runs the full check periodically (like we already do).

This would probably reduce the number of the annoying 429 Too Many Requests errors.

mark-i-m:
Closes #886

In this PR, we change the CI setup to only check files that have changed in the
most recent git commit. This means that changes in PRs are less likely to get
blocked on unrelated breakage (it's still possible if that breakage happens to
be in the same file).

On the other hand, we do want to periodically check all files. Currently, there
is a cron job that runs every day. We change the CI script to check if the
Travis run is a cron job and if so do a full link check. The hope is that by
having the cron job run often enough and preserving the linkcheck cache, we can
continue to find breakage relatively soon while avoiding the 429 Too Many
Requests issues we've been seeing.

This is currently blocked on a new mdbook-linkcheck release.

### DIFF
--- a/ci/linkcheck.sh
+++ b/ci/linkcheck.sh
@@ -10,7 +10,7 @@ if [ "$TRAVIS_EVENT_TYPE" = "cron" ] ; then # running in cron job
   echo "Doing full link check."
 elif [ "$CI" = "true" ] ; then # running in PR CI build
   if [ -z "$TRAVIS_COMMIT_RANGE" ]; then
-    echo "error: unexpected state: COMMIT_RANGE must be non-empty in CI"
+    echo "error: unexpected state: TRAVIS_COMMIT_RANGE must be non-empty in CI"
     exit 1
   fi
 
--- a/ci/linkcheck.sh
+++ b/ci/linkcheck.sh
@@ -9,7 +9,7 @@ if [ "$TRAVIS_EVENT_TYPE" = "cron" ] ; then # running in cron job
 
   echo "Doing full link check."
 elif [ "$CI" = "true" ] ; then # running in PR CI build
-  if [ -n "$TRAVIS_COMMIT_RANGE" ]; then
+  if [ -z "$TRAVIS_COMMIT_RANGE" ]; then
     echo "error: unexpected state: COMMIT_RANGE must be non-empty in CI"
     exit 1
   fi
--- a/ci/linkcheck.sh
+++ b/ci/linkcheck.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+
+if [ "$TRAVIS_EVENT_TYPE" = "cron" ] ; then # running in cron job
+  FLAGS=""
+
+  echo "Doing full link check."
+elif [ "$CI" = "true" ] ; then # running in PR CI build
+  CHANGED_FILES=$(git diff --name-only $TRAVIS_COMMIT_RANGE | tr '\n' ' ')
+  FLAGS="-f $CHANGED_FILES"
+
+  echo "Checking files changed in $TRAVIS_COMMIT_RANGE: $CHANGED_FILES"
+else # running locally
+  COMMIT_RANGE=master...
+  CHANGED_FILES=$(git diff --name-only $COMMIT_RANGE | tr '\n' ' ')
+  FLAGS="-f $CHANGED_FILES"
+
+  echo "Checking files changed in $COMMIT_RANGE: $CHANGED_FILES"
+fi
+
+exec mdbook-linkcheck $FLAGS -- $TRAVIS_BUILD_DIR
--- a/ci/linkcheck.sh
+++ b/ci/linkcheck.sh
@@ -9,6 +9,11 @@ if [ "$TRAVIS_EVENT_TYPE" = "cron" ] ; then # running in cron job
 
   echo "Doing full link check."
 elif [ "$CI" = "true" ] ; then # running in PR CI build
+  if [ -n "$TRAVIS_COMMIT_RANGE" ]; then
+    echo "error: unexpected state: COMMIT_RANGE must be non-empty in CI"
+    exit 1
+  fi
+
   CHANGED_FILES=$(git diff --name-only $TRAVIS_COMMIT_RANGE | tr '\n' ' ')
   FLAGS="-f $CHANGED_FILES"
 
--- a/ci/linkcheck.sh
+++ b/ci/linkcheck.sh
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+set -ev
+set -o pipefail
+
 # https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
 if [ "$TRAVIS_EVENT_TYPE" = "cron" ] ; then # running in cron job
   FLAGS=""
--- a/ci/linkcheck.sh
+++ b/ci/linkcheck.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 
+# https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
 if [ "$TRAVIS_EVENT_TYPE" = "cron" ] ; then # running in cron job
   FLAGS=""
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ before_install:
 install:
 - source ~/.cargo/env || true
 - cargo install mdbook --version '^0.4.3'
-- cargo install mdbook-linkcheck --version '^0.7.0'
+- cargo install mdbook-linkcheck --git https://github.com/Michael-F-Bryan/mdbook-linkcheck
 script:
 - git checkout -b ci
 - git rebase origin/master
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ before_install:
 install:
 - source ~/.cargo/env || true
 - cargo install mdbook --version '^0.4.3'
-- cargo install mdbook-linkcheck --git https://github.com/Michael-F-Bryan/mdbook-linkcheck
+- cargo install mdbook-linkcheck --git https://github.com/Michael-F-Bryan/mdbook-linkcheck --rev 14441d77646d58cea8ffc32fde9ea33b2bedb1a2
 script:
 - git checkout -b ci
 - git rebase origin/master
--- a/book.toml
+++ b/book.toml
@@ -14,6 +14,7 @@ enable = true
 level = 0
 
 [output.linkcheck]
+command = "../../ci/linkcheck.sh"
 follow-web-links = true
 exclude = [ "crates\\.io", "gcc\\.godbolt\\.org", "youtube\\.com", "youtu\\.be", "dl\\.acm\\.org", "cs\\.bgu\\.ac\\.il", "www\\.amazon\\.com", "www\\.rustaceans\\.org" ]
 cache-timeout = 86400


ci/linkcheck.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+if [ "$TRAVIS_EVENT_TYPE" = "cron" ] ; then
+  echo "Doing full link check."
+  FLAGS=""
+else
+  echo "Checking files changed in commit range: $TRAVIS_COMMIT_RANGE"
+  CHANGED_FILES=$(git diff-tree --no-commit-id --name-only -r $TRAVIS_COMMIT_RANGE | tr '\n' ' ')
```
jyn514, line 8: Won't this break when running linkcheck locally? It would be nice to use `$(git merge-base HEAD master)` when `TRAVIS_COMMIT_RANGE` is unset.

ci/linkcheck.sh
```diff
@@ -1,14 +1,14 @@
...
-  echo "Doing full link check."
   FLAGS=""
-else
-  echo "Checking files changed in commit range: $TRAVIS_COMMIT_RANGE"
-  CHANGED_FILES=$(git diff-tree --no-commit-id --name-only -r $TRAVIS_COMMIT_RANGE | tr '\n' ' ')
 
-  echo "Checking only changed files: $CHANGED_FILES"
+  echo "Doing full link check."
+else
+  CHANGED_FILES=$(git diff-tree --no-commit-id --name-only -r master..HEAD | tr '\n' ' ')
```
jyn514, line 8: Sorry, I wasn't clear - this won't work in travis because it will be the master of _your_ fork, not the master of the upstream codebase. So in CI it still needs to use TRAVIS_COMMIT_RANGE. https://gitlab.com/YottaDB/DB/YDB/-/merge_requests/840#note_418423834

mark-i-m: Even though we rebase first?

https://github.com/rust-lang/rustc-dev-guide/blob/d4581a56ae7b16c258b7e43b21d9caacc66e9ff4/.travis.yml#L18

jyn514: Hmm, are you sure `origin` here is `rust-lang/rustc-dev-guide` and not the fork? If so this seems fine.

mark-i-m: That said, I'm not super confident...

ci/linkcheck.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+if [ "$TRAVIS_EVENT_TYPE" = "cron" ] ; then
+  FLAGS=""
+
+  echo "Doing full link check."
+else
+  CHANGED_FILES=$(git diff-tree --no-commit-id --name-only -r master..HEAD | tr '\n' ' ')
```
camelid, line 8: Hmm, I think you want `master...HEAD` (three dots, not two) so that it shows a diff from the point where the branches diverged. What you have right now will diff from the tip of master.

```suggestion
  CHANGED_FILES=$(git diff-tree --no-commit-id --name-only -r master...HEAD | tr '\n' ' ')
```

ci/linkcheck.sh
```diff
@@ -0,0 +1,14 @@
...
+  FLAGS="-f $CHANGED_FILES"
+
+  echo "Checking files changed in master..HEAD: $CHANGED_FILES"
```
camelid, line 11: Same here.

```suggestion
  echo "Checking files changed in master...HEAD: $CHANGED_FILES"
```

camelid (changes requested):

camelid:
```console
$ cargo install mdbook-linkcheck --version '^0.7.1'

    Updating crates.io index

error: could not find `mdbook-linkcheck` in registry `https://github.com/rust-lang/crates.io-index` with version `^0.7.1`

The command "cargo install mdbook-linkcheck --version '^0.7.1'" failed and exited with 101 during .
```

Are we waiting on an mdbook-linkcheck release?

mark-i-m:
@camelid Yes, and also good catch on the endpoint... I misread the docs.

ci/linkcheck.sh
mark-i-m: I just changed it to origin/master, so I think this should work properly, but let's just watch the first few PRs and make sure that the right set of files is checked.

camelid:
We could use a commit hash version of mdbook-linkcheck until there's a point release; that way we can test to make sure the setup is working properly. What do you think?

ci/linkcheck.sh
mark-i-m: Hmm... well, it seems like it didn't work...

mark-i-m:
Hmm... somehow we are still messing up the git commit ranges, so there is something I don't understand. However, hopefully the current version works correctly.

ci/linkcheck.sh
```diff
@@ -1,14 +1,20 @@
...
-else
-  CHANGED_FILES=$(git diff-tree --no-commit-id --name-only -r origin/master... | tr '\n' ' ')
+elif [ "$CI" = "true" ] ; then # running in PR CI build
+  CHANGED_FILES=$(git diff-tree --no-commit-id --name-only -r $TRAVIS_COMMIT_RANGE | tr '\n' ' ')
   FLAGS="-f $CHANGED_FILES"
 
-  echo "Checking files changed in origin/master...: $CHANGED_FILES"
+  echo "Checking files changed in $TRAVIS_COMMIT_RANGE: $CHANGED_FILES"
+else # running locally
+  COMMIT_RANGE=master..
```
jyn514, line 13:
```suggestion
  COMMIT_RANGE=$(git merge-base master HEAD)..
```

ci/linkcheck.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+if [ "$TRAVIS_EVENT_TYPE" = "cron" ] ; then # running in cron job
+  FLAGS=""
+
+  echo "Doing full link check."
+elif [ "$CI" = "true" ] ; then # running in PR CI build
+  CHANGED_FILES=$(git diff-tree --no-commit-id --name-only -r $TRAVIS_COMMIT_RANGE | tr '\n' ' ')
```
camelid, line 8: This works locally for me:

```console
❯ git diff --name-only upstream/master...
.travis.yml
book.toml
ci/linkcheck.sh
```

ci/linkcheck.sh
```diff
@@ -0,0 +1,21 @@
...
+  echo "Doing full link check."
+elif [ "$CI" = "true" ] ; then # running in PR CI build
+  COMMIT_RANGE=$(echo $TRAVIS_COMMIT_RANGE | sed 's/\.\.\./../g')
+  CHANGED_FILES=$(git diff-tree --no-commit-id --name-only -r $COMMIT_RANGE | tr '\n' ' ')
+  FLAGS="-f $CHANGED_FILES"
+
+  echo "Checking files changed in $TRAVIS_COMMIT_RANGE: $CHANGED_FILES"
+else # running locally
+  COMMIT_RANGE=master..
+  CHANGED_FILES=$(git diff-tree --no-commit-id --name-only -r $COMMIT_RANGE | tr '\n' ' ')
```
camelid, line 15:
```suggestion
  COMMIT_RANGE=origin/master...
  CHANGED_FILES=$(git diff --name-only $COMMIT_RANGE | tr '\n' ' ')
```

camelid: `git diff-tree` doesn't seem to be working locally, while `git diff` is.

mark-i-m:
@jyn514 @camelid The latest commit pulls out the files I would expect it to... Now the question is: is it correct? You both suggested simpler alternatives. Could you explain why they are more correct than what we have now?

camelid:
`git diff` is the plain diff command, and we're trying to do diffing, so that seems correct. Also it works locally for me, while `diff-tree` does not (it did before though...). Also, you should have three dots `...` so that it does it from the merge-base instead of from the tip of master. `jyn514`'s recommendation should work as well, but it's more verbose.

ci/linkcheck.sh
mark-i-m: Interesting... it looks like `git diff` parses ranges with different syntax from `git diff-tree`...

@camelid Any reason for including `origin/master`, rather than just `master`? For me, my local master branch is more up-to-date than my origin/master branch, so using `origin/master` makes it think that nearly every file has changed...

ci/linkcheck.sh
mark-i-m: Why?

ci/linkcheck.sh
camelid: Ah, I forgot that that's for the local build only. I'll update my suggestion :)

ci/linkcheck.sh
camelid: The change to use `...` means this should no longer be necessary.

mark-i-m:
Indeed, it does seem to work properly this way.

mark-i-m:
Ok, I've squashed... I'm actually ok with just merging this, even though it will cause us to use master mdbook-linkcheck.

camelid:
Do you want to pin to a particular commit though?

mark-i-m:
Ah, yeah, that's a good idea.

camelid:
I think this is ready! 🎉

LeSeulArtichaut (approved):
This is clear improvement for the CI workflow, though it would be awesome if we could somehow get rid of GitHub's ratelimits.
I don't really feel comfortable approving the shell script though, I'd prefer if someone else reviews it

LeSeulArtichaut:
@mark-i-m @camelid If you are confident the script works I think we should merge this

camelid:
I wouldn't say I'm "confident" - more than I'm pretty sure it'll work, and we have a regular cronjob to check all the links anyway. Maybe @jyn514 could review the script?

camelid:
Also: worst-case scenario, we have a couple broken links :)

LeSeulArtichaut:
Right. I also think this is pretty low-risk, but since we have very few PRs open right now, this can wait few hours/days for someone else to review

ci/linkcheck.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+if [ "$TRAVIS_EVENT_TYPE" = "cron" ] ; then # running in cron job
```
jyn514, line 3:
```suggestion
# https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
if [ "$TRAVIS_EVENT_TYPE" = "cron" ] ; then # running in cron job
```

ci/linkcheck.sh
```diff
@@ -0,0 +1,20 @@
...
+  FLAGS=""
+
+  echo "Doing full link check."
+elif [ "$CI" = "true" ] ; then # running in PR CI build
+  CHANGED_FILES=$(git diff --name-only $TRAVIS_COMMIT_RANGE | tr '\n' ' ')
```
jyn514, line 8: > TRAVIS_COMMIT_RANGE: The range of commits that were included in the push or pull request. (Note that this is empty for builds triggered by the initial commit of a new branch.)

After reading over the docs I think this means for 'push' runs, not 'pr' runs, so this is probably fine, but can you add an assert just in case?
```suggestion
  if ! [ "$TRAVIS_EVENT_TYPE" = push ] || [ -n "$TRAVIS_COMMIT_RANGE" ]; then
    echo "error: unexpected state: COMMIT_RANGE must be non-empty in CI"
    exit 1
  fi
  CHANGED_FILES=$(git diff --name-only $TRAVIS_COMMIT_RANGE | tr '\n' ' ')
```

ci/linkcheck.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
```
jyn514, line 1:
```suggestion
#!/usr/bin/env bash

set -ev
set -o pipefail
```

mark-i-m: What do these do?

ci/linkcheck.sh
mark-i-m: I'm not sure I understand the concern... wouldn't `[ -n "$TRAVIS_COMMIT_RANGE" ]` be sufficient?

ci/linkcheck.sh
jyn514: `set -e` exits on error. `set -v` prints each command as it's executed, which makes it easier to debug. `pipefail` counts the command as failing if any command in pipeline fails; the default is it will have the status of the last command in the pipeline, which can ignore errors.

ci/linkcheck.sh
jyn514: No, because it's expected that this will be empty on the first push in a new branch. So people will get spurious emails about the PR failing even though nothing went wrong.

mark-i-m: Hmm... As I understand it, neither the range nor the environment variable will be empty for push or pr runs... that is, if you either create a new PR or push commits to a PR, TRAVIS_COMMIT_RANGE will exist and contain a non-empty range with all changes back to the merge-base with master...

Thus, I believe the `! [ "$TRAVIS_EVENT_TYPE" = push ]` is incorrect?

jyn514: I'm not _sure_ that's what the docs say, but this will fail noisily if it fails so I'm fine with removing the `push` check.

camelid: Do we even need to use `$TRAVIS_COMMIT_RANGE`? Can't we just use `git diff --name-only master...` like for running locally?

jyn514: @camelid no, because we have no guarentee `master` on travis is up-to-date with master on github. We don't even know if the remote is `rust-lang/` or the contributors fork (I think normally the fork?)

mark-i-m: Ok, I have added the check.

jyn514:
```
error: unexpected state: COMMIT_RANGE must be non-empty in CI
```
I'm glad we added the assert! Not sure why it's failing though.

ci/linkcheck.sh
```diff
@@ -0,0 +1,29 @@
...
+set -o pipefail
+
+# https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
+if [ "$TRAVIS_EVENT_TYPE" = "cron" ] ; then # running in cron job
+  FLAGS=""
+
+  echo "Doing full link check."
+elif [ "$CI" = "true" ] ; then # running in PR CI build
+  if [ -n "$TRAVIS_COMMIT_RANGE" ]; then
+    echo "error: unexpected state: COMMIT_RANGE must be non-empty in CI"
```
jyn514, line 13: oh oops, you meant to add a `!`. But at that point we might as well use `-z`:
```suggestion
  if [ -z "$TRAVIS_COMMIT_RANGE" ]; then
    echo "error: unexpected state: COMMIT_RANGE must be non-empty in CI"
```
```
    String operators:
    
      -z STRING      True if string is empty.
    
      -n STRING
         STRING      True if string is not empty.
```

ci/linkcheck.sh
```diff
@@ -0,0 +1,29 @@
...
+set -ev
+set -o pipefail
+
+# https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
+if [ "$TRAVIS_EVENT_TYPE" = "cron" ] ; then # running in cron job
+  FLAGS=""
+
+  echo "Doing full link check."
+elif [ "$CI" = "true" ] ; then # running in PR CI build
+  if [ -n "$TRAVIS_COMMIT_RANGE" ]; then
```
camelid, line 12: I'm pretty sure this is backwards:

```suggestion
  if [ -z "$TRAVIS_COMMIT_RANGE" ]; then
```

ci/linkcheck.sh
camelid, line 13:
```suggestion
    echo "error: unexpected state: TRAVIS_COMMIT_RANGE must be non-empty in CI"
```

camelid (changes requested):

ci/linkcheck.sh
camelid: Oops, I left the same comment :)

jyn514:
Thanks so much for tackling this!

PR #915, opened 2020-10-07
Talk about git submodules

LeSeulArtichaut:
Helps with #850.
cc @jyn514

### DIFF
--- a/src/git.md
+++ b/src/git.md
@@ -73,6 +73,30 @@ the same, with some steps skipped:
  2. Make, stage, and commit your additional changes just like before.
  3. Push those changes to your fork: `git push`.
 
+### Quick note about submodules
+
+When updating your local repository with `git pull`, you may notice that sometimes
+Git says you have modified some files that you have never edited. For example,
+running `git status` gives you something like (note the `new commits` mention):
+
+```
+On branch master
+Your branch is up to date with 'origin/master'.
+
+Changes not staged for commit:
+  (use "git add <file>..." to update what will be committed)
+  (use "git restore <file>..." to discard changes in working directory)
+	modified:   src/tools/cargo (new commits)
+	modified:   src/tools/rls (new commits)
+	modified:   src/tools/rustfmt (new commits)
+
+no changes added to commit (use "git add" and/or "git commit -a")
+```
+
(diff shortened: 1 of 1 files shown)


src/git.md
```diff
@@ -73,6 +73,30 @@ the same, with some steps skipped:
...
+	modified:   src/tools/cargo (new commits)
+	modified:   src/tools/rls (new commits)
+	modified:   src/tools/rustfmt (new commits)
+
+no changes added to commit (use "git add" and/or "git commit -a")
+```
+
+These changes are not changes to files: they are changes to submodules
+(more on this later). To get rid of those, run `git submodule update` (or run any
+`x.py` command which will automatically update the submodules).
```
jyn514, line 98:
```suggestion
`x.py` command, which will automatically update the submodules).
```
Before it sounded like you had to know which commands would update the modules, but in reality any command will work.

src/git.md
```diff
@@ -73,6 +73,30 @@ the same, with some steps skipped:
...
+
```
jyn514, line 99: Github suggestions don't like nested triple quotes, but can you add this?

Note that there is currently a bug if you use `x.py`, worktrees, and submodules. If you run into an error like
```
error: failed to read `/home/joshua/rustc-worktree/src/tools/miri/cargo-miri/Cargo.toml`

Caused by:
  No such file or directory (os error 2)
```
it's not anything you did wrong. There is a workaround at [#77620](https://github.com/rust-lang/rust/issues/77620#issuecomment-705228229).

jyn514 (commented):
cc @thomcc - does this make you a little more comfortable with submodules?

src/git.md
camelid: @jyn514 Something I learned: you _can_ nest triple backticks if you use triple tildes as the outer delimiter:

~~~suggestion
```
foo
```
~~~

^ just for demonstration, don't apply!

jyn514: Works great, thanks!

LeSeulArtichaut: :o That's super helful, thank you ❤️

jyn514 (approved):
LGTM, we can make improvements in follow-ups.

PR #919, opened 2020-10-10
Update installation instructions for `mdbook-linkcheck`

LeSeulArtichaut:
Following up #913, cc @mark-i-m @jyn514


jyn514 (approved):
Seems fine, but ideally we would get linkcheck to publish soon. @Michael-F-Bryan is there any way I can help with that?

LeSeulArtichaut:
> Seems fine, but ideally we would get linkcheck to publish soon.

But it's a shame to have "broken" installation instructions until then

PR #920, opened 2020-10-11
Use official mdbook-linkcheck version

camelid:
We were using an alpha version before, pinned to a Git hash, so that we
could access a new feature (Michael-F-Bryan/mdbook-linkcheck#42).

Now there is a point release though (0.7.1), so we can update to that!

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,10 +103,10 @@ We use `mdbook-linkcheck` to validate URLs included in our documentation. To per
 > cargo install mdbook-linkcheck
 ```
 
-You will need `mdbook` version `>= 0.3.5` and `mdbook-linkcheck` version `>= 0.7.1` to check links.
+You will need recent versions of `mdbook` and `mdbook-linkcheck` to check links.
 `linkcheck` will be run automatically when you build with the instructions in the section above.
 
-**Please note**: You may receive errors like the following when link checks are active on local `mdbook` builds:
+**Please note**: You may receive errors like the following when link-checking is enabled on local `mdbook` builds:
 
 ```
 error: The server responded with 429 Too Many Requests for "https://github.com/rust-lang/rust/tree/master/src/tools/compiletest"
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -100,11 +100,10 @@ New-Item -Path .git/hooks/pre-commit -ItemType HardLink -Value <absolute_path/to
 We use `mdbook-linkcheck` to validate URLs included in our documentation. To perform link checks, uncomment the `[output.linkcheck]` field in the `book.toml` configuration file and install `mdbook-linkcheck` with:
 
 ```bash
-> cargo install mdbook-linkcheck --git https://github.com/Michael-F-Bryan/mdbook-linkcheck --rev 14441d77646d58cea8ffc32fde9ea33b2bedb1a2
+> cargo install mdbook-linkcheck
 ```
 
-Note that we use an alpha version of `mdbook-linkcheck` to be able to use a feature that hasn't landed in a release yet.
-You will also need `mdbook` version `>= 0.3.5`.
+You will need `mdbook` version `>= 0.3.5` and `mdbook-linkcheck` version `>= 0.7.1` to check links.
```
camelid, line 106: This was here before, but are we sure it still works on `mdbook 0.3.5`? Maybe we should just say "you will need a recent version".

README.md
```diff
@@ -100,11 +100,10 @@ New-Item -Path .git/hooks/pre-commit -ItemType HardLink -Value <absolute_path/to
...
 `linkcheck` will be run automatically when you build with the instructions in the section above.
 
 **Please note**: You may receive errors like the following when link checks are active on local `mdbook` builds:
```
jyn514, line 109: Is this still relevant now that it's not checking all links?

README.md
jyn514: Sounds good to me.

README.md
mark-i-m: I think it's still theoretically possible (e.g. if a file has a ton of links or if you touch a bunch of files). I thinks it's ok to keep this.

README.md
mark-i-m: We could also just update the number to the most recent version...

README.md
camelid: Yes, I thought of removing it too, but kept it for the reasons mark mentioned.

README.md
camelid: But then it will become out of date :)

jyn514 (approved):

PR #921, opened 2020-10-12
Added description of word Scrutinee to the glossary

### LINKED ISSUE #796: Documenting what "scrutinee" mean specifically
I see the word "scrutinee" used in Rust source, but I haven't never known its meaning.
So it is hard to remember and know when/why use that word.

1c3t3a:
Closes #796.

### DIFF
--- a/src/appendix/glossary.md
+++ b/src/appendix/glossary.md
@@ -64,6 +64,7 @@ query <div id="query"/>                  |  Perhaps some sub-computation during 
 recovery <div id="recovery"/>            |  Recovery refers to handling invalid syntax during parsing (e.g. a missing comma) and continuing to parse the AST. This avoid showing spurious errors to the user (e.g. showing 'missing field' errors when the struct definition contains errors).
 region <div id="region"/>                |  Another term for "lifetime" often used in the literature and in the borrow checker.
 rib <div id="rib"/>                      |  A data structure in the name resolver that keeps track of a single scope for names. ([see more](../name-resolution.html))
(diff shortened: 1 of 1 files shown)


src/appendix/glossary.md
```diff
@@ -64,6 +64,7 @@ query <div id="query"/>                  |  Perhaps some sub-computation during
 recovery <div id="recovery"/>            |  Recovery refers to handling invalid syntax during parsing (e.g. a missing comma) and continuing to parse the AST. This avoid showing spurious errors to the user (e.g. showing 'missing field' errors when the struct definition contains errors).
 region <div id="region"/>                |  Another term for "lifetime" often used in the literature and in the borrow checker.
 rib <div id="rib"/>                      |  A data structure in the name resolver that keeps track of a single scope for names. ([see more](../name-resolution.html))
+Scrutinee                                |  A scrutinee is the expression that is matched on in `match` expressions and similar pattern matching constructs. For example, in `match x { A => 1, B => 2 }`, the expression `x` is the scrutinee.
```
JohnTitor, line 67:
```suggestion
scrutinee                                |  A scrutinee is the expression that is matched on in `match` expressions and similar pattern matching constructs. For example, in `match x { A => 1, B => 2 }`, the expression `x` is the scrutinee.
```

I don't think this capitalization is necessary?

JohnTitor (approved):
Thanks!

jyn514 (approved):

PR #922, opened 2020-10-16
Fix a link on git.md

JohnTitor:


jyn514 (approved):
Oops, good catch, thanks.

PR #926, opened 2020-10-20
Remove outdated build rustdoc --stage 1

pickfire:
It builds with stage 1 by default.

r? @jyn514 for correctness


jyn514 (approved):
The only time these might  be different is if you set `doc-stage` or `profile = "library"` (possibly using x.py setup). But if you're working on rustdoc itself I wouldn't expect you to do either of those things; if you opt-out like that I'm ok with saying you should opt back in.

PR #938, opened 2020-10-27
Add missing closing '/' in tags

camelid:


src/appendix/glossary.md
```diff
@@ -16,7 +16,7 @@ cx <div id="cx"/>                        |  We tend to use "cx" as an abbreviati
 ctxt <div id="ctxt"/>                    |  We also use "ctxt" as an abbreviation for context, e.g. [`TyCtxt`](TyCtxt). See also [cx](#cx) or [tcx](#tcx).
 DAG <div id="dag"/>                      |  A directed acyclic graph is used during compilation to keep track of dependencies between queries. ([see more](../queries/incremental-compilation.html))
 data-flow analysis <div id="data-flow"/> |  A static analysis that figures out what properties are true at each point in the control-flow of a program; see [the background chapter for more](./background.html#dataflow).
-DeBruijn Index <div id="debruijn">       |  A technique for describing which binder a variable is bound by using only integers. It has the benefit that it is invariant under variable renaming. ([see more](./background.md#what-is-a-debruijn-index))
+DeBruijn Index <div id="debruijn"/>      |  A technique for describing which binder a variable is bound by using only integers. It has the benefit that it is invariant under variable renaming. ([see more](./background.md#what-is-a-debruijn-index))
```
jyn514, line 19: I thought `div` was not a self-closing tag. What does this do?

camelid: Not sure, but that's what all the others do 🤷

jyn514: All of these are being recovered by the HTML parser as invalid syntax.

self-closing

I think we should a) close this tag, and b) change it to `<a href="#" id="debruijn>` so you can click on it; right now you have to type it into your browser manually.

camelid: You mean change it to:

```html
<a href="#debruijn" id="debruijn">{...}</a>
```

?

camelid: Vim regex for the win!

```
%s/^\(.*\) <div id="\([^"]\+\)"\/>/<a href="#\2" id="\2">\1<\/a>/g
```

camelid: I'm not sure if having self-linking anchors is really worth it here. It's kind of busy in the output and in the source:

image

```markdown
# Glossary

Term                                     | Meaning
-----------------------------------------|--------
<a href="#arena" id="arena">arena/arena allocation</a> |  An _arena_ is a large memory buffer from which other memory allocations are made. This style of allocation is called _arena allocation_. See [this chapter](../memory.md) for more info.
<a href="#ast" id="ast">AST</a>                      |  The abstract syntax tree produced by the `rustc_ast` crate; reflects user syntax very closely.
<a href="#binder" id="binder">binder</a>                |  A "binder" is a place where a variable or type is declared; for example, the `<T>` is a binder for the generic type parameter `T` in `fn foo<T>(..)`, and \|`a`\|` ...` is a binder for the parameter `a`. See [the background chapter for more](./background.html#free-vs-bound).
<a href="#body-id" id="body-id">BodyId</a>               |  An identifier that refers to a specific body (definition of a function or constant) in the crate. See [the HIR chapter for more](../hir.html#identifiers-in-the-hir).
<a href="#bound-var" id="bound-var">bound variable</a>     |  A "bound variable" is one that is declared within an expression/term. For example, the variable `a` is bound within the closure expression \|`a`\|` a * 2`. See [the background chapter for more](./background.html#free-vs-bound)
<a href="#codegen" id="codegen">codegen</a>              |  The code to translate MIR into LLVM IR.
<a href="#codegen-unit" id="codegen-unit">codegen unit</a>    |  When we produce LLVM IR, we group the Rust code into a number of codegen units (sometimes abbreviated as CGUs). Each of these units is processed by LLVM independently from one another, enabling parallelism. They are also the unit of incremental re-use. ([see more](../backend/codegen.md))
<a href="#completeness" id="completeness">completeness</a>    |  A technical term in type theory, it means that every type-safe program also type-checks. Having both soundness and completeness is very hard, and usually soundness is more important. (see "soundness").
<a href="#cfg" id="cfg">control-flow graph</a>       |  A representation of the control-flow of a program; see [the background chapter for more](./background.html#cfg)
<a href="#ctfe" id="ctfe">CTFE</a>                    |  Short for Compile-Time Function Evaluation, this is the ability of the compiler to evaluate `const fn`s at compile time. This is part of the compiler's constant evaluation system. ([see more](../const-eval.html))
<a href="#cx" id="cx">cx</a>                        |  We tend to use "cx" as an abbreviation for context. See also `tcx`, `infcx`, etc.
```

It would also take a lot of effort to get the aligning spaces right in the source :/

What about just `<span id="{id}">{...}</span>`?

jyn514: I'd be ok with `<span>`, that will still take a lot of time if you want to align the | though.

camelid:
Closing in favor of #939.

src/appendix/glossary.md
camelid: It actually takes no effort since all of the rows add the net same number of characters:

- Remove `<div id="{id}"/>`
- Add `<span id="{id}">`
- Add `</span>`

camelid: Whereas adding a link adds an additional instance of `{id}`, which makes the lengths vary.

PR #939, opened 2020-10-27
Fix broken tags in glossary

camelid:
- Fix broken tags in glossary
- Put `TyCtxt` at the right position


jyn514 (approved):
Thanks for separating this into different commits.

PR #941, opened 2020-11-02
Use `provide_both` link instead of `provide(_extern)`

JohnTitor:


JohnTitor:
The links are broken, caused by #77795 (cc @bjorn3).

mark-i-m (approved):
Thanks!

PR #950, opened 2020-11-11
Add a section on data-flow convergence

camelid:
r? @oli-obk

### DIFF
--- a/src/mir/dataflow.md
+++ b/src/mir/dataflow.md
@@ -89,7 +89,21 @@ statement must be considered separately from the left-hand side.
 
 ### Convergence
 
-TODO
+Your analysis must converge to "fixpoint", otherwise it will run forever.
+Converging to fixpoint is just another way of saying "reaching equilibrium".
+In order to reach equilibrium, your analysis must obey some laws. One of the
+laws it must obey is that the bottom value joined with some other value equals
+the second value. Or, as an equation:
+
+> *bottom* join *x* = *x*
+
+Another law is that your analysis must have a "top value" such that
+
+> *top* join *x* = *top*
+
+Having a top value ensures that your semilattice has a finite height, and the
+law state above ensures that once the dataflow state reaches top, it will no
+longer change (the fixpoint will be top).
 
 ## Inspecting the Results of a Dataflow Analysis
 
--- a/src/mir/dataflow.md
+++ b/src/mir/dataflow.md
@@ -109,6 +109,32 @@ longer change (the fixpoint will be top).
     state. Each basic block's entry state is initialized to bottom before the
     analysis starts.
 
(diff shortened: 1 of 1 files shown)


src/mir/dataflow.md
```diff
@@ -89,7 +89,21 @@ statement must be considered separately from the left-hand side.
 
 ### Convergence
 
-TODO
+Your analysis must converge to "fixpoint", otherwise it will run forever.
+Converging to fixpoint is just another way of saying "reaching equilibrium".
+In order to reach equilibrium, your analysis must obey some laws. One of the
+laws it must obey is that the bottom value joined with some other value equals
+the second value. Or, as an equation:
```
jyn514, line 96: Why is a bottom value necessary? I see that having a top is important, since the top is a fixpoint, but what is the bottom used for?

camelid: What would be the initial state of the analysis if you didn't have a bottom value? The bottom value *is* that initial state.

jyn514: Got it, thanks - I didn't realize 'bottom' meant 'initial state'. Could you add that?

camelid: Sure, good idea. Although I think the bottom value is a bit more general than 'initial state'. Not 100% sure though.

oli-obk:
Thanks, this lgtm from a perspective of someone who already has a moderate understanding of what dataflow does. I'm not sure it will be very helpful to someone new to dataflow without a concrete example. What do you think about adding an example for finding out whether there is a path in the control flow graph to the current location from a call to `transmute`? Or some other example that requires little extra state that is easy to explain with examples?

camelid:
Sure, I can do that. I think it would be a good idea to make this section more approachable to beginners in general. The reason I didn't add an example is because of this paragraph at the top of the page:

> However, this documentation is not a general introduction to dataflow analysis. It is merely a description of the framework used to define these analyses in rustc. It assumes that the reader is familiar with the core ideas as well as some basic terminology, such as "transfer function", "fixpoint" and "lattice".

camelid:
I'm just going to provide an example of what the analysis domain would look like, since explaining the algorithm is a much bigger involvement.

By the way, I'm in the process of writing a blog post to explain how the algorithm works in practice. Perhaps we could link to it from this page?

oli-obk:
> By the way, I'm in the process of writing a blog post to explain how the algorithm works in practice. Perhaps we could link to it from this page?

I think you can open a PR to link to your blog post when it's up and we'll decide then. Seems fine to me tho

oli-obk (approved):
Thanks! This makes this section much more accessible

jyn514 (approved):

PR #951, opened 2020-11-11
Add a section on identifiers in the MIR

camelid:
r? @oli-obk

By the way, let me know if there are other people I can ping for review so I
don't assign every MIR PR to you :)

@rustbot modify labels: waiting-on-review

### DIFF
--- a/src/identifiers.md
+++ b/src/identifiers.md
@@ -12,7 +12,7 @@ A [`NodeId`] is an identifier number that uniquely identifies an AST node within
 a crate. Every node in the AST has its own [`NodeId`], including top-level items
 such as structs, but also individual statements and expressions.
 
-However, because they are absolute within in a crate, adding or removing a single
+However, because they are absolute within a crate, adding or removing a single
 node in the AST causes all the subsequent [`NodeId`]s to change. This renders
 [`NodeId`]s pretty much useless for incremental compilation, where you want as
 few things as possible to change.
@@ -31,14 +31,17 @@ The HIR uses a bunch of different identifiers that coexist and serve different p
   the crate the definition comes from, and a [`DefIndex`] which identifies the definition
   within the crate. Unlike [`NodeId`]s, there isn't a [`DefId`] for every expression, which
   makes them more stable across compilations.
+
 - A [`LocalDefId`] is basically a [`DefId`] that is known to come from the current crate.
   This allows us to drop the [`CrateNum`] part, and use the type system to ensure that
   only local definitions are passed to functions that expect a local definition.
+
 - A [`HirId`] uniquely identifies a node in the HIR of the current crate. It is composed
   of two parts: an `owner` and a `local_id` that is unique within the `owner`. This
   combination makes for more stable values which are helpful for incremental compilation.
   Unlike [`DefId`]s, a [`HirId`] can refer to [fine-grained entities][Node] like expressions,
   but stays local to the current crate.
+
 - A [`BodyId`] identifies a HIR [`Body`] in the current crate. It is currenty only
   a wrapper around a [`HirId`]. For more info about HIR bodies, please refer to the
   [HIR chapter][hir-bodies].
@@ -62,14 +65,20 @@ See the [HIR chapter][hir-map] for more detailed information.
 
 - [`BasicBlock`] identifies a *basic block*. It points to an instance of
   [`BasicBlockData`].
+
 - [`Local`] identifies a local variable in a function. Its associated data is in
   [`LocalDecl`].
+
 - [`Field`] identifies a struct's, union's, or enum variant's field. It is used
   as a "projection" in [`Place`].
+
 - [`SourceScope`] identifies a name scope in the original source code.
+
 - [`Promoted`] identifies a promoted constant (related to const evaluation).
+
 - [`GlobalId`] identifies a global variable: a `const`, a `static`, a `const fn`
   where all arguments are [zero-sized types], or a promoted constant.
+
 - [`Location`] represents the location in the MIR of a statement or terminator.
   It identifies the block (using [`BasicBlock`]) and the index of the statement
   or terminator in the block.
@@ -82,6 +91,6 @@ See the [HIR chapter][hir-map] for more detailed information.
 [`Place`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/struct.Place.html
 [`SourceScope`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/struct.SourceScope.html
 [`Promoted`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/struct.Promoted.html
-[`GlobalId`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/struct.GlobalId.html
+[`GlobalId`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/interpret/struct.GlobalId.html
 [`Location`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/struct.Location.html
 [zero-sized types]: https://doc.rust-lang.org/nomicon/exotic-sizes.html#zero-sized-types-zsts
--- a/src/identifiers.md
+++ b/src/identifiers.md
@@ -72,9 +72,12 @@ See the [HIR chapter][hir-map] for more detailed information.
 - [`Field`] identifies a struct's, union's, or enum variant's field. It is used
   as a "projection" in [`Place`].
 
-- [`SourceScope`] identifies a name scope in the original source code.
+- [`SourceScope`] identifies a name scope in the original source code. Used for
+  diagnostics and for debuginfo in debuggers.
 
-- [`Promoted`] identifies a promoted constant (related to const evaluation).
+- [`Promoted`] identifies a promoted constant within another item (related to
+  const evaluation). Note: it is unique only locally within the item;
+  [`GlobalId`] must be used if you need to *globally* identify the promoted.
 
 - [`GlobalId`] identifies a global variable: a `const`, a `static`, a `const fn`
   where all arguments are [zero-sized types], or a promoted constant.
--- a/src/identifiers.md
+++ b/src/identifiers.md
@@ -64,16 +64,20 @@ See the [HIR chapter][hir-map] for more detailed information.
 ## In the MIR
 
 - [`BasicBlock`] identifies a *basic block*. It points to an instance of
-  [`BasicBlockData`].
+  [`BasicBlockData`], which can be retrieved by indexing into
+  [`Body::basic_blocks()`] (note that you must call a function; the field is
+  private).
 
 - [`Local`] identifies a local variable in a function. Its associated data is in
-  [`LocalDecl`].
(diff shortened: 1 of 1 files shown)


src/identifiers.md
```diff
@@ -60,4 +60,27 @@ See the [HIR chapter][hir-map] for more detailed information.
...
+- [`BasicBlock`] identifies a *basic block*. It points to an instance of
+  [`BasicBlockData`].
+- [`Local`] identifies a local variable in a function. Its associated data is in
+  [`LocalDecl`].
+- [`Field`] identifies a struct's, union's, or enum variant's field. It is used
+  as a "projection" in [`Place`].
+- [`SourceScope`] identifies a name scope in the original source code.
+- [`Promoted`] identifies a promoted constant (related to const evaluation).
+- [`GlobalId`] identifies a global variable: a `const`, a `static`, a certain
+  kind of `const fn`, or a promoted constant.
```
camelid, line 72: I don't know much about these four; I was just trying to guess what they do based on the limit docs and looking through the source code. So let me know if I got it wrong!

src/identifiers.md
```diff
@@ -60,4 +60,27 @@ See the [HIR chapter][hir-map] for more detailed information.
...
+- [`Location`] represents the location in the MIR of a statement or terminator.
+  It identifies the block (using [`BasicBock`]) and the index of the statement
```
camelid, line 74: "BasicBock" :)

src/identifiers.md
jyn514: What does 'certain kind of const fn' mean?

jyn514:
> By the way, let me know if there are other people I can ping for review so I
don't assign every MIR PR to you :)

I think RalfJung and jonas-schievink would both be good reviewers.

camelid:
> > By the way, let me know if there are other people I can ping for review so I
> > don't assign every MIR PR to you :)
> 
> I think RalfJung and jonas-schievink would both be good reviewers.

Yeah, I just wasn't sure if they're open to reviewing dev-guide PRs.

camelid:
That's weird... CI passed but there's a broken link: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/struct.GlobalId.html

src/identifiers.md
camelid: Quoting the docs for `GlobalId`:

> Uniquely identifies one of the following:
> - A constant
> - A static
> - A const fn where all arguments (if any) are zero-sized types

I didn't want to make the docs here too long and I was trying to limit the amount of duplication, but it probably makes sense to reword this.

mark-i-m:
It might be due to caching. I wonder if we can turn that off now...

oli-obk:
we do have wg-mir-opt that you can (I think you can?) ping, but I have no problem getting pinged and potentially forwarding to someone else.

src/identifiers.md
```diff
@@ -60,4 +63,34 @@ See the [HIR chapter][hir-map] for more detailed information.
...
+- [`BasicBlock`] identifies a *basic block*. It points to an instance of
+  [`BasicBlockData`].
+
+- [`Local`] identifies a local variable in a function. Its associated data is in
+  [`LocalDecl`].
+
+- [`Field`] identifies a struct's, union's, or enum variant's field. It is used
+  as a "projection" in [`Place`].
+
+- [`SourceScope`] identifies a name scope in the original source code.
```
oli-obk, line 75: This is also used for debug info in debuggers. In contrast with any `Span`s which are only used for diagnostics

src/identifiers.md
```diff
@@ -60,4 +63,34 @@ See the [HIR chapter][hir-map] for more detailed information.
...
+
+- [`Promoted`] identifies a promoted constant (related to const evaluation).
```
oli-obk, line 77: It identifies a promoted *within* another item. So you always need a `GlobalId` to be able to identify a promoted across MIR bodies

camelid:
@oli-obk Could you re-review this? I think it's ready!

src/identifiers.md
```diff
@@ -60,4 +63,37 @@ See the [HIR chapter][hir-map] for more detailed information.
 
 ## In the MIR
 
-**TODO**
+- [`BasicBlock`] identifies a *basic block*. It points to an instance of
+  [`BasicBlockData`].
```
LeSeulArtichaut, line 67: Maybe?
```suggestion
  [`BasicBlockData`] in a body which can be retrieved with [`Body::basic_blocks`].
```

src/identifiers.md
```diff
@@ -60,4 +63,37 @@ See the [HIR chapter][hir-map] for more detailed information.
...
+
+- [`Local`] identifies a local variable in a function. Its associated data is in
+  [`LocalDecl`].
```
LeSeulArtichaut, line 70: Same here?
```suggestion
  [`LocalDecl`] which can be retrieved with [`Body::local_decls`].
```

src/identifiers.md
```diff
@@ -60,4 +63,37 @@ See the [HIR chapter][hir-map] for more detailed information.
...
+
+- [`Field`] identifies a struct's, union's, or enum variant's field. It is used
+  as a "projection" in [`Place`].
+
+- [`SourceScope`] identifies a name scope in the original source code. Used for
+  diagnostics and for debuginfo in debuggers.
```
LeSeulArtichaut, line 76: We should probably also mention `SourceScopeData` and how to get it here?

src/identifiers.md
```diff
@@ -60,4 +63,37 @@ See the [HIR chapter][hir-map] for more detailed information.
...
+
+- [`Promoted`] identifies a promoted constant within another item (related to
+  const evaluation). Note: it is unique only locally within the item;
```
LeSeulArtichaut, line 79: If I understand correctly, the basic unit in MIR is `Body`?
```suggestion
  const evaluation). Note: it is unique only locally within the body;
```

LeSeulArtichaut (commented):
As someone who's never really touched MIR, this makes things clearer to me! :heart:
I think it would be worth it to also mention how to use those identifiers (i.e. get the `XData` with `X` through `Body::x`), but I'm not really sure...

In the HIR part of the chapter I had only written:
> These identifiers can be converted into one another through the [HIR map][map].
See the [HIR chapter][hir-map] for more detailed information.

So 🤷‍♂️... @jyn514 @oli-obk what do you think?

jyn514:
cc @robinhundt

src/identifiers.md
camelid: Hmm, well the promoted constant *is* a `Body`. I think all items are `Body`s, but not all `Body`s are items. Basically, a `Body` represents a chunk of code that's executed together, which may be extracted from a function's `Body` as a promoted constant. So in this case, I think item is correct. Though I would like to hear from @oli-obk if I'm right here :)

oli-obk: I immediately thought this is correct, but writing up my response I realized that inline-constants (so the new `const { 42 }` expressions), are not items, but is an `AnonConst`, which has a `DefId`. Though we can just treat these as "anonymous const item" and then everything fits again.

So... the general problem is that you need a `DefId` + `Promoted` pair to identify a specific promoted. So if you have something which does not have `DefId` but has a `Body`, then you cannot have `Promoted`s in it. Thus, you cannot have promoteds within other promoteds, even if a promoted could refer to another promoted from another item or even the same item.

src/identifiers.md
oli-obk: It can be retrived by indexing `Body::local_decls`

src/identifiers.md
camelid: Yes, thanks for catching that!

src/identifiers.md
camelid: @oli-obk So should I change the documentation somehow? If so, I'm not fully sure what you are requesting I add.

oli-obk: Not sure. @LeSeulArtichaut does my comment above make it clearer? If so, do you have any suggestion on how to best put this into the text?

LeSeulArtichaut: I think we could either have a footnote with your comment, just leave it as is, or add something like
```suggestion
  const evaluation). Note: it is unique only locally within the item, so it should always be associated with a `DefId`;
```

oli-obk: That snippet sounds like a good middle ground between explaining it in full and just leaving it as it is.

oli-obk (approved):

src/identifiers.md
```diff
@@ -60,4 +63,45 @@ See the [HIR chapter][hir-map] for more detailed information.
...
+- [`Field`] identifies a struct's, union's, or enum variant's field. It is used
+  as a "projection" in [`Place`].
+
+- [`SourceScope`] identifies a name scope in the original source code. Used for
+  diagnostics and for debuginfo in debuggers. It points to an instance of
+  [`SourceScopeData`], which can be retrieved by indexing into
+  [`Body.source_scopes`].
+
+- [`Promoted`] identifies a promoted constant within another item (related to
+  const evaluation). Note: it is unique only locally within the item, so it should be associated with a `DefId`;
```
LeSeulArtichaut, line 83: You're going to have a CI error complaining because this line is too long

src/identifiers.md
```diff
@@ -60,4 +63,45 @@ See the [HIR chapter][hir-map] for more detailed information.
...
+  [`GlobalId`] must be used if you need to *globally* identify the promoted.
```
camelid, line 84: @oli-obk The change you applied confuses me: the text now says

> Note: it is unique only locally within the item, so it should be associated with a `DefId`;
> [`GlobalId`] must be used if you need to *globally* identify the promoted.

So which one is it? It says that you should associate it with a `DefId`, but it also says use `GlobalId`.

oli-obk: `GlobalId` is the combination of `DefId` + `Substs` + `Option<Promoted>`. You can identify a promoted with just a `DefId` and `Promoted`, but you may not be able to compute its value without the `Substs`, because it's still polymorphic.

camelid: So what should I change this sentence to?

oli-obk: I guess something like `must be used if you need a globally unique identifier for the monomorphic version of a promoted`? Is that filled with too many weird words or do you think someone who gets to this section the first time will be able to understand it?

camelid: Hmm, but didn't you say that you can identify a promoted with just `DefId` and `Promoted`, and you only need `GlobalId` in order to actually compute the value?

oli-obk: My mental model was always that `GlobalId` identifies the monomorphic version of sth, `DefId` identifies a definition (potentially polymorphic) and `(DefId, Promoted)` identifies a promoted within a definition (also potentially polymorphic). But "identifying a monomorphic version" may not really be a thing...

camelid: You know what, how about we just say "`GlobalId` gives you more information (TODO)" and come back to this later? That way we can get the rest of it merged :)

LeSeulArtichaut: I personally wouldn't mind, especially given that this PR itself removes a giant **TODO** :)

src/identifiers.md
LeSeulArtichaut, line 83: Also this line failed in CI because it's too long

LeSeulArtichaut (approved):

LeSeulArtichaut:
Thanks for your work on this @camelid :)

PR #954, opened 2020-11-11
Turn off caching for normal CI runs

mark-i-m:
Currently, we use caching for all CI runs with a timeout of 1 day. However,
recently, we switched to testing only changed files for non-cron CI jobs.

This PR additionally nixes the cache for non-cron CI jobs, so that we fully
check all links in the relevant files.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ script:
 - git rebase origin/master
 - git log --oneline | head -n 10
 - mdbook build
-- ([ -e /tmp/cache.json ] && mv /tmp/cache.json book/linkcheck/cache.json) || echo "No cache to restore."
+- { [ -e /tmp/cache.json ] && mv /tmp/cache.json book/linkcheck/cache.json; } || echo "No cache to restore."
 notifications:
   email: false
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,6 @@ script:
 - git rebase origin/master
 - git log --oneline | head -n 10
 - mdbook build
-after_script:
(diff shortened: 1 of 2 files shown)


camelid:
What do you mean by "caching" in this case?

.travis.yml
```diff
@@ -18,7 +18,6 @@ script:
 - git rebase origin/master
 - git log --oneline | head -n 10
 - mdbook build
-after_script:
 - ([ -e /tmp/cache.json ] && mv /tmp/cache.json book/linkcheck/cache.json) || echo "No cache to restore."
```
jyn514, line 21: This spawns a new subshell.
```suggestion
- { [ -e /tmp/cache.json ] && mv /tmp/cache.json book/linkcheck/cache.json; } || echo "No cache to restore."
```

jyn514:
@camelid see `cache-timeout` under https://github.com/Michael-F-Bryan/mdbook-linkcheck/#configuration

mark-i-m:
@camelid mdbook-linkcheck keeps a cache of recently validated links. If a link was checked in the last 24 hours and found valid, we don't check it again. We have a cron job that runs everyday. The cache allows the cron to avoid the 429 errors in most cases, while still ensuring that most broken links get found within a couple of days.

camelid:
Why would that mean that a *newly-added* broken link is not detected?

mark-i-m:
@camelid It happens that that particular link is also used in `src/const-eval.md`. So I guess it is probably cached from there.

jyn514 (approved):

mark-i-m:
It looks like github is waiting for a travis run that doesn't exist? I'm going to close this an make a new PR.

Issue #892, opened 2020-09-25
Update logging for `tracing`


ehuss:
It looks like the compiler moved to the `tracing` crate. I think at a minimum, the section at https://rustc-dev-guide.rust-lang.org/compiler-debugging.html?highlight=log#getting-logging-output should be updated to mention that.

I don't actually know why it was changed, what the differences are, or what the motivation was. https://github.com/rust-lang/compiler-team/issues/331 and https://github.com/rust-lang/rust/pull/74726 don't really say.  Is there anything I need to know about the differences? Are there any examples of when using spans or `#[instrument]` would be useful? Should messages use structured fields (and how or why is that useful)?

jyn514:
@oli-obk do you have time to work on this?

oli-obk:
We are not using spans or `#[instrument]` yet. There are no differences in usage yet. Before we can start using awesome tracing features we'll need to benchmark them and agree on a span rendering strategy. https://github.com/rust-lang/rust/pull/75143 explores span rendering, please look at it and leave opinions.

But yea, some minimal updating of the logging section would be a good thing. I'll try to find some time, but won't complain if someone grabs it

igaray:
I'm currently working on the survey translations but in a few days I'll free up and could take this :) Won't claim it in case someone wants to do it right away.

camelid:
I fixed this in #997.

PR #955, opened 2020-11-11
Update link in compiler-debugging.md

PoignardAzur:
Link didn't point to latest version of `env_logger` crate.

Add an explanation to disambiguate the RUSTC_LOG syntax.

### DIFF
--- a/src/compiler-debugging.md
+++ b/src/compiler-debugging.md
@@ -149,10 +149,10 @@ $ # Cool, now I have a backtrace for the error
 These crates are used in compiler for logging:
 
 * [log]
-* [env-logger]: check the link to see the full `RUSTC_LOG` syntax
+* [env-logger]
 
 [log]: https://docs.rs/log/0.4.6/log/index.html
-[env-logger]: https://docs.rs/env_logger/0.4.3/env_logger/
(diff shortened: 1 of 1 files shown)


jyn514 (approved):
Seems fine, but just so you know this section will be changing a lot soon since we switched to tracing ( https://github.com/rust-lang/rustc-dev-guide/issues/892)

src/compiler-debugging.md
```diff
@@ -149,10 +149,10 @@ $ # Cool, now I have a backtrace for the error
 These crates are used in compiler for logging:
 
 * [log]
-* [env-logger]: check the link to see the full `RUSTC_LOG` syntax
+* [env-logger]
 
 [log]: https://docs.rs/log/0.4.6/log/index.html
-[env-logger]: https://docs.rs/env_logger/0.4.3/env_logger/
+[env-logger]: https://docs.rs/env_logger/0.8.1/env_logger/
```
jyn514, line 155: Let's use docs.rs to get the latest version so we don't have to keep updating it.
```suggestion
[env-logger]: https://docs.rs/env_logger
```

PoignardAzur:
Aw. And I spent, like, 20 minutes bikeshedding it too.

PR #960, opened 2020-11-15
add string interning wikipedia link

hosseind88:


src/overview.md
```diff
@@ -32,7 +32,10 @@ we'll talk about that later.
 - The token stream passes through a higher-level lexer located in
   [`rustc_parse`] to prepare for the next stage of the compile process. The
   [`StringReader`] struct is used at this stage to perform a set of validations
-  and turn strings into interned symbols (_interning_ is discussed later).
+  and turn strings into interned symbols (_interning_ is discussed later). 
+  string interning is a method of storing only one copy of each distinct 
+  string value, which must be immutable.
+  [read more](https://en.wikipedia.org/wiki/String_interning)
```
jyn514, line 38:
```suggestion
  [String interning] is a way of storing only one immutable copy of each distinct string value.
```
and then add a reference link at the bottom:
```
  [String interning]: https://en.wikipedia.org/wiki/String_interning)
```

jyn514: You didn't add the reference at the bottom. It should go near the other links.

hosseind88: you mean add `[String interning]: https://en.wikipedia.org/wiki/String_interning)` under `read more`?

LeSeulArtichaut: Please put it with the other links for the paragraph: see lines 101-115

jyn514: You undid the previous changes when you added the link.

hosseind88: yeah but now that change is back

jyn514: I mean this change: https://github.com/rust-lang/rustc-dev-guide/pull/960/files#r523762736
It's disappeared.

hosseind88: fixed

jyn514: Can you please only show 'string interning' once instead of twice? Right now there are 5 lines related to it, I would rather have 1 line like it was before. Or in other words, my suggestion should replace the first change, not add to it.

hosseind88: ok sure

jyn514 (approved):

PR #967, opened 2020-11-23
Document when errors should have an associated error code

camelid:
I don't know if there's an official guideline (that's why we need these docs!);
this is just what sounded reasonable to me.

r? @estebank

### DIFF
--- a/src/diagnostics.md
+++ b/src/diagnostics.md
@@ -73,15 +73,20 @@ codes.
 The descriptions are written in Markdown, and all of them are linked in the
 [`rustc_error_codes`] crate.
 
-As a general rule, errors should have an associated error code if the problem is
-complicated enough or frequent enough that it would be helpful to have a
-long-form explanation. For example, borrow-check and type-check errors tend to
-have error codes, but many parse errors do not. If you are not sure, just ask
-your reviewer!
+As a general rule, give an error a code (with an associated explanation) if the
+explanation would give more information than the error itself. A lot of the time
+it's better to put all the information in the emitted error itself. However,
+sometimes that would make the error verbose or there are too many possible
+triggers to include useful information for all cases in the error, in which case
+it's a good idea to add an explanation.[^estebank]
+As always, if you are not sure, just ask your reviewer!
+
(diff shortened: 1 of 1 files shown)


jyn514:
Hmm, I don't know if this is a *guideline* so much as describing the current situation. I would say when in doubt, always add an error code (I've even made PRs adding error codes to errors that didn't have one before: https://github.com/rust-lang/rust/pull/76143).

camelid:
> I don't know if there's an official guideline (that's why we need these docs!);
> this is just what sounded reasonable to me.

@jyn514 That's what I was saying: this *isn't* some official guideline :)

jyn514:
Right, yes, but I'm saying we shouldn't suggest it as a policy. This makes it seem like you *shouldn't* add error codes for parse errors, which I'm not sure is true.

estebank:
@jyn514 parse errors in particular are tricky. I find that error codes work best when they are targeted and specific, meaning there is a small number of triggers for it and their description can give you enough context to fix the problem. There are some cases where the same error code has a bunch of solutions (E0277, E0308, for example), but for the most part that's fine. Parse errors in particular have a combination of things that we proactively handle (hey! missing semicolon here!) but the vast majority of them are kind of random and can be reached through multiple different kinds of unrelated mistakes. We _could_ introduce a new generic error code for the general case of "E0666: parse error", and then take every case we proactively handle and give it its own error code. That would be fine, but _I_ have instead focused on putting all the info that _would_ be in the error code index into the error itself. When the index entry wouldn't have any more information, I don't bother creating a new code. Maybe _that_ would be a good rule of thumb: give an error a code if the the index entry would give more info than the error itself which we haven't either because of verbosity or because we can't detect all the possible triggers.

jyn514:
> Maybe that would be a good rule of thumb: give an error a code if the the index entry would give more info than the error itself which we haven't either because of verbosity or because we can't detect all the possible triggers.

Sounds great, thanks!

camelid:
@estebank I changed the advice to something similar to what you suggested. Should be ready now!

estebank (approved):

PR #968, opened 2020-11-23
Document how to modify feature gates

camelid:
How to remove them and how to rename them.

cc <https://github.com/rust-lang/rust/pull/79336#discussion_r528627318>
cc @varkor

### DIFF
--- a/src/feature-gates.md
+++ b/src/feature-gates.md
@@ -1,0 +1,65 @@
+# Feature Gates
+
+This chapter is intended to provide basic help for modifying feature gates. See
+["Stability in code"][stability-section] for help with *adding* feature gates.
+
+[stability-section]: ./implementing_new_features.md#stability-in-code
+
+
+## Removing a feature gate
+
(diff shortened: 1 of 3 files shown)


src/feature-gates.md
```diff
@@ -0,0 +1,65 @@
+# Feature Gates
+
+This chapter is intended to provide basic help for modifying feature gates. See
+["Stability in code"][stability-section] for help with *adding* feature gates.
```
jyn514, line 4: Could we have "stability in code" link here instead, so all the info is one place?

camelid: Done. I thought of doing that before, but wasn't sure. Good suggestion!

jyn514 (approved):
LGTM unless you want to wait for varkor to review.

camelid:
I'm fine either way. Worst comes to worse, we can always update the docs :)

I pushed a new commit at the same time you submitted your review though, so you might want to take another look.

varkor:
Looks good to me too 👍

PR #970, opened 2020-11-24
Update bootstrap documentation with the new envvars for RUSTFLAGS

g2p:
An update to the documentation for passing RUSTFLAGS to stages.
Follows https://github.com/rust-lang/rust/pull/63722


jyn514 (approved):
Good catch, thanks!

PR #977, opened 2020-11-29
Add notes about running the UI test suite

bstrie:
As requested by @jyn514 .

### DIFF
--- a/src/tests/running.md
+++ b/src/tests/running.md
@@ -59,43 +59,6 @@ Likewise, you can test a single file by passing its path:
 ./x.py test src/test/ui/const-generics/const-test.rs
 ```
 
-### Run the UI test suite
-
-The UI test suite is special in that it tests the command-line output
-of the compiler, e.g. the wording and formatting of warnings and errors.
-Every UI test file such as `foo.rs` will have a corresponding file
-`foo.stderr` describing its expected output.
-
-Similar to any other test suite, the UI test suite can be run with:
-
-```bash
-./x.py test src/test/ui
-```
-
-If you have made a change to the way that the compiler formats its output,
-it would be extraordinarily tedious to have to reformat every `*.stderr`
-file manually. Fortunately there exists a flag that will allow you to
-"bless" the current output of the compiler as the expected output.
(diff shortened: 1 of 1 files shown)


src/tests/running.md
```diff
@@ -59,6 +59,43 @@ Likewise, you can test a single file by passing its path:
...
+If you have made a change to the way that the compiler formats its output,
+it would be extraordinarily tedious to have to reformat every `*.stderr`
+file manually. Fortunately there exists a flag that will allow you to
+"bless" the current output of the compiler as the expected output.
+When using this flag with the test runner, any UI tests that would
+ordinarily fail will instead have their `*.stderr` files overwritten
+with whatever output the compiler produces:
+
+```bash
+./x.py test src/test/ui --bless
```
jyn514, line 84: Most of this info is already present in the '## Editing and updating the reference files' section below. Can you move it to that section so it's not duplicated?

src/tests/running.md
```diff
@@ -59,6 +59,43 @@ Likewise, you can test a single file by passing its path:
...
+```
+
+Some UI tests will have different output depending on which "mode" that
+the compiler is in. Specifically, the compiler may have different output
+depending on whether the feature "non-lexical lifetimes" (NLL) is enabled.
+Any UI test such as `foo.rs` whose output differs with NLL enabled will
+have both a `foo.stderr` file and a `foo.nll.stderr` file.
+By default, the UI test suite will not be run in NLL mode.
```
jyn514, line 92: This makes it sound like `nll` is the only mode, but there are other modes like `noopt` and `migrate`; plus any revisions will also have their own stderr files. Could you change it to use nll as an example instead?
```suggestion
the compiler is in. For example, "non-lexical lifetimes" (NLL) mode or a [revision](./adding.html#revisions) will have a different output file than the default.
```

bstrie:
Ah yes, I forgot that the document already talked about --bless. I didn't see anything that my section would add over the original so I just removed it. I've moved the compare-modes bit into its own new section at the bottom and reworded it.

jyn514 (approved):
Looks great, thanks!

PR #978, opened 2020-12-01
Document lang items

MikailBag:
I'm not sure where to put this page, selected the simplest place.


src/lang-items.md
```diff
@@ -0,0 +1,289 @@
...
+The `rustc` compiler has certain pluggable operations, that is,
+functionality that isn't hard-coded into the language, but is
+implemented in libraries, with a special marker to tell the compiler
+it exists. The marker is the attribute `#[lang = "..."]` and there are
+various different values of `...`, i.e. various different 'lang
+items'.
+
+For example, `Box` pointers require two lang items, one for allocation
+and one for deallocation. A freestanding program that uses the `Box`
+sugar for dynamic allocations via `malloc` and `free`:
```
jyn514, line 13: Hmm, I'm not sure this is a great example because you can do this more easily with `#[global_allocator]`. Maybe use `#[lang = "start"]` instead? I think you can use that to have a graphical entrypoint instead of a console entrypoint on Windows. It would also simplify the example a lot I think - this shouldn't be a section on how to write an allocator ;)

src/lang-items.md
```diff
@@ -0,0 +1,289 @@
...
+#[lang = "panic_impl"]
+#[no_mangle]
+pub extern fn rust_begin_panic(info: &PanicInfo) -> ! {
+    unsafe { intrinsics::abort() }
+}
+```
+
+In many cases, you may need to manually link to the `compiler_builtins` crate
+when building a `no_std` binary. You may observe this via linker error messages
+such as "```undefined reference to `__rust_probestack'```".
```
jyn514, line 174: I don't think this section belongs in the dev-guide. I think there's already some info about this on https://os.phil-opp.com/freestanding-rust-binary/, and I think that's a better place - this doesn't tell you how the compiler works, just gives an example of a no_std executable, which seems like language documentation. If you want to add it to the reference, it might be a better fit there.

src/lang-items.md
```diff
@@ -0,0 +1,289 @@
...
+  - `drop_in_place`: `libcore/ptr.rs`
+  - `clone`: `libcore/clone.rs`
+  - `copy`: `libcore/marker.rs`
+  - `send`: `libcore/marker.rs`
+  - `sized`: `libcore/marker.rs`
+  - `unsize`: `libcore/marker.rs`
+  - `sync`: `libcore/marker.rs`
+  - `phantom_data`: `libcore/marker.rs`
+  - `discriminant_kind`: `libcore/marker.rs`
+  - `freeze`: `libcore/marker.rs`
```
jyn514, line 289: This will *very* rapidly get out of date. Since you can easily rederive it with `rg 'lang = "'`, I don't think it needs to go in the guide.

jyn514 (commented):
Thanks for writing it up! There's definitely useful info here, but I'm not sure the dev-guide is the right place for it.

src/lang-items.md
camelid: (Also the filenames are wrong: it should be `library/core/src/...` not `libcore/...`)

MikailBag:
Oh, I should have made it clear that I'm not author of this text. It's a bit actualized page from the unstable book. 

@nikomatsakis proposed moving it to rustc-dev-guide. I think documentation of std implementation details fits rustc-dev-guide. Is there a better place?

jyn514:
@spastorino do you have opinions on this?

spastorino:
@jyn514 I like the idea of having content about lang items in the guide and this is great but as you pointed out we may want to improve some things. I don't have any further comment than what you've already stated.

src/lang-items.md
jyn514: Maybe it's better to link to https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/lang_items/enum.LangItem.html instead? That will always be up to date. It won't have the files in `library` but you can grep for those pretty easily.

src/lang-items.md
```diff
@@ -197,6 +159,14 @@ the screen. While the language item's name is `panic_impl`, the symbol name is
 Finally, a `eh_catch_typeinfo` static is needed for certain targets which
 implement Rust panics on top of C++ exceptions.
 
+## Well-known paths
+
+In several cases compiler finds specific item not by `lang` attribute. Instead
+item path is hardcored. For example compiler assumes `Iterator` trait to be
```
jyn514, line 165:
```suggestion
item path is hardcoded. For example compiler assumes `Iterator` trait to be
```
I didn't know this, thanks!

src/lang-items.md
```diff
@@ -197,6 +159,14 @@ the screen. While the language item's name is `panic_impl`, the symbol name is
...
+available as `core::iter::Iterator`. This only happens when item is required
+on early compilation stages (for example `Iterator` is used in for loops
```
jyn514, line 167:
```suggestion
in early compilation stages (for example `Iterator` is used in `for` loops
```

src/lang-items.md
```diff
@@ -0,0 +1,259 @@
...
+pub extern fn rust_begin_panic(info: &PanicInfo) -> ! {
+    unsafe { intrinsics::abort() }
+}
+```
+
+In many cases, you may need to manually link to the `compiler_builtins` crate
+when building a `no_std` binary. You may observe this via linker error messages
+such as "```undefined reference to `__rust_probestack'```".
+
+## More about the language items
```
camelid, line 138: Usually they're referred to without 'the'.

```suggestion
## More about language items
```

src/lang-items.md
```diff
@@ -0,0 +1,259 @@
...
+`rust_begin_panic`.
+
+Finally, a `eh_catch_typeinfo` static is needed for certain targets which
+implement Rust panics on top of C++ exceptions.
+
+## Well-known paths
+
+In several cases compiler finds specific item not by `lang` attribute. Instead
+item path is hardcored. For example compiler assumes `Iterator` trait to be
+available as `core::iter::Iterator`. This only happens when item is required
```
camelid, line 166:
```suggestion
available as `core::iter::Iterator`. This only happens when an item is required
```

src/lang-items.md
```diff
@@ -0,0 +1,259 @@
...
+compiler. When a panic happens, this controls the message that's displayed on
+the screen. While the language item's name is `panic_impl`, the symbol name is
+`rust_begin_panic`.
+
+Finally, a `eh_catch_typeinfo` static is needed for certain targets which
+implement Rust panics on top of C++ exceptions.
+
+## Well-known paths
+
+In several cases compiler finds specific item not by `lang` attribute. Instead
```
camelid, line 164:
```suggestion
In several cases, the compiler finds a specific item without using a `lang` attribute. Instead, the
```

(you may have to re-wrap the lines)

src/lang-items.md
```diff
@@ -0,0 +1,259 @@
...
+item path is hardcored. For example compiler assumes `Iterator` trait to be
```
camelid, line 165:
```suggestion
item path is hardcoded. For example, the compiler assumes that the `Iterator` trait will be
```

src/lang-items.md
```diff
@@ -0,0 +1,259 @@
...
+  - `eq`: `library/core/src/cmp.rs`
+  - `ord`: `library/core/src/cmp.rs`
+- Functions
+  - `fn`: `library/core/src/ops/function.rs`
+  - `fn_mut`: `library/core/src/ops/function.rs`
+  - `fn_once`: `library/core/src/ops/function.rs`
+  - `generator_state`: `library/core/src/ops/generator.rs`
+  - `generator`: `library/core/src/ops/generator.rs`
+- Opting out
+  - `unsafe_cell` (relaxes pointer provenance rules, allowing const-to-mut casts): `library/core/src/cell.rs`
```
camelid, line 245: Btw, you need to wrap this line.

src/lang-items.md
```diff
@@ -197,6 +159,14 @@ the screen. While the language item's name is `panic_impl`, the symbol name is
...
 implement Rust panics on top of C++ exceptions.
 
+## Well-known paths
+
+In several cases compiler finds specific item not by `lang` attribute. Instead
+item path is hardcored. For example compiler assumes `Iterator` trait to be
+available as `core::iter::Iterator`. This only happens when item is required
+on early compilation stages (for example `Iterator` is used in for loops
+desugaring).
+
```
camelid, line 169: Actually, based on [looking at the output of `rustc +nightly -Z unpretty=hir`](https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/why.20are.20.60Some.60.20and.20.60None.60.20language.20items.3F/near/219286210), `core::iter::Iterator` is *not* hardcoded when desugaring `for` loops. Do you know of a place where the language item's path *is* hardcoded, or should we just remove this section?

camelid (changes requested):

src/lang-items.md
lcnr: afaik we now try to use `QPath::LangItem` everywhere instead of using fully qualified paths.

https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/hir/enum.QPath.html#variant.LangItem

I think this is probably out of date

MikailBag: Yeah, you're right, the compiler no longer hardcodes paths. IIRC it was true some time ago, so the compiler was improved!

I'll verify if there are any well-known paths. If no I'll delete this section. Thanks!

jyn514:
Hi @MikailBag, it's been a while - are you still planning to work on this?

nikomatsakis (commented):
This seems pretty useful to document indeed. I was expecting though to read more about how the code *in the compiler* can gain access to lang items and some of the patterns we have around that. 

Another interesting angle: sometimes, when you have compiler errors in libcore, it's useful to create a standalone binary / library for reproducing the failure. That might be a good thing to mention or even use as the motivating case.

I think we should land this PR, but it'd be ideal to tweak it to have more of a "rustc-dev" angle.

MikailBag:
Sorry for the huge delay.

I've just turned this PR into an MVP that only contains a small introduction and a list of lang items.

I think that examples should be better added in follow-up PRs because my compiler knowledge (and English writing skills :smile:) is not enough for writing a good example.

src/lang-items.md
```diff
@@ -0,0 +1,133 @@
...
+Many such lang items can be implemented only in one sensible way, such as
+`add` (`trait core::ops::Add`) or `future_trait`
+(`trait core::future::Future`). 
+
+Others can be overriden to achieve some
+specific goals.
+
+For example, later sections describe how to control
+your binary startup or override panic implementation.
```
jyn514, line 19: This is no longer true.
```suggestion
Others can be overridden to achieve some specific goals; for example, you can control your binary's entrypoint.
```

src/lang-items.md
```diff
@@ -0,0 +1,133 @@
...
+Features provided by lang items include:
+
+- overloadable operators via traits: the traits corresponding to the
+  `==`, `<`, dereferencing (`*`) and `+` (etc.) operators are all
+  marked with lang items; those specific four are `eq`, `ord`,
+  `deref`, and `add` respectively.
+- stack unwinding and general failure; the `eh_personality`,
+  `panic` and `panic_bounds_checks` lang items.
+- the traits in `std::marker` used to indicate types of
+  various kinds; lang items `send`, `sync` and `copy`.
```
jyn514, line 30:
```suggestion
- the traits in `std::marker` used to indicate properties of
  types that are used by the compiler; lang items `send`, `sync` and `copy`.
```

src/lang-items.md
```diff
@@ -0,0 +1,133 @@
...
+  - `drop_in_place`: `library/core/src/ptr.rs`
+  - `clone`: `library/core/src/clone.rs`
+  - `copy`: `library/core/src/marker.rs`
+  - `send`: `library/core/src/marker.rs`
+  - `sized`: `libcrary/ore/msrc/arker.rs`
+  - `unsize`: `library/core/src/marker.rs`
+  - `sync`: `library/core/src/marker.rs`
+  - `phantom_data`: `libcore/marker.rs`
+  - `discriminant_kind`: `library/core/src/marker.rs`
+  - `freeze`: `library/core/src/marker.rs`
```
jyn514, line 133: I still think this is not a good fit because it will get outdated so rapidly. Maybe instead you could recommend `rg '#\[lang' library` or looking at https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/lang_items/enum.LangItem.html ?

camelid: Yeah, I was going to suggest that it would be better to have people look at the docs. And suggesting `rg -F '#[lang' library/` is probably a good idea too.

jyn514:
Hi @MikailBag, are you planning to follow-up on this?

MikailBag:
Given that I don't have enough understanding of how this page *should* look, I'm going to close this PR for now.

PR #987, opened 2020-12-10
All links to rustdoc-internals.md

aDotInTheVoid:


src/rustdoc-internals.md
```diff
@@ -213,3 +213,13 @@ $ python3 -m http.server -d build/[YOUR ARCH]/doc
 
 Now you can browse your documentation just like you would if it was hosted
 on the internet. For example, the url for `std` will be `/std/".
```
camelid, line 215: I know you didn't add this, but could you fix this?

```suggestion
on the internet. For example, the url for `std` will be `/std/`.
```

jyn514 (approved):
Seems useful, thanks :)

src/rustdoc-internals.md
jyn514: I think this can be fixed in a follow-up.

PR #989, opened 2020-12-12
Document `error-pattern` header

### LINKED ISSUE #986: Document // error-pattern
Example: https://github.com/rust-lang/rust/pull/76143/files#diff-23706e664a2a576e6e031e0be43b04ff0f1916b93c2358b518d67e53aa92fbc3L1

It's used when you don't have a span available in the error but still want to test the output.

_Originally posted by @jyn514 in https://github.com/rust-lang/rust/pull/79621#discussion_r537715698_

JohnTitor:
Fixes #986

### DIFF
--- a/src/tests/adding.md
+++ b/src/tests/adding.md
@@ -207,6 +207,9 @@ source.
 * `needs-sanitizer-{address,leak,memory,thread}` - indicates that test
   requires a target with a support for AddressSanitizer, LeakSanitizer,
   MemorySanitizer or ThreadSanitizer respectively.
+* `error-pattern` checks the diagnostics just like the `ERROR` annotation
+  without specifying error line. This is useful when the error doesn't give
+  any span.
 
 [`header.rs`]: https://github.com/rust-lang/rust/tree/master/src/tools/compiletest/src/header.rs
 [bless]: ./running.md#editing-and-updating-the-reference-files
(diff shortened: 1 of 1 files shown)


src/tests/adding.md
```diff
@@ -291,6 +294,37 @@ fn main() {
 //~| ERROR this pattern has 1 field, but the corresponding tuple struct has 3 fields [E0023]
 ```
 
+#### When error line cannot be specified
+
+Let's this about this test:
```
LeSeulArtichaut, line 299: Typo?

src/tests/adding.md
```diff
@@ -291,6 +294,37 @@ fn main() {
...
+```rust,ignore
+fn main() {
+    let a: *const [_] = &[1, 2, 3];
+    unsafe {
+        let _b = (*a)[3];
+    }
+}
+```
+
+We want to ensure this shows "index out of bounds" but we cannot use the `ERROR` annotation
```
LeSeulArtichaut, line 310: Maybe specify "at runtime" to explain why this has the `run-fail`? I'm not sure this is really necessary though.
```suggestion
We want to ensure this shows "index out of bounds" at runtime, but we cannot use the `ERROR` annotation
```

LeSeulArtichaut (approved):
Awesome, I didn't know this functionnality at all! ❤️
I left a few nits

src/tests/adding.md
JohnTitor: Ah, copied/pasted headers from the original test but `run-fail` may be confusing. Just removed it.

src/tests/adding.md
JohnTitor: Yeah, it is!

LeSeulArtichaut:
Thanks!

PR #998, opened 2020-12-30
Fix broken link for `BoundRegion`

JohnTitor:
Caused by https://github.com/rust-lang/rust/pull/80163.

### DIFF
--- a/src/early-late-bound.md
+++ b/src/early-late-bound.md
@@ -89,7 +89,7 @@ Moreover, a reference to a late-bound lifetime is written `^0.a`:
 - The `0` is the index; it identifies that this lifetime is bound in the
   innermost binder (the `for`).
 - The `a` is the "name"; late-bound lifetimes in rustc are identified by a
-  "name" -- the [`BoundRegion`] struct. This struct can contain a
+  "name" -- the [`BoundRegionKind`] enum. This enum can contain a
   [`DefId`][defid] or it might have various "anonymous" numbered names. The
   latter arise from types like `fn(&u32, &u32)`, which are equivalent to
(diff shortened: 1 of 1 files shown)


src/early-late-bound.md
```diff
@@ -89,7 +89,7 @@ Moreover, a reference to a late-bound lifetime is written `^0.a`:
 - The `0` is the index; it identifies that this lifetime is bound in the
   innermost binder (the `for`).
 - The `a` is the "name"; late-bound lifetimes in rustc are identified by a
-  "name" -- the [`BoundRegion`] struct. This struct can contain a
+  "name" -- the [`BoundRegionKind`] enum. This enum can contain a
```
camelid, line 92: It seems to me that we should still be linking to `BoundRegion` – it's just a struct now, rather than an enum.

JohnTitor: I think it's more straightforward to use `BoundRegionKind` like `ExprKind` or similar, since we use it `Kind` thing mainly.

LeSeulArtichaut: Given that `BoundRegion` is only a wrapper around `BoundRegionKind`, and that we talk about what *kind* of name it can be in this paragraph, I think linking to `BoundRegionKind` makes more sense

LeSeulArtichaut (approved):

src/early-late-bound.md
camelid: In the future `BoundRegion` may have more fields than just `kind`, so wouldn't we want to link people to that?

camelid: Oops, I was too late 😄

LeSeulArtichaut: I'd argue that if this is the case this chapter will need updating anyway

PR #1003, opened 2021-01-04
Add link to rust-analyzer docs on getting started page

rylev:


jyn514 (approved):
Looks good to me since it's just adding a link to the existing text. In general I've been wanting to rethink how we do the intro though, there's a lot packed in there.

PR #1005, opened 2021-01-04
Make instructions for configuring compiler for debug more explicit

rylev:
I made the mistake of not updating the config file on this machine I'm using which has never built rustc before. This was an easy fix (especially since I knew what the issue was), but I think this information should be more prominently placed.

### DIFF
--- a/src/compiler-debugging.md
+++ b/src/compiler-debugging.md
@@ -56,7 +56,8 @@ stack backtrace:
   37: rustc_driver::run_compiler
 ```
 
-If you set `debug = true`, you will get line numbers for the stack trace. Then the backtrace will look like this:
+If you set `debug = true`, you will get line numbers for the stack trace. 
+Then the backtrace will look like this:
 
 ```text
 stack backtrace:
--- a/src/compiler-debugging.md
+++ b/src/compiler-debugging.md
@@ -14,11 +14,7 @@ set `debug = true` in your config.toml.
 
 Setting `debug = true` turns on many different debug options (e.g., `debug-assertions`, 
 `debug-logging`, etc.) which can be individually tweaked if you want to, but many people 
-simply set `debug = true`. Check out the comments in `config.toml.example` for more info.
-
-In some cases, just setting these options will not trigger a rebuild,
-so if you changed it and you already have a compiler built, you might
-want to call `x.py clean` to force a rebuild.
+simply set `debug = true`. Check out the comments in config.toml.example for more info.
 
 You will need to rebuild the compiler once you've changed any configuration options. 
 
--- a/src/compiler-debugging.md
+++ b/src/compiler-debugging.md
@@ -60,9 +60,7 @@ stack backtrace:
   37: rustc_driver::run_compiler
 ```
 
-If you set `debug = true`, you will get line numbers for the stack trace. 
-`debuginfo-level = 1` will also add line numbers, but `debug = true` gives 
-full debug info. Then the backtrace will look like this:
+If you set `debug = true`, you will get line numbers for the stack trace. Then the backtrace will look like this:
 
 ```text
 stack backtrace:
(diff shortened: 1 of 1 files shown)


src/compiler-debugging.md
```diff
@@ -7,6 +7,21 @@ advice about specific parts of the compiler (e.g. the [Queries Debugging and
...
+By default, rustc is built without most debug information. To enable debug info, 
+set `debug = true` in your config.toml. 
+
+Setting `debug = true` turns on many different debug options (e.g., `debug-assertions`, 
+`debug-logging`, etc.) which can be individually tweaked if you want to, but many people 
+simply set `debug = true`. Check out the comments in `config.toml.example` for more info.
+
+In some cases, just setting these options will not trigger a rebuild,
+so if you changed it and you already have a compiler built, you might
+want to call `x.py clean` to force a rebuild.
```
jyn514, line 21: Hmm, if this happens it's an x.py bug. Have you seen this happen or are you just moving the existing test?

src/compiler-debugging.md
```diff
@@ -45,10 +60,9 @@ stack backtrace:
   37: rustc_driver::run_compiler
 ```
 
-If you want line numbers for the stack trace, you can enable `debug = true` in
-your config.toml and rebuild the compiler (`debuginfo-level = 1` will also add
-line numbers, but `debug = true` gives full debuginfo). Then the backtrace will
-look like this:
+If you set `debug = true`, you will get line numbers for the stack trace. 
+`debuginfo-level = 1` will also add line numbers, but `debug = true` gives 
+full debug info. Then the backtrace will look like this:
```
jyn514, line 65: This is out of date.
```suggestion
If you set `debug = true`, you will get line numbers for the stack trace. Then the backtrace will look like this:
```
I also want to add
```
`debuginfo-level = 2` will also add information about local variables when using `gdb`, but is not enabled by default.
```
somewhere, but I'm not sure this is the best place.

jyn514:
> I made the mistake of not updating the config file on this machine I'm using which has never built rustc before. This was an easy fix (especially since I knew what the issue was), but I think this information should be more prominently placed.

x.py should nag you to run `x.py setup` if you don't have a config file - did you overlook the message maybe? Happy to hear suggestions about how to make that easier to find.

src/compiler-debugging.md
rylev: Just moving the existing text.

rylev:
@jyn514 I'm actually not in the habit of using `./x.py setup` so this is mostly my fault. 

FWIW I find the flow information from this change to be better than what it currently is.

src/compiler-debugging.md
```diff
@@ -7,6 +7,21 @@ advice about specific parts of the compiler (e.g. the [Queries Debugging and
 Testing chapter](./incrcomp-debugging.html) or the [LLVM Debugging
 chapter](./backend/debugging.md)).
 
+## Configuring the compiler
+
+By default, rustc is built without most debug information. To enable debug info, 
+set `debug = true` in your config.toml.
```
camelid, line 13:
```suggestion
set `debug = true` in your `config.toml`.
```

src/compiler-debugging.md
jyn514: Ok. @Mark-Simulacrum am I right in thinking that if x.py doesn't rebuild after setting debug = true that's a bug? I don't want to recommend clearing the whole cache if it's not necessary.

Mark-Simulacrum: Hm, yes, this does indeed sound like a bug. I am rather surprised though.

jyn514: @rylev Let's remove this text then until someone actually encounters this bug. It sounds like this hasn't happened in practice.

src/compiler-debugging.md
rylev: It seems config.toml is not in backticks anhywhere else in the file. I'll remove them from `config.toml.example`

src/compiler-debugging.md
rylev: Done!

spastorino (approved):

PR #1013, opened 2021-01-09
Add context to dated information

### LINKED ISSUE #924: Add context to "may be out of date" sections
Many places in the guide mention "this section may be out of date", such as in [the parser](https://rustc-dev-guide.rust-lang.org/the-parser.html). This is unfortunately quite ambiguous because the writer usually has not contextualized it with a "when" or "why", and so the reader does not know how much to trust or mistrust the information. Such comments should be marked with enough data to provide a reasonable amount of context, possibly including but not limited to

0) a reason for why the data may be out of date other than "change"... change is a constant across the project
1) a time the comment was added (visible in git blame but not on the website)
2) a link to a relevant WG, tracking issue, or similar, that may provide further explanation for the change process

igaray:
Searching for references to "future" and "currently", these non-contextualized (context-free? isolated?) sections have been identified and modified. 

In addition, the About This Guide and Contributing sections have been updated to reflect our policy of trying to contextualize future additions to the guide.

### DIFF
--- a/src/contributing.md
+++ b/src/contributing.md
@@ -410,7 +410,8 @@ There are issues for beginners and advanced compiler devs alike!
 
 Just a few things to keep in mind:
 
-- Please limit line length to 100 characters. This is enforced by CI, and you can run the checks locally with `ci/check_line_lengths.sh`.
+- Please limit line length to 100 characters. This is enforced by CI, and you can run the checks
+  locally with `ci/check_line_lengths.sh`.
 - When contributing text to the guide, please contextualize the information with some time period
   and/or a reason so that the reader knows how much to trust or mistrust the information.
   Aim to provide a reasonable amount of context, possibly including but not limited to:
--- a/src/contributing.md
+++ b/src/contributing.md
@@ -401,16 +401,28 @@ You can also use `rustdoc` directly to check small fixes. For example,
 `rustdoc src/doc/reference.md` will render reference to `doc/reference.html`.
 The CSS might be messed up, but you can verify that the HTML is right.
 
+### Contributing to rustc-dev-guide
+
 Additionally, contributions to the [rustc-dev-guide][rdg] are always welcome.
 Contributions can be made directly at [the rust-lang/rustc-dev-guide
 repo][rdgrepo]. The issue tracker in that repo is also a great way to find
 things that need doing. There are issues for beginners and advanced compiler
 devs alike!
 
-When contributing text to the guide, please contextualize the information in some time period.
-E.g. instead of writing _"Currently, ..."_, or _"In the future, ..."_, consider writing
-_"As of January 2021, ..."_.
+Just a few things to keep in mind:
 
+- Please limit line length to 100 characters.
+- When contributing text to the guide, please contextualize the information with some time period
+  and/or a reason so that the reader knows how much to trust or mistrust the information.
+  Aim to provide a reasonable amount of context, possibly including but not limited to:
+  - A reason for why the data may be out of date other than "change", as change is a constant across
+    the project.
+  - A date the comment was added, e.g. instead of writing _"Currently, ..."_, or
+    _"In the future, ..."_, consider writing _"As of January 2021, ..."_.
+    Try to format the date as `<MONTH> <YEAR>` to ease search.
+  - A link to a relevant WG, tracking issue, `rustc` rustdoc page, or similar, that may provide
+    further explanation for the change process or a way to verify that the information is not
+    outdated.
 
 [rdg]: https://rustc-dev-guide.rust-lang.org/
 [rdgrepo]: https://github.com/rust-lang/rustc-dev-guide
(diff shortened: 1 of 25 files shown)


src/parallel-rustc.md
```diff
@@ -1,9 +1,10 @@
 # Parallel Compilation
 
 Most of the compiler is not parallel. This represents an opportunity for
-improving compiler performance. Much effort has been put into parallelizing
-`rustc`, but it is still pretty early days for this work. There is a lot of
-design and correctness work that needs to be done.
+improving compiler performance.
+
+As of December 2020, work on explicitly parallelizing the compiler has stalled.
```
jyn514, line 6:
```suggestion
As of January 2021, work on explicitly parallelizing the compiler has stalled.
```

jyn514 (approved):

rylev:
Is there someway to make the data references machine parseable? We could then write a simple tool to remind us when references are ~6 months old, and we should check it they’re still true.

src/salsa.md
```diff
@@ -7,8 +7,10 @@ want to watch [Salsa In More
 Depth](https://www.youtube.com/watch?v=i_IhACacPRY), also by Niko
 Matsakis.
 
-> Salsa is not used directly in rustc, but it is used extensively for
-> rust-analyzer and may be integrated into the compiler in the future.
+> As of January 2021, although Salsa is inspired by (among other things9 rustc's query system,
+> it is not used directly in rustc.
+> It _is_ used in chalk, and extensively in `rust-analyzer`, but there no medium or long-term
```
rylev, line 12:
```suggestion
> It _is_ used in chalk and extensively in `rust-analyzer`, but there are no medium or long-term
```

igaray:
> Is there someway to make the data references machine parseable? We could then write a simple tool to remind us when references are ~6 months old, and we should check it they’re still true.

I think this would fall under the general problem of automating keeping documentation and code in sync, and I would love to have a conversation about that.

camelid:
> Is there someway to make the data references machine parseable? We could then write a simple tool to remind us when references are ~6 months old, and we should check it they’re still true.

If we make sure that they are always of the form "as of {month} {year}" then we could use a simple regex to extract them.

igaray:
> > Is there someway to make the data references machine parseable? We could then write a simple tool to remind us when references are ~6 months old, and we should check it they’re still true.
> 
> If we make sure that they are always of the form "as of {month} {year}" then we could use a simple regex to extract them.

Added [this bit](https://github.com/rust-lang/rustc-dev-guide/pull/1013/files#diff-6fd31aec287ce7566f057de0f165da0436e86f79786be6d057508c2b6cdb62aeR412)

src/contributing.md
```diff
@@ -401,16 +401,28 @@ You can also use `rustdoc` directly to check small fixes. For example,
 `rustdoc src/doc/reference.md` will render reference to `doc/reference.html`.
 The CSS might be messed up, but you can verify that the HTML is right.
 
+### Contributing to rustc-dev-guide
+
 Additionally, contributions to the [rustc-dev-guide][rdg] are always welcome.
```
jyn514, line 406: 'Additionally' seems a little odd at the start of a section IMO.

src/contributing.md
```diff
@@ -401,16 +401,28 @@ You can also use `rustdoc` directly to check small fixes. For example,
...
 things that need doing. There are issues for beginners and advanced compiler
 devs alike!
 
-When contributing text to the guide, please contextualize the information in some time period.
-E.g. instead of writing _"Currently, ..."_, or _"In the future, ..."_, consider writing
-_"As of January 2021, ..."_.
-
+Just a few things to keep in mind:
+
+- Please limit line length to 100 characters.
```
jyn514, line 414:
```suggestion
- Please limit line length to 100 characters. This is enforced by CI, and you can run the checks locally with `ci/check_line_lengths.sh`.
```

src/contributing.md
```diff
@@ -401,16 +401,28 @@ You can also use `rustdoc` directly to check small fixes. For example,
...
+- When contributing text to the guide, please contextualize the information with some time period
+  and/or a reason so that the reader knows how much to trust or mistrust the information.
+  Aim to provide a reasonable amount of context, possibly including but not limited to:
+  - A reason for why the data may be out of date other than "change", as change is a constant across
+    the project.
+  - A date the comment was added, e.g. instead of writing _"Currently, ..."_, or
+    _"In the future, ..."_, consider writing _"As of January 2021, ..."_.
```
jyn514, line 421: I don't necessarily agree with this one. 'In the future' specifically means 'some indeterminate time' and you can't quantify it. ('currently' should definitely be replaced though.)

src/salsa.md
```diff
@@ -7,9 +7,9 @@ want to watch [Salsa In More
 Depth](https://www.youtube.com/watch?v=i_IhACacPRY), also by Niko
 Matsakis.
 
-> As of January 2021, although Salsa is inspired by (among other things9 rustc's query system,
+> As of January 2021, although Salsa is inspired by (among other things rustc's query system,
```
jyn514, line 10:
```suggestion
> As of January 2021, although Salsa is inspired by (among other things) rustc's query system,
```

PR #1015, opened 2021-01-10
Point to `HirId` instead of `NodeId`

camelid:
cc https://rust-lang.zulipchat.com/#narrow/stream/182449-t-compiler.2Fhelp/topic/changing.20error.20message/near/222198371

r? @jyn514

### DIFF
--- a/src/identifiers.md
+++ b/src/identifiers.md
@@ -29,7 +29,7 @@ The HIR uses a bunch of different identifiers that coexist and serve different p
 - A [`DefId`], as the name suggests, identifies a particular definition, or top-level
   item, in a given crate. It is composed of two parts: a [`CrateNum`] which identifies
   the crate the definition comes from, and a [`DefIndex`] which identifies the definition
-  within the crate. Unlike [`NodeId`]s, there isn't a [`DefId`] for every expression, which
+  within the crate. Unlike [`HirId`]s, there isn't a [`DefId`] for every expression, which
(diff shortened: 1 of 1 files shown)


src/identifiers.md
```diff
@@ -29,7 +29,7 @@ The HIR uses a bunch of different identifiers that coexist and serve different p
 - A [`DefId`], as the name suggests, identifies a particular definition, or top-level
   item, in a given crate. It is composed of two parts: a [`CrateNum`] which identifies
   the crate the definition comes from, and a [`DefIndex`] which identifies the definition
-  within the crate. Unlike [`NodeId`]s, there isn't a [`DefId`] for every expression, which
+  within the crate. Unlike [`HirId`]s, there isn't a [`DefId`] for every expression, which
```
jyn514, line 32: At this point the guide hasn't discussed HirIds yet. Do you want to move that section above DefIds? It's just a couple bullet points below.

camelid: There's an infinite cycle: the `DefId` bullet refer to `HirId` and the `HirId` bullet refers to `DefId` :)

Also, `HirId` refers (indirectly) to `LocalDefId`, which is above it and below `DefId`. I'm fine to move it though; I don't think it matters too much either way.

jyn514: Fair enough, this seems fine then.

jyn514 (approved):

PR #1017, opened 2021-01-10
Update 'Workspace structure' section

camelid:


src/compiler-src.md
```diff
@@ -15,18 +15,24 @@ look at the structure of the contents of the rust-lang/rust repo.
...
+The repository consists of three main directories:
+
+- `compiler/` contains the source code for `rustc`. It consists of many crates
+  that together make up the compiler.
+
+- `library/` contains the standard libraries (`core`, `alloc`, `std`,
+`proc_macro`, etc.).
+
+- `src/` contains the source code for the build system, tools like clippy and
+  cargo, language docs, etc. This directory is currently being broken up to be
```
jyn514, line 34: Is it being broken up? I'm not aware of any plans for that.

JohnTitor, line 34: We could also mention that rustdoc is here since we mention it at the beginning of this section?

> The rust-lang/rust repository consists of a single large cargo workspace containing the compiler, the standard libraries (core, alloc, std, proc_macro, etc), and rustdoc, along with the build system and a bunch of tools and submodules for building a full Rust distribution.

camelid: That's what the original text said, so I kept it. It might be out of date.

camelid: Done.

camelid: Done.

src/compiler-src.md
```diff
@@ -15,18 +15,23 @@ look at the structure of the contents of the rust-lang/rust repo.
...
-directory is currently being broken up to be less monolithic. There is also a
-`library/` directory, where the standard libraries (`core`, `alloc`, `std`,
-`proc_macro`, etc) live.
+The repository consists of three main directories:
+
+- `compiler/` contains the source code for `rustc`. It consists of many crates
+  that together make up the compiler.
+
+- `library/` contains the standard libraries (`core`, `alloc`, `std`,
+`proc_macro`, etc.).
```
jyn514, line 31:
```suggestion
- `library/` contains the standard libraries (`core`, `alloc`, `std`,
`proc_macro`, etc.), as well as the Rust runtime (`backtrace`, `rtstartup`, `lang_start`).
```

jyn514 (approved):

PR #1019, opened 2021-01-11
Add a note about UI test entry limitation

JohnTitor:
https://github.com/rust-lang/rust/pull/79959 added a tidy check to ensure the number of UI test entries. This PR documents it and adds a note not to place more tests on `src/test/ui` and `src/test/ui/issues` as possible.

### DIFF
--- a/src/tests/adding.md
+++ b/src/tests/adding.md
@@ -66,6 +66,8 @@ directory that helps identify what piece of code is being tested here
 If you've tried and cannot find a more relevant place,
 the test may be added to `src/test/ui/issues/`.
 Still, **do include the issue number somewhere**.
+But please avoid putting your test there as possible since that
+directory has too many tests and it causes poor semantic organization.
 
 When writing a new feature, **create a subdirectory to store your
 tests**. For example, if you are implementing RFC 1234 ("Widgets"),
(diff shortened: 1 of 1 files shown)


src/tests/adding.md
```diff
@@ -391,6 +393,21 @@ and so forth.
...
 
+We now have a ton of UI tests and some directories have too many entries.
+This is a problem because it isn't editor/IDE friendly and GitHub UI won't
+show more than 1000 entries. To resolve it and organize semantic structure,
+we have a tidy check to ensure the number of entries is less than 1000.
+However, since `src/test/ui` (UI test root directory) and
+`src/test/ui/issues` directories have more than 1000 entries,
+we set a different limit for each directories. So, please
+avoid putting a new test there as possible and try to find a more relevant place.
+For example, if your test is related to closure, you should put it to
```
camelid, line 404:
```suggestion
avoid putting a new test there and try to find a more relevant place.
For example, if your test is related to closures, you should put it in
```

JohnTitor: Thanks! Addressed it.

jyn514 (approved):
This seems good to me - there's a little duplication about how `src/test/ui` is big both here and on line 56 that would be nice to merge, but I'm also fine with merging as-is.

PR #1026, opened 2021-01-16
improved the readablity of a line

ashosive:
On the _Cloning and Building_ section on the page https://rustc-dev-guide.rust-lang.org/getting-started.html#cloning-and-building, the following line can be made better readable : 

> clone these immediately, but the build tool will automatically clone and sync them (more later).` 

In this PR, I have added the following : 

> clone these immediately, but the build tool will automatically clone and sync them (more on this later).` 

I believe this makes the text better readable and more precise technically.


jyn514 (approved):
Thanks!

PR #1029, opened 2021-01-20
Add note about regressions using `--bless`

ThePuzzlemaker:
Using `--bless` on tests without checking the output can cause regressions, some of which may confuse people and may take a long time to track down. It's just generally best practice to make sure the failures in UI tests are expected and not from a regression.

### DIFF
--- a/src/tests/running.md
+++ b/src/tests/running.md
@@ -120,6 +120,10 @@ to automatically adjust the `.stderr`, `.stdout` or `.fixed` files of
 all tests. Of course you can also target just specific tests with the
 `--test-args your_test_name` flag, just like when running the tests.
 
+Please note that it is **VERY IMPORTANT** that you check the output of        
+failed tests before using `--bless` on them as this may cause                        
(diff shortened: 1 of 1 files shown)


src/tests/running.md
```diff
@@ -120,6 +120,10 @@ to automatically adjust the `.stderr`, `.stdout` or `.fixed` files of
 all tests. Of course you can also target just specific tests with the
 `--test-args your_test_name` flag, just like when running the tests.
 
+Please note that it is **VERY IMPORTANT** that you check the output of        
+failed tests before using `--bless` on them as this may cause                        
+regressions and make other peoples' jobs **a lot harder**!
```
camelid, line 125: I don't know if this is necessarily the right way to talk about it:

1. I almost always use `--bless` and then check the Git diff for any unexpected changes, so "you should only use `--bless` after you check" is not always true.
2. Reviewers will also check the changes to test output, so I don't think we should emphasize it quite this much.

camelid (changes requested):

PR #1032, opened 2021-01-21
Add missing period

camelid:


src/backend/updating-llvm.md
```diff
@@ -109,7 +109,7 @@ through each in detail.
    [llvm]
    download-ci-llvm = false
    ```
-   in config.toml
+   in `config.toml`.
```
jyn514, line 112: We decided not to add this earlier: https://github.com/rust-lang/rustc-dev-guide/pull/1005#discussion_r551793601. Has anything changed since then?

camelid: Ah, I forgot about that. I'll change this to just add a period.

jyn514 (approved):

PR #1044, opened 2021-01-29
Add chapter on libs and metadata.

### LINKED ISSUE #468: Introduce something about rmeta and metadata in general

ehuss:
This adds a new chapter on library files and metadata, and how these things interact.

This leans a bit heavy from a Cargo/build-tools perspective, but I think this information is useful.  I think it is difficult to hit a good balance of too-much or too-little detail, since things tend to change a lot over time. Not sure if I hit the mark there.

I think this is pretty accurate, but I am not involved with compiler development a lot, so there may be some things that are incomplete or wrong.

Closes #468 (I think?)

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -145,6 +145,7 @@
     - [Debugging LLVM](./backend/debugging.md)
     - [Backend Agnostic Codegen](./backend/backend-agnostic.md)
     - [Implicit Caller Location](./backend/implicit-caller-location.md)
+- [Libraries and Metadata](./backend/libs-and-metadata.md)
 - [Profile-guided Optimization](./profile-guided-optimization.md)
 - [LLVM Source-Based Code Coverage](./llvm-coverage-instrumentation.md)
 - [Sanitizers Support](./sanitizers.md)
--- a/src/backend/libs-and-metadata.md
+++ b/src/backend/libs-and-metadata.md
@@ -1,6 +1,6 @@
 # Libraries and Metadata
 
-When the compiler sees a reference to an extern crate, it needs to load some
+When the compiler sees a reference to an external crate, it needs to load some
 information about that crate. This chapter gives an overview of that process,
 and the supported file formats for crate libraries.
 
@@ -9,8 +9,8 @@ and the supported file formats for crate libraries.
 A crate dependency can be loaded from an `rlib`, `dylib`, or `rmeta` file. A
 key point of these file formats is that they contain `rustc`-specific
 [*metadata*](#metadata). This metadata allows the compiler to discover enough
-information about the extern crate to understand the items it contains, which
-macros it exports, and *much* more.
+information about the external crate to understand the items it contains,
+which macros it exports, and *much* more.
 
 ### rlib
 
@@ -134,13 +134,13 @@ can have the same crate name, so they need to be disambiguated).
 ## Crate loading
 
 Crate loading can have quite a few subtle complexities. During [name
-resolution], when an extern crate is referenced (via an `extern crate` or
+resolution], when an external crate is referenced (via an `extern crate` or
 path), the resolver uses the [`CrateLoader`] which is responsible for finding
 the crate libraries and loading the [metadata] for them. After the dependency
 is loaded, the `CrateLoader` will provide the information the resolver needs
 to perform its job (such as expanding macros, resolving paths, etc.).
 
-To load each extern crate, the `CrateLoader` uses a [`CrateLocator`] to
+To load each external crate, the `CrateLoader` uses a [`CrateLocator`] to
 actually find the correct files for one specific crate. There is some great
 documentation in the [`locator`] module that goes into detail on how loading
 works, and I strongly suggest reading it to get the full picture.
--- a/src/backend/libs-and-metadata.md
+++ b/src/backend/libs-and-metadata.md
@@ -1,0 +1,192 @@
+# Libraries and Metadata
+
+When the compiler sees a reference to an extern crate, it needs to load some
+information about that crate. This chapter gives an overview of that process,
+and the supported file formats for crate libraries.
+
+## Libraries
+
+A crate dependency can be loaded from an `rlib`, `dylib`, or `rmeta` file. A
+key point of these file formats is that they contain `rustc`-specific
+[*metadata*](#metadata). This metadata allows the compiler to discover enough
+information about the extern crate to understand the items it contains, which
+macros it exports, and *much* more.
+
+### rlib
+
+An `rlib` is an [archive file], which is similar to a tar file. This file
+format is specific to `rustc`, and may change over time. This file contains:
+
+* Object code, which is the result of code generation. This is used during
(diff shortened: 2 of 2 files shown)


ehuss:
BTW, I had no idea where to stick this in the SUMMARY hierarchy.  I think "backend" makes sense, but I'm uncertain.

camelid:
I've wanted to see info about metadata and rustlibs in the dev-guide for a
while, so thank you!

src/SUMMARY.md
```diff
@@ -145,6 +145,7 @@
     - [Debugging LLVM](./backend/debugging.md)
     - [Backend Agnostic Codegen](./backend/backend-agnostic.md)
     - [Implicit Caller Location](./backend/implicit-caller-location.md)
+- [Libraries and Metadata](./backend/libs-and-metadata.md)
```
camelid, line 148: Perhaps instead put this under "High-level compiler architecture"? That's where queries and serialization are discussed, which are somewhat related.

src/backend/libs-and-metadata.md
```diff
@@ -0,0 +1,192 @@
+# Libraries and Metadata
+
+When the compiler sees a reference to an extern crate, it needs to load some
```
camelid, line 3: Do you want to say "external crate" instead since `extern crate` is almost never used explicitly?

src/backend/libs-and-metadata.md
```diff
@@ -0,0 +1,192 @@
...
+### rlib
+
+An `rlib` is an [archive file], which is similar to a tar file. This file
+format is specific to `rustc`, and may change over time. This file contains:
+
+* Object code, which is the result of code generation. This is used during
+  regular linking. There is a separate `.o` file for each [codegen unit]. The
+  codegen step can be skipped with the [`-C
+  linker-plugin-lto`][linker-plugin-lto] CLI option, which means each `.o`
+  file will only contain LLVM bitcode, which can be used when using LTO.
```
jyn514, line 24: You say bitcode can be used for LTO three times, maybe only say it twice?
```suggestion
  file will only contain LLVM bitcode.
```

src/backend/libs-and-metadata.md
```diff
@@ -0,0 +1,192 @@
...
+
+### rlib
+
+An `rlib` is an [archive file], which is similar to a tar file. This file
+format is specific to `rustc`, and may change over time. This file contains:
+
+* Object code, which is the result of code generation. This is used during
+  regular linking. There is a separate `.o` file for each [codegen unit]. The
+  codegen step can be skipped with the [`-C
+  linker-plugin-lto`][linker-plugin-lto] CLI option, which means each `.o`
```
jyn514, line 23: Why does rustc reuse .o as the extension for bitcode files? I would expect .ll or something.

src/backend/libs-and-metadata.md
```diff
@@ -0,0 +1,192 @@
...
+[LLVM bitcode]: https://llvm.org/docs/BitCodeFormat.html
+[Link Time Optimization]: https://llvm.org/docs/LinkTimeOptimization.html
+[codegen unit]: ../backend/codegen.md
+[embed-bitcode]: https://doc.rust-lang.org/rustc/codegen-options/index.html#embed-bitcode
+[linker-plugin-lto]: https://doc.rust-lang.org/rustc/codegen-options/index.html#linker-plugin-lto
+
+### dylib
+
+A `dylib` is a platform-specific shared library. It includes the `rustc`
+[metadata] in a special section called `.rustc` in a compressed format.
```
jyn514, line 45: I would clarify that section is being used as jargon here.
```suggestion
[metadata] in a special link section called `.rustc` in a compressed format.
```

src/backend/libs-and-metadata.md
```diff
@@ -0,0 +1,192 @@
...
+package.
+
+Here are a few highlights of things it contains:
+
+* The version of the `rustc` compiler. The compiler will refuse to load files
+  from any other version.
+* The [Strict Version Hash](#strict-version-hash) (SVH). This helps ensure the
+  correct dependency is loaded.
+* The [Crate Disambiguator](#crate-disambiguator). This is a hash used for
+  various things to disambiguate between different crates of the same name.
```
jyn514, line 76:
```suggestion
  to disambiguate between different crates of the same name.
```

src/backend/libs-and-metadata.md
```diff
@@ -0,0 +1,192 @@
...
+### Strict Version Hash
+
+The Strict Version Hash ([SVH], also known as the "crate hash") is a 64-bit
+hash that is used to ensure that the correct crate dependencies are loaded. It
+is possible for a directory to contain multiple copies of the same dependency
+built with different settings, or built from different sources. The crate
+loader will skip any crates that have the wrong SVH.
+
+The SVH is also used for the [incremental compilation] session filename,
+though that usage is mostly historic.
```
jyn514, line 99: What's used instead? I'm not sure what you mean by historic here.

jyn514 (commented):
This is super helpful, thank you!

src/backend/libs-and-metadata.md
camelid: What do you mean? My understanding is this talking about object code (not LLVM bitcode), so `.o` is correct.

src/backend/libs-and-metadata.md
camelid: (You'll need to delete the "for" on the previous line as well.)

src/backend/libs-and-metadata.md
jyn514: See the line below, this is reused for bitcode too: https://github.com/rust-lang/rustc-dev-guide/pull/1044/files#diff-3a3e84f49881b6db90d4538e80369c04098f7a6fe6d20cca0ccad9e46fead011R24

camelid: Ah, it confused me that your comment only extended until the line before :)

ehuss: In the past, the bitcode was stored as a separate, compressed file in the archive (with the `.bc.z` extension).  This was changed to simplify the implementation and to match the behavior of clang, which always places the bitcode in the `.o` file.

src/backend/libs-and-metadata.md
ehuss: In the past, `rustc` would use the incremental directories of dependencies for resolving queries, and would use the SVH in the directory name to ensure it was querying the correct dependency. That was removed (it no longer looks for dependencies), and AFAIK it no longer looks at the SVH in the filename.  I think it could be replaced with a random number and have the same effect (it uses a random number before it knows the svh, it gets renamed at the end).  

The incremental naming convention is described at https://doc.rust-lang.org/nightly/nightly-rustc/rustc_incremental/persist/fs/index.html.

jyn514 (approved):

src/backend/libs-and-metadata.md
camelid: Kind of like `.o` for "output" :)

PR #1055, opened 2021-02-15
part-4-intro.md rewrite - oliver

kw-fn:


src/part-4-intro.md
```diff
@@ -1,12 +1,20 @@
...
-representation, inference, and checking of types, the trait system, and the
-borrow checker. These analyses do not happen as one big pass or set of
-contiguous passes. Rather, they are spread out throughout various parts of the
-compilation process and use different intermediate representations. For example,
-type checking happens on the HIR, while borrow checking happens on the MIR.
-Nonetheless, for the sake of presentation, we will discuss all of these
-analyses in this part of the guide.
+This part discusses analyses which the compiler uses to check various properties
+of source code and to inform later compilation stages. Typically the analyses
+prestented here is what people mean when they refer to "Rust's type system".
```
camelid, line 5:
```suggestion
presented here is what people mean when they refer to "Rust's type system".
```

Also, the "the analyses ... is what people mean" sounds a bit funny, but so does using "are". Is there a better phrase this?

src/part-4-intro.md
```diff
@@ -1,12 +1,20 @@
...
+This includes inference and checking of types, the trait and type systems
+representation, and the borrow checker generally. These analyses are separated
```
camelid, line 7:
```suggestion
The analyses include inference and type-checking, trait solving, and borrow-checking.
This section also discusses rustc's representation of types. These analyses are separated
```

src/part-4-intro.md
```diff
@@ -1,12 +1,20 @@
...
+throughout various parts of the compilation process and use meaningfully
+different Intermediate Representations (IR). For example, type checking happens
+on the HIR, while borrow checking happens on the MIR.
+
+This section of the guide discusses all of these analyses which are the heart of
+Rust's novel approach to ownership which balances type system *expressivity with
+usability*, and which represents the integration of decades of programming
```
camelid, line 14:
```suggestion
Rust's novel approach to ownership, which balances type system *expressivity* with
*usability*, and which represents the integration of decades of programming
```

camelid (commented):
Why is this being rewritten? I see nothing wrong with how it was before, and the new version seems to have information that belongs somewhere else (like talking about "evolving Rust towards its essence").

kw-fn:
I could have added a more descriptive note to the PR originally.

This update is meant to advanced a more descriptive explanation of what the following section ultimately provides the reader through its examination of fundamentals of Rust's type system

camelid:
I don't know, I just don't feel like this adds that much. Personally I prefer how it was before. Maybe others disagree though?

jyn514:
I agree with @Camelid, this seems more like a brochure advertising rust than a guide with technical info.

kw-fn:
I don't know about brochures or advertising and I think that is a rather crude generalization of the contents as proposed

kw-fn:
I would prefer if feedback could be constructive, I'm not interested in defending this PR

rylev:
Thanks for contributing 😊I agree with the others that the content added is not really appropriate for the guide. If you're contributing to Rust, you already know about these guarantees. In a sense, you're already "bought in" to the idea of Rust. This guide should be used for providing information on understanding compiler internals, not on promoting the ideas of Rust as a language.

rylev:
I'm going to close this since it seems there's no desire to have content related to Rust's promises as a language in this section.

PR #1060, opened 2021-02-17
Update running.md

bstrie:


jyn514 (approved):
Thanks!

For posterity: we had a conversation about this on #contribute in discord, the `compile-fail` directory no longer exists and is mostly folded into `src/test/ui`. See https://github.com/rust-lang/rust/pull/80453.

PR #1043, opened 2021-01-28
Update some out-of-date information

camelid:
Fixes #1040. (And some other stuff.)

PR #1064, opened 2021-02-20
Point to main branch on llvm-project's links

JohnTitor:
Their default branch is now main. Other failures should be addressed by #1043.


jyn514 (approved):
Thanks!

PR #1066, opened 2021-02-21
Fix date-check comment formatting

camelid:
Fix date-check comment formatting

I realized that the format I had been previously using for date-checks

    (as of January 2021 )

left a space before the closing parenthesis in the rendered output,
which looks weird.

I could fix this by removing the space between `2021` and `<!-- ...`,
but that looks weird in the source. So I ended up deciding on changing
it to

    (as of  January 2021)


JohnTitor (approved):
Looks reasonable to me 👍

PR #1067, opened 2021-02-21
Fix formatting of more date references

camelid:
It turns out I missed several in #1066.

### DIFF
--- a/src/type-inference.md
+++ b/src/type-inference.md
@@ -71,8 +71,8 @@ inference works, or perhaps this blog post on
 
 [Unification in the Chalk project]: http://smallcultfollowing.com/babysteps/blog/2017/03/25/unification-in-chalk-part-1/
 
-All told, the inference context stores four kinds of inference variables (as of
-<!-- date: 2018-01 --> January 2018):
+All told, the inference context stores four kinds of inference variables
+(as of <!-- date: 2018-01 --> January 2018):
 
 - Type variables, which come in three varieties:
   - General type variables (the most common). These can be unified with any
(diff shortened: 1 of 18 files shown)


src/type-inference.md
```diff
@@ -71,8 +71,8 @@ inference works, or perhaps this blog post on
 
 [Unification in the Chalk project]: http://smallcultfollowing.com/babysteps/blog/2017/03/25/unification-in-chalk-part-1/
 
-All told, the inference context stores four kinds of inference variables (as of
-<!-- date: 2018-01 --> January 2018):
+All told, the inference context stores four kinds of inference variables
+(as of <!-- date: 2018-01 --> January 2018):
```
camelid, line 75: It turns out that mdBook freaks out when a line begins with an HTML comment:

image

Not sure if it's a bug or expected behavior.

camelid: It generates this HTML:

```html
<p>All told, the inference context stores four kinds of inference variables (as of</p>
<!-- date: 2018-01 --> January 2018):
```

Probably pulldown-cmark is the one doing that.

JohnTitor (approved):
Seems you still missed the below:

https://github.com/rust-lang/rustc-dev-guide/blob/c4c0f33a1d1bfad621b15b9dd797d0ed25d70ac1/src/tests/intro.md#L129-L130

Otherwise LGTM (I wonder if we could somehow detect with date-check 🤔).

camelid:
> Seems you still missed the below:
> 
> https://github.com/rust-lang/rustc-dev-guide/blob/c4c0f33a1d1bfad621b15b9dd797d0ed25d70ac1/src/tests/intro.md#L129-L130
> 
> Otherwise LGTM (I wonder if we could somehow detect with date-check 🤔).

Great catch, thanks!

PR #1068, opened 2021-02-21
Fix broken links in query.md

matsujika:
r? @jyn514


LeSeulArtichaut (approved):
So queries have finally been extracted from `rustc_middle`? Nice

Thanks for the PR!

PR #1071, opened 2021-02-28
Cleanup rustdoc chapters a bit

camelid:
* Rename "The walking tour of rustdoc" to "Rustdoc overview",
  which I think is a more accurate name
* Use same name in sidebar TOC as in chapter text
* Make links between the two chapters prominent
* Convert a few `.html` links to `.md`

Probably we should just merge the chapters, but leaving that for later.

### DIFF
--- a/src/rustdoc.md
+++ b/src/rustdoc.md
@@ -1,17 +1,20 @@
-# The walking tour of rustdoc
+# Rustdoc overview
 
 Rustdoc actually uses the rustc internals directly. It lives in-tree with the
 compiler and standard library. This chapter is about how it works.
 For information about Rustdoc's features and how to use them, see
 the [Rustdoc book](https://doc.rust-lang.org/nightly/rustdoc/).
+For more details about how rustdoc works, see the ["Rustdoc internals" chapter].
+
+["Rustdoc internals" chapter]: ./rustdoc-internals.md
 
 Rustdoc is implemented entirely within the crate [`librustdoc`][rd]. It runs
 the compiler up to the point where we have an internal representation of a
 crate (HIR) and the ability to run some queries about the types of items. [HIR]
 and [queries] are discussed in the linked chapters.
 
-[HIR]: ./hir.html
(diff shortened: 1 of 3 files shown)


src/rustdoc.md
```diff
@@ -1,17 +1,20 @@
...
 
 Rustdoc is implemented entirely within the crate [`librustdoc`][rd]. It runs
 the compiler up to the point where we have an internal representation of a
 crate (HIR) and the ability to run some queries about the types of items. [HIR]
 and [queries] are discussed in the linked chapters.
 
-[HIR]: ./hir.html
-[queries]: ./query.html
+[HIR]: ./hir.md
+[queries]: ./query.md
```
jyn514, line 17: What's the difference between these styles of links?

jyn514 (approved):

src/rustdoc.md
camelid: My understanding is that `./foo.md` is link-checked but `./foo.html` is not.

camelid: Or maybe it's that `./foo.md` is backend-independent? (So e.g. LaTeX could be added.)

https://rust-lang.github.io/mdBook/for_developers/preprocessors.html?highlight=link#preprocessors

jyn514: Hmm, ok. It would be nice to find out how the markdown backend is preprocessing it at some point.

igaray: The `mdbook-linkcheck` preprocessor uses the markdown scanner provided by a library called [linkcheck](https://crates.io/crates/linkcheck), which [does](https://github.com/Michael-F-Bryan/linkcheck/blob/master/src/scanners/markdown.rs#L30) scan all links, internal markdown, html, even images. 

`mdbook-linkcheck` can be configured to skip external web links but I don't think we want that as have it explicitly set to [true](https://github.com/rust-lang/rustc-dev-guide/blob/master/book.toml#L22).

Issue #940, opened 2020-11-02
Switch from Travis CI to GHA?


JohnTitor:
See https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/RIP.20travis and https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing :(

rylev:
Would switching effect the number of runners that can be used by rust-lang/rust? I wasn't sure if runners were partitioned by org or by repo. I'd be happy to do this if people think it would be an improvement.

JohnTitor:
@rylev IIRC the limit on the number of jobs is divided between rust-lang/rust and the rest (that's why we have the rust-lang-ci org), and the number is enough to use GHA here. There's no concern for a while so I think it's okay to go ahead, I'm happy to review a PR :)

PR #1073, opened 2021-03-02
Switch from Travis to GHA

JohnTitor:
Fixes https://github.com/rust-lang/rustc-dev-guide/issues/940

Some scripts are taken from https://github.com/rust-lang/std-dev-guide and https://github.com/rust-lang/simpleinfra. Some actions log can be found here: https://github.com/JohnTitor/rustc-dev-guide/actions

TODO: We should also switch required status from Travis to GHA.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,86 @@
...
+        with:
+            toolchain: nightly
+            override: true
+
+      - name: Install Dependencies
+        if: steps.mdbook-cache.outputs.cache-hit != 'true'
+        run: |
+          cargo install mdbook --version ${{ env.MDBOOK_VERSION }}
+          cargo install mdbook-linkcheck --version ${{ env.MDBOOK_LINKCHECK_VERSION }}
+          cargo install mdbook-toc --version ${{ env.MDBOOK_TOC_VERSION }}
```
JohnTitor, line 62: I haven't come up with a good way to use `--version ^x.y.z` while using actions/cache.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,86 @@
...
+          # Set 0 to rebase later.
+          fetch-depth: 0
+
+      - name: Read .env
+        run: |
+          . ./.env
+          echo "MDBOOK_VERSION=${MDBOOK_VERSION}" >> $GITHUB_ENV
+          echo "MDBOOK_LINKCHECK_VERSION=${MDBOOK_LINKCHECK_VERSION}" >> $GITHUB_ENV
+          echo "MDBOOK_TOC_VERSION=${MDBOOK_TOC_VERSION}" >> $GITHUB_ENV
+          echo "DEPLOY_DIR=${DEPLOY_DIR}" >> $GITHUB_ENV
```
jyn514, line 27: Why did you choose to make this a separate file, rather than putting the versions directly in the CI workflow?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,86 @@
...
+          cargo install mdbook-linkcheck --version ${{ env.MDBOOK_LINKCHECK_VERSION }}
+          cargo install mdbook-toc --version ${{ env.MDBOOK_TOC_VERSION }}
+
+      - name: Check build
+        if: github.event_name != 'schedule'
+        run: |
+          git switch -c ci
+          git fetch origin master
+          git rebase origin/master
+          git log --oneline | head -n 10
```
jyn514, line 70: Why does this rebase over master? Doesn't github do that automatically? What happens on merge conflicts?

ci/linkcheck.sh
```diff
@@ -3,22 +3,18 @@
...
   echo "Doing full link check."
   set -x
 elif [ "$CI" = "true" ] ; then # running in PR CI build
-  if [ -z "$TRAVIS_COMMIT_RANGE" ]; then
-    echo "error: unexpected state: TRAVIS_COMMIT_RANGE must be non-empty in CI"
-    exit 1
-  fi
 
-  CHANGED_FILES=$(git diff --name-only $TRAVIS_COMMIT_RANGE | tr '\n' ' ')
+  CHANGED_FILES=$(git diff --name-only origin/master... | tr '\n' ' ')
```
jyn514, line 14: Are you sure this is correct? I think `origin` may not always be up to date in the pipeline. You can use `github.event.pull_request.base.sha` instead: https://stackoverflow.com/a/61861763/7669110

.github/workflows/ci.yml
JohnTitor: I just copy-pasted from the travis script, this is no longer required now?

.github/workflows/ci.yml
JohnTitor: Initially, I had two config files and placing values to each file was annoying. Then I forgot to revert that change after cleaning up. Thanks for pointing out :)

ci/linkcheck.sh
JohnTitor: I think it works with `git fetch origin master` but your suggestion seems better!

.github/workflows/ci.yml
JohnTitor: (IIRC we introduced it to avoid broken link failures that are already fixed on master.)

.github/workflows/ci.yml
jyn514: This seems fine to me.

.github/workflows/ci.yml
jyn514: I know that rust-lang/rust runs CI after rebasing against master; I don't *think* they do that explicitly and github handles it automatically. Not sure though.

jyn514:
Personally I think this is good as is and we can fix any bugs as they arise :) maybe @spastorino or @LeSeulArtichaut want to take a look though?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,84 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+  schedule:
+    # Run at 18:00 UTC every day
+    - cron: '0 18 * * *'
```
camelid, line 10: Is this the link-check cronjob?

.github/workflows/ci.yml
camelid: I guess that means we'll have to manually update these tools more often?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,84 @@
...
+
+      - name: Cache linkcheck
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/book/linkcheck
+          key: ${{ runner.os }}-${{ hashFiles('./book/linkcheck') }}
+
+      - name: Check line lengths
+        if: github.event_name != 'push'
```
camelid, line 43: Why doesn't this run on push?

ci/linkcheck.sh
```diff
@@ -3,22 +3,22 @@
...
+    echo "error: unexpected state: BASE_SHA must be non-empty in CI"
     exit 1
   fi
 
-  CHANGED_FILES=$(git diff --name-only $TRAVIS_COMMIT_RANGE | tr '\n' ' ')
+  CHANGED_FILES=$(git diff --name-only $BASE_SHA... | tr '\n' ' ')
   FLAGS="--no-cache -f $CHANGED_FILES"
 
-  echo "Checking files changed in $TRAVIS_COMMIT_RANGE: $CHANGED_FILES"
+  echo "Checking files changed from $BASE_SHA: $CHANGED_FILES"
```
camelid, line 21: nit: I think "since" is better here:

```suggestion
  echo "Checking files changed since $BASE_SHA: $CHANGED_FILES"
```

.github/workflows/ci.yml
LeSeulArtichaut: That sounds surprising to me, but a quick search for 'rebase' in rust's `src/ci` found no result. 🤷‍♂️

.github/workflows/ci.yml
LeSeulArtichaut: From what I understand, all the steps in the file are executed when master or a PR is pushed to, and at a regular interval (cron job). Is that correct?

camelid: I think you're partially correct, but there are several instances of https://github.com/rust-lang/rustc-dev-guide/pull/1073#discussion_r585800357.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,84 @@
...
+        run: |
+          git switch -c ci
+          git fetch origin master
+          git rebase origin/master
+          git log --oneline | head -n 10
+          mdbook build
+
+      - name: Deploy to gh-pages
+        if: github.event_name == 'push' && github.repository == 'rust-lang/rustc-dev-guide'
```
LeSeulArtichaut, line 72: Does this run if I was to ever push a patch branch in `rust-lang/rustc-dev-guide` by accident? Can it cause loops with the deploy branch?

.github/workflows/ci.yml
LeSeulArtichaut: Is it worth it to introduce some automation to update those dependencies?

camelid: Well, I guess we could might be able to get dependabot setup, but I'm not sure if it's worth doing it here.

.github/workflows/ci.yml
JohnTitor: IIRC this was introduced when we did huge refactorings on librustc, which causes a ton of link failures. Now the broken rate is low, so we could remove rebasing, I think.

.github/workflows/ci.yml
JohnTitor: > Is this the link-check cronjob?

Yeah, it's the same as what we have on travis.

image

.github/workflows/ci.yml
JohnTitor: I'm not sure we should block the deployment by the line length failure. It will be caught by the cron job anyway.

.github/workflows/ci.yml
JohnTitor: Generally no. This should be triggered only when we push to the master branch so `patch-1` branch or something else doesn't hurt AFAIK. But if you tweak the condition, it may.

.github/workflows/ci.yml
camelid: Will the line-length check run for PRs? I was confused about whether `push` was just for pushes to master, or if it included PR pushes as well.

.github/workflows/ci.yml
JohnTitor: AFAIK dependabot only updates Cargo.toml or Cargo.lock so it doesn't help us. But an automation job or something else would be great as follow-up work.

> I guess that means we'll have to manually update these tools more often?

Yes, we _can_ still use `--version ^x.y.z` specifying but it doesn't update the mdbook and their tools version due to the cache key.

.github/workflows/ci.yml
JohnTitor: > Will the line-length check run for PRs? I was confused about whether push was just for pushes to master, or if it included PR pushes as well.

`push` here means just for master as we restrict it here: https://github.com/JohnTitor/rustc-dev-guide/blob/4c70da4512e1019a4deb0de9160791965b73a6a8/.github/workflows/ci.yml#L4-L6

For instance, I pushed this branch but the workflow wasn't triggered.
https://github.com/JohnTitor/rustc-dev-guide/tree/doesnt-trigger-event

.github/workflows/ci.yml
rylev: Is there a reason to not have this in a separate yml file inside of workflows?

JohnTitor: > Is there a reason to not have this in a separate yml file inside of workflows?

I don't think it's worth having duplicate workflows with multiple files.

LeSeulArtichaut:
Mhh, btw we won't be able to merge this without disabling the required CI pass from Travis.

Is this waiting on something else?

JohnTitor:
> Is this waiting on something else?

There should not be any blocker on this PR, I think. But yeah, we have to tweak the required status.
@Mark-Simulacrum may I ask you?

Mark-Simulacrum (approved):

Mark-Simulacrum:
I'll need to figure out how to merge - it looks like GitHub is not seeing the CI/ci workflow yet, maybe it needs to run on a non-PR first. I think maybe the easiest thing is to just override the protections, merge into master, and then go back and try and configure things.

Mark-Simulacrum:
Updated the requirements.

PR #1076, opened 2021-03-08
Add stub about the THIR

### LINKED ISSUE #1075: Update documentation about THIR
With rust-lang/rust#82495, the chapters touching THIR will probably need to be updated.
cc https://github.com/rust-lang/project-thir-unsafeck/issues/6

LeSeulArtichaut:
With the ongoing changes to the THIR (rust-lang/rust#82495, but [project-thir-unsafeck](https://github.com/rust-lang/project-thir-unsafeck) in general), it might deserve its own chapter. Here's a stub to begin with.

Helps with rust-lang/project-thir-unsafeck#6, blocked on rust-lang/rust#82495.
Closes #1075.


src/thir.md
```diff
@@ -0,0 +1,39 @@
...
+  is what is done with the HIR).
+- besides making the types of all nodes available, the THIR also has additional
+  desugaring compared to the HIR. For example, automatic references and dereferences
+  are made explicit, and method calls and overloaded operators are converted into
+  plain function calls. Destruction scopes are also made explicit.
+
+[HIR]: ./hir.md
+
+The THIR lives in [`rustc_mir_build::thir`][thir]. To construct a `thir::Expr`,
+you can use the `build_thir` function, passing in the memory arena where the THIR
```
camelid, line 34: - nit: `thir` is a pretty vague link target, so it seems possible that we could have incorrect link targets in the future
- It might be nice to link to `thir::Expr` and `build_thir` both so people can quickly read more about them and so that we know if their names change

src/thir.md
```diff
@@ -0,0 +1,39 @@
...
+- like the MIR, the THIR only represents bodies, i.e. "executable code"; this includes
+  function bodies, but also `const` initializers, for example. Consequently, the THIR
+  has no representation for items like `struct`s or `trait`s.
+- a body of THIR is only stored temporarily and is dropped as soon as it's no longer
+  needed, as opposed to being stored until the end of the compilation process (which
+  is what is done with the HIR).
+- besides making the types of all nodes available, the THIR also has additional
+  desugaring compared to the HIR. For example, automatic references and dereferences
+  are made explicit, and method calls and overloaded operators are converted into
+  plain function calls. Destruction scopes are also made explicit.
```
camelid, line 29: nit: Each of these list items is a full paragraph, so it might be better to capitalize the first sentence of each (I also replaced "a" with "each" in the second list item) and use "loose" spacing:

```suggestion
- Like the MIR, the THIR only represents bodies, i.e. "executable code"; this includes
  function bodies, but also `const` initializers, for example. Consequently, the THIR
  has no representation for items like `struct`s or `trait`s.

- Each body of THIR is only stored temporarily and is dropped as soon as it's no longer
  needed, as opposed to being stored until the end of the compilation process (which
  is what is done with the HIR).

- Besides making the types of all nodes available, the THIR also has additional
  desugaring compared to the HIR. For example, automatic references and dereferences
  are made explicit, and method calls and overloaded operators are converted into
  plain function calls. Destruction scopes are also made explicit.
```

camelid (commented):
Looks good!

camelid:
By the way, should this close #1075 or is there more to do?

LeSeulArtichaut:
I think this can close #1075

jyn514 (approved):
Thanks!

LeSeulArtichaut:
rust-lang/rust#82495 hasn't been merged yet (there is a test failure that magically appeared out of nowhere :confused:)

jyn514:
Oh oops - do you want me to revert the change? It sounds like 82495 will be merged shortly so I think it's ok for the docs to be out of sync temporarily.

LeSeulArtichaut:
Let's see. The PR was retried, so if it fails again I'll file a revert

LeSeulArtichaut:
#82495 has been merged now, so this is sorted out

PR #1077, opened 2021-03-09
Update cmake version in prerequisites.md

wesleywiser:
LLVM requires at least cmake 3.13.4.

https://www.llvm.org/docs/CMake.html

Corresponding PR in https://github.com/rust-lang/rust/pull/82937


LeSeulArtichaut (approved):
Thanks!

PR #1086, opened 2021-03-11
Complete unfinished statement

spastorino:
I guess like any other is what the writer was trying to say. cc @jyn514


jyn514 (approved):
Oops, thanks :laughing:

PR #1089, opened 2021-03-11
Some tweaks for GHA

### LINKED ISSUE #1085: GitHub Actions workflow triggered on push to master branch of fork
Workflow run is here: https://github.com/camelid/rustc-dev-guide/runs/2082979294

This is a bug and was introduced by the new GHA CI.

cc @JohnTitor

JohnTitor:
Fixes #1085

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,7 @@ on:
 
 jobs:
   ci:
+    if: github.repository == 'rust-lang/rustc-dev-guide'
     runs-on: ubuntu-latest
     env:
       MDBOOK_VERSION: 0.4.7
@@ -63,7 +64,7 @@ jobs:
         run: mdbook build
 
       - name: Deploy to gh-pages
-        if: github.event_name == 'push' && github.repository == 'rust-lang/rustc-dev-guide'
+        if: github.event_name == 'push'
         run: |
           touch "${{ env.DEPLOY_DIR }}/.nojekyll"
           cp CNAME "${{ env.DEPLOY_DIR }}"
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-[![Travis CI badge](https://api.travis-ci.com/rust-lang/rustc-dev-guide.svg?branch=master)](https://travis-ci.com/github/rust-lang/rustc-dev-guide)
+[![CI](https://github.com/rust-lang/rustc-dev-guide/actions/workflows/ci.yml/badge.svg)](https://github.com/rust-lang/rustc-dev-guide/actions/workflows/ci.yml)
 
 
 This is a collaborative effort to build a guide that explains how rustc
--- a/ci/linkcheck.sh
+++ b/ci/linkcheck.sh
@@ -28,4 +28,4 @@ else # running locally
   echo "Checking files changed in $COMMIT_RANGE: $CHANGED_FILES"
 fi
 
-exec mdbook-linkcheck $FLAGS -- $GITHUB_WORKSPACE
+exec mdbook-linkcheck $FLAGS
--- a/ci/linkcheck.sh
+++ b/ci/linkcheck.sh
@@ -28,4 +28,4 @@ else # running locally
   echo "Checking files changed in $COMMIT_RANGE: $CHANGED_FILES"
 fi
 
-exec mdbook-linkcheck $FLAGS -- $TRAVIS_BUILD_DIR
+exec mdbook-linkcheck $FLAGS -- $GITHUB_WORKSPACE
--- a/.github/workflows/date-check.yml
+++ b/.github/workflows/date-check.yml
@@ -10,6 +10,7 @@ on:
 
 jobs:
   date-check:
+    if: github.repository == 'rust-lang/rustc-dev-guide'
     runs-on: ubuntu-latest
 
     steps:


.github/workflows/ci.yml
```diff
@@ -23,6 +23,7 @@ jobs:
         with:
           # linkcheck needs the base commit.
           fetch-depth: 0
+    if: github.repository == 'rust-lang/rustc-dev-guide'
```
JohnTitor, line 26: @camelid We can add the if condition here.

README.md
```diff
@@ -1,4 +1,4 @@
-[![Travis CI badge](https://api.travis-ci.com/rust-lang/rustc-dev-guide.svg?branch=master)](https://travis-ci.com/github/rust-lang/rustc-dev-guide)
+[![CI](https://github.com/rust-lang/rustc-dev-guide/actions/workflows/ci.yml/badge.svg)](https://github.com/rust-lang/rustc-dev-guide/actions/workflows/ci.yml)
```
JohnTitor, line 1: NOTE: This catches all the events, e.g., scheduled and push. Do we want to catch only one event?

ci/linkcheck.sh
```diff
@@ -28,4 +28,4 @@ else # running locally
   echo "Checking files changed in $COMMIT_RANGE: $CHANGED_FILES"
 fi
 
-exec mdbook-linkcheck $FLAGS -- $TRAVIS_BUILD_DIR
+exec mdbook-linkcheck $FLAGS -- $GITHUB_WORKSPACE
```
JohnTitor, line 31: Do we still need this?

.github/workflows/ci.yml
```diff
@@ -12,6 +12,7 @@ on:
 jobs:
   ci:
     runs-on: ubuntu-latest
+    if: github.repository == 'rust-lang/rustc-dev-guide'
     env:
```
camelid, line 16: nit: For this and for `date-check.yml` I think it would be better to have the `if` before `runs-on` because right now it's in between `runs-on` and `env` which are in the same category.

camelid:
This looks good to me, but I don't have much experience with GHA, so it might be nice to have someone else review as well.

.github/workflows/ci.yml
```diff
@@ -11,6 +11,7 @@ on:
 
 jobs:
   ci:
+    if: github.repository == 'rust-lang/rustc-dev-guide'
```
rylev, line 14: The whole issue about this run on forks confuses me. Actions shouldn't run on forks by default...

README.md
rylev: This shouldn't be affected by the date checker, and it should only be for events on the default branch, so I think it's fine. Why are we running this on a cron anyway?

ci/linkcheck.sh
rylev: I'm not sure we do. Can we try removing this completely?

rylev (commented):
I have some questions...

.github/workflows/ci.yml
JohnTitor: This still avoids running workflows by mistakes.

ci/linkcheck.sh
JohnTitor: Okay, will remove it.

README.md
JohnTitor: > Why are we running this on a cron anyway?

We can catch the broken links ASAP with the daily check.

.github/workflows/ci.yml
jyn514: How would you run it by mistake? You'd have to modify the workflow file, right? If you're intentionally changing CI I don't think we should make it harder.

camelid: > Actions shouldn't run on forks by default...

Yeah, I feel confused about that. Is there a setting somewhere in the config that changed the behavior?

JohnTitor: > You'd have to modify the workflow file, right? If you're intentionally changing CI I don't think we should make it harder.

No, without this, workflows can be triggered even if you don't edit the config file. If you pull upstream commits and push to fork's master, for example, the workflows will run on your fork.

I just forked rust-lang/docs.rs and the actions' config looks like:
image

Actually, workflows are currently triggered on @camelid's and my fork repos, that's why I made this change.

camelid: > I just forked rust-lang/docs.rs and the actions' config looks like:
> image

But my fork of docs.rs (or any other repo) never runs GHA...

JohnTitor: Ah, sorry, it was confusing. You have to click this to enable it actually:
image

I just mean it _can be_ triggered without editing workflows/*.yml

jyn514: I'm really confused though - if someone explicitly enables ci on their fork, why wouldn't we want it to run?

camelid: Hmm, maybe the issue is that workflows were added *after* we made our forks, where as for other repos they were added before we forked?

jyn514: I don't understand though - what here is a bug? Everything @JohnTitor's shown looks to me as if it's working as expected.

camelid: The issue is that by default the rustc-dev-guide GHA runs on forks when you push to your fork's master. No other repo I know of does that.

JohnTitor: > Hmm, maybe the issue is that workflows were added after we made our forks, where as for other repos they were added before we forked?

Oh I didn't know that behavior, hmm, then it should be more helpful than I thought. So this would be a problem for all the contributors who forked this repo before we introduced GHA if it's true, and it'll cause the notifications and eat the concurrent job limits.

Anyway, I don't see any negative effects by this so I'd like to add this stopper for further issues.

ci/linkcheck.sh
```diff
@@ -27,5 +27,3 @@ else # running locally
 
   echo "Checking files changed in $COMMIT_RANGE: $CHANGED_FILES"
 fi
-
-exec mdbook-linkcheck $FLAGS -- $TRAVIS_BUILD_DIR
```
camelid, line 31: Wait, why was this whole line removed? Without this line, this script is useless.

.github/workflows/ci.yml
camelid: Interestingly, this just happened to me on my fork of the Rust playground. It hasn't happened before. I wonder if this is a new GitHub behavior?

ci/linkcheck.sh
JohnTitor: Ouch, my bad, fixed.

.github/workflows/ci.yml
JohnTitor: So, does someone still have a concern about this? I think this is the only blocker here but I don't see any disadvantage.

jyn514: I'm still a little confused why this is necessary, but I don't push to master branches on forks, so if you think it's useful I'm fine merging it.

jyn514 (approved):
LGTM - @rylev did you have any more comments on https://github.com/rust-lang/rustc-dev-guide/pull/1089/files#r597581782 ?

jyn514:
I think this is ok as-is - @rylev if you have more comments we can address them in follow-ups.

PR #1092, opened 2021-03-23
Suggest using `git range-diff`

camelid:
I have found it to be very helpful when rebasing.

### DIFF
--- a/src/git.md
+++ b/src/git.md
@@ -217,6 +217,7 @@ there are no glaring errors.
 Once you're all done fixing the conflicts, you need to stage the files that had
 conflicts in them via `git add`. Afterwards, run `git rebase --continue` to let
 Git know that you've resolved the conflicts and it should finish the rebase.
+
 Once the rebase has succeeded, you'll want to update the associated branch on
 your fork with `git push --force-with-lease`.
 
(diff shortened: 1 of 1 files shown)


src/git.md
```diff
@@ -217,8 +217,19 @@ there are no glaring errors.
...
+
+Once the rebase has succeeded, you may want to review the changes between your
+old branch and your new one. You can do that with the `git range-diff master
+@{u} HEAD` command, which can be very helpful when you had a complicated rebase
+and you want to make sure you changed the right things. Unlike in regular Git
+diffs, you'll see a `-` or `+` next to another `-` or `+` in the range-diff
+output. The marker on the left indicates a change between the old branch and the
+new branch, and the marker on the right indicates a change you've committed. So,
+you can think of a range-diff as a "diff of diffs" since it shows you the
+differences between your old diff and your new diff.
```
jyn514, line 229: This is in the tutorial section ("how do I rebase at all"). I would prefer to move it to "advanced rebasing".

jyn514 (commented):
This is a great suggestion, thanks!

jyn514:
I actually suggested that github do this by default, but I don't know how to get in contact with them. The 'issues' repo seems to be mostly ignored. https://github.com/isaacs/github/issues/1834

camelid:
> I actually suggested that github do this by default, but I don't know how to get in contact with them. The 'issues' repo seems to be mostly ignored. [isaacs/github#1834](https://github.com/isaacs/github/issues/1834)

Yeah, I have no idea how to contact them either.

src/git.md
camelid: I guess it is a somewhat advanced feature, though I use it on pretty much every rebase. The problem with putting it in the advanced section is I worry people won't see it.

Even though it can be a bit confusing to use the command at first, I think it makes rebasing much easier.

On the other hand, I don't want to overwhelm people and make them think it's necessary before doing any rebasing.

So, do you still think it should move?

jyn514: Yes, I think so. I think the initial guide should be limited to things people strictly need to rebase. I'm not too worried about people not seeing it - or I guess, I think it's not a big deal if they don't.

camelid: Maybe we could add a link to the range-diff section from the tutorial section so people can look at it if they want?

jyn514: The section is about 2 paragraphs directly below, it will be the next thing you see if you keep reading. I'm not sure why you think it's hard to find.

camelid: I think I just feel like range-diff has been extremely helpful and I want to make sure people see it, even if it's not that far down. But I'm okay with skipping a link 👍

src/git.md
```diff
@@ -263,6 +264,26 @@ You also may want to squash just the last few commits together, possibly
 because they only represent "fixups" and not real changes. For example,
 `git rebase --interactive HEAD~2` will allow you to edit the two commits only.
 
+### `git range-diff`
+
+After completing a complicated rebase, or even a relatively simple one, you may
+want to review the changes between your old branch and your new one. You can do
+that with the `git range-diff master @{u} HEAD` command, which can be very
```
jyn514, line 271: Maybe say that `@{u}` means the upstream branch? I had to figure it out by experimenting.

camelid: I added a bunch more detail, let me know what you think.

src/git.md
```diff
@@ -263,6 +264,67 @@ You also may want to squash just the last few commits together, possibly
 because they only represent "fixups" and not real changes. For example,
 `git rebase --interactive HEAD~2` will allow you to edit the two commits only.
 
+### `git range-diff`
+
+After completing a complicated rebase, or even a relatively simple one, you may
```
jyn514, line 269:
```suggestion
After a rebase, you may
```

src/git.md
```diff
@@ -263,6 +264,67 @@ You also may want to squash just the last few commits together, possibly
...
+want to review the changes between your old branch and your new one. You can do
+that with the `git range-diff master @{upstream} HEAD` command, which can be
+very helpful when you had a complicated rebase and you want to make sure you
+changed the right things.
```
jyn514, line 273:
```suggestion
that with `git range-diff master @{upstream} HEAD`.
```

jyn514 (approved):

jyn514:
Hi @camelid, are you planning to follow up on this?

src/git.md
camelid: Or maybe:

```suggestion
After completing a rebase, and before pushing up your changes,
```

Because `git range-diff master @{u} HEAD` only works before you push.

src/git.md
camelid: How about something like this:

```suggestion
that with the `git range-diff master @{upstream} HEAD` command, which can help
you make sure that you merged the upstream changes correctly.
```

camelid:
> Hi @camelid, are you planning to follow up on this?

Yes, somehow I missed that this was waiting on me!

src/git.md
jyn514: Sounds good to me :)

src/git.md
jyn514: I don't see what that adds? We already said "you may want to review the changes" above.

camelid: Yeah, I see what you mean 👍

src/git.md
```diff
@@ -263,6 +264,65 @@ You also may want to squash just the last few commits together, possibly
 because they only represent "fixups" and not real changes. For example,
 `git rebase --interactive HEAD~2` will allow you to edit the two commits only.
 
+### `git range-diff`
+
+After completing a rebase, and before pushing up your changes, want to review
```
jyn514, line 269:
```suggestion
After completing a rebase, and before pushing up your changes, you may want to review
```

jyn514 (approved):
r=me with the typo fixed, thanks for working on this :)

src/git.md
camelid: Thanks, good catch!

PR #1095, opened 2021-03-28
Update rustc-driver-*.rs examples

Undxxx:
remove the unused fields and change unknown field

### DIFF
--- a/examples/rustc-driver-interacting-with-the-ast.rs
+++ b/examples/rustc-driver-interacting-with-the-ast.rs
@@ -15,8 +15,8 @@ extern crate rustc_span;
 use rustc_ast_pretty::pprust::item_to_string;
 use rustc_errors::registry;
 use rustc_session::config;
-use rustc_session::config::PpMode::PpmSource;
-use rustc_session::config::PpSourceMode::PpmExpanded;
+//use rustc_session::config::PpMode::PpmSource;
+//use rustc_session::config::PpSourceMode::PpmExpanded;
 use rustc_span::source_map;
 use std::path;
 use std::process;
@@ -46,7 +46,7 @@ fn main() {
         output_file: None,
         file_loader: None,
         stderr: None,
-        crate_name: None,
+        //crate_name: None,
         lint_caps: rustc_hash::FxHashMap::default(),
         register_lints: None,
         override_queries: None,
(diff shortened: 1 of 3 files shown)


examples/rustc-driver-interacting-with-the-ast.rs
```diff
@@ -57,8 +57,8 @@ fn main() {
         compiler.enter(|queries| {
             // TODO: add this to -Z unpretty
             let ast_krate = queries.parse().unwrap().take();
-            let ast_krate_mod = ast_krate.module;
-            for item in ast_krate_mod.items {
+            //let ast_krate_mod = ast_krate.module;
```
JohnTitor, line 60:
```suggestion
```

examples/rustc-driver-interacting-with-the-ast.rs
```diff
@@ -46,7 +46,7 @@ fn main() {
         output_file: None,
         file_loader: None,
         stderr: None,
-        crate_name: None,
+        //crate_name: None,
         lint_caps: rustc_hash::FxHashMap::default(),
```
JohnTitor, line 50: We now need this field for more recent nightly.

```suggestion
        lint_caps: rustc_hash::FxHashMap::default(),
        parse_sess_created: None,
```

JohnTitor (changes requested):
Thanks for the PR! You want to run this with the recent nightly, right? Then we have to change L78 to `let def_id = tcx.hir().local_def_id(item.hir_id());`.

Furthermore, it'd be great to mention the specific nightly version like `<!-- date: xx -->` to keep the examples working on the docs.

JohnTitor:
And other examples may also be stalled, we should check it.

examples/rustc-driver-interacting-with-the-ast.rs
```diff
@@ -15,8 +15,8 @@ extern crate rustc_span;
 use rustc_ast_pretty::pprust::item_to_string;
 use rustc_errors::registry;
 use rustc_session::config;
-use rustc_session::config::PpMode::PpmSource;
-use rustc_session::config::PpSourceMode::PpmExpanded;
+//use rustc_session::config::PpMode::PpmSource;
+//use rustc_session::config::PpSourceMode::PpmExpanded;
```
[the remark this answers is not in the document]
Undxxx: > And other examples may also be stalled, we should check it.

I have removed the unused lines and added my version to the head of the files, thanks!

examples/rustc-driver-interacting-with-the-ast.rs
JohnTitor: While we're on it, could you use `rustc 1.53.0-nightly (9b0edb7fd 2021-03-27)` instead? At least, we have one breaking change, see the above.

JohnTitor (approved):
Thanks!

PR #1097, opened 2021-03-29
Explain the new valtree system for type level constants.

oli-obk:
Dependent on getting https://github.com/rust-lang/rust/pull/83234 merged first.

cc @rust-lang/wg-const-eval

### DIFF
--- a/src/const-eval.md
+++ b/src/const-eval.md
@@ -66,8 +66,9 @@ level constants, as it is not clear how they should be represented, because thei
 is unknown. Similarly there is no way to represent raw pointers, as addresses are unknown at
 compile-time and thus we cannot make any assumptions about them. References on the other hand
 *can* be represented, as equality for references is defined as equality on their value, so we
-ignore their address and just look at the backing value. We must make sure that the pointer value
-of the references are not observable. We thus encode `&42` exactly like `42`. Any conversion from
+ignore their address and just look at the backing value. We must make sure that the pointer values
+of the references are not observable at compile time. We thus encode `&42` exactly like `42`.
+Any conversion from
 valtree back to codegen constants must reintroduce an actual indirection. At codegen time the
 addresses may be deduplicated between multiple uses or not, entirely depending on arbitrary
 optimization choices.
--- a/src/const-eval.md
+++ b/src/const-eval.md
@@ -20,17 +20,62 @@ Additionally constant evaluation can be used to reduce the workload or binary
 size at runtime by precomputing complex operations at compiletime and only
 storing the result.
 
+All uses of constant evaluation can either be categorized as "influencing the type system"
+(array lengths, enum variant discriminants, const generic parameters), or as solely being
+done to precompute expressions to be used at runtime.
+
 Constant evaluation can be done by calling the `const_eval_*` functions of `TyCtxt`.
 They're the wrappers of the `const_eval` query.
+
+* `const_eval_global_id_for_typeck` evaluates a constant to a valtree,
+  so the result value can be further inspected by the compiler.
+* `const_eval_global_id` evaluate a constant to an "opaque blob" containing its final value;
+  this is only useful for codegen backends and the CTFE evaluator engine itself.
+* `eval_static_initializer` specifically computes the initial values of a static.
+  Statics are special; all other functions do not represent statics correctly
+  and have thus assertions preventing their use on statics.
 
 The `const_eval_*` functions use a [`ParamEnv`](./param_env.html) of environment
 in which the constant is evaluated (e.g. the function within which the constant is used)
 and a [`GlobalId`]. The `GlobalId` is made up of an `Instance` referring to a constant
 or static or of an `Instance` of a function and an index into the function's `Promoted` table.
 
-Constant evaluation returns a [`EvalToConstValueResult`] with either the error, or a
-representation of the constant. `static` initializers are always represented as
-[`miri`](./miri.html) virtual memory allocations (via [`ConstValue::ByRef`]).
+Constant evaluation returns an [`EvalToValTreeResult`] for type system constants or
+[`EvalToConstValueResult`] with either the error, or a representation of the constant.
+
+Constants for the type system are encoded in "valtree representation". The `ValTree` datastructure
+allows us to represent
+
+* arrays,
+* many structs,
+* tuples,
+* enums and,
+* most primitives.
+
+The basic rule for
+being permitted in the type system is that every value must be uniquely represented. In other
+words: a specific value must only be representable in one specific way. For example: there is only
+one way to represent an array of two integers as a `ValTree`:
+`ValTree::Branch(&[ValTree::Leaf(first_int), ValTree;:Leaf(second_int)])`.
+Even though theoretically a `[u32; 2]` could be encoded in a `u64` and thus just be a
+`ValTree::Leaf(bits_of_two_u32)`, that is not a legal construction of `ValTree`
+(and is very complex to do, so it is unlikely anyone is tempted to do so).
+
+These rules also mean that some values are not representable. There can be no `union`s in type
+level constants, as it is not clear how they should be represented, because their active variant
+is unknown. Similarly there is no way to represent pointers, as addresses are unknown at
+compile-time and thus we cannot make any assumptions about them. References on the other hand
+*can* be represented, as equality for references is defined as equality on their value, so we
+ignore their address and just look at the backing value. We must make sure that the pointer value
+of the references are not observable. We thus encode `&42` exactly like `42`. Any conversion from
+valtree back to codegen constants must reintroduce an actual indirection. At codegen time the
+addresses may be deduplicated between multiple uses or not, entirely depending on arbitrary
+optimization choices.
+
+As a consequence, all decoding of `ValTree` must happen by matching on the type first and making
+decisions depending on that. The value itself gives no useful information without the type that
+belongs to it.
+
 Other constants get represented as [`ConstValue::Scalar`]
 or [`ConstValue::Slice`] if possible. This means that the `const_eval_*`
 functions cannot be used to create miri-pointers to the evaluated constant.
@@ -42,4 +87,5 @@ If you need the value of a constant inside Miri, you need to directly work with
 [`ConstValue::Slice`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/interpret/value/enum.ConstValue.html#variant.Slice
 [`ConstValue::ByRef`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/interpret/value/enum.ConstValue.html#variant.ByRef
 [`EvalToConstValueResult`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/interpret/error/type.EvalToConstValueResult.html
+[`EvalToValTreeResult`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/interpret/error/type.EvalToValTreeResult.html
 [`const_to_op`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_const_eval/interpret/struct.InterpCx.html#method.const_to_op
--- a/src/const-eval.md
+++ b/src/const-eval.md
@@ -63,7 +63,7 @@ Even though theoretically a `[u32; 2]` could be encoded in a `u64` and thus just
 
 These rules also mean that some values are not representable. There can be no `union`s in type
 level constants, as it is not clear how they should be represented, because their active variant
-is unknown. Similarly there is no way to represent pointers, as addresses are unknown at
+is unknown. Similarly there is no way to represent raw pointers, as addresses are unknown at
 compile-time and thus we cannot make any assumptions about them. References on the other hand
 *can* be represented, as equality for references is defined as equality on their value, so we
 ignore their address and just look at the backing value. We must make sure that the pointer value


src/const-eval.md
```diff
@@ -20,17 +20,44 @@ Additionally constant evaluation can be used to reduce the workload or binary
...
+Constant evaluation returns a [`EvalToValTreeResult`] (for type system constants) or [`EvalToConstValueResult`] with either the error, or a
+representation of the constant.
+
+Constants for the type system are encoded in "valtree representation". The `ValTree` datastructure
+allows us to represent arrays, many structs, tuples, enums and most primitives. The basic rule for
+being permitted in the type system is that every value must be uniquely represented. In other words:
+a specific value must only be representable in one specific way. For example: there is only one way
+to represent an array of two integers as a `ValTree`: `ValTree::Branch(&[ValTree::Leaf(first_int), ValTree;:Leaf(second_int)])`.
+Even though theoretically a `[u32; 2]` could  be encoded in a `u64` and thus just be a `ValTree::Leaf(bits_of_two_u32)`, that
+is not a legal construction of `ValTree` (and is so complex to do, that it is unlikely to tempt anyone to do so).
```
lcnr, line 47:
```suggestion
is not a legal construction of `ValTree` (and is so complex to do, so it is unlikely anyone is tempted to do so).
```
the wording seems a bit weird to me

src/const-eval.md
```diff
@@ -20,17 +20,44 @@ Additionally constant evaluation can be used to reduce the workload or binary
...
+These rules also mean that some values are not representable. There can be no `union`s in type level
+constants, as it is not clear how they should be represented, because their active variant is unknown.
+Similarly there is no way to represent pointers, as addresses are unknown at compile-time and thus we
+cannot make any assumptions about them. References on the other hand can be represented, as equality
+for references is defined as equality on their value, so we ignore their address and just look at the
+backing value. This means that there is no difference in encoding for `&42` and `42`.
+As a consequence, all decoding of `ValTree` must happen by matching on the type first and making decisions
+depending on that. The value itself gives no useful information without the type that belongs to it.
+One notable oddity is `&str` representation. There is no sized equivalent of it, so unlike slices we cannot
+choose to represent them as their sized variant (slices are represented as arrays). `&str` thus has
```
lcnr, line 57: Is this a correct implication? I assume that we do so for perf reasons, I don't expect there to be any semantic reason we can't represent `&str` in the same way as `&[u8]`.

lcnr (approved):

src/const-eval.md
```diff
@@ -20,17 +20,44 @@ Additionally constant evaluation can be used to reduce the workload or binary
 size at runtime by precomputing complex operations at compiletime and only
 storing the result.
 
+All uses of constant evaluation can either be categorized as "influencing the type system"
+(array lengths, enum variant discriminants, const generic parameters), or as solely being
```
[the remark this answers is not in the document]
oli-obk: > do discriminants influence the type system in any way?

as tags are computed from discriminants, yes, the tag is part of the representation.

src/const-eval.md
oli-obk: well... there is that tiny difference described in the text, and I'm using it to justify the existance of a custom variant for `&str`. But you are right, the main reason is performance, but it didn't actually do much difference in my tests (at best going from an 18% regression to a 17% regression), so I may revert it in the future once I have a better handle on the perf issues

src/const-eval.md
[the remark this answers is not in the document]
lcnr: yeah, how does the tag feel into the type system? It feels like I am missing something here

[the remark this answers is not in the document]
oli-obk: Well... You can technically transmute an enum into a layout equivalent datastructure and then use the discriminant to change an array length. But in a less whacky (and unsound) way, if you put a non-c-like enum into a const generic, the tag will end up in the ValTree and thus in the parameters of the const generic.

Enums are encoded by adding their tag as a field before all variant fields. Alternatively we could encode enums by having an `Option<VariantIdx>` on `ValTree::Branches`. There is no strong reason to prefer either, except that enums are rare right now, so we'd be pessimizing non-enum aggregates with an unused `None` field.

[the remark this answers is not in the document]
lcnr: ah, yeah :thinking: if we encode the discriminant in the val tree we need to do it this way. Using the `VariantIdx` might be *cleaner* in the long run, but I don't really think it matters too much.

src/const-eval.md
```diff
@@ -20,17 +20,44 @@ Additionally constant evaluation can be used to reduce the workload or binary
...
+Similarly there is no way to represent pointers, as addresses are unknown at compile-time and thus we
+cannot make any assumptions about them. References on the other hand can be represented, as equality
+for references is defined as equality on their value, so we ignore their address and just look at the
+backing value. This means that there is no difference in encoding for `&42` and `42`.
+As a consequence, all decoding of `ValTree` must happen by matching on the type first and making decisions
+depending on that. The value itself gives no useful information without the type that belongs to it.
+One notable oddity is `&str` representation. There is no sized equivalent of it, so unlike slices we cannot
+choose to represent them as their sized variant (slices are represented as arrays). `&str` thus has
+its own `ValTree` variant `Str`. The advantage of using a custom variant is that we are able to translate
+parser/AST/HIR string literals without any conversion as we use the same (`Symbol`) representation.
```
camelid, line 59: Could you add some paragraph breaks here? Right now it's one long block of text that is a bit hard to read.

src/const-eval.md
```diff
@@ -20,17 +20,44 @@ Additionally constant evaluation can be used to reduce the workload or binary
...
-[`miri`](./miri.html) virtual memory allocations (via [`ConstValue::ByRef`]).
+Constant evaluation returns a [`EvalToValTreeResult`] (for type system constants) or [`EvalToConstValueResult`] with either the error, or a
+representation of the constant.
+
+Constants for the type system are encoded in "valtree representation". The `ValTree` datastructure
+allows us to represent arrays, many structs, tuples, enums and most primitives. The basic rule for
+being permitted in the type system is that every value must be uniquely represented. In other words:
+a specific value must only be representable in one specific way. For example: there is only one way
+to represent an array of two integers as a `ValTree`: `ValTree::Branch(&[ValTree::Leaf(first_int), ValTree;:Leaf(second_int)])`.
+Even though theoretically a `[u32; 2]` could  be encoded in a `u64` and thus just be a `ValTree::Leaf(bits_of_two_u32)`, that
```
camelid, line 46: nit: There's an extra space here.

```suggestion
Even though theoretically a `[u32; 2]` could be encoded in a `u64` and thus just be a `ValTree::Leaf(bits_of_two_u32)`, that
```

camelid (commented):
This looks really good! I left some more stylistic comments (can't comment much on the content since I'm learning about valtrees as I'm reading this :). Totally fine to skip some of the suggestions if you prefer the way it is now :)

src/const-eval.md
```diff
@@ -20,17 +20,50 @@ Additionally constant evaluation can be used to reduce the workload or binary
...
+
+Constants for the type system are encoded in "valtree representation". The `ValTree` datastructure
+allows us to represent arrays, many structs, tuples, enums and most primitives. The basic rule for
+being permitted in the type system is that every value must be uniquely represented. In other
+words: a specific value must only be representable in one specific way. For example: there is only
+one way to represent an array of two integers as a `ValTree`:
+`ValTree::Branch(&[ValTree::Leaf(first_int), ValTree;:Leaf(second_int)])`.
+Even though theoretically a `[u32; 2]` could be encoded in a `u64` and thus just be a
+`ValTree::Leaf(bits_of_two_u32)`, that is not a legal construction of `ValTree`
+(and is so complex to do, so it is unlikely anyone is tempted to do so).
```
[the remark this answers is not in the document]
wesleywiser: I think this "so" was meant as a synonym of "very".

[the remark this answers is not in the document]
camelid: Ah, maybe "very" would be clearer then?

src/const-eval.md
```diff
@@ -20,17 +20,50 @@ Additionally constant evaluation can be used to reduce the workload or binary
...
+compile-time and thus we cannot make any assumptions about them. References on the other hand
+can be represented, as equality for references is defined as equality on their value, so we
+ignore their address and just look at the backing value.
+This means that there is no difference in encoding for `&42` and `42`.
+
+As a consequence, all decoding of `ValTree` must happen by matching on the type first and making
+decisions depending on that. The value itself gives no useful information without the type that
+belongs to it. One notable oddity is `&str` representation. There is no sized equivalent of it,
+so unlike slices we cannot choose to represent them as their sized variant (slices are represented
+as arrays). `&str` thus has its own `ValTree` variant `Str`. The advantage of using a custom
```
RalfJung, line 63: I don't follow this argument. Even if there is no sized version of the type, we could still representing an `&str` as a single branch with many leaves below it (one per byte, for example). The only issue with that is performance.

src/const-eval.md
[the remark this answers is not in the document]
RalfJung: Maybe "affect the type system" is the wrong terminology... it's more about "does the compiler work on these values, or does it just treat them as opaque blobs".

Patterns, at least the subset of them that are considered for exhaustiveness checking, are another example of values that the compiler works on, and that hence do/should use valtrees.

src/const-eval.md
```diff
@@ -20,17 +20,50 @@ Additionally constant evaluation can be used to reduce the workload or binary
...
 storing the result.
 
+All uses of constant evaluation can either be categorized as "influencing the type system"
+(array lengths, enum variant discriminants, const generic parameters), or as solely being
+done to precompute expressions to be used at runtime.
+
 Constant evaluation can be done by calling the `const_eval_*` functions of `TyCtxt`.
 They're the wrappers of the `const_eval` query.
 
+`static` initializers must use the `eval_static_initializer` function. All other functions
```
RalfJung, line 30: This comes out of the blue... and it is strange that we are naming this function out of all of them. Given teh big paragraph about valtrees, I'd have expected something like(names are made up as I am not always sure what the real thing is called)

> Constant evaluation can be done by calling the `const_eval_*` functions of `TyCtxt`. They're the wrappers of the `const_eval` query.
> * `const_eval_valtree` evaluates a constant to a valtree, so the result value can be further inspected by the compiler.
> * `const_eval_const` evaluate a constant to an "opaque blob" containing its final value; this is only useful for codegen backends and the CTFE evaluator engine itself.
> `eval_static_initializer` specifically computes the initial values of a static. Statics are special; all other functions do not represent statics correctly and have thus assertions preventing their use on statics.

However, the static special case should also disappear once valtrees are properly done, right? All of this "normalization" magic should not be needed any more.

src/const-eval.md
RalfJung: Looks like @lcnr had already brought this up before... I think the text should be changed; currently, it makes an argument X to support Y but X really has no bearing on Y, so this is confusing.

src/const-eval.md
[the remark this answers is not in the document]
camelid: (Leaving a suggestion so you can apply on GitHub if you like.)

```suggestion
(and is very complex to do, so it is unlikely anyone is tempted to do so).
```

jyn514:
I think @oli-obk is on break - @RalfJung would you be interested in following up on this? NBD if not, it can wait until they get back.

RalfJung:
Given that https://github.com/rust-lang/rust/pull/83234 didn't land yet, I think it might be a bit early to take this one over.

tshepang:
closing as [the PR it documents is also closed](https://github.com/rust-lang/rust/pull/83234#issuecomment-934273521)... feel free to re-open when that PR is revived

lcnr:
valtrees have been merged in https://github.com/rust-lang/rust/pull/96591

going to check if this document is still up to date

src/const-eval.md
```diff
@@ -20,17 +20,50 @@ Additionally constant evaluation can be used to reduce the workload or binary
...
+`ValTree::Branch(&[ValTree::Leaf(first_int), ValTree;:Leaf(second_int)])`.
+Even though theoretically a `[u32; 2]` could be encoded in a `u64` and thus just be a
+`ValTree::Leaf(bits_of_two_u32)`, that is not a legal construction of `ValTree`
+(and is so complex to do, so it is unlikely anyone is tempted to do so).
+
+These rules also mean that some values are not representable. There can be no `union`s in type
+level constants, as it is not clear how they should be represented, because their active variant
+is unknown. Similarly there is no way to represent pointers, as addresses are unknown at
+compile-time and thus we cannot make any assumptions about them. References on the other hand
+can be represented, as equality for references is defined as equality on their value, so we
```
lcnr, line 55: we definitely have to mention here that the pointer value of references must not be observable during ctfe for this to be sound

src/const-eval.md
lcnr: this is still the case

JohnTitor:
@oli-obk Friendly-ping, could you resolve merge conflicts and address the above comments? Once these are done, I think we could merge this!

oli-obk:
done

src/const-eval.md
```diff
@@ -20,17 +20,62 @@ Additionally constant evaluation can be used to reduce the workload or binary
...
+words: a specific value must only be representable in one specific way. For example: there is only
+one way to represent an array of two integers as a `ValTree`:
+`ValTree::Branch(&[ValTree::Leaf(first_int), ValTree;:Leaf(second_int)])`.
+Even though theoretically a `[u32; 2]` could be encoded in a `u64` and thus just be a
+`ValTree::Leaf(bits_of_two_u32)`, that is not a legal construction of `ValTree`
+(and is very complex to do, so it is unlikely anyone is tempted to do so).
+
+These rules also mean that some values are not representable. There can be no `union`s in type
+level constants, as it is not clear how they should be represented, because their active variant
+is unknown. Similarly there is no way to represent pointers, as addresses are unknown at
```
lcnr, line 66: should still use raw pointers here i think

PR #1104, opened 2021-04-04
Mention CI build of LLVM in build instruction

AngelicosPhosphoros:


src/building/how-to-build-and-run.md
```diff
@@ -33,6 +33,9 @@ settings (and possibly others, such as `llvm.ccache`):
 
 ```toml
 [llvm]
+# Set `true` if you don't want build LLVM from source
```
camelid, line 36: I think it would be better to say something like

> Whether to download LLVM from CI instead of building it locally

for consistency with the other comments.

jyn514: The other comments are taken directly from config.toml, if you want to be consistent this should be too: https://github.com/rust-lang/rust/blob/0b417ab5cdfdedffd74fb22cf22d27033c851304/config.toml.example#L33

AngelicosPhosphoros: I fixed.

> The other comments are taken directly from config.toml, if you want to be consistent this should be too: https://github.com/rust-lang/rust/blob/0b417ab5cdfdedffd74fb22cf22d27033c851304/config.toml.example#L33

jyn514 (approved):
This is great, thanks!

PR #1108, opened 2021-04-09
Change wording a bit: `module` => `crate`

WaffleLapkin:
It seems more correct


jyn514 (commented):
Yup, this is correct: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=2b4b485550640037746702beeed53a6d. Thanks :)

jyn514 (approved):

PR #1109, opened 2021-04-09
Improve formatting and update info in "method lookup" section

camelid:


src/method-lookup.md
```diff
@@ -6,18 +6,11 @@ file provides an overview of the process. More detailed notes are in
...
-```
-
-into a more explicit [UFCS] form:
-
-```rust,ignore
-Trait::method(ADJ(receiver), ...) // for a trait call
-ReceiverType::method(ADJ(receiver), ...) // for an inherent method call
-```
+- `Trait::method(ADJ(receiver), ...)` for a trait call
+- `ReceiverType::method(ADJ(receiver), ...)` for an inherent method call
```
jyn514, line 13: Hmm, this doesn't seem any more clear to me? Why did you change it?

camelid: I think it makes it clearer that it's showing one of two options. The way it was before made it seem like it turned the one line `self.` form into two lines of an explicit form. Also, I think it looks better and is clearer when the comment is turned into proper Markdown text.

Here's a before and after, for comparison:

image

image

jyn514 (approved):

PR #1114, opened 2021-04-26
more RA config suggestions

RalfJung:
Cc https://github.com/rust-analyzer/rust-analyzer/issues/8667 @matklad

### DIFF
--- a/src/building/suggested.md
+++ b/src/building/suggested.md
@@ -34,7 +34,9 @@ you can write:
     "rust-analyzer.rustfmt.overrideCommand": [
       "./build/TARGET_TRIPLE/stage0/bin/rustfmt"
     ],
-    "editor.formatOnSave": true
+    "editor.formatOnSave": true,
+    "rust-analyzer.cargo.runBuildScripts": false,
+    "rust-analyzer.procMacro.enable": false,
 }
 ```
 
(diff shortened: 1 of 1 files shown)


src/building/suggested.md
```diff
@@ -34,7 +34,9 @@ you can write:
     "rust-analyzer.rustfmt.overrideCommand": [
       "./build/TARGET_TRIPLE/stage0/bin/rustfmt"
     ],
-    "editor.formatOnSave": true
+    "editor.formatOnSave": true,
+    "rust-analyzer.cargo.runBuildScripts": false,
+    "rust-analyzer.procMacro.enable": false,
```
jyn514, line 39: Doesn't JSON disallow trailing commas?
```suggestion
    "rust-analyzer.procMacro.enable": false
```

jyn514 (approved):
Thanks!

src/building/suggested.md
RalfJung: It works for me in vscode configs...

RalfJung:
Weird, this branch was merged but my commit does not show up in the repo? This is confusing the heck out of my "detect which branches can be deleted" tooling.^^

jyn514:
Hmm, I think I might have used "squash and merge" instead of rebasing it. Not sure if there's  a way  to detect that automatically.

RalfJung:
Ah yeah that would explain it. One more reason I prefer proper merges. :) But obviously I cannot tell you how to run a repo you maintain.

PR #1115, opened 2021-04-26
Remove outdated "ctags" section 

andoriyu:
Right now, it might be unclear that Rust Analyzer works fine if you land on "ctags" page. `ctag` manual also outdated and no longer useful, so I'm removing it from documentation to avoid any confusion.


src/building/suggested.md
```diff
@@ -63,6 +63,33 @@ Task] instead:
 
 [Build Task]: https://code.visualstudio.com/docs/editor/tasks
 
+## Configuring `ctags`
+
+One of the challenges with rustc is that the RLS can't handle it, since it's a
+bootstrapping compiler. This makes code navigation difficult. One solution is to
+use `ctags`.
```
jyn514, line 70: This is no longer relevant since RA is suggested right above.

Actually, I wonder if it still makes sense to suggest CTags at all. RA works decently with vim as far as I know, I would like to avoid suggesting multiple different IDEs if possible.

andoriyu: That's what I was thinking. Maybe removing it completely makes more sense.

jyn514: @nikomatsakis @mark-i-m do you still use ctags at all? If so, do you think they make sense to recommend now that rust-analyzer is a thing?

src/building/suggested.md
```diff
@@ -63,6 +63,33 @@ Task] instead:
...
+and does have built-in Rust support.
+
+The following script can be used to set up Exuberant Ctags:
+[https://github.com/nikomatsakis/rust-etags][etags].
+
+`ctags` integrates into emacs and vim quite easily. The following can then be
+used to build and generate tags:
+
+```console
+$ rust-ctags src/lib* && ./x.py build <something>
```
camelid, line 84: If we end up keeping the ctags docs, this should be changed because `src/lib*` will not match anything in the compiler or standard library (only rustdoc).

camelid: Based on looking at the script, I think that would require multiple invocations.

src/building/suggested.md
mark-i-m: My primary use of ctags is for the [vim tagbar](https://github.com/majutsushi/tagbar.git) pluggin, and RA doesn't currently work with that tmk... there are other pluggins that do, I think, but I haven't spent a lot of time looking.

nikomatsakis: I do not use ctags anymore, I'm all in on VSCode + rust-analyzer at this point.

nikomatsakis: I'd be ok with removing discussion of this hack.

jyn514: > My primary use of ctags is for the vim tagbar pluggin, and RA doesn't currently work with that tmk... there are other pluggins that do, I think, but I haven't spent a lot of time looking.

Gotcha. I think it's ok to leave that out here - people using vim *and* RA *and* the tagbar is probably not super common, and there's still docs at https://github.com/universal-ctags/ctags in case people are still using it.

rylev:
@andoriyu 👋 are you still planning on finishing this PR?

andoriyu:
> @andoriyu 👋 are you still planning on finishing this PR? 

Oh, totally forgotten about it, I thought it's already merged. Not sure what do i need to change about it to be honest.

rylev:
@andoriyu if you're happy with it, I'll give it a review and try to get you feedback or merge it if it's ready to go.

andoriyu:
> @andoriyu if you're happy with it, I'll give it a review and try to get you feedback or merge it if it's ready to go. 

It's ready to go

rylev:
@andoriyu reading over the previous reviews - it seems that there was consensus that removing the section on ctags is the right move since it's not really needed any more due to improvements in rust-analyzer support. What do you think? Should the section be removed? Can you do that change?

jyn514:
Hi @andoriyu, are you planning to follow up on this work?

andoriyu:
> 
> 
> Hi @andoriyu, are you planning to follow up on this work?

Yes, my bad, I thought i've already done it.Give me like a week to find time.

andoriyu:
@rylev took awhile, but it's done.

jyn514 (approved):
Thanks!

PR #1117, opened 2021-04-28
Update build instructions for rustdoc

dns2utf8:
Hi all

I updated the build instructions with all the things I learned while building the docs.
Thank you @GuillaumeGomez for helping me!

Cheers,
Stefan

### DIFF
--- a/src/rustdoc.md
+++ b/src/rustdoc.md
@@ -34,7 +34,7 @@ does is call the `main()` that's in this crate's `lib.rs`, though.)
 
 [bin]: https://github.com/rust-lang/rust/tree/master/src/tools/rustdoc
 
-## Fast builds
+## Fast builds (for tier 1 targets)
 
 Preconfigure your `/config.toml`:
 
@@ -43,12 +43,7 @@ Preconfigure your `/config.toml`:
 # Global Settings
 # =============================================================================
 
-# Use different pre-set defaults than the global defaults.
-#
-# See `src/bootstrap/defaults` for more information.
-# Note that this has no default value (x.py uses the defaults in `config.toml.example`).
 #profile = <none>
-#
 # make sure that you do not have a profile set
 
 # =============================================================================
@@ -56,19 +51,6 @@ Preconfigure your `/config.toml`:
 # =============================================================================
 [llvm]
 
-# Whether to use Rust CI built LLVM instead of locally building it.
-#
-# Unless you're developing for a target where Rust CI doesn't build a compiler
-# toolchain or changing LLVM locally, you probably want to set this to true.
-#
-# This is false by default so that distributions don't unexpectedly download
-# LLVM from the internet.
-#
-# All tier 1 targets are currently supported; set this to `"if-supported"` if
-# you are not sure whether you're on a tier 1 target.
-#
-# We also currently only support this when building LLVM for the build triple.
-#
 # Note that many of the LLVM options are not currently supported for
 # downloading. Currently only the "assertions" option can be toggled.
 download-ci-llvm = true
@@ -79,8 +61,6 @@ Now build `rustdoc` from stage 2 and build the documentation of the standard lib
 ```bash
 x.py doc std --stage 2 --open
 ```
-
-**Note:** `--open` fixes build errors with `libtest`.
 
 ## Cheat sheet
 
--- a/src/rustdoc.md
+++ b/src/rustdoc.md
@@ -74,7 +74,7 @@ Preconfigure your `/config.toml`:
 download-ci-llvm = true
 ```
 
-Now build ´rustdoc´ from stage 2 and build the documentation of the standard library:
+Now build `rustdoc` from stage 2 and build the documentation of the standard library:
 
 ```bash
 x.py doc std --stage 2 --open
--- a/src/rustdoc.md
+++ b/src/rustdoc.md
@@ -34,6 +34,54 @@ does is call the `main()` that's in this crate's `lib.rs`, though.)
 
 [bin]: https://github.com/rust-lang/rust/tree/master/src/tools/rustdoc
 
+## Fast builds
+
+Preconfigure your `/config.toml`:
+
+```toml
+# =============================================================================
+# Global Settings
+# =============================================================================
+
+# Use different pre-set defaults than the global defaults.
+#
+# See `src/bootstrap/defaults` for more information.
+# Note that this has no default value (x.py uses the defaults in `config.toml.example`).
+#profile = <none>
+#
+# make sure that you do not have a profile set
+
+# =============================================================================
+# Tweaking how LLVM is compiled
+# =============================================================================
+[llvm]
+
+# Whether to use Rust CI built LLVM instead of locally building it.
+#
+# Unless you're developing for a target where Rust CI doesn't build a compiler
+# toolchain or changing LLVM locally, you probably want to set this to true.
+#
+# This is false by default so that distributions don't unexpectedly download
+# LLVM from the internet.
+#
+# All tier 1 targets are currently supported; set this to `"if-supported"` if
+# you are not sure whether you're on a tier 1 target.
+#
+# We also currently only support this when building LLVM for the build triple.
+#
+# Note that many of the LLVM options are not currently supported for
+# downloading. Currently only the "assertions" option can be toggled.
+download-ci-llvm = true
+```
+
+Now build ´rustdoc´ from stage 2 and build the documentation of the standard library:
+
+```bash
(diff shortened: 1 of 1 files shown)


src/rustdoc.md
```diff
@@ -34,6 +34,54 @@ does is call the `main()` that's in this crate's `lib.rs`, though.)
...
+#
+# Note that many of the LLVM options are not currently supported for
+# downloading. Currently only the "assertions" option can be toggled.
+download-ci-llvm = true
+```
+
+Now build ´rustdoc´ from stage 2 and build the documentation of the standard library:
+
+```bash
+x.py doc std --stage 2 --open
```
GuillaumeGomez, line 80: Normally, `stage 1` is supposed to be working. Maybe open an issue on the rust repository about it?

src/rustdoc.md
```diff
@@ -34,6 +34,54 @@ does is call the `main()` that's in this crate's `lib.rs`, though.)
...
+```
+
+**Note:** `--open` fixes build errors with `libtest`.
```
GuillaumeGomez, line 83: Normally no, it's not supposed to be linked in any way. I think it's `stage 2` here which fixed it.

src/rustdoc.md
```diff
@@ -34,6 +34,54 @@ does is call the `main()` that's in this crate's `lib.rs`, though.)
...
+# you are not sure whether you're on a tier 1 target.
+#
+# We also currently only support this when building LLVM for the build triple.
+#
+# Note that many of the LLVM options are not currently supported for
+# downloading. Currently only the "assertions" option can be toggled.
+download-ci-llvm = true
+```
+
+Now build ´rustdoc´ from stage 2 and build the documentation of the standard library:
```
JohnTitor, line 77:
```suggestion
Now build `rustdoc` from stage 2 and build the documentation of the standard library:
```

src/rustdoc.md
```diff
@@ -34,6 +34,54 @@ does is call the `main()` that's in this crate's `lib.rs`, though.)
...
+#
+# See `src/bootstrap/defaults` for more information.
+# Note that this has no default value (x.py uses the defaults in `config.toml.example`).
+#profile = <none>
+#
+# make sure that you do not have a profile set
+
+# =============================================================================
+# Tweaking how LLVM is compiled
+# =============================================================================
```
JohnTitor, line 56:
```suggestion
```

The comment here doesn't seem to be related to the topic, let's shrink it.

src/rustdoc.md
dns2utf8: I am rerunning this to verify, my gut feeling agrees with you

src/rustdoc.md
```diff
@@ -34,6 +34,54 @@ does is call the `main()` that's in this crate's `lib.rs`, though.)
 
 [bin]: https://github.com/rust-lang/rust/tree/master/src/tools/rustdoc
 
+## Fast builds
+
+Preconfigure your `/config.toml`:
+
+```toml
```
jyn514, line 41: This is all way overkill. I would suggest `x.py setup tools` and nothing else.

src/rustdoc.md
jyn514: I would rather just recommend  --stage 1.

src/rustdoc.md
jyn514: What do you mean by build errors? Documenting with stage 1 works fine for me locally.

jyn514 (changes requested):

GuillaumeGomez:
@jyn514 In his case, it's triggering errors about "libtest not found". Which is why I recommended him to try with `--stage 2` and to open an issue.

jyn514:
@GuillaumeGomez that's fine, but we should help @dns2utf8 get --stage 1 working instead of recommending a command that takes twice as long.

src/rustdoc.md
dns2utf8: I did rebuild it with the current `master` something fixed the build process there.
Now `stage 1` works too, but it is ~38min on an 8 core CPU.

dns2utf8:
In my opinion building from `stage 2` and the CI artefacts should be possible in ~5min or less, but my current test gave me ~23min. So not that great either

dns2utf8:
The build error I was getting:
```
rust/build/x86_64-unknown-linux-gnu/stage1/bin/rustdoc: error while loading shared libraries: http://libtest-767351f2ae2fb418.so: cannot open shared object file: No such file or directory
error: could not document `core`
```
image

I fixed this by building the docs with the master branch at 855c2d130fb70da1643cf8f696c7aad7537aef34 and then switching to [my branch that has mostly JavaScript/CSS differences](https://github.com/jsha/rust/pull/1)

src/rustdoc.md
dns2utf8: I replied in the bottom, sorry for the confusion

src/rustdoc.md
dns2utf8: I had "libraries" configured  before and the setup would not touch the file anymore, so I decided to do it by hand. Is this intended behaviour?

dns2utf8:
Maybe I misunderstood the build process, @jyn514
I was under the impression that building the compiler goes from stage 0 -> 1 -> 2 -> completely finished. Where as it takes the longest to build from stage 0 and the least from 2. Or is it the other way round?

dns2utf8:
I usually prefere a CSS over a JS solution, but the scrolling behaves very strange on my machine. 
On Firefox 88 (Arch Linux) the scroll margin lets the selection out, but it works on chromium -> is it a firefox bug or does one of my addons interfere? 

https://user-images.githubusercontent.com/739070/116457474-f9ffaf00-a852-11eb-83a0-9c8f53fa2832.mp4

src/rustdoc.md
jyn514: Ok, then I would recommend `profile = "tools"` instead - that applies the latest changes automatically and also means you don't need to copy/paste 20 lines of config.

src/rustdoc.md
jyn514: Stage 1 should never be slower than stage 2. I saw below that stage2 took you about 23 minutes - did you maybe have rustdoc already compiled or something?

src/rustdoc.md
dns2utf8: I reduced the amount of text, but would leave some comments

src/rustdoc.md
dns2utf8: I enabled `[llvm] download-ci-llvm = true` in `/config.toml` so I did not compile LLVM if I understand correctly

jyn514: Ok, cool. Let's change this to --stage 1 then (or just omit the stage, the default is correct).

src/rustdoc.md
dns2utf8: Are the implications for the "tools" profile documented somewhere? I would like to link to it

jyn514: All the profiles are documented in bootstrap itself, or otherwise it would get out of date: https://github.com/rust-lang/rust/blob/master/src/bootstrap/defaults/config.tools.toml

jyn514:
> In my opinion building from `stage 2` and the CI artefacts should be possible in ~5min or less, but my current test gave me ~23min. So not that great either

It should be less than ~3 minutes if you set download-rustc = true. profile = tools does that by default.

jyn514:
Hi @dns2utf8, are you planning to work on this?

src/rustdoc.md
```diff
@@ -42,7 +70,7 @@ does is call the `main()` that's in this crate's `lib.rs`, though.)
   * If you've used `rustup toolchain link local /path/to/build/$TARGET/stage1`
     previously, then after the previous build command, `cargo +local doc` will
     Just Work.
-* Use `./x.py doc --stage 1 library/std` to use this rustdoc to generate the
+* Use `./x.py doc --stage 1 std` to use this rustdoc to generate the
```
camelid, line 73: This change seems incorrect to me.

GuillaumeGomez: No, it's valid. I added the support for shorter lib path a while ago. :)

JohnTitor:
Triage: I'm going to close this as inactive, but feel free to open another PR if anyone is still interested in this change (I think https://github.com/rust-lang/rustc-dev-guide/pull/1389/files#diff-d9ae61aa6e37d7cc365c1ae1c66c801e2741026fbc752d4628dc34f3c5ce9290 updated some instructions but the "Fast builds (for tier 1 targets)" section might be still useful). Thanks for contributing!

PR #1125, opened 2021-05-09
Remove `--stage 1` argument from cheat sheet

Stupremee:
Requested by @jyn514 on [zulip](https://rust-lang.zulipchat.com/#narrow/stream/266220-rustdoc/topic/compiling.20rustdoc.20lib.2Fstd/near/238041043)

### DIFF
--- a/src/rustdoc-internals.md
+++ b/src/rustdoc-internals.md
@@ -245,7 +245,7 @@ server. To test these features locally, you can run a local HTTP server, like
 this:
 
 ```bash
-$ ./x.py doc library/std --stage 1
+$ ./x.py doc library/std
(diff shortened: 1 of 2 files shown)


jyn514:
For context, `--stage 1` is currently broken when `download-rustc` is set, which is set by default with `profile = "tools"` (the one x.py setup recommends for rustdoc). `doc library/std` *does* do the wrong thing when you don't have a config.toml at all (it documents with stage 0), but at least it doesn't give an error which hopefully should make people less confused.

jyn514:
@Stupremee can you also update https://rustc-dev-guide.rust-lang.org/rustdoc-internals.html#testing-locally ?

LeSeulArtichaut (approved):

PR #1126, opened 2021-05-10
Unified CPU Requirements

hi-rustin:
See: https://github.com/rust-lang/rustc-dev-guide/blob/master/src/getting-started.md#system-requirements


JohnTitor (approved):
Good catch, thanks! I'm not sure if "2 cores" is still valid nowadays but I'm going to sync with it for now.

PR #1127, opened 2021-05-12
not all tools require waiting for a nightly release before they can be fixed

RalfJung:
For example, Miri installs a recent toolchain directly from the master branch using `rustup-toolchain-install-master`.

This fixes the docs so they are no longer incorrect for Miri. I do not know what other tools do.

### DIFF
--- a/src/contributing.md
+++ b/src/contributing.md
@@ -272,12 +272,16 @@ tests.
 That would mean that, in the default state, you couldn't update the compiler without first
 fixing miri, rls and the other tools that the compiler builds.
 
-Luckily, a feature was [added to Rust's build](https://github.com/rust-lang/rust/issues/45861)
-to make all of this easy to handle. The idea is that we allow these tools to be "broken",
-so that the rust-lang/rust build passes without trying to build them, then land the change
-in the compiler, wait for a nightly, and go update the tools that you broke. Once you're done
-and the tools are working again, you go back in the compiler and update the tools
-so they can be distributed again.
+Luckily, a feature was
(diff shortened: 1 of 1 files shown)


jyn514 (commented):
This change looks good and correct, but IMO this whole section makes very little sense because it never mentions the word "submodule"; none of these concerns apply to rustdoc or clippy. Do you mind mentioning it near the start of the section somewhere?

>  I do not know what other tools do.

RLS uses rustfmt as a library, which I believe in turn uses nightly releases, making the text here correct. @xanewok may know more. I'm not aware of other tool submodules at this time.

RalfJung:
Does something like this match what you have in mind?

jyn514 (approved):

jyn514:
This is great, thanks! If you can fix the merge conflict I think this is ready to go :)

RalfJung:
FWIW, merge conflicts would be much less likely if the source was formatted with one-sentence-per-line... line-wrapping with a fixed width tends to amplify conflicts by a lot, IMO making it a bad choice for text formatting.

jyn514:
> FWIW, merge conflicts would be much less likely if the source was formatted with one-sentence-per-line... line-wrapping with a fixed width tends to amplify conflicts by a lot, IMO making it a bad choice for text formatting.

Can you open an issue for that? I'm not opposed, but changing it would be an enormous diff.

RalfJung:
I resolved the conflict. I hope I did not erase any prior changes, but it seems to have just added Miri so my changes subsume the changes on master.

jyn514:
Yes, the prior changes were just removing the reference to rustfmt.

PR #1128, opened 2021-05-13
Document -Zunpretty=thir-tree

Smittyvb:
This flag is useful for debugging the THIR.

### DIFF
--- a/src/thir.md
+++ b/src/thir.md
@@ -9,6 +9,10 @@ The THIR ("Typed High-Level Intermediate Representation"), previously called HAI
 [it may also soon be used for unsafety checking][thir-unsafeck] as a replacement
 for the current MIR unsafety checker.
 
+
+You can get a human-readable repersentation of the THIR by passing the `-Zunpretty=thir-tree` flag
+to `rustc`.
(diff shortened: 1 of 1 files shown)


src/thir.md
```diff
@@ -9,6 +9,10 @@ The THIR ("Typed High-Level Intermediate Representation"), previously called HAI
 [it may also soon be used for unsafety checking][thir-unsafeck] as a replacement
 for the current MIR unsafety checker.
 
+
+You can get a human-readable repersentation of the THIR by passing the `-Zunpretty=thir-tree` flag
```
LeSeulArtichaut, line 13: I wouldn't necessarily say it's "human readable" though. Maybe
```suggestion
You can get a debug representation of the THIR by passing the `-Zunpretty=thir-tree` flag
```

LeSeulArtichaut (commented):
Good idea to document this, though I'm probably going to break it with #83842.

I think I would prefer if this was further down in the chapter, maybe at the end?

LeSeulArtichaut (approved):
Thanks!

PR #1129, opened 2021-05-18
update rustfmt references to reflect change from submod to subtree

calebcartwright:
rustfmt was recently changed from a submodule to a subtree (https://github.com/rust-lang/rust/pull/82208) so this accordingly updates a few rustfmt submodule references

cc @RalfJung and @Xanewok, as I largely did a lazy replacement of rustfmt wording in submodule-related sections with miri or rls since those seem to be the most used. Please let me know if any of those changes are off base though for your respective tools

### DIFF
--- a/src/contributing.md
+++ b/src/contributing.md
@@ -269,7 +269,7 @@ the compiler. At the same time, you can't get your changes merged into the compi
 because the rust-lang/rust build won't pass until those tools build and pass their
 tests.
 
-That means that, in the default state, you can't update the compiler without first
+That would mean that, in the default state, you couldn't update the compiler without first
 fixing miri, rls and the other tools that the compiler builds.
 
 Luckily, a feature was [added to Rust's build](https://github.com/rust-lang/rust/issues/45861)
--- a/src/contributing.md
+++ b/src/contributing.md
@@ -169,7 +169,8 @@ the issue in question.
 As a developer to this repository, you don't have to treat the following external projects
 differently from other crates that are directly in this repo:
 
-* Clippy
+* [Clippy](https://github.com/rust-lang/rust-clippy)
+* [rustfmt](https://github.com/rust-lang/rustfmt)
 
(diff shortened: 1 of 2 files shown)


src/contributing.md
```diff
@@ -270,7 +270,7 @@ because the rust-lang/rust build won't pass until those tools build and pass the
 tests.
 
 That means that, in the default state, you can't update the compiler without first
```
RalfJung, line 272:
```suggestion
That would mean that, in the default state, you couldn't update the compiler without first
```
I still find this sentence very confusing, but now at least it is clearly non-factual.
(I also realize this is pre-existing.)

src/git.md
```diff
@@ -347,9 +346,9 @@ you might want to get used to the main concepts of Git before reading this secti
 
 The `rust-lang/rust` repository uses [Git submodules] as a way to use other
 Rust projects from within the `rust` repo. Examples include Rust's fork of
-`llvm-project` and many devtools such as `cargo`, `rust-analyzer` and `rustfmt`.
+`llvm-project` and many devtools such as `cargo`, `rust-analyzer` and `RLS`.
```
RalfJung, line 349:
```suggestion
`llvm-project` and many devtools such as `cargo`, `rust-analyzer` and `rls`.
```
I assume? It's lower-case most everywhere else.

RalfJung:
LGTM. :)

LeSeulArtichaut (approved):
Thanks! I'll merge with Ralf's nit fixed

src/contributing.md
calebcartwright: Yes! I'm so glad I'm not the only one that was confused by this. It gave me a brief scare initially because I didn't realize until I'd gotten a few paragraphs in that it was actually describing a past or hypothetical state, but using present-tense prose. Applied in ed0c82bc91a5052d92a07580778759b00c537cc6

src/git.md
calebcartwright: done!

PR #1131, opened 2021-05-26
Describe the difference of rustc_lint vs rustc_lint_defs.

ehuss:
I was a little uncertain about the relationship between these crates.


jyn514:
I think @Aaron1011 introduced rustc_lint_defs, he may have an opinion.

Aaron1011 (approved):

PR #1139, opened 2021-06-11
We use HIR to do type inference, trait solving and type checking

spastorino:


src/overview.md
```diff
@@ -68,10 +68,12 @@ we'll talk about that later.
...
   AST.  This involves a lot of desugaring of things like loops and `async fn`.
-- We use the HIR to do [type inference]. This is the process of automatic
-  detection of the type of an expression.
-- **TODO: Maybe some other things are done here? I think initial type checking
-  happens here? And trait solving?**
+- We use the HIR to do [type inference] (the process of automatic
+  detection of the type of an expression), [trait solving] (the process
+  of pairing up an impl with each reference to a trait) and [type
+  checking] (the process of converting the types found in the HIR
+  (hir::Ty), which represent the syntactic things that the user wrote,
```
JohnTitor, line 75:
```suggestion
  (`hir::Ty`), which represent the syntactic things that the user wrote,
```

src/overview.md
```diff
@@ -68,10 +68,12 @@ we'll talk about that later.
...
+  into the internal representation used by the compiler (Ty<'tcx>)).
```
JohnTitor, line 76:
```suggestion
  into the internal representation used by the compiler (`Ty<'tcx>`)).
```

JohnTitor (approved):
LGTM with some nits.

src/overview.md
```diff
@@ -68,10 +68,12 @@ we'll talk about that later.
 - We then take the AST and [convert it to High-Level Intermediate
   Representation (HIR)][hir]. This is a compiler-friendly representation of the
   AST.  This involves a lot of desugaring of things like loops and `async fn`.
-- We use the HIR to do [type inference]. This is the process of automatic
-  detection of the type of an expression.
-- **TODO: Maybe some other things are done here? I think initial type checking
-  happens here? And trait solving?**
+- We use the HIR to do [type inference] (the process of automatic
+  detection of the type of an expression), [trait solving] (the process
+  of pairing up an impl with each reference to a trait) and [type
```
camelid, line 73: nit: We usually use the Oxford comma:

```suggestion
  of pairing up an impl with each reference to a trait), and [type
```

src/overview.md
```diff
@@ -68,10 +68,12 @@ we'll talk about that later.
...
+  checking] (the process of converting the types found in the HIR
```
flodiebold, line 74: that doesn't seem right? the sentence in the parentheses describes type collection, not type checking.

spastorino: Ohh you're right, I've done a quick copy and paste.

What about something like (see below) ...

src/overview.md
```diff
@@ -68,10 +68,12 @@ we'll talk about that later.
...
-- We use the HIR to do [type inference]. This is the process of automatic
-  detection of the type of an expression.
-- **TODO: Maybe some other things are done here? I think initial type checking
-  happens here? And trait solving?**
+- We use the HIR to do [type inference] (the process of automatic
+  detection of the type of an expression), [trait solving] (the process
+  of pairing up an impl with each reference to a trait), and [type
+  checking] (the process of converting the types found in the HIR
+  (`hir::Ty`), which represent the syntactic things that the user wrote,
+  into the internal representation used by the compiler (`Ty<'tcx>`)).
```
spastorino, line 76:
```suggestion
  into the internal representation used by the compiler (`Ty<'tcx>`), and use that information to
  verify the type safety of a program based on analysis of a program's text).
```

Or something like that, feel free to suggest something better :)

flodiebold: Seems good to me, except the `use` should be `using` :slightly_smiling_face:

spastorino:
This should be now ready.

LeSeulArtichaut (approved):
Looks good to me as well

PR #1141, opened 2021-06-11
Document how to run unit tests

### LINKED ISSUE #1138: Document how to run unit tests
If you want to run `thin_vec` unit tests you may try to do something like ...

`./x.py test compiler/rustc_data_structures/src/thin_vec/tests.rs`

But that's not possible.

The right way is ...

`./x.py test compiler/rustc_data_structures/ --test-args thin_vec`

We should add this information to https://rustc-dev-guide.rust-lang.org/tests/running.html and possibly discuss a bit about why unit tests are not supported using a single file. Maybe we should also support that through x.py.

JohnTitor:
Fixes #1138

### DIFF
--- a/src/tests/running.md
+++ b/src/tests/running.md
@@ -92,6 +92,20 @@ tests for components you did not change at all.
 build; therefore, while the tests **usually** work fine with stage 1,
 there are some limitations.
 
+## Run unit tests on the compiler/library
+
+You may want to run unit tests on a specific file with following:
+
+```bash
+./x.py test compiler/rustc_data_structures/src/thin_vec/tests.rs
+```
+
+But unfortunately, it's impossible. You should invoke following instead:
+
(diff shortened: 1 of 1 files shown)


src/tests/running.md
```diff
@@ -92,6 +92,20 @@ tests for components you did not change at all.
...
 
+## Run unit tests on the compiler/library
+
+You may want to run unit tests on a specific file with following:
+
+```bash
+./x.py test compiler/rustc_data_structures/src/thin_vec/tests.rs
+```
+
+But unfortunately, it's impossible. You should invoke following instead:
```
JohnTitor, line 103: Searched the issue tracker a bit but I couldn't find any issues related to this, maybe should we open a new issue and add a link here?

spastorino (approved):

PR #1143, opened 2021-06-12
Change the feature used as an example of stabilizing lib features

lf-:
The previous one was not following the process we're documenting: it was a PR, which an FCP was started on after the fact. This doesn't match what the stated process is: we should instead link to an issue where t-libs started an fcp, then a pr was written later.

### DIFF
--- a/src/stability.md
+++ b/src/stability.md
@@ -85,7 +85,10 @@ To stabilize a feature, follow these steps:
    - Add the appropriate labels: `@rustbot modify labels: +T-libs`.
    - Link to the tracking issue and say "Closes #XXXXX".
 
-You can see an example of stabilizing a feature at [#81656](https://github.com/rust-lang/rust/issues/81656).
+You can see an example of stabilizing a feature with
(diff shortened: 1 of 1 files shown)


src/stability.md
```diff
@@ -85,7 +85,7 @@ To stabilize a feature, follow these steps:
    - Add the appropriate labels: `@rustbot modify labels: +T-libs`.
    - Link to the tracking issue and say "Closes #XXXXX".
 
-You can see an example of stabilizing a feature at [#75132](https://github.com/rust-lang/rust/pull/75132).
+You can see an example of stabilizing a feature at [#81656](https://github.com/rust-lang/rust/issues/81656).
```
JohnTitor, line 88: The section talks about how we create a PR so we should link to the PR, not the issue.

lf-: I've added a link to both and clarified the wording so it points to all steps of the process.

JohnTitor (approved):
Alright LGTM, thanks!

PR #1145, opened 2021-06-21
Update "Inference variables" section

### LINKED ISSUE #1116: Check if type variable section needs to be updated
This section on type variables was written in 2018 (or earlier), and we should check to make sure it is up to date. That would likely involve asking the compiler team on Zulip (in `#t-compiler`).

**`src/type-inference.md`**

> All told, the inference context stores four kinds of inference variables
> (as of  January 2018):
>
> - Type variables, which come in three varieties:
>   - General type variables (the most common). These can be unified with any
>     type.
>   - Integral type variables, which can only be unified with an integral type,
>     and arise from an integer literal expression like `22`.
>   - Float type variables, which can only be unified with a float type, and
>     arise from a float literal expression like `22.0`.
> - Region variables, which represent lifetimes, and arise all over the place.

cc #1121

JohnTitor:
Fixes #1116
Const variable was introduced in rust-lang/rust#59008.

### DIFF
--- a/src/type-inference.md
+++ b/src/type-inference.md
@@ -71,8 +71,8 @@ inference works, or perhaps this blog post on
 
 [Unification in the Chalk project]: http://smallcultfollowing.com/babysteps/blog/2017/03/25/unification-in-chalk-part-1/
 
-All told, the inference context stores four kinds of inference variables
-(as of <!-- date: 2018-01 --> January 2018):
+All told, the inference context stores five kinds of inference variables
+(as of <!-- date: 2021-06 --> June 2021):
 
 - Type variables, which come in three varieties:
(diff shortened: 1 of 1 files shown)


jackh726 (approved):

LeSeulArtichaut (approved):
:+1:

Btw, maybe we should point out that these are stored in [`InferCtxtInner`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_infer/infer/struct.InferCtxtInner.html), but that's not a problem for this PR.

tshepang:
Is this still current? How do I find the current status (closest I could find is InferCtxtInner, but it does not make things clear)?

PR #1147, opened 2021-06-23
Correct the wrong serial number

hi-rustin:
the number of the parts should be seven if we count "Appendices" as one.

### DIFF
--- a/src/about-this-guide.md
+++ b/src/about-this-guide.md
@@ -9,7 +9,8 @@ There are six parts to this guide:
 1. [Building and debugging `rustc`][p1]: Contains information that should be
    useful no matter how you are contributing, about building, debugging,
    profiling, etc.
-2. [Contributing to `rustc`][p1-5]: Contains information that should be useful
+   
+   [Contributing to `rustc`][p1-5]: Contains information that should be useful
    no matter how you are contributing, about procedures for contribution, using git
    and Github, stabilizing features, etc.
 2. [High-Level Compiler Architecture][p2]: Discusses the high-level
--- a/src/about-this-guide.md
+++ b/src/about-this-guide.md
@@ -9,26 +9,26 @@ There are seven parts to this guide:
 1. [Building and debugging `rustc`][p1]: Contains information that should be
    useful no matter how you are contributing, about building, debugging,
    profiling, etc.
-2. [Contributing to `rustc`][contributing]: Contains information that should be useful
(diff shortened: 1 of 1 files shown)


src/about-this-guide.md
```diff
@@ -9,7 +9,8 @@ There are six parts to this guide:
 1. [Building and debugging `rustc`][p1]: Contains information that should be
    useful no matter how you are contributing, about building, debugging,
    profiling, etc.
-2. [Contributing to `rustc`][p1-5]: Contains information that should be useful
+   
+   [Contributing to `rustc`][p1-5]: Contains information that should be useful
```
JohnTitor, line 13: Hmm, the number of the parts should be seven if we count "Appendices" as one.

JohnTitor (changes requested):

JohnTitor:
Also, could you update the [p1-p5] link to ./contributing.html?

src/about-this-guide.md
hi-rustin: Fixed.

src/about-this-guide.md
```diff
@@ -4,26 +4,26 @@ This guide is meant to help document how rustc – the Rust compiler –
...
 development.
 
-There are six parts to this guide:
+There are seven parts to this guide:
 
 1. [Building and debugging `rustc`][p1]: Contains information that should be
    useful no matter how you are contributing, about building, debugging,
    profiling, etc.
-2. [Contributing to `rustc`][p1-5]: Contains information that should be useful
+2. [Contributing to `rustc`][contributing]: Contains information that should be useful
```
JohnTitor, line 12: nit: I'd like to keep the consistency of the links here. I mean, the chapter name vs. the chapter number. I'm fine either way.

hi-rustin: Actually I'm not quite sure which chapter p1-5 is referring to? I'm not quite sure what  chapter number to use in this place.

JohnTitor: I guess it was considered as a subchapter? It explains why it said "six chapters". But now it has its own chapter so it should make sense to call it p2.

hi-rustin: Fixed.

JohnTitor (approved):
Thanks! Still, part-n-intro.md is confusing but it's out of scope on this PR and we can discuss on another issue.

PR #1152, opened 2021-07-01
Links from rustc-dev-guide to std-dev-guide

joshtriplett:


LeSeulArtichaut (approved):
:+1:

LeSeulArtichaut:
Side note: if the library team has a guide for the standard library, maybe we'll want to remove some chapters like https://rustc-dev-guide.rust-lang.org/stability.html from the rustc-dev-guide.

rylev:
I added an issue to consider removing those chapters https://github.com/rust-lang/rustc-dev-guide/issues/1156

PR #1173, opened 2021-07-20
Link directly to stabilization report comments

### LINKED ISSUE #1172: Link to actual stabilization reports instead of just PRs that contain those reports
The section on [stabilization reports](https://rustc-dev-guide.rust-lang.org/stabilization_guide.html#write-a-stabilization-report), links to two PRs that supposedly contain those reports. The second of these PRs contains hundreds of comments and it takes quite a while to expand those comments and find the actual report. We should link directly to them.

camelid:
Fixes #1172.

Previously, readers had to scroll through tons of comments to find the
report.

### DIFF
--- a/src/stabilization_guide.md
+++ b/src/stabilization_guide.md
@@ -58,10 +58,16 @@ The report should contain:
   and describe the feature's behavior on encountering edge cases.
 - Links to the documentations (the PRs we have made in the
   previous steps).
-- Any other relevant information(Examples of such reports can
-  be found in rust-lang/rust#44494 and rust-lang/rust#28237).
+- Any other relevant information.
 - The resolutions of any unresolved questions if the stabilization
   is for an RFC.
+
+Examples of stabilization reports can be found in
+[rust-lang/rust#44494][report1] and [rust-lang/rust#28237][report2] (these links
+will bring you directly to the comment containing the stabilization report).
+
+[report1]: https://github.com/rust-lang/rust/issues/44494#issuecomment-360191474
+[report2]: https://github.com/rust-lang/rust/issues/28237#issuecomment-363374130
 
 ## FCP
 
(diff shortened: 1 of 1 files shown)


src/stabilization_guide.md
```diff
@@ -58,18 +58,24 @@ The report should contain:
...
   previous steps).
-- Any other relevant information(Examples of such reports can
-  be found in rust-lang/rust#44494 and rust-lang/rust#28237).
+- Any other relevant information.
 - The resolutions of any unresolved questions if the stabilization
   is for an RFC.
 
+Examples of stabilization reports can be found in
+[rust-lang/rust#44494][report1] and [rust-lang/rust#28237][report2] (these links
+will bring you directly to the comment containing the stabilization report).
```
jyn514, line 67: I appreciate you doing the work :) but I don't think it needs to be called out specifically in the docs.
```suggestion
[rust-lang/rust#44494][report1] and [rust-lang/rust#28237][report2].
```

jyn514 (changes requested):

src/stabilization_guide.md
camelid: The reason I included that phrase was so people wouldn't be daunted by seeing an issue number and thinking they'd have to scroll through a ton of comments. I also thought of writing something like

> Examples of stabilization reports can be found [here](#) and [here](#)

but I thought it might be better to keep the issue numbers. I'm okay with removing that phrase if you don't think it'll be daunting to people though :)

Let me know if you'd still like me to remove it.

jyn514: I think removing it is fine - they wouldn't have reason to think ahead of time that the comment chain is super long. But I don't feel super strongly about it.

jyn514 (approved):

PR #1176, opened 2021-07-28
git.md: Fix No-Merge Policy link

xphoniex:
This is supposed to be a link but since it's using brackets for both URL and name, it's obviously not working.


jyn514 (approved):
Thanks!

PR #1180, opened 2021-08-05
allow to quickly edit a page directly on github

tshepang:


jyn514:
Hmm, how does this work? I built the book locally but I don't see an "edit" button. The github link in the top right just goes to the repository root, which is already enabled before this PR.
image

jyn514:
Looks like this was added in https://github.com/rust-lang/mdBook/pull/1506, and the first release since then was 0.4.8: https://lib.rs/crates/mdbook/versions. Can you also update `MDBOOK_VERSION` in `.github/workflows/ci.yml` so this will work properly on github pages?

jyn514 (approved):

PR #1181, opened 2021-08-05
Fixed team responsible for stabilization

YohDeadfall:
As @m-ou-se pointed out in https://github.com/rust-lang/rust/pull/86879#issuecomment-892896170 `T-libs-api` should be used instead of `T-libs` for stabilization purposes.


LeSeulArtichaut (approved):
:+1:

PR #1182, opened 2021-08-05
Add description of -opt-bisect-limit LLVM option

bcantrill:
This adds a description of the `-opt-bisect-limit` LLVM option to the (very helpful!) doc on debugging LLVM. (We found this option to be essential in honing in on an LLVM optimization issue -- but it's not easy to find if you don't already know about it!)

### DIFF
--- a/src/backend/debugging.md
+++ b/src/backend/debugging.md
@@ -56,6 +56,17 @@ different stages during compilation, which is sometimes useful. One just needs
 to convert the bitcode files to `.ll` files using `llvm-dis` which should be in
 the target local compilation of rustc.
 
+If you are seeing incorrect behavior due to an optimization pass, a very handy
+LLVM option is `-opt-bisect-limit`, which takes an integer denoting the index
+value of the highest pass to run.  Index values for taken passes are stable
+from run to run; by coupling this with software that automates bisecting the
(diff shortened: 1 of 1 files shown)


src/backend/debugging.md
```diff
@@ -56,6 +56,17 @@ different stages during compilation, which is sometimes useful. One just needs
 to convert the bitcode files to `.ll` files using `llvm-dis` which should be in
 the target local compilation of rustc.
 
+If you are seeing incorrect behavior due to an optimization pass, a very handy
+LLVM option is `-opt-bisect-limit`, which takes an integer denoting the index
+value of the highest pass to run.  Index values for taken passes are stable
+from run to run; by coupling this with software that automates bisecting the
```
rylev, line 62: Should we mention software that can be used for this?

rylev (approved):
Looks great! Thank you ❤️

spastorino (approved):

src/backend/debugging.md
bcantrill: Sorry -- I missed this comment!  Yeah, the "software that automates bisecting" is really a euphemism for: "the gnarly script that you, dear reader, are about to write." ;)  Happy to spell that out, if with slightly less punchy language..

PR #1183, opened 2021-08-06
Expand THIR section with more details

Smittyvb:
This adds some more details about how the THIR works. It also adds as an example what some THIR looks like, along with some annotations I added.

Contributes to https://github.com/rust-lang/project-thir-unsafeck/issues/6. (cc @LeSeulArtichaut)

### DIFF
--- a/src/thir.md
+++ b/src/thir.md
@@ -4,10 +4,11 @@
 
 The THIR ("Typed High-Level Intermediate Representation"), previously called HAIR for
 "High-Level Abstract IR", is another IR used by rustc that is generated after
-[type checking]. It is (as of <!-- date: 2021-03 --> March 2021) only used for
-[MIR construction] and [exhaustiveness checking], but
-[it may also soon be used for unsafety checking][thir-unsafeck] as a replacement
-for the current MIR unsafety checker.
+[type checking]. It is (as of <!-- date: 2021-08 --> August 2021) only used for
+[MIR construction] and [exhaustiveness checking]. There is also
+[an experimental unsafety checker][thir-unsafeck] that operates on the THIR as a replacement for
+the current MIR unsafety checker, and can be used instead of the MIR unsafety checker by passing
+the `-Z thir-unsafeck` flag to `rustc`.
 
 [type checking]: ./type-checking.md
 [MIR construction]: ./mir/construction.md
@@ -19,8 +20,9 @@ the types have been filled in, which is possible after type checking has complet
 But it has some other interesting features that distinguish it from the HIR:
 
 - Like the MIR, the THIR only represents bodies, i.e. "executable code"; this includes
-  function bodies, but also `const` initializers, for example. Consequently, the THIR
-  has no representation for items like `struct`s or `trait`s.
+  function bodies, but also `const` initializers, for example. Specifically, all [body owners] have
+  THIR created. Consequently, the THIR has no representation for items like `struct`s or `trait`s.
+  References to those items are represented as `HirId`s in the THIR.
 
 - Each body of THIR is only stored temporarily and is dropped as soon as it's no longer
   needed, as opposed to being stored until the end of the compilation process (which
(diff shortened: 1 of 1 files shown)


src/thir.md
```diff
@@ -19,8 +20,10 @@ the types have been filled in, which is possible after type checking has complet
 But it has some other interesting features that distinguish it from the HIR:
 
 - Like the MIR, the THIR only represents bodies, i.e. "executable code"; this includes
-  function bodies, but also `const` initializers, for example. Consequently, the THIR
-  has no representation for items like `struct`s or `trait`s.
+  function bodies, but also `const` initializers, for example. Specifically, everything with a
+  `DefId` can (and does) have a THIR body created. Consequently, the THIR
+  has no representation for items like `struct`s or `trait`s. References to those items are
+  represented as `HirId`s in the THIR.
```
LeSeulArtichaut, line 26: This isn't actually true, `struct`s and `trait`s do have `DefId`s, and the THIR (like the HIR) references them through a `DefId`. If you want to link to a list of all things that have a body you could use [`BodyOwnerKind`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/enum.BodyOwnerKind.html).

src/thir.md
```diff
@@ -31,6 +34,11 @@ But it has some other interesting features that distinguish it from the HIR:
   are made explicit, and method calls and overloaded operators are converted into
   plain function calls. Destruction scopes are also made explicit.
 
+- Statements, expressions, and match arms are stored separately. Statements in the `stmts` array
+  reference expressions by their index in the `exprs` array. Expressions reference subexpressions by
+  their index in the `exprs` array. Match experssions reference arms by their index in the
+  `arms` array.
```
LeSeulArtichaut, line 40: This should mention the `{Expr,Stmt,Arm}Id` types. Also, I think you only need to provide one example.

src/thir.md
```diff
@@ -40,7 +48,176 @@ which is useful to keep peak memory in check. Having a THIR representation of
 all bodies of a crate in memory at the same time would be very heavy.
 
 You can get a debug representation of the THIR by passing the `-Zunpretty=thir-tree` flag
-to `rustc`.
+to `rustc`. Here is how a function with just the statement `let x = 1 + 2;` gets represented in
+annotated THIR:
```
LeSeulArtichaut, line 52: This wording is kind of confusing, you can understand it as "annotated THIR" being a separate concept. I don't think we need to mention that annotations have been added
```suggestion
THIR:
```

LeSeulArtichaut (commented):
I know there are plans to have more text IRs in the guide. Just in case: @spastorino does this fit with the overall plans for the guide?

src/thir.md
```diff
@@ -4,10 +4,11 @@
 
 The THIR ("Typed High-Level Intermediate Representation"), previously called HAIR for
 "High-Level Abstract IR", is another IR used by rustc that is generated after
-[type checking]. It is (as of <!-- date: 2021-03 --> March 2021) only used for
-[MIR construction] and [exhaustiveness checking], but
-[it may also soon be used for unsafety checking][thir-unsafeck] as a replacement
-for the current MIR unsafety checker.
+[type checking]. It is (as of <!-- date: 2021-08 --> August 2021) only used for
+[MIR construction] and [exhaustiveness checking]. There is also
+[an experimental unsafety checker][thir-unsafeck] that operates on the THIR as a replacement for
```
rylev, line 9: I'm curious why there is an unsafety checker in THIR that aims to replace the one in MIR. Could we add something about that?

LeSeulArtichaut: See rust-lang/compiler-team#402. TL;DR is that unsafety checking is expected to match the syntax, but MIR doesn't match syntax closely which led to issues like
```rs
// deref of raw ptr should be unsafe even though the result is unused,
// but passes with MIR unsafeck
*ptr;
```

We can still handle cases like this in MIR (by using e.g. fake reads), but checking unsafety in THIR should be more natural and require less/no hacks.

src/thir.md
```diff
@@ -19,8 +20,9 @@ the types have been filled in, which is possible after type checking has complet
 But it has some other interesting features that distinguish it from the HIR:
 
 - Like the MIR, the THIR only represents bodies, i.e. "executable code"; this includes
-  function bodies, but also `const` initializers, for example. Consequently, the THIR
-  has no representation for items like `struct`s or `trait`s.
+  function bodies, but also `const` initializers, for example. Specifically, all [body owners] have
+  THIR created. Consequently, the THIR has no representation for items like `struct`s or `trait`s.
+  References to those items are represented as `HirId`s in the THIR.
```
LeSeulArtichaut, line 25: This isn't correct. `HirId`s only exist for the local crate, but we need to be able to reference items from other crates as well. I'm not sure this sentence is really needed though.
```suggestion
```

LeSeulArtichaut (commented):
Just one change and this looks good to me

jyn514 (approved):

PR #1190, opened 2021-08-18
`ty::Unevaluated`: dealing with unused substs

lcnr:
blocked on https://github.com/rust-lang/rust/pull/87280

still not too happy with what I wrote here, pls review :heart:

### DIFF
--- a/src/constants.md
+++ b/src/constants.md
@@ -1,0 +1,60 @@
+# Constants in the type system
+
+Constants used in the type system are represented as [`ty::Const`].
+The variants of their [`ty::ConstKind`] mostly mirror the variants of [`ty::TyKind`]
+with the two *additional* variants being `ConstKind::Value` and `ConstKind::Unevaluated`.
+
+
+## Unevaluated constants
+
+*This section talks about what's happening with `feature(const_generics)` enabled.
+On stable we do not yet supply any generic parameters to anonymous constants,
+avoiding most of the issues mentioned here.*
+
+Unless a constant is either a simple literal, e.g. `[u8; 3]` or `foo::<{ 'c' }>()`,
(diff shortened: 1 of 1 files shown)


jyn514 (commented):
This is sort of like lazy normalization, but only for constants, right? What made it easier for constants than for all types?

lcnr:
It's a bit stronger than "full lazy norm", in that you can't look at the generic args of constants at all, but also quite a bit simpler, as it doesn't really interact with our type system.

The biggest reason this is easier is that the affected parts of const generics are still unstable, so we don't have to worry about causing some subtle breakage. @matthewjasper did try to implement lazy norm and afaict found some annoying regressions: https://gist.github.com/matthewjasper/358245b033a1b7bd158acfe0c08d0d09

lcnr:
rust-lang/rust#87280 now landed so this is ready for review/merge.

cc @nikomatsakis

src/constants.md
```diff
@@ -0,0 +1,65 @@
...
+
+To deal with this, we intend to look at the generic parameters explicitly mentioned
+by the constants and then search the predicates of its parents to figure out which
+of the other generic parameters are reachable by our constant.
+
+**TODO**: Expand this section once the parameter filtering is implemented.
+
+As constants can be part of their parents `where`-clauses, we mention unevaluated
+constants in their parents predicates. It is therefore necessary to mention unevaluated
+constants without being able to first compute the generic parameters
```
nikomatsakis, line 52:
```suggestion
constants before we have computed the generic parameters
```

nikomatsakis (approved):
This seems good. I left one suggestion.

lcnr:
opened #1213

PR #1194, opened 2021-08-27
Fix warning "Renderer command uses a path relative to the renderer output directory ..."

steffahn:
Output of `mdbook build` currently includes
```
2021-08-27 22:01:01 [WARN] (mdbook::renderer): Renderer command `../../ci/linkcheck.sh` uses
a path relative to the renderer output directory `/PATH_TO/rustc-dev-guide/book/linkcheck`.
This was previously accepted, but has been deprecated. Relative executable paths should be
relative to the book root.
```

This PR aims to fix that.

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -19,7 +19,7 @@ enable = true
 level = 0
 
 [output.linkcheck]
-command = "../../ci/linkcheck.sh"
+command = "ci/linkcheck.sh"
 follow-web-links = true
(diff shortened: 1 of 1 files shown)


ehuss (approved):
This was changed back in January in mdbook 0.4.6 due to platform inconsistencies, so any version since then should work.

JohnTitor (approved):
Good catch, thanks!

PR #1197, opened 2021-08-28
Fix typo, clarify backtick wording, and use inline code

camelid:


src/diagnostics.md
```diff
@@ -449,8 +449,8 @@ The possible values of [`Applicability`][appl] are:
 
 - `MachineApplicable`: Can be applied mechanically.
 - `HasPlaceholders`: Cannot be applied mechanically because it has placeholder
-  text in the suggestions. For example, "Try adding a type: \`let x:
-  \<type\>\`".
+  text in the suggestions. For example: ``try adding a type: `let x:
+  <type>` ``.
```
JohnTitor, line 453: Seems this adds trailing whitespace, is there any way to avoid it?

camelid: Not that I'm aware of; I think it's a limitation of Markdown. If someone has a suggestion for how to get rid of it, I'm happy to try it!

jyn514 (approved):

PR #1209, opened 2021-09-19
Add a paragraph on installing `ninja-build` in order to build LLVM

daira:
refs https://github.com/rust-lang/rust/issues/89091

Signed-off-by: Daira Hopwood <daira@jacaranda.org>

### DIFF
--- a/src/building/how-to-build-and-run.md
+++ b/src/building/how-to-build-and-run.md
@@ -52,6 +52,10 @@ If you have already built `rustc` and you change settings related to LLVM, then 
 execute `rm -rf build` for subsequent configuration changes to take effect. Note that `./x.py
 clean` will not cause a rebuild of LLVM.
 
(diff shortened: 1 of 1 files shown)


the8472:
ninja is already listed in the [prerequisites](https://rustc-dev-guide.rust-lang.org/building/prerequisites.html). And there it lists either gnu make *or* ninja. Links to install instructions should probably go there.

daira:
Hmm. It says "ninja, or GNU make 3.81 or later", and I have GNU make, which is not sufficient by default.

jyn514:
@daira GNU make would have worked if you had ninja = false in config.toml. We shouldn't remove those docs.

jyn514:
I'm going to close this for now - I think improving errors from x.py is better than making the guide even longer, and it's not clear to me that this would be easy to find if you're building for the first time.

PR #1211, opened 2021-09-19
Recommend `rustfmt --edition=2021`

camelid:
Blocked on rust-lang/rust#89103.

### DIFF
--- a/src/building/suggested.md
+++ b/src/building/suggested.md
@@ -33,7 +33,7 @@ you can write: <!-- date: 2021-09 --><!-- the date comment is for the edition be
     ],
     "rust-analyzer.rustfmt.overrideCommand": [
         "./build/TARGET_TRIPLE/stage0/bin/rustfmt",
-        "--edition=2018"
+        "--edition=2021"
     ],
     "editor.formatOnSave": true,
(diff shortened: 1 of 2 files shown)


camelid:
No longer blocked!

camelid:
The stdlib is still on Rust 2018, but I don't *think* that should cause any issues since IIUC the syntax should still be compatible with Rust 2021.

jyn514:
I think we should wait until all parts of rust-lang/rust are converted. 2021 has lex errors that 2018 doesn't. (Have you tested running --edition 2021 on master?)

camelid:
> I think we should wait until all parts of rust-lang/rust are converted. 2021 has lex errors that 2018 doesn't. (Have you tested running --edition 2021 on master?)

No I have not, but sure, it's fine to wait until the migration is complete. I don't think Rust 2021 _introduces_ any syntax that would cause errors on Rust 2018, so I guess it should be fine to stick to Rust 2018 formatting for now :+1:

Mark-Simulacrum:
FWIW, I'm not actually sure there's any new syntax in 2021 that would end up *requiring* rustfmt to be configured to 2021 rather than 2018. We should still do it, but I agree that slow playing it seems right.

PR #1214, opened 2021-09-20
Update link for NoLandingPads

xphoniex:
Fix broken link for NoLandingPads struct

### DIFF
--- a/src/mir/visitor.md
+++ b/src/mir/visitor.md
@@ -41,7 +41,7 @@ A very simple example of a visitor can be found in [`NoLandingPads`].
 That visitor doesn't even require any state: it just visits all
 terminators and removes their `unwind` successors.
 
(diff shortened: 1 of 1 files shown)


spastorino:
I think we would need to check out this change https://github.com/rust-lang/rust/pull/86155 and adapt the guide accordingly. cc @alexcrichton

alexcrichton:
Ah I don't know what these links are referring to but yeah the `NoLandingPads` structure has been removed so documentation referencing it would need to be updated rather than simply updating the links.

xphoniex:
another reference [here](https://rustc-dev-guide.rust-lang.org/mir/optimizations.html): 

> pick a small optimization (such as [no_landing_pads](https://github.com/rust-lang/rust/blob/master/compiler/rustc_mir/src/transform/no_landing_pads.rs))

PR #1219, opened 2021-09-25
📢 be louder about the user needing to put their own target triple in the config

spookyvision:
be louder about the user needing to put their own target triple in the config

### DIFF
--- a/src/building/suggested.md
+++ b/src/building/suggested.md
@@ -32,7 +32,7 @@ you can write: <!-- date: 2021-09 --><!-- the date comment is for the edition be
         "--json-output"
     ],
     "rust-analyzer.rustfmt.overrideCommand": [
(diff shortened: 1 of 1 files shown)


JohnTitor:
Hey @spookyvision, could you resolve the above comment and merge conflict? Once they're done, we could merge this PR.

spookyvision:
> Hey @spookyvision, could you resolve the above comment and merge conflict? Once they're done, we could merge this PR.

whoops, didn't see this notification, sorry! I specifically added those emojis because of a conversation in discord, where a person had missed the fact they needed to fill it out. So just adding a `$` maybe won't resolve it for them, maybe depending on whether that character changes syntax highlighting in their editor…

JohnTitor:
Agree that emojis are more visible, but `$FOO` usually means a variable and AFAIK it's widely used. Also, #1250 added an additional note, I think it's helpful enough here.

PR #1223, opened 2021-10-02
Document tracing awesomeness

oli-obk:
cc @hawkw anything cool that I missed?


GuillaumeGomez (approved):

src/tracing.md
```diff
@@ -0,0 +1,190 @@
...
+`INFO` logs, make sure that `debug-logging=true` is turned on in your
+config.toml.
+
+## Logging etiquette and conventions
+
+Because calls to `debug!` are removed by default, in most cases, don't worry
+about adding "unnecessary" calls to `debug!` and leaving them in code you
+commit - they won't slow down the performance of what we ship, and if they
+helped you pinning down a bug, they will probably help someone else with a
+different one.
```
petrochenkov, line 168: This stuff pollutes the code and will usually *not* help anyone outside of this specific debugging session.
So I personally remove it from code that I write before making a PR, and ask others to remove it before merge in code that I review.

oli-obk: I've often had to recompile with new tracing statements somewhere in order to figure out my problem. I do clean up overly noisy ones, but keeping some in is preferable to me. Also: this section existed previously in the guide, it's not new.

I don't know how to draft a good set of instructions that hit the right balance between "keep everything in" and "don't keep any instrumentation in" beyond letting authors and reviewers figure this out amongst themselves.

src/tracing.md
```diff
@@ -0,0 +1,190 @@
...
+One thing to be **careful** of is **expensive** operations in logs.
+
+If in the module `rustc::foo` you have a statement
+
+```Rust
+debug!(x = ?random_operation(tcx));
+```
+
+Then if someone runs a debug `rustc` with `RUSTC_LOG=rustc::bar`, then
+`random_operation()` will run.
```
hawkw, line 186: this seems incorrect: if the `DEBUG` level is enabled for `rustc::bar`, but not for `rustc::foo`, then `random_operation` will not be evaluated. `random_operation` will only be evaluated if `RUSTC_LOG=rustc::foo`. 

if we recorded `random_operation` in `rustc::foo::bar`, and someone turned on `RUSTC_LOG=rustc::foo`, hoping to get debug logs from `rustc::foo::baz`, _then_ `random_operation` would be evaluated.

i think cautioning against expensive, or _especially_ potentially-crashy operations in log messages is definitely important, but the current example isn't actually accurate...

src/tracing.md
```diff
@@ -0,0 +1,190 @@
...
+Because calls to `debug!` are removed by default, in most cases, don't worry
+about adding "unnecessary" calls to `debug!` and leaving them in code you
+commit - they won't slow down the performance of what we ship, and if they
+helped you pinning down a bug, they will probably help someone else with a
+different one.
+
+It may be preferrable to use `trace!` over `debug!` for very noisy logs.
+
+A loosely followed convention is to use `#[instrument(level = "debug")]` in
+favour of `debug!("foo(...)")` at the start of a function `foo`.
```
hawkw, line 173: it may be worth adding a link to https://docs.rs/tracing-attributes/0.1.17/tracing_attributes/attr.instrument.html for details on how to override the `#[instrument]` attribute's behavior (e.g. adding additional fields, skipping arguments, etc)?

src/tracing.md
```diff
@@ -0,0 +1,190 @@
...
+Within functions, prefer `debug!(?variable.field)` over `debug!("xyz = {:?}", variable.field)`
+and `debug!(bar = ?var.method(arg))` over `debug!("bar = {:?}", var.method(arg))`.
```
hawkw, line 175: it might be worth linking to https://docs.rs/tracing/0.1.28/tracing/#recording-fields for details on the macro syntax for recording field values?

src/tracing.md
```diff
@@ -0,0 +1,190 @@
...
+
+For example, to get the `debug!` output and higher for a specific module, you
+can run the compiler with `RUSTC_LOG=path::to::module=debug rustc my-file.rs`.
+All `debug!` output will then appear in standard error.
+
+Note that you can use a partial path and the filter will still work. For
+example, if you want to see `info!` output from only
+`rustdoc::passes::collect_intra_doc_links`, you could use
+`RUSTDOC_LOG=rustdoc::passes::collect_intra_doc_links=info` *or* you could use
+`RUSTDOC_LOG=rustdoc::passes::collect_intra=info`.
```
hawkw, line 78: this is a great summary of how to practically *use* `tracing-subscriber`'s filtering. it might be worth linking to the `tracing-subscriber` docs, here: https://docs.rs/tracing-subscriber/0.2.24/tracing_subscriber/filter/struct.EnvFilter.html#directives as well, which provides a more general overview of the syntax?

src/tracing.md
```diff
@@ -0,0 +1,190 @@
...
+
+If you are developing rustdoc, use `RUSTDOC_LOG` instead. If you are developing
+Miri, use `MIRI_LOG` instead. You get the idea :)
+
+See the [`tracing`] crate's docs, and specifically the docs for [`debug!`] to
```
hawkw, line 83: it looks like the `[`tracing`]` link reference was defined in the other file these docs used to be in, but now isn't defined in this file? i think the link is broken now.

hawkw (commented):
this is great, and I'm really happy to see that `tracing` is useful enough for `rustc` development to deserve its own documentation :)

i had a few small suggestions: mostly, there are some places where it might be helpful to link to `tracing`'s own documentation for additional details. also, i noticed one inaccuracy that might be worth correcting.

overall though, this looks great! :)

src/tracing.md
jyn514: @oli-obk maybe you meant to write "if someone runs rustc with `RUSTC_LOG=rustc`?

oli-obk: this is preexisting and I only skimmed it. will reread properly

jyn514: Oh, is this copied from somewhere? Is there a way we could reduce the duplication between this and the other section? It seems silly to have two separate tutorials on how logging works ...

jyn514: Oh I see, it's moved from src/compiler-debugging.md. Yeah this is fine once you fix it to be accurate then :)

oli-obk: I made this obvious by adding a commit that just does the move before the rest of this PR

spastorino (approved):
This is great, will leave the final decision to merge for @hawkw or @jyn514 given that they had some thoughts

PR #1228, opened 2021-10-09
Fix `rustc_mir` related links

Lee-Janggun:
* With https://github.com/rust-lang/rust/pull/80522, `rustc_mir` has been split into 5 crates, effectively invalidating all related links
* This PR fixes such links, with a few issues.

**The "main" MIR page**
* As seen in https://github.com/rust-lang/rustc-dev-guide/blob/master/src/overview.md, there is a direct link to the `rustc_mir` crate, which has been split into 5. As the split crates do not have a representative for "the MIR crate", I was unable to make a choice 
* The other page with such link is https://github.com/rust-lang/rustc-dev-guide/blob/master/src/mir/index.md  

**NoLandingPads**
* Some docs have links to the now removed `NoLandingPads` optimization. Similar to above, I was unable to make a choice here and left it as is.
* The places where such is mentioned that I have found are four
  * https://github.com/rust-lang/rustc-dev-guide/blob/master/src/mir/debugging.md  
  * https://github.com/rust-lang/rustc-dev-guide/blob/master/src/mir/optimizations.md
  * https://github.com/rust-lang/rustc-dev-guide/blob/master/src/mir/passes.md
  * https://github.com/rust-lang/rustc-dev-guide/blob/master/src/mir/visitor.md
* Related: #1214, https://github.com/rust-lang/rust/pull/86155.

**Hardcoded Versions**
* In `src/sanitizers.md`, there are a few hardcoded links to docs version 1.55.0 that point to the `rustc_mir` crate. I wasn't sure how to change this so I left it as is.

### DIFF
--- a/src/mir/index.md
+++ b/src/mir/index.md
@@ -18,8 +18,8 @@ graphs and desugaring), you may enjoy the
 ## Introduction to MIR
 
 MIR is defined in the [`compiler/rustc_middle/src/mir/`][mir] module, but much of the code
-that manipulates it is found in [`compiler/rustc_mir_build`][mirmanip_build], 
-[`compiler/rustc_mir_transform`][mirmanip_transform], and 
+that manipulates it is found in [`compiler/rustc_mir_build`][mirmanip_build],
+[`compiler/rustc_mir_transform`][mirmanip_transform], and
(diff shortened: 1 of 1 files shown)


JohnTitor:
You could do via a comment: @rustbot label: -waiting-on-review +waiting-on-author

PR #1229, opened 2021-10-10
Fix instructions to get `rust-analyzer` to work

ghost:
This fixed it for me.

### DIFF
--- a/src/building/suggested.md
+++ b/src/building/suggested.md
@@ -33,7 +33,7 @@ you can write: <!-- date: 2021-09 --><!-- the date comment is for the edition be
     ],
     "rust-analyzer.rustfmt.overrideCommand": [
         "./build/TARGET_TRIPLE/stage0/bin/rustfmt",
-        "--edition=2018"
+        "--edition=2021"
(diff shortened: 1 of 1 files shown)


r00ster91:
Oh, my bad.
I agree, I think it'd be nice to have this fixed because the error that rust-analyzer gives for this is not very helpful and probably confuses other contributors too.
```
[ERROR rust_analyzer::reload] failed to switch workspaces: rust-analyzer failed to load workspace: Failed to read Cargo metadata from Cargo.toml file /home/roost/Code/rust/Cargo.toml, cargo 1.57.0-nightly (d56b42c54 2021-09-27): Failed to run `cargo metadata --manifest-path /home/roost/Code/rust/Cargo.toml`: `cargo metadata` exited with an error: error: failed to load manifest for workspace member `/home/roost/Code/rust/src/tools/rust-installer`

Caused by:
  failed to read `/home/roost/Code/rust/src/tools/rust-installer/Cargo.toml`

Caused by:
  No such file or directory (os error 2)
```

PR #1237, opened 2021-10-20
Describe how to trigger perf runs

### LINKED ISSUE #1236: Document how to run perf on PRs
Would be nice to add `@bors try @rust-timer queue` somewhere in the guide.

The current place where we can find this is in https://github.com/rust-lang/rustc-perf/blob/master/collector/README.md?plain=1#L49 but I guess it would be nice to have it in the guide somewhere.

lqd:
This PR adds the command to request perf runs on PRs.

Fixes #1236

### DIFF
--- a/src/profiling.md
+++ b/src/profiling.md
@@ -5,7 +5,10 @@ This section talks about how to profile the compiler and find out where it spend
 Depending on what you're trying to measure, there are several different approaches:
 
 - If you want to see if a PR improves or regresses compiler performance:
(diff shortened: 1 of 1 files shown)


jyn514:
> And I feel like that it's more of instruction for team members rather than (new) contributors so forge or something else would be better

There are already quite a lot of docs that aren't meant for new contributors, I'm not sure how that's related to it being on forge or not. Most of the profiling section is for existing contributors.

PR #1240, opened 2021-10-27
Describe drop elaboration

ecstatic-morse:
I've focused on the "big picture" rather than trying to describe the existing pass from the bottom up. There are some subtleties (mostly around unwinding) that I don't fully understand. I did not implement the existing pass, so it's possible that I've missed something, but I think what I've described has been described correctly.

This (or something like it) was requested as part of the stabilization of rust-lang/rust#73255. I've included a section on the relationship between drop elaboration and const-checking which assumes that `#![feature(const_precise_live_drops)]` is merged.

cc @nikomatsakis @cramertj

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -116,6 +116,7 @@
     - [Opaque Types](./opaque-types-type-alias-impl-trait.md)
 - [Pattern and Exhaustiveness Checking](./pat-exhaustive-checking.md)
 - [MIR dataflow](./mir/dataflow.md)
+- [Drop elaboration](./mir/drop-elaboration.md)
 - [The borrow checker](./borrow_check.md)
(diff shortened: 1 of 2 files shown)


camelid:
@nikomatsakis I assigned you on this, but feel free to reassign!

nikomatsakis (approved):
This was a good read! It made me want to dig a bit deeper into the code-- I think I'd like some examples of how the bits are propagated -- but I definitely think it's good enough already to be useful to people who are coming to read the code.

nikomatsakis:
I believe that the stabilization in question is not proceeding anymore, so I think we can go ahead and merge this. I don't think it's quite at the level I would want to serve as a kind of 'rough language spec', but that's not required just now.

PR #1242, opened 2021-10-28
Some minor adjustments to the diagnostic documentation

xFrednet:
As the title says, some minor adjustments. I've also updated the backtick example, as mdBook currently doesn't render them correctly. (It's expected that this originated from `mdbook-toc`, the related issue is: badboy/mdbook-toc#23)


jyn514 (approved):
Thanks!

PR #1244, opened 2021-11-02
Make subtree docs an overview and add link to clippy docs for details

calebcartwright:
Have found the branch consideration to be a crucial piece when dealing with subtrees and think it warrants being called out explicitly here as well. The Clippy docs are fantastic, so worth including a direct link as well

### DIFF
--- a/src/contributing.md
+++ b/src/contributing.md
@@ -172,32 +172,34 @@ differently from other crates that are directly in this repo:
 * [Clippy](https://github.com/rust-lang/rust-clippy)
 * [rustfmt](https://github.com/rust-lang/rustfmt)
 
-They are just regular files and directories. This is in contrast to `submodule` dependencies
-(see below for those). Only tool authors will actually use any operations here.
+In contrast to `submodule` dependencies
+(see below for those), the `subtree` dependencies are just regular files and directories which can 
+be updated in tree. However, enhancements, bug fixes, etc. specific to these tools should be filed 
+against the tools directly in their respective upstream repositories.
 
 #### Synchronizing a subtree
 
-There are two synchronization directions: `subtree push` and `subtree pull`.
+Periodically the changes made to subtree based dependencies need to be synchronized between this 
+repository and the upstream tool repositories.
 
-```
-git subtree push -P src/tools/clippy git@github.com:your-github-name/rust-clippy sync-from-rust
-```
+Subtree synchronizations are typically handled by the respective tool maintainers. Other users 
+are welcome to submit synchronization PRs, however, in order to do so you you will need to modify 
+your local git installation and follow a very precise set of instructions. 
+These instructions are documented, along with several useful tips and tricks, in the 
+[syncing subtree changes][clippy-sync-docs] section in Clippy's Contributing guide. 
+The instructions are applicable for use with any subtree based tool, just be sure to 
+use the correct corresponding subtree directory and remote repository. 
 
-takes all the changes that
-happened to the copy in this repo and creates commits on the remote repo that match the local
-changes. Every local commit that touched the subtree causes a commit on the remote repo, but is
-modified to move the files from the specified directory to the tool repo root.
+The synchronization process goes in two directions: `subtree push` and `subtree pull`.
 
-Make sure to not pick the `master` branch on the tool repo, so you can open a normal PR to the tool
-to merge that subrepo push.
+A `subtree push` takes all the changes that happened to the copy in this repo and creates commits 
+on the remote repo that match the local changes. Every local 
+commit that touched the subtree causes a commit on the remote repo, but 
+is modified to move the files from the specified directory to the tool repo root.
 
-```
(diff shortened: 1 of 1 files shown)


src/contributing.md
```diff
@@ -189,7 +189,10 @@ changes. Every local commit that touched the subtree causes a commit on the remo
 modified to move the files from the specified directory to the tool repo root.
 
 Make sure to not pick the `master` branch on the tool repo, so you can open a normal PR to the tool
-to merge that subrepo push.
+to merge that subrepo push. It is also recommended that the branch 
+you use as the push target should either be a brand new branch, 
+or first delete the branch if targeting an existing branch, 
+because changes cannot be fast forwarded.
```
pierwill, line 195: The "either... or..." in the second sentence is little confusing. Maybe something like this would work?

```suggestion
to merge that subrepo push. It is also recommended that the branch 
you use as the push target should either be a brand new branch, 
or a branch that exists on the remote.
If the latter, you will need to delete the branch locally, because changes cannot be fast forwarded.
```

calebcartwright: It's not just the local branch, remote needs to be deleted as well

src/contributing.md
```diff
@@ -188,8 +188,12 @@ happened to the copy in this repo and creates commits on the remote repo that ma
 changes. Every local commit that touched the subtree causes a commit on the remote repo, but is
 modified to move the files from the specified directory to the tool repo root.
 
+The target branch on the tool repo should be chosen carefully. 
 Make sure to not pick the `master` branch on the tool repo, so you can open a normal PR to the tool
-to merge that subrepo push.
+to merge that subrepo push. It is also recommended that you create 
+a brand new branch as the push target, however, 
+if you'd prefer to use an existing branch you will typically need to delete it first
+because changes cannot be fast forwarded.
```
jyn514, line 196: "changes cannot be fast forwarded" sounds like a quote of the git error message, but I don't think it's particularly informative - maybe say "subtree pushes rewrite history, so using an existing branch will require you to force push"?

calebcartwright: You can't force push these changes which is part of the problem and why existing branches can't be used.

I'm open to changing the text again if folks think it can be more clear, but I'm also trying to stick to the pattern of not explaining everything inline in the guide (which also skips crucial pieces of information mandatory for performing the syncs, e.g. patching git subtree itself), especially since I think the now-linked Clippy docs cover everything at that more detailed level.

From my perspective this section has historically just called out some of the major "gotchas", and the target branch is one of those "gotchas" that's been missing here. I'm hesitant to start trying to explain the various "whys" in more detail especially if doing so introduces ambiguity around the "what" actually needs to be conveyed.

jyn514: Hmm, ok, can you just rephrase to say "git will give an error that changes cannot ..." instead? I think that's more clear than just the phrase alone, especially since git's error isnt very good IMO.

calebcartwright: Do you feel like a statement to the effect of "do this because if you don't, git will give you this error" is all that helpful for users?

Again I'm open to changes, but here's the piece I'm struggling with... the existing material that's in the guide under https://rustc-dev-guide.rust-lang.org/contributing.html?highlight=subtree#synchronizing-a-subtree is not in a complete/sufficient state. Any user that attempts to do subtree sync via just the steps documented here will run into all sorts of issues and will never be able to perform the sync successfully.

Additionally, for the items that _are_ documented here, most directions/guidance don't drill into their respective why/rationale (e.g. no mention of why subtree vs. submodule, why a prefix is always needed on both push and pull, why "rebasing merge commits is bad in general", etc.). A complete description of the rationale/necessity for why a clean/new branch is needed on pushes would require a relatively lengthy and in-the-weeds description which wouldn't really match the surround content/level of detail. Similarly, trying to give a high level short summary doesn't seem to be working either/has the potential to introduce more confusion.

I'm starting to feel like the best path may be to cut out most of this section and just point over to the clippy docs which do cover the full, reproducible steps that have to be followed, with details around the "why" for edification, and various important tips and tricks that are useful when actually dealing with these activities.

Otherwise, I feel like this section would need to be greatly expanded to provide a holistic guide, and would be largely (if not entirely) duplicative of the existing clippy docs

jyn514: > I'm starting to feel like the best path may be to cut out most of this section and just point over to the clippy docs which do cover the full, reproducible steps that have to be followed, with details around the "why" for edification, and various important tips and tricks that are useful when actually dealing with these activities.

If you're willing to work on that, I think it would be a good idea :) otherwise I'm fine merging this as-is.

calebcartwright: 👍 sounds like a plan

I can update this later today/tomorrow. I don't think there's an urgent need for the concept in the originally proposed text so I'm content with repurposing this PR for the above approach, though also fine opening a follow up if you want to go ahead and merge.

jyn514: Let's use this PR so you have less rebasing / git magic to worry about :)

jyn514 (approved):
This is great, thank you!

PR #1249, opened 2021-11-04
Document more compiletest headers.

ehuss:
This adds some more documentation for compiletest headers.

### DIFF
--- a/src/tests/adding.md
+++ b/src/tests/adding.md
@@ -215,9 +215,57 @@ source.
 * `error-pattern` checks the diagnostics just like the `ERROR` annotation
   without specifying error line. This is useful when the error doesn't give
   any span.
+* `incremental` runs the test with the `-C incremental` flag and an empty
+  incremental directory. This should be avoided when possible; you should use
+  an *incremental mode* test instead. Incremental mode tests support running
+  the compiler multiple times and verifying that it can load the generated
+  incremental cache. This flag is for specialized circumstances, like checking
+  the interaction of codegen unit partitioning with generating an incremental
+  cache.
+* `aux-build` is used to compile additional crates to link. Just pass it the
+  name of the source file. The source file should be in a directory called
+  `auxiliary` beside the test file. The aux crate will be built as a dylib if
+  possible (unless on a platform that does not support them, or
+  `no-prefer-dynamic` is specified in the aux file). The `-L` flag is used to
+  find the extern crates.
+* `aux-crate` is very similar to `aux-build`, however it uses the `--extern`
+  flag to link to the extern crate. That allows you to specify the additional
+  syntax of the `--extern` flag, such as renaming a dependency. For example,
+  `// aux-crate:foo=bar.rs` will compile `auxiliary/bar.rs` and add make it
+  available under then name `foo` within the test. This is similar to how
+  Cargo does dependency renaming.
+* `no-prefer-dynamic` will force an auxiliary crate to be built as an rlib
+  instead of a dylib. When specified in a test, it will remove the use of `-C
+  prefer-dynamic`. This can be useful in a variety of circumstances. For
+  example, it can prevent a proc-macro from being built with the wrong crate
+  type. Or if your test is specifically targeting behavior of other crate
+  types, it can be used to prevent building with the wrong crate type.
+* `force-host` will force the test to build for the host platform instead of
+  the target. This is useful primarily for auxiliary proc-macros, which need
+  to be loaded by the host compiler.
+* `pretty-mode` for pretty-print tests specifies the mode it should run in.
+  The default is `normal` if not specified.
+* `pretty-compare-only` causes a pretty test to only compare the
+  pretty-printed output. It will not try to compile the expanded output to
+  typecheck it. This is needed for a pretty-mode that does not expand to valid
+  rust, or for other situations where the expanded output cannot be compiled.
+* `pretty-expanded` allows a pretty test to also run with
+  `-Zunpretty=expanded` as a final step, and will also try to compile the
+  resulting output (without codegen). This is needed because not all code can
+  be compiled after being expanded. Pretty tests should specify this if they
+  can. More history about this may be found in [#23616].
+* `pp-exact` is used to ensure a pretty-print test results in specific output.
+  If specified without a value, then it means the pretty-print output should
+  match the original source. If specified with a value, as in `//
+  pp-exact:foo.pp`, will ensure that that pretty-printed output matches the
+  contents of the given file. Otherwise, if `pp-exact` is not specified, then
+  the pretty-printed output will be pretty-printed one more time, and the
+  output of the two pretty-printing rounds will be compared to ensure that the
+  pretty-printed output converges to a steady state.
 
 [`header.rs`]: https://github.com/rust-lang/rust/tree/master/src/tools/compiletest/src/header.rs
 [bless]: ./running.md#editing-and-updating-the-reference-files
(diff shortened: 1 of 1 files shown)


src/tests/adding.md
```diff
@@ -215,9 +215,57 @@ source.
...
+* `no-prefer-dynamic` will force an auxiliary crate to be built as an rlib
+  instead of a dylib. When specified in a test, it will remove the use of `-C
+  prefer-dynamic`. This can be useful in a variety of circumstances. For
+  example, it can prevent a proc-macro from being built with the wrong crate
+  type. Or if your test is specifically targeting behavior of other crate
+  types, it can be used to prevent building with the wrong crate type.
+* `force-host` will force the test to build for the host platform instead of
+  the target. This is useful primarily for auxiliary proc-macros, which need
+  to be loaded by the host compiler.
+* `pretty-mode` for pretty-print tests specifies the mode it should run in.
```
pierwill, line 246:
```suggestion
* `pretty-mode` specifies the mode pretty-print tests should run in.
```

src/tests/adding.md
```diff
@@ -215,9 +215,57 @@ source.
...
+  The default is `normal` if not specified.
+* `pretty-compare-only` causes a pretty test to only compare the
+  pretty-printed output. It will not try to compile the expanded output to
+  typecheck it. This is needed for a pretty-mode that does not expand to valid
+  rust, or for other situations where the expanded output cannot be compiled.
```
pierwill, line 251:
```suggestion
  Rust, or for other situations where the expanded output cannot be compiled.
```

What's an example of a mode that doesn't expand to valid Rust?

src/tests/adding.md
```diff
@@ -215,9 +215,57 @@ source.
...
+* `pretty-expanded` allows a pretty test to also run with
+  `-Zunpretty=expanded` as a final step, and will also try to compile the
```
pierwill, line 253:
```suggestion
  `-Zunpretty=expanded` as a final step. It will also try to compile the
```

ehuss:
Thanks for taking a look! I have applied the suggestions.

src/tests/adding.md
```diff
@@ -215,9 +220,57 @@ source.
...
+  incremental cache. This flag is for specialized circumstances, like checking
+  the interaction of codegen unit partitioning with generating an incremental
+  cache.
+* `aux-build` is used to compile additional crates to link. Just pass it the
+  name of the source file. The source file should be in a directory called
+  `auxiliary` beside the test file. The aux crate will be built as a dylib if
+  possible (unless on a platform that does not support them, or
+  `no-prefer-dynamic` is specified in the aux file). The `-L` flag is used to
+  find the extern crates.
+* `aux-crate` is very similar to `aux-build`, however it uses the `--extern`
```
camelid, line 236: nit:

```suggestion
* `aux-crate` is very similar to `aux-build`; however, it uses the `--extern`
```

src/tests/adding.md
```diff
@@ -215,9 +220,57 @@ source.
...
+  flag to link to the extern crate. That allows you to specify the additional
+  syntax of the `--extern` flag, such as renaming a dependency. For example,
+  `// aux-crate:foo=bar.rs` will compile `auxiliary/bar.rs` and add make it
```
camelid, line 239:
```suggestion
  `// aux-crate:foo=bar.rs` will compile `auxiliary/bar.rs` and make it
```

src/tests/adding.md
```diff
@@ -215,9 +220,57 @@ source.
...
+* `no-prefer-dynamic` will force an auxiliary crate to be built as an rlib
+  instead of a dylib. When specified in a test, it will remove the use of `-C
+  prefer-dynamic`. This can be useful in a variety of circumstances. For
+  example, it can prevent a proc-macro from being built with the wrong crate
+  type. Or if your test is specifically targeting behavior of other crate
+  types, it can be used to prevent building with the wrong crate type.
+* `force-host` will force the test to build for the host platform instead of
+  the target. This is useful primarily for auxiliary proc-macros, which need
+  to be loaded by the host compiler.
+* `pretty-mode` specifies the mode pretty-print tests should run in.
```
camelid, line 251: IIRC, all tests `>=check-pass` are also run as pretty-print tests. Is that correct? If so, does `pretty-mode` affect those too?

src/tests/adding.md
```diff
@@ -215,9 +220,57 @@ source.
...
+  The default is `normal` if not specified.
+* `pretty-compare-only` causes a pretty test to only compare the
+  pretty-printed output. It will not try to compile the expanded output to
+  typecheck it. This is needed for a pretty-mode that does not expand to valid
+  Rust, or for other situations where the expanded output cannot be compiled.
+* `pretty-expanded` allows a pretty test to also run with
+  `-Zunpretty=expanded` as a final step. It will also try to compile the
+  resulting output (without codegen). This is needed because not all code can
+  be compiled after being expanded. Pretty tests should specify this if they
```
camelid, line 260: Maybe add a brief description of what kind of code can't be compiled after being expanded?

src/tests/adding.md
```diff
@@ -215,9 +220,57 @@ source.
...
+  can. More history about this may be found in [#23616].
+* `pp-exact` is used to ensure a pretty-print test results in specific output.
+  If specified without a value, then it means the pretty-print output should
+  match the original source. If specified with a value, as in `//
+  pp-exact:foo.pp`, will ensure that that pretty-printed output matches the
```
camelid, line 265:
```suggestion
  pp-exact:foo.pp`, it will ensure that the pretty-printed output matches the
```

camelid (commented):
I'm learning about most of these headers for the first time, so I didn't check for accuracy, but it looks good to me aside from that. I left a few nits.

Would it make sense to alphabetize this list to make it easier to skim?

Also, is there someone who can check to make sure the descriptions are accurate?

src/tests/adding.md
ehuss: Unless I'm mistaken, the pretty-print checks that ran on all tests were removed several years ago. The only pretty-print tests now are those in the `src/test/pretty` directory.

There are a bunch of `// pretty-expanded` comment headers sprinkled across the testsuite, but those are vestigial.

ehuss:
Thanks for taking a look!

> Would it make sense to alphabetize this list to make it easier to skim?

As mentioned on Zulip ([here](https://rust-lang.zulipchat.com/#narrow/stream/196385-t-compiler.2Fwg-rustc-dev-guide/topic/Expanding.20on.20writing.20tests/near/260312488)), I'd like to work on reorganizing the testing chapter. As part of that, I'd like to break this list up into different topics (for example, the pretty-print headers, debugger headers, incremental, etc.).  I'd like to get this merged before I start on that.

src/tests/adding.md
camelid: Ah, I didn't realize that. Good to know, thanks.

camelid:
> Thanks for taking a look!

Sure, thanks for making this PR :)

> As mentioned on Zulip (here), I'd like to work on reorganizing the testing chapter. As part of that, I'd like to break this list up into different topics (for example, the pretty-print headers, debugger headers, incremental, etc.). I'd like to get this merged before I start on that.

Sounds good.

camelid:
@petrochenkov could you review this to check accuracy? I'm not super familiar with all the different headers.

jyn514:
@camelid same here, I think this is good enough to merge and we can fix it later if necessary

jyn514 (approved):

PR #1256, opened 2021-11-14
Fix compare-mode documentation

rukai:
I got the list of possible compare-modes from here: https://github.com/rust-lang/rust/blob/65b3c853178b04c5cbe41f26e6a1b1b90d7b8cc6/src/tools/compiletest/src/common.rs#L120
And I determined that compare-modes are separate to revisions from observing the behavior of tests like: https://github.com/rust-lang/rust/blob/56ebd57960efd1d96c142b456add67b5257c21fb/src/test/ui/regions/regions-outlives-projection-container-wc.rs

### DIFF
--- a/src/tests/running.md
+++ b/src/tests/running.md
@@ -186,8 +186,17 @@ To run the UI test suite in NLL mode, one would use the following:
 ./x.py test src/test/ui --compare-mode=nll
 ```
 
-Other examples of compare-modes are "noopt", "migrate", and
-[revisions](./adding.html#revisions).
+The possible compare modes are:
+
+* nll - currently nll is implemented in migrate mode, this option runs with true nll.
+* polonius
+* chalk
+* split-dwarf
+* split-dwarf-single
+
+Note that compare modes are seperate to [revisions](./adding.html#revisions).
+All revisions are tested when running `./x.py test src/test/ui`,
+however compare-modes must be manually run individually via the `--compare-mode` flag.
 
 ## Running tests manually
 


src/tests/running.md
```diff
@@ -186,8 +186,17 @@ To run the UI test suite in NLL mode, one would use the following:
...
+The possible compare modes are:
+
+* nll - currently nll is implemented in migrate mode, this option runs with true nll.
+* polonius
+* chalk
+* split-dwarf
+* split-dwarf-single
+
+Note that compare modes are seperate to [revisions](./adding.html#revisions).
+All revisions are tested when running `x./py test src/test/ui`,
```
jyn514, line 198:
```suggestion
All revisions are tested when running `./x.py test src/test/ui`,
```

jyn514 (commented):
Thanks for the fix! This looks good to me other than a little typo :)

rukai:
Thanks! should be good to go now.

jyn514 (approved):

src/tests/running.md
```diff
@@ -186,8 +186,17 @@ To run the UI test suite in NLL mode, one would use the following:
...
-[revisions](./adding.html#revisions).
+The possible compare modes are:
+
+* nll - currently nll is implemented in migrate mode, this option runs with true nll.
+* polonius
+* chalk
+* split-dwarf
+* split-dwarf-single
+
+Note that compare modes are seperate to [revisions](./adding.html#revisions).
```
spastorino, line 197:
```suggestion
Note that compare modes are separate to [revisions](./adding.html#revisions).
```

rukai: ah.
I'll make a follow up PR to fix it.

PR #1258, opened 2021-11-15
Unify `x.py` usage

ken-matsui:
There are various expressions of `x.py`: x.py, `x.py`, `./x.py`, x.py cmd, `x.py cmd`, and `./x.py cmd`.
These make us confused, so I unified them to `x.py` or `./x.py cmd`.


JohnTitor (approved):
Looks good, thanks!

PR #1260, opened 2021-11-15
Fix some more links

JohnTitor:
Separating from #1259 because I'm not 100% sure if `compute_hir_hash` is the right link.
cc @cjgillot could you review this?

### DIFF
--- a/src/building/bootstrapping.md
+++ b/src/building/bootstrapping.md
@@ -56,7 +56,7 @@ because one must first build the new compiler with an older compiler
 and then use that to build the new compiler with itself.
 For development, you usually only want the `stage1` compiler,
 which you can build with `./x.py build library/std`.
-See [Building the Compiler](/building/how-to-build-and-run.html#building-the-compiler).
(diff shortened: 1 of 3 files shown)


src/building/bootstrapping.md
```diff
@@ -56,7 +56,7 @@ because one must first build the new compiler with an older compiler
 and then use that to build the new compiler with itself.
 For development, you usually only want the `stage1` compiler,
 which you can build with `./x.py build library/std`.
-See [Building the Compiler](/building/how-to-build-and-run.html#building-the-compiler).
+See [Building the Compiler](./how-to-build-and-run.html#building-the-compiler).
```
camelid, line 59: Pre-existing, but while you're at it:

```suggestion
See [Building the Compiler](./how-to-build-and-run.md#building-the-compiler).
```

camelid (approved):
r=me on the non-crate-hash changes. Let's wait for @cjgillot for the crate hash change though.

jyn514:
@camelid I think we should merge this in the meantime and we can change the link later if necessary.

jyn514 (approved):

PR #1262, opened 2021-11-24
Update examples for latest compiler version

simon-perriard:
I used 1.58.0 nigthly to compile but compilation should be supported since 1.56.0 nightly


jyn514 (approved):
Thanks!

PR #1264, opened 2021-11-27
Improve 'Running tests manually' section

rukai:
I started rustc development by running tests without ever directly invoking the compiler.
Because of this the `Running tests manually` section didnt work for me and it wasnt clear what I had to do to fix it.
I added a link to the prereq section to avoid similar occurrences in the future.

I also added a note explaining how rebuilding rust will affect the toolchain added to rustup.

### DIFF
--- a/src/building/how-to-build-and-run.md
+++ b/src/building/how-to-build-and-run.md
@@ -212,6 +212,11 @@ host: x86_64-unknown-linux-gnu
 release: 1.48.0-dev
 LLVM version: 11.0
 ```
+
+The rustup toolchain points to the specified toolchain compiled in your `build` directory,
+so the rustup toolchain will be updated whenever `x.py build` or `x.py test` are run for
+that toolchain/stage.
+
 ## Other `x.py` commands
 
 Here are a few other useful `x.py` commands. We'll cover some of them in detail
--- a/src/tests/running.md
+++ b/src/tests/running.md
@@ -200,8 +200,10 @@ however compare-modes must be manually run individually via the `--compare-mode`
 
 ## Running tests manually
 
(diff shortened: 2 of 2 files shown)


src/building/how-to-build-and-run.md
```diff
@@ -212,6 +212,11 @@ host: x86_64-unknown-linux-gnu
 release: 1.48.0-dev
 LLVM version: 11.0
 ```
+
+The rustup toolchain points to the specified toolchain compiled in your `build` directory,
+so the rustup toolchain will be updated whenever `x.py build` or `x.py test` are run for
+that toolchain/stage.
```
jyn514, line 218: Hmm, do you think this wasn't clear before? It's definitely true but this section is already kind of long.

jyn514 (commented):
Good call on the link in the testing section, thanks for adding it :)

src/building/how-to-build-and-run.md
rukai: It wasnt clear to me so I had to test it out to verify how it worked.

I think in particular it might be surprising that `x.py test` can affect the rustup toolchain.
I thought that this wasnt the case for cargo, but I just tested it and `cargo test` does overwrite the output of `cargo build`!

So if its too long I could shorten it to:
```
Note that both `x.py build` and `x.py test` will update the linked rustup toolchain.
```

Otherwise if this is all obvious then im happy to remove it entirely.
I totally agree that saying obvious things is wasting everyone's bandwidth.

jyn514: > It wasnt clear to me so I had to test it out to verify how it worked.

👍 Seems good to add this bit then.

jyn514 (approved):

PR #1271, opened 2021-12-05
Update rustdoc internals

jsha:
Use current paths when discussing source files.
Update cheat sheet section with download-rustc.
Add "use cases" section.

### DIFF
--- a/src/rustdoc.md
+++ b/src/rustdoc.md
@@ -36,30 +36,141 @@ does is call the `main()` that's in this crate's `lib.rs`, though.)
 
 ## Cheat sheet
 
+* Run `./x.py setup tools` before getting started. This will configure `x.py`
+  with nice settings for developing rustdoc and other tools, including
+  downloading a copy of rustc rather than building it.
 * Use `./x.py build` to make a usable
   rustdoc you can run on other projects.
   * Add `library/test` to be able to use `rustdoc --test`.
-  * If you've used `rustup toolchain link local /path/to/build/$TARGET/stage1`
-    previously, then after the previous build command, `cargo +local doc` will
-    Just Work.
+  * Run `rustup toolchain link stage2 build/$TARGET/stage2` to add a
+    custom toolchain called `stage2` to your rustup environment. After
+    running that, `cargo +stage2 doc` in any directory will build with
+    your locally-compiled rustdoc.
 * Use `./x.py doc library/std` to use this rustdoc to generate the
   standard library docs.
-  * The completed docs will be available in `build/$TARGET/doc/std`, though the
-    bundle is meant to be used as though you would copy out the `doc` folder to
-    a web server, since that's where the CSS/JS and landing page are.
+  * The completed docs will be available in `build/$TARGET/doc/std`.
+  * If you want to copy those docs to a webserver, copy all of
+    `build/$TARGET/doc`, since that's where the CSS, JS, fonts, and landing
+    page are.
 * Use `./x.py test src/test/rustdoc*` to run the tests using a stage1 rustdoc.
   * See [Rustdoc internals] for more information about tests.
-* Most of the HTML printing code is in `html/format.rs` and `html/render.rs`.
+
+## Code structure
+
+* All paths in this section are relative to `src/librustdoc` in the rust-lang/rust repository.
+* Most of the HTML printing code is in `html/format.rs` and `html/render/mod.rs`.
   It's in a bunch of `fmt::Display` implementations and supplementary
   functions.
 * The types that got `Display` impls above are defined in `clean/mod.rs`, right
   next to the custom `Clean` trait used to process them out of the rustc HIR.
-* The bits specific to using rustdoc as a test harness are in `test.rs`.
+* The bits specific to using rustdoc as a test harness are in
+  `doctest.rs`.
 * The Markdown renderer is loaded up in `html/markdown.rs`, including functions
   for extracting doctests from a given block of Markdown.
-* The tests on rustdoc *output* are located in `src/test/rustdoc`, where
+* The tests on the structure of rustdoc HTML output are located in `src/test/rustdoc`, where
   they're handled by the test runner of rustbuild and the supplementary script
   `src/etc/htmldocck.py`.
-* Tests on search index generation are located in `src/test/rustdoc-js`, as a
+
+## Tests
+
+* All paths in this section are relative to `src/test` in the rust-lang/rust repository.
+* Tests on search index generation are located in `rustdoc-js`, as a
   series of JavaScript files that encode queries on the standard library search
   index and expected results.
+* Tests on the "UI" of rustdoc (the terminal output it produces when run) are in
+  `rustdoc-ui`
+* Tests on the "GUI" of rustdoc (the HTML, JS, and CSS as rendered in a browser)
+  are in `rustdoc-gui`. These use a [NodeJS tool called
+  browser-UI-test](https://github.com/GuillaumeGomez/browser-UI-test/) that uses
+  puppeteer to run tests in a headless browser and check rendering and
+  interactivity.
+
+## Constraints
+
+We try to make rustdoc work reasonably well with JavaScript disabled, and when
+browsing local files. We support
+[these browsers](https://rust-lang.github.io/rfcs/1985-tiered-browser-support.html#supported-browsers).
+
+Supporting local files (`file:///` URLs) brings some surprising restrictions.
+Certain browser features that require secure origins, like `localStorage` and
+Service Workers, don't work reliably. We can still use such features but we
+should make sure pages are still usable without them.
+
+## Multiple runs, same output directory
+
+Rustdoc can be run multiple times for varying inputs, with its output set to the
+same directory. That's how cargo produces documentation for dependencies of the
+current crate. It can also be done manually if a user wants a big
+documentation bundle with all of the docs they care about.
+
+HTML is generated independently for each crate, but there is some cross-crate
+information that we update as we add crates to the output directory:
+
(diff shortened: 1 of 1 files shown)


camelid:
@jyn514 says on Zulip:

> please don't merge until I get a chance to write them down, I'll try to do that tomorrow

src/rustdoc.md
```diff
@@ -36,30 +36,174 @@ does is call the `main()` that's in this crate's `lib.rs`, though.)
...
-* Use `./x.py build` to make a usable
-  rustdoc you can run on other projects.
+* Copy config.toml.example (in the root of your Rust checkout to config.toml,
+  add add the lines below. This will speed up build times dramatically by not
+  building rustc locally. And skipping minifcation will make debugging JS
+  easier.
+    download-rustc = true
+    download-ci-llvm = true
+    skip-rebuild = true
+    docs-minification = false
```
camelid, line 46: For one, this should use inline code with `` ```toml ``. But this is handled by using `./x.py setup tools`, so beginners should already be setup. I don't think we really need this here?

src/rustdoc.md
```diff
@@ -36,30 +36,174 @@ does is call the `main()` that's in this crate's `lib.rs`, though.)
...
-* The tests on rustdoc *output* are located in `src/test/rustdoc`, where
+* The tests on the structure of rustdoc HTML output are located in `src/test/rustdoc`, where
   they're handled by the test runner of rustbuild and the supplementary script
   `src/etc/htmldocck.py`.
 * Tests on search index generation are located in `src/test/rustdoc-js`, as a
   series of JavaScript files that encode queries on the standard library search
   index and expected results.
+* Tests on the "UI" of rustdoc (the terminal output it produces when run) are in
+  `src/tests/rustdoc-ui`
+* Tests on the "GUI" of rustdoc are in `src/tests/rustdoc-gui`. These use a
```
camelid, line 79: Probably worth defining GUI too.

src/rustdoc.md
```diff
@@ -36,30 +36,174 @@ does is call the `main()` that's in this crate's `lib.rs`, though.)
...
-* Most of the HTML printing code is in `html/format.rs` and `html/render.rs`.
+
+## Code structure
+
+* Most of the HTML printing code is in `src/librustdoc/html/format.rs` and
+  `src/librustdoc/html/render/mod.rs`.
   It's in a bunch of `fmt::Display` implementations and supplementary
   functions.
-* The types that got `Display` impls above are defined in `clean/mod.rs`, right
+* The types that got `Display` impls above are defined in `src/librustdoc/clean/mod.rs`, right
```
camelid, line 65: Do we really need to prefix _everything_ with `src/librustdoc`? I'd rather add one paragraph at the beginning of the section saying that rustdoc is in `src/librustdoc`, and then use relative paths here.

src/rustdoc.md
```diff
@@ -36,30 +36,174 @@ does is call the `main()` that's in this crate's `lib.rs`, though.)
...
+
+By default `cargo doc` will generate documentation for a crate and all of its
+dependencies. That can result in a very large documentation bundle, with a large
+(and slow) search corpus. The Cargo flag `--no-deps` inhibits that behavior and
+generates docs for just the crate.
+
+### Self-hosted project docs
+
+Some projects like to host their own documentation. For example:
+https://docs.serde.rs/. This is easy to do by locally generating docs, and
```
camelid, line 170:
```suggestion
<https://docs.serde.rs>. This is easy to do by locally generating docs, and
```

src/rustdoc.md
```diff
@@ -36,30 +36,174 @@ does is call the `main()` that's in this crate's `lib.rs`, though.)
...
+ - Static files (JS and CSS) are renamed so they unique identifiers (including a
+   hash) for each version, and stored independently of the crate docs.
+
+Documentation on docs.rs is always generated for a single crate at a time, so
+the search and sidebar functionality don't include dependencies of the current
+crate.
+
+### Locally generated docs
+
+Develops run `cargo doc --open` in crates they have checked out locally to see
```
camelid, line 155: "Developers" seems to mean "developers of rustdoc", but that's not what you mean AFAICT:

```suggestion
Users can run `cargo doc --open` in crates they have checked out locally to see
```

src/rustdoc.md
```diff
@@ -36,30 +36,174 @@ does is call the `main()` that's in this crate's `lib.rs`, though.)
...
+storage and display a navigation bar at the top:
+
+ - The `<body>` tag is renamed to a `<div>` with the id `rustdoc_body_wrapper`
+   so docs.rs can have its own body tag.
+ - Static files (JS and CSS) are renamed so they unique identifiers (including a
+   hash) for each version, and stored independently of the crate docs.
+
+Documentation on docs.rs is always generated for a single crate at a time, so
+the search and sidebar functionality don't include dependencies of the current
+crate.
```
camelid, line 151: I don't know if this should really be in the rustdoc chapter; it's more relevant for the docs.rs team.

src/rustdoc.md
```diff
@@ -36,30 +36,174 @@ does is call the `main()` that's in this crate's `lib.rs`, though.)
...
+
+There are a few major use cases for rustdoc that you should keep in mind when
+working on it:
+
+### Standard library docs
+
+These are published at https://doc.rust-lang.org/std as part of the Rust release
+process. Stable releases are also uploaded to specific versioned URLs like
+https://doc.rust-lang.org/1.57.0/std/. Beta and nightly docs are published to
+https://doc.rust-lang.org/beta/std/ and https://doc.rust-lang.org/nightly/std/.
```
camelid, line 123: I think none of these links will render as `<a>` tags. You need to add `<` and `>` around them.

src/rustdoc.md
jyn514: It would be nice to recommend `x.py setup tools` here; but yeah, I agree we shouldn't recommend custom configs. @jsha could you add `docs-minification = false` to `src/bootstrap/defaults/tools.toml`? I don't think `skip-rebuild` is a good option to set though, that just runs with completely the wrong version (and shouldn't be necessary with `download-ci-llvm` anyway).

src/rustdoc.md
jyn514: The bit about the `body` tag is overkill, but I think this is definitely something we should have in the guide - right now there's no tests for whether static files should be `invocation-specific` or `toolchain-specific`, so people need to know the difference (and the difference only matters for docs.rs).

src/rustdoc.md
```diff
@@ -36,30 +36,174 @@ does is call the `main()` that's in this crate's `lib.rs`, though.)
...
+--html-after-content FILES
+                files to include inline between the content and
+                </body> of a rendered Markdown file or generated
+                documentation
+--markdown-before-content FILES
+                files to include inline between <body> and the content
+                of a rendered Markdown file or generated documentation
+--markdown-after-content FILES
+                files to include inline between the content and
+                </body> of a rendered Markdown file or generated
```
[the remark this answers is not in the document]
camelid: It also seems likely to get out of date quickly.

jsha:
Great feedback, thanks! I've responded to it all. I also updated some of the sentences in `cheat sheet` so they consistently use the imperative mood.

jyn514 (approved):
This looks good to me :) I think there's a little more detail in the "use cases" section than necessary but everything else is amazing and I don't want to block it.

src/rustdoc.md
```diff
@@ -36,30 +36,141 @@ does is call the `main()` that's in this crate's `lib.rs`, though.)
...
 * Use `./x.py build` to make a usable
   rustdoc you can run on other projects.
   * Add `library/test` to be able to use `rustdoc --test`.
-  * If you've used `rustup toolchain link local /path/to/build/$TARGET/stage1`
-    previously, then after the previous build command, `cargo +local doc` will
-    Just Work.
+  * Run `rustup toolchain link local /path/to/build/$TARGET/stage1` to add a
+    custom toolchain called `local` to your rustup environment. After running
+    that, `cargo +local doc` in any directory will build with your
+    locally-compiled rustdoc.
```
camelid, line 48: FWIW, this doesn't work correctly with download-rustc on. Instead, you need to use stage2.

We don't have to address that here, but perhaps change the toolchain name from `local` to `stage1` so it's easier to tell the two stages apart?

jyn514: Yeah, I think this should be using the stage2 directory as well since it recommended download-rustc.

jsha: Fixed, thanks!

src/rustdoc.md
```diff
@@ -36,30 +36,141 @@ does is call the `main()` that's in this crate's `lib.rs`, though.)
...
+Rustdoc can be run multiple times for varying inputs, with its output set to the
+same directory. That's how cargo produces documentation for dependencies of the
+current crate. It can also be done manually if a user wants a big
+documentation bundle with all of the docs they care about.
+
+HTML is generated independently for each crate, but there is some cross-crate
+information that we update as we add crates to the output directory:
+
+ - crates`<SUFFIX>`.js holds a list of all crates in the output directory.
+ - search-index`<SUFFIX>`.js holds a list of all searchable items.
```
camelid, line 110: Hmm, seems like it'd be better to just use `` `crates<SUFFIX>.js` ``. I imagine the way it is now will render in a weird way.

jsha: Updated!

camelid:
I haven't looked at all the changes, but nothing stands out to me from skimming, so feel free to merge if you want @jyn514 :)

jyn514:
Thanks jsha for working on this!

PR #1272, opened 2021-12-05
Improve documentation on r?

rukai:
* Whenever its mentioned that the bot randomly assigns someone I added that its based on the files you touch.
    + I think this is an important detail because otherwise, at least personally, I worry about trying to assign to an appropriate user but knowing that there is some intelligence behind the bots decision is reassuring.
* I added documentation for assigning to a group/team.
   + This is a useful feature but isnt documented anywhere.
   + I'm not sure which file describing `r?` I should put it in but contributing.md seemed like the one most suitable for  extensive documentation

### DIFF
--- a/src/getting-started.md
+++ b/src/getting-started.md
@@ -398,7 +398,7 @@ When you open a PR on the `rust-lang/rust` repo, a bot called `@rust-highfive` w
 automatically assign a reviewer to the PR based on which files you changed.
 The reviewer is the person that will approve the PR to be tested and merged.
 If you want a specific reviewer (e.g. a team member you've been working with),
-you can specifically request them by writing `r? @user` (e.g. `r? @eddyb`) in
+you can specifically request them by writing `r? @user` (e.g. `r? @jyn514`) in
 either the original post or a followup comment
 (you can see [this comment][r?] for example).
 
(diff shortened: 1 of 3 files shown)


src/getting-started.md
```diff
@@ -396,11 +395,12 @@ highlights, but there are a lot more details, which we will link to below.
...
 When you open a PR on the `rust-lang/rust` repo, a bot called `@rust-highfive` will
-automatically assign a reviewer to the PR. The reviewer is the person that will
-approve the PR to be tested and merged. If you want a specific reviewer (e.g. a
-team member you've been working with), you can specifically request them by
-writing `r? @user` (e.g. `r? @eddyb`) in either the original post or a followup
-comment (you can see [this comment][r?] for example).
+automatically assign a reviewer to the PR based on which files you changed.
+The reviewer is the person that will approve the PR to be tested and merged.
+If you want a specific reviewer (e.g. a team member you've been working with),
+you can specifically request them by writing `r? @user` (e.g. `r? @eddyb`) in
```
jyn514, line 401: Let's not use eddyb as an example, they already get too many notifications.
```suggestion
you can specifically request them by writing `r? @user` (e.g. `r? @jyn514`) in
```

jyn514 (approved):
This is great, thanks!

src/getting-started.md
camelid: FWIW, eddyb was the existing example.

PR #1273, opened 2021-12-08
Clarify cargo fallback behavior for rustup link

couchand:
When building and testing on a new system (without an installed `nightly` toolchain), the default behavior of `rustup link` can seem a bit surprising, namely that it uses **stable** `cargo`. This change adds a note to the section on `rustup link` to help clarify what's happening and how to fix it.

I think it would also make sense to include a note about how to get `rustup link` to use a cargo compiled with `./x.py build cargo`, but I haven't yet figured out how to do that.

### DIFF
--- a/src/building/how-to-build-and-run.md
+++ b/src/building/how-to-build-and-run.md
@@ -217,6 +217,11 @@ The rustup toolchain points to the specified toolchain compiled in your `build` 
 so the rustup toolchain will be updated whenever `x.py build` or `x.py test` are run for
 that toolchain/stage.
 
+**Note:** the toolchain we've built does not include `cargo`.  In this case, `rustup` will
+fall back to using `cargo` from the installed `nightly`, `beta`, or `stable` toolchain
+(in that order).  If you need to use unstable `cargo` flags, be sure to run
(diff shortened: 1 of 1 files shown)


couchand:
I was just alerted to the docs for `rustup` on this issue [here](https://rust-lang.github.io/rustup/concepts/toolchains.html#custom-toolchains), perhaps it's worth linking there?

jyn514:
@couchand yup, linking to the rustup docs seems good :)

JohnTitor:
@couchand Triage ping, could you add a link mentioned above?

couchand:
@jyn514 @JohnTitor thank you for your feedback/poke.  Updated!

jyn514 (approved):
There should probably be some way to build with in-tree cargo, but that shouldn't block documenting the current behavior. Thanks!

couchand:
> There should probably be some way to build with in-tree cargo

That sounds right to me?  I don't have the context, though.  Offhand, I don't really even have the context for my own PR! :upside_down_face:

JohnTitor (approved):
Thanks!

PR #1274, opened 2021-12-13
Fix some broken links

Badel2:


JohnTitor (approved):
Note that we suppose to see the docs on rustc-dev-guide.rust-lang.org so the links aren't actually _broken_. But IIRC @camelid prefer `.md` over `.html` so there's no reason to reject, I think. Thanks for contributing :)

camelid:
Yeah, I think there's some advantage to using `.md`, though I don't quite remember what it is.

Badel2:
Oh, I see. I found a direct link to https://github.com/rust-lang/rustc-dev-guide/blob/master/src/profiling.md somewhere, so I assumed that this guide was meant to be read using github.

PR #1276, opened 2021-12-15
Move date-check comment to fix Markdown syntax

camelid:
The inline code wasn't being rendered correctly.


jyn514:
I don't understand - why didn't it work before? Why does this change fix it?

camelid:
Quirks of Markdown–HTML interaction: https://github.com/rust-lang/rustc-dev-guide/pull/1067#discussion_r579744499

jyn514 (approved):
Hmm, ok - can you file a bug upstream with pulldown?

camelid:
> Hmm, ok - can you file a bug upstream with pulldown?

Given that GitHub also renders it weirdly, it may not be a bug.

PR #1277, opened 2021-12-20
remove rustfix item in test intro

wangkirin:
since  [https://github.com/rust-lang/rust/pull/88387 ](https://github.com/rust-lang/rust/pull/88387)    , version `1.56.0` ,  the `rustfix` test suite have been removed , this PR fix it to avoid confusion


JohnTitor (approved):
Good call, thanks!

PR #1287, opened 2022-01-15
Clean up section about FCPs

camelid:
* I've never heard the term "pFCP" used before, so spell it out as
  "proposed final comment period"

* Fix some unclear things (like implying that you need `r+` rights to
  start an FCP)

* Improve the chapter's title, and make it consistent with its TOC entry

* Make some other assorted minor cleanups and improvements

### DIFF
--- a/src/implementing_new_features.md
+++ b/src/implementing_new_features.md
@@ -1,43 +1,42 @@
-# Implement New Feature
+# Implementing new features
 
 When you want to implement a new significant feature in the compiler,
 you need to go through this process to make sure everything goes
 smoothly.
 
-## The @rfcbot (p)FCP process
+## The @rfcbot FCP process
 
 When the change is small and uncontroversial, then it can be done
-with just writing a PR and getting r+ from someone who knows that
+with just writing a PR and getting an r+ from someone who knows that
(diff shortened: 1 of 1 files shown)


src/implementing_new_features.md
```diff
@@ -79,7 +78,6 @@ feature can be changed, or the feature might be completely
 rewritten or removed. Features are not supposed to gain tenure
 by being unstable and unchanged for a year.
 
-<a name = "tracking-issue"></a>
```
camelid, line 82: Removing this anchor will technically break links, but it seems unlikely that people are linking to this specific section that much.

src/implementing_new_features.md
```diff
@@ -111,7 +109,6 @@ for stabilization in a checklist, e.g.,
 - [ ] Stabilize the feature. ([See instructions on rustc-dev-guide.](https://rustc-dev-guide.rust-lang.org/stabilization_guide.html#stabilization-pr))
 ```
 
-<a name="stability-in-code"></a>
```
camelid, line 114: Removing this shouldn't break anything because I believe the ID already matches the ID generated for the heading.

spastorino (approved):

PR #1296, opened 2022-02-06
method-lookup.md improvements

mattheww:
Here are some proposed improvements to `method-lookup.md`, based on reading the current implementation (mostly `/rustc_typeck/src/check/method/probe.rs`).


JohnTitor (approved):
LGTM, thanks!

PR #1298, opened 2022-02-09
Correction, building stage3 compiler

raoulstrackx:
The [bootstrapping chapter](https://github.com/rust-lang/rustc-dev-guide/blame/master/src/building/bootstrapping.md#L185) states:

> `build --stage N compiler/rustc` **does not** build the stage N compiler:
> instead it builds the stage N+1 compiler _using_ the stage N compiler.

That is in conflict with the "How to build and run" chapter that states:

> `./x.py build --stage 2` – builds the stage2 compiler

I believe the former is correct.

### DIFF
--- a/src/building/how-to-build-and-run.md
+++ b/src/building/how-to-build-and-run.md
@@ -241,7 +241,7 @@ in other sections:
 - Building things:
   - `./x.py build` – builds everything using the stage 1 compiler,
     not just up to `std`
-  - `./x.py build --stage 2` – builds the stage2 compiler, along with `std` and
+  - `./x.py build --stage 2` – builds everything with the stage 2 compiler including
     `rustdoc` (which doesn't take too long)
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
By default, --stage 2 builds only a stage 2 compiler. Without full-bootstrap = true, the artifacts that *would* be a stage3 compiler are merely duplicates of the stage 2 compiler. Since there's essentially no reason typically to think about 'stage 3', I think it's likely more confusing for the documentation to reference it in this one place than to be mildly inconsistent with a different section.

raoulstrackx:
I see your point in trying to avoid references to stage 3. What would you think of rephrasing it as:

> `./x.py build --stage 2` – builds everything with the stage 2 compiler including `rustdoc` (which doesn't take too long)

It doesn't reference the stage 3 compiler, is easier to relate to the previous command one line up, and is completely correct.

PR #1301, opened 2022-02-17
Edit "Queries" chapter

pierwill:


src/query.md
```diff
@@ -1,58 +1,56 @@
...
+<!-- toc -->
+
 As described in [the high-level overview of the compiler][hl], the Rust compiler
 is still (as of <!-- date: 2021-07 --> July 2021) transitioning from a
-traditional "pass-based" setup to a "demand-driven" system. **The Compiler Query
-System is the key to our new demand-driven organization.** The idea is pretty
-simple. You have various queries that compute things about the input – for
-example, there is a query called `type_of(def_id)` that, given the [def-id] of
+traditional "pass-based" setup to a "demand-driven" system. The compiler query
+system is the key to rustc's demand-driven organization.
```
camelid, line 8: I think it might still make sense to have this in bold.

src/query.md
```diff
@@ -1,58 +1,56 @@
...
+The idea is pretty simple. Instead of entirely independent passes
+(parsing, type-checking, etc.), various smaller *queries*
```
camelid, line 10:
```suggestion
(parsing, type-checking, etc.), a set of function-like *queries*
```

src/query.md
```diff
@@ -1,58 +1,56 @@
...
+compute information about the input source. For example,
+there is a query called `type_of(def_id)` that, given the [`DefId`] of
```
camelid, line 12:
```suggestion
there is a query called `type_of` that, given the [`DefId`] of
```

src/query.md
```diff
@@ -1,58 +1,56 @@
...
 some item, will compute the type of that item and return it to you.
 
-[def-id]: appendix/glossary.md#def-id
+[`DefId`]: appendix/glossary.md#def-id
```
camelid, line 15: I think we should link to the docs here; that's what we usually do.

src/query.md
```diff
@@ -308,3 +303,12 @@ More discussion and issues:
 [GitHub issue #42633]: https://github.com/rust-lang/rust/issues/42633
 [Incremental Compilation Beta]: https://internals.rust-lang.org/t/incremental-compilation-beta/4721
 [Incremental Compilation Announcement]: https://blog.rust-lang.org/2016/09/08/incremental.html
+
+[^1]: The ["Incremental Compilation in Detail](queries/incremental-compilation-in-detail.md) chapter gives a more
+in-depth description of what queries are and how they work.
+If you intend to write a query of your own, this is a good read.
```
camelid, line 309: Can you move the footnote to near the reference site and rename `^1` to something explanatory? And the same for the second footnote.

src/query.md
```diff
@@ -1,58 +1,56 @@
...
 [hl]: ./compiler-src.md
 
-Query execution is **memoized** – so the first time you invoke a
+Query execution is *memoized*. The first time you invoke a
 query, it will go do the computation, but the next time, the result is
 returned from a hashtable. Moreover, query execution fits nicely into
-**incremental computation**; the idea is roughly that, when you do a
-query, the result **may** be returned to you by loading stored data
-from disk (but that's a separate topic we won't discuss further here).
+*incremental computation*; the idea is roughly that, when you do a
```
camelid, line 21:
```suggestion
*incremental computation*; the idea is roughly that, when you invoke a
```

src/query.md
```diff
@@ -1,58 +1,56 @@
...
-compiler (for example, generating MIR) work exactly like this.
-
-### Incremental Compilation in Detail
-
-The [Incremental Compilation in Detail][query-model] chapter gives a more
-in-depth description of what queries are and how they work.
-If you intend to write a query of your own, this is a good read.
-
-[query-model]: queries/incremental-compilation-in-detail.md
+Although this vision is not fully realized, large sections of
```
camelid, line 39:
```suggestion
Although this vision is not fully realized, large sections of the
```

src/query.md
```diff
@@ -1,58 +1,56 @@
...
+compiler (for example, generating [MIR](mir/)) currently work exactly like this.
```
camelid, line 40:
```suggestion
compiler (for example, generating [MIR](./mir/)) currently work exactly like this.
```

src/query.md
```diff
@@ -1,58 +1,56 @@
...
 
 ### Invoking queries
 
-To invoke a query is simple. The tcx ("type context") offers a method
-for each defined query. So, for example, to invoke the `type_of`
+Invoking a query is simple. The [`TyTcx`] ("type context") struct offers a method
```
camelid, line 44: You'll need to update the link ref name too.

```suggestion
Invoking a query is simple. The [`TyCtxt`] ("type context") struct offers a method
```

src/query.md
```diff
@@ -162,13 +160,13 @@ they define both a `provide` and a `provide_extern` function, through
 [rustc_metadata]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_metadata/index.html
 [wasm_import_module_map]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_codegen_ssa/back/symbol_export/fn.wasm_import_module_map.html
 
-### Adding a new kind of query
+### Adding a new query
 
-So suppose you want to add a new kind of query, how do you do so?
-Well, defining a query takes place in two steps:
+Suppose you want to add a new query: how do you do so?
```
camelid, line 165:
```suggestion
How do you add a new query?
```

src/query.md
```diff
@@ -190,8 +188,9 @@ rustc_queries! {
...
 
 Queries are grouped into categories (`Other`, `Codegen`, `TypeChecking`, etc.).
-Each group contains one or more queries. Each query definition is broken up like
-this:
+Each group contains one or more queries.
+
+A query definition has the following form:
 
 ```rust,ignore
 query type_of(key: DefId) -> Ty<'tcx> { ... }
```
camelid, line 196: The `^^` for "query keyword" should point to the whole word.

src/query.md
```diff
@@ -236,9 +231,9 @@ So, to add a query:
 
 #### Query structs and descriptions
 
-For each kind, the `rustc_queries` macro will generate a "query struct"
+For each query, the `rustc_queries` macro will generate a "query struct"
 named after the query. This struct is a kind of a place-holder
```
camelid, line 235:
```suggestion
named after the query. This struct is a kind of placeholder
```

camelid (approved):
Thanks for doing this! Sorry for leaving so many comments...

src/query.md
pierwill: I generally don't like using bold face for emphasis in regular prose. Let me try something else here...

pierwill: Actually, I'm thinking of leaving this as-is, if that's alright. 😇

pierwill:
Thanks for reviewing, @camelid!

src/query.md
camelid: Eh, not a big deal either way. I slightly lean toward bold in this case, but it's fine that you left it out.

PR #1302, opened 2022-02-17
Edit glossary

pierwill:
Adds back-ticks to a few terms that appear in code.

Clarify that `tcx` is a convention.

### DIFF
--- a/src/appendix/glossary.md
+++ b/src/appendix/glossary.md
@@ -75,7 +75,7 @@ Term                                                  | Meaning
 <span id="substs">substs</span>                &nbsp; |  The substitutions for a given generic type or item (e.g. the `i32`, `u32` in `HashMap<i32, u32>`).
 <span id="sysroot">sysroot</span>              &nbsp; |  The directory for build artifacts that are loaded by the compiler at runtime. ([see more](../building/bootstrapping.html#what-is-a-sysroot))
 <span id="tag">tag</span>                      &nbsp; |  The "tag" of an enum/generator encodes the [discriminant](#discriminant) of the active variant/state.  Tags can either be "direct" (simply storing the discriminant in a field) or use a ["niche"](#niche).
-<span id="tcx">`tcx`</span>                    &nbsp; |  Conventional variable name for the "typing context" (`TyCtxt`), main data structure of the compiler. ([see more](../ty.md))
+<span id="tcx">`tcx`</span>                    &nbsp; |  Standard variable name for the "typing context" (`TyCtxt`), main data structure of the compiler. ([see more](../ty.md))
 <span id="lifetime-tcx">`'tcx`</span>          &nbsp; |  The lifetime of the allocation arenas used by `TyCtxt`. Most data interned during a compilation session will use this lifetime with the exception of HIR data which uses the `'hir` lifetime. ([see more](../ty.md))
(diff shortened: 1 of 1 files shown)


src/appendix/glossary.md
```diff
@@ -75,14 +75,14 @@ Term                                                  | Meaning
 <span id="substs">substs</span>                &nbsp; |  The substitutions for a given generic type or item (e.g. the `i32`, `u32` in `HashMap<i32, u32>`).
 <span id="sysroot">sysroot</span>              &nbsp; |  The directory for build artifacts that are loaded by the compiler at runtime. ([see more](../building/bootstrapping.html#what-is-a-sysroot))
 <span id="tag">tag</span>                      &nbsp; |  The "tag" of an enum/generator encodes the [discriminant](#discriminant) of the active variant/state.  Tags can either be "direct" (simply storing the discriminant in a field) or use a ["niche"](#niche).
-<span id="tcx">tcx</span>                      &nbsp; |  The "typing context" (`TyCtxt`), main data structure of the compiler. ([see more](../ty.md))
+<span id="tcx">`tcx`</span>                    &nbsp; |  Commonly used variable name for the "typing context" (`TyCtxt`), main data structure of the compiler. ([see more](../ty.md))
```
camelid, line 78: "Commonly used" makes it sound like you can use other names too. You really should use tcx always, so while it's not automatically enforced, it's pretty much a rule.

camelid:
Have you checked that these backticks render correctly? I think you'll need to use `<code>`.

pierwill:
> Have you checked that these backticks render correctly? I think you'll need to use <code> HTML tags.

They worked correctly locally.

src/appendix/glossary.md
```diff
@@ -75,14 +75,14 @@ Term                                                  | Meaning
 <span id="substs">substs</span>                &nbsp; |  The substitutions for a given generic type or item (e.g. the `i32`, `u32` in `HashMap<i32, u32>`).
 <span id="sysroot">sysroot</span>              &nbsp; |  The directory for build artifacts that are loaded by the compiler at runtime. ([see more](../building/bootstrapping.html#what-is-a-sysroot))
 <span id="tag">tag</span>                      &nbsp; |  The "tag" of an enum/generator encodes the [discriminant](#discriminant) of the active variant/state.  Tags can either be "direct" (simply storing the discriminant in a field) or use a ["niche"](#niche).
-<span id="tcx">tcx</span>                      &nbsp; |  The "typing context" (`TyCtxt`), main data structure of the compiler. ([see more](../ty.md))
+<span id="tcx">`tcx`</span>                    &nbsp; |  Conventional variable name for the "typing context" (`TyCtxt`), main data structure of the compiler. ([see more](../ty.md))
```
camelid, line 78: Maybe too much of a nitpick, but what about "standard"?

pierwill: Maybe "Standard name" instead of "variable name"? (It's a common field name as well...)

camelid: "Standard name" is confusing because it sounds like it means the name of the type too (which is in fact `TyCtxt`). I think "variable name" can cover field name as well. After all, usually it's not a **vari**able, but an immutable binding anyway ;)

camelid (approved):

PR #1307, opened 2022-02-17
Edit the "Compiler Source Code" chapter

pierwill:
Makes various edits for clarity, style, readability, and formatting.

### DIFF
--- a/src/compiler-src.md
+++ b/src/compiler-src.md
@@ -96,7 +96,7 @@ by the whole compiler (e.g. [`rustc_span`]). The very early parts of the
 compilation process (e.g. parsing and the AST) depend on only these.
 
 After the AST is constructed, the compiler's [query system][query]
-gets set up.The query system is set up in a clever way using function
+gets set up. The query system is set up in a clever way using function
 pointers. This allows us to break dependencies between crates, allowing more
 parallel compilation.
 The query system is defined in [`rustc_middle`], so nearly all
--- a/src/compiler-src.md
+++ b/src/compiler-src.md
@@ -96,7 +96,9 @@ by the whole compiler (e.g. [`rustc_span`]). The very early parts of the
 compilation process (e.g. parsing and the AST) depend on only these.
 
 After the AST is constructed, the compiler's [query system][query]
-gets set up.[^query-setup]
+gets set up.The query system is set up in a clever way using function
+pointers. This allows us to break dependencies between crates, allowing more
+parallel compilation.
 The query system is defined in [`rustc_middle`], so nearly all
 subsequent parts of the compiler depend on this crate. It is a really large
 crate, leading to long compile times. Some efforts have been made to move stuff
@@ -124,10 +126,6 @@ compilation to completion.
 [query]: ./query.md
 
 [orgch]: ./overview.md
-
-[^query-setup]: The query system is set up in a clever way using function
-pointers. This allows us to break dependencies between crates, allowing more
-parallel compilation.
 
 ## rustdoc
 
--- a/src/compiler-src.md
+++ b/src/compiler-src.md
@@ -3,7 +3,10 @@
 <!-- toc -->
 
 Now that we have [seen what the compiler does](./overview.md), let's take a
-look at the structure of the contents of the rust-lang/rust repo.
+look at the structure of the [`rust-lang/rust`] repository, where the rustc
+source code lives.
+
+[`rust-lang/rust`]: https://github.com/rust-lang/rust
 
 ## Workspace structure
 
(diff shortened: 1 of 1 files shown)


src/compiler-src.md
```diff
@@ -16,22 +19,11 @@ The repository consists of three main directories:
...
   that together make up the compiler.
-
 - `library/` contains the standard libraries (`core`, `alloc`, `std`,
   `proc_macro`, `test`), as well as the Rust runtime (`backtrace`, `rtstartup`,
   `lang_start`).
-
 - `src/` contains the source code for rustdoc, clippy, cargo, the build system,
-  language docs, etc.
-
-## Standard library
```
pierwill, line 27: My feeling is that this section isn't needed, but I'm happy to keep it if people want.

src/compiler-src.md
```diff
@@ -16,22 +19,11 @@ The repository consists of three main directories:
 
 - `compiler/` contains the source code for `rustc`. It consists of many crates
   that together make up the compiler.
-
```
camelid, line 19: These blank lines add more space between the list elements, which I think makes this list easier to read. Can you re-add them?

src/compiler-src.md
```diff
@@ -16,22 +19,11 @@ The repository consists of three main directories:
...
-## Standard library
-
-The standard library crates are all in `library/`. They have intuitive names
-like `std`, `core`, `alloc`, etc.  There is also `proc_macro`, `test`, and
-other runtime libraries.
-
-This code is fairly similar to most other Rust crates except that it must be
-built in a special way because it can use unstable features.
+  compiler tests, language docs, etc.
```
camelid, line 26: That was a big omission 😅

src/compiler-src.md
camelid: I'd rather keep it.

src/compiler-src.md
```diff
@@ -101,12 +93,9 @@ At the very bottom of the dependency tree are a handful of crates that are used
 by the whole compiler (e.g. [`rustc_span`]). The very early parts of the
 compilation process (e.g. parsing and the AST) depend on only these.
 
-Pretty soon after the AST is constructed, the compiler's [query system][query]
-gets set up.  The query system is set up in a clever way using function
-pointers. This allows us to break dependencies between crates, allowing more
-parallel compilation.
-
-However, since the query system is defined in [`rustc_middle`], nearly all
+After the AST is constructed, the compiler's [query system][query]
```
camelid, line 96: Some things happen between AST construction and query system setup I think.

src/compiler-src.md
```diff
@@ -101,12 +93,9 @@ At the very bottom of the dependency tree are a handful of crates that are used
...
+gets set up.[^query-setup]
```
camelid, line 97: IMO too many footnotes is more distracting than just having the text inline.

src/compiler-src.md
pierwill: Moved this below, the `##` headings are in the same order as the list.

src/compiler-src.md
pierwill: Too many footnotes is certainly distracting! In this case, I didn't think this implementation detail belongs in the main body text of a "Big picture" section.

src/compiler-src.md
pierwill: 👍 Moved this below, so the ## headings are in the same order as the list.

src/compiler-src.md
camelid: I agree in some ways, but the implementation detail informs the code structure, so it still seems potentially like it belongs.

src/compiler-src.md
pierwill: Can we address missing content in a follow-up?

camelid: Sure, but I'd rather you revert the wording change since it now sounds like it's asserting that nothing happens in between.

src/compiler-src.md
```diff
@@ -101,12 +95,11 @@ At the very bottom of the dependency tree are a handful of crates that are used
 by the whole compiler (e.g. [`rustc_span`]). The very early parts of the
 compilation process (e.g. parsing and the AST) depend on only these.
 
-Pretty soon after the AST is constructed, the compiler's [query system][query]
-gets set up.  The query system is set up in a clever way using function
+After the AST is constructed, the compiler's [query system][query]
+gets set up.The query system is set up in a clever way using function
```
camelid, line 99:
```suggestion
gets set up. The query system is set up in a clever way using function
```

src/compiler-src.md
spastorino: @pierwill can you address this comment?

spastorino:
@pierwill should we merge this after addressing [@camelid ](https://github.com/rust-lang/rustc-dev-guide/pull/1307#discussion_r809543217)

src/compiler-src.md
pierwill: Maybe 

```suggestion
After the AST is constructed and other early analysis is done, the compiler's [query system][query]
```

pierwill:
> @pierwill should we merge this after addressing [#1307 (comment)](https://github.com/rust-lang/rustc-dev-guide/pull/1307#discussion_r809543217)

That'd be great! 🙏 Offered a suggestion there.

src/compiler-src.md
camelid: Seems good enough to me 👍

spastorino (approved):
Looks good to me r? @camelid

JohnTitor (approved):
Applied the suggestion on https://github.com/rust-lang/rustc-dev-guide/pull/1307#discussion_r856503052.
I'm going to merge, @camelid if you have any further comments, let us know!

PR #1309, opened 2022-02-20
Replace "Shell" highlighting with "Bash" highlighting

steffahn:
(compare https://highlightjs.readthedocs.io/en/latest/supported-languages.html for a list of languages and their identifiers/aliases)

Motivation: This code block was mishighlighted: https://rustc-dev-guide.rust-lang.org/git.html#i-see--head

### DIFF
--- a/src/closure.md
+++ b/src/closure.md
@@ -34,7 +34,7 @@ fn main() {
 Let's say the above is the content of a file called `immut.rs`. If we compile
 `immut.rs` using the following command. The [`-Z dump-mir=all`][dump-mir] flag will cause
 `rustc` to generate and dump the [MIR][mir] to a directory called `mir_dump`.
-```console
+```sh
 > rustc +stage1 immut.rs -Z dump-mir=all
 ```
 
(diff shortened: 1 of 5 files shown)


src/closure.md
```diff
@@ -34,7 +34,7 @@ fn main() {
 Let's say the above is the content of a file called `immut.rs`. If we compile
 `immut.rs` using the following command. The [`-Z dump-mir=all`][dump-mir] flag will cause
 `rustc` to generate and dump the [MIR][mir] to a directory called `mir_dump`.
-```console
+```sh
```
camelid, line 37: I believe `console` is correct for all the ones that start with `>` or `$`.

src/diagnostics.md
```diff
@@ -423,7 +423,7 @@ err.emit();
 
 This might emit an error like
 
-```console
+```sh
```
camelid, line 426:
```suggestion
```text
```

camelid (commented):
Most of these changes seem wrong to me. If there are prompts (`>` or `$`) in the code blocks, they should use `console`. If they don't have prompts, I think `` ```text `` is the most correct unless they have shell syntax like `#`.

steffahn:
> If there are prompts (`>` or `$`) in the code blocks, they should use `console`

Oh… I didn’t look close enough to notice that the `>` or `$` are actually highlighted (hard to spot in ayu). I thought there’d be no benefit at all of using `shell` over `bash` highlighting.

I guess I’ll have to actually read through these then, instead of just proposing a simple "find and replace" commit. Have got no time for that anymore today, I’ll come back to this PR on a later date.

JohnTitor:
@steffahn Ping, do you have some time to work on now?

JohnTitor:
Closing for inactivity, feel free to open a new PR when you have some time to work on, thanks!

PR #1311, opened 2022-02-20
Add known-bug header.

ehuss:
The `known-bug` header was added in https://github.com/rust-lang/rust/pull/93953.

### DIFF
--- a/src/tests/ui.md
+++ b/src/tests/ui.md
@@ -398,15 +398,15 @@ test won't work properly with that override.
 
 The `known-bug` header may be used for tests that demonstrate a known bug that
 has not yet been fixed.
-Adding tests for known bugs is helpful for several reasons:
+Adding tests for known bugs is helpful for several reasons, including:
 
-1. Maintains a functional test that can be conveniently reused when the bug is fixed.
-2. Provides a sentinel that will fail if the bug is incidentally fixed.
+1. Maintaining a functional test that can be conveniently reused when the bug is fixed.
+2. Providing a sentinel that will fail if the bug is incidentally fixed.
    This can alert the developer so they know that the associated issue has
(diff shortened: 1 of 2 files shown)


src/tests/ui.md
```diff
@@ -394,6 +394,21 @@ The `ignore-pass` header can be used to ignore the `--pass` CLI flag if the
...
+## Known bugs
+
+The `known-bug` header may be used for tests that demonstrate a known bug that
+has not yet been fixed.
+Adding tests for known bugs is helpful for several reasons:
+
+1. Maintains a functional test that can be conveniently reused when the bug is fixed.
+2. Provides a sentinel that will fail if the bug is incidentally fixed.
+   This can alert the developer so they know that the associated issue has
+   been fixed and can possibly be closed.
```
pierwill, line 406:
```suggestion
Adding tests for known bugs is helpful for several reasons, including:

1. Maintaining a functional test that can be conveniently reused when the bug is fixed.
2. Providing a sentinel that will fail if the bug is incidentally fixed.
   This can alert the developer so they know that the associated issue has
   been fixed and can possibly be closed.
```

src/tests/ui.md
```diff
@@ -394,6 +394,21 @@ The `ignore-pass` header can be used to ignore the `--pass` CLI flag if the
...
+
+Do not include [error annotations](#error-annotations) in a test with `known-bug`.
+Otherwise, the test should include the normal headers and stdout/stderr files.
```
pierwill, line 409:
```suggestion
The test should still include other normal headers and stdout/stderr files.
```

pierwill (approved):
Looks good! Just a few suggestions. Maybe we could link to an example as well?

PR #1312, opened 2022-02-20
Move epigraph to beginning of section

pierwill:
It's conventional to have quotations like this at the beginning of the section they relate to.


camelid (commented):
I'm unsure if the quote is for this section or for the whole page. Also, it's perhaps intended to be a parting thought rather than an epigraph.

pierwill:
Thanks @camelid. I'll leave this open for a bit in case others have thoughts on this.

PR #1314, opened 2022-02-23
Fix typo "memoized" → "memorized"

WowbaggersLiquidLunch:


pierwill:
Thanks for the PR @WowbaggersLiquidLunch. I believe "memoized" is the correct term here, so I'll close this PR. See https://en.wikipedia.org/wiki/Memoization.

PR #1315, opened 2022-02-24
Fix typo

teobouvard:


src/building/bootstrapping.md
```diff
@@ -218,7 +218,7 @@ Stage N `std` is pretty much necessary for any useful work with the stage N comp
 Without it, you can only compile programs with `#![no_core]` -- not terribly useful!
 
 The reason these need to be different is because they aren't necessarily ABI-compatible:
-there could be a new layout optimizations, changes to MIR, or other changes
+there could be new layout optimizations, changes to the MIR, or other changes
```
camelid, line 221: I think it's correct without the "the".

teobouvard: I was originally only fixing the "a", but found that adding "the" in the following sounded nicer. As you prefer !

camelid: Not a big deal either way, but I think it's slightly more standard to do it without the "the".

teobouvard: I'll remove it then

camelid (approved):
Thanks!

PR #1317, opened 2022-02-26
Send people doing *library* stabilizations over to the std-dev-guide

scottmcm:
Because sending people doing library stabilization to `feature_gate.rs` will just result in confusion.

### DIFF
--- a/src/stabilization_guide.md
+++ b/src/stabilization_guide.md
@@ -84,6 +84,9 @@ decision is to stabilize, we proceed to do the actual code modification.
 
 ## Stabilization PR
 
+*This is for stabilizing language features.  If you are stabilizing a library
+feature, see [the stabilization chapter of the std dev guide][std-guide-stabilization] instead.*
+
 Once we have decided to stabilize a feature, we need to have
 a PR that actually makes that stabilization happen. These kinds
 of PRs are a great way to get involved in Rust, as they take
@@ -95,6 +98,24 @@ require steps beyond what this guide talks about.
 
 Note: Before we stabilize any feature, it's the rule that it
 should appear in the documentation.
+
+### Determining the stabilization version
+
+The version in which the feature will be stabilized *must* match
+the value of [the `src/version` file in `master`][src-version] when the PR is merged.
+
+It's worth checking [the version schedule on the Forge][forge-versions] to see whether
+changes are coming soon.  You'll usually use the version labelled "Nightly".
+That's two versions higher than the current stable release,
+as what's currently in beta will be the next stable release,
+and any change you're making now will be in the one after that.
+
+No PR is merged instantly, so you'll want to be careful around release time.
+The version bump happens [the Friday before][forge-release-process] the stable release,
+not the same time as the release.  So if you're opening a PR shortly before then,
+be prepared to update the version, or consider just opening it for one version
+higher than the current nightly, with a note saying not to merge until
+after the upcoming version bump.
 
 ### Updating the feature-gate listing
 
@@ -119,14 +140,7 @@ When it is done, it should look like:
 // note that we changed this
 ```
 
-Note that, the version number is updated to be the version number
-of the stable release where this feature will appear. This can be
-found by consulting [the forge](https://forge.rust-lang.org/), which will guide
-you the next stable release number. You want to add 1 to that,
-because the code that lands today will become go into beta on that
-date, and then become stable after that. So, at the time of this
-writing, the next stable release (i.e. what is currently beta) was
-1.30.0, hence I wrote 1.31.0 above.
+(The version here is the one discussed in the previous section.)
 
 ### Removing existing uses of the feature-gate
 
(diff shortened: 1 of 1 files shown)


src/stabilization_guide.md
```diff
@@ -183,6 +183,73 @@ if self.tcx.sess.features.borrow().pub_restricted && something { /* XXX */ }
...
+You'll need to change that to a `#[stable]` attribute with a version:
+
+```rust,ignore
+#[stable(feature = "total_cmp", since = "1.61.0")]
+```
+
+Note that, the version number is updated to be the version number
+of the stable release where this feature will appear. This can be
+found by consulting [the forge](https://forge.rust-lang.org/#current-release-versions).
+Specifically, you'll want to use the version labelled "Nightly".
```
ehuss, line 215: Although this is a perfectly fine way to check the version, it might be more reliable to consult [`src/version`](https://github.com/rust-lang/rust/blob/master/src/version) on master.  That can help avoid confusion with the version bump that happens on the Friday before release.

I might also suggest *predicting* the version that it will land in.  If it looks like it might land around the Friday before release, then there is risk that it will miss the release window.

Finally, I might suggest putting this discussion of determining the version into a dedicated section and removing the duplicate discussion above with lang stabilizations. For the most part, the process is the same between them. (And the lang one is not written as clearly.)

It is not uncommon that PRs are approved with the wrong version in them, and then there is confusion if it isn't caught, or triggers busywork for backports and such.  It would be nice to be a little more careful about that.

ehuss (commented):
I'm sometimes a little confused why some things are documented here and some are in the [std dev guide](https://std-dev-guide.rust-lang.org/feature-lifecycle/stabilization.html).  Would it make sense to try to avoid the duplication?  What are the guidelines for what is here vs there?

scottmcm:
> I'm sometimes a little confused why some things are documented here and some are in the [std dev guide](https://std-dev-guide.rust-lang.org/feature-lifecycle/stabilization.html).

Oh.  Well, that would be because I didn't know the std dev guide existed, so looked in the wrong place 😥

I'll change this to just have a pointer to that, and send a PR over there with the examples.

scottmcm:
Ok, I've re-done the update here.

PR to std-dev-guide to happen in future.

src/stabilization_guide.md
```diff
@@ -96,6 +99,24 @@ require steps beyond what this guide talks about.
...
 
+### Determining the stabilization version
+
+The version in which the feature will be stabilized *must* match
+the value of [the `src/version` file in `master`] when the PR is merged.
+
+It's worth checking [the version schedule on the Forge] to see whether
+changes are coming soon.  You'll usually use the version labelled "Nightly".
+That's two versions higher than the current stable release,
+as what's currently in beta will be the next stable release,
```
pierwill, line 110:
```suggestion
"Nightly" is two versions higher than the current stable release,
since what's currently in beta will be the next stable release,
```

src/stabilization_guide.md
```diff
@@ -96,6 +99,24 @@ require steps beyond what this guide talks about.
...
+and any change you're making now will be in the one after that.
```
pierwill, line 111:
```suggestion
and any changes you're making now will be in the one after that.
```

scottmcm:
Ping for a review here, please.

https://github.com/rust-lang/std-dev-guide/pull/32 is open to add good guidance there, so it'd be nice to help it be found.

src/stabilization_guide.md
```diff
@@ -184,6 +198,10 @@ if something { /* XXX */ }
 ```
 
 [rust-lang/rust#32409]: https://github.com/rust-lang/rust/issues/32409
+[the stabilization chapter of the std dev guide]: https://std-dev-guide.rust-lang.org/feature-lifecycle/stabilization.html
+[the `src/version` file in `master`]: https://github.com/rust-lang/rust/blob/master/src/version
+[the version schedule on the Forge]: https://forge.rust-lang.org/#current-release-versions
+[the Friday before]: https://forge.rust-lang.org/release/process.html
```
camelid, line 204: These link reference names (e.g., `the stabilization chapter of the std dev guide`) are long and prone to breaking if wording is changed. Could you change them to be pithier?

E.g.: `std-guide-stabilization`, `src-version`, `release-process`.

(You'll have to add manual `[...]` link references to the text too.)

scottmcm: Good idea.  I've added pithy names to these, and added the `(...)` link references.

(The `[...]` was a typo, right?  My syntax highlighter thinks this is right.)

camelid: No, it should be `[...]`. For example, `[look at this](src-version)` will link to `src-version` as if it is a URL. On the other hand, `[look at this][src-version]` will use the URL value of the definition `[src-version]: https://example.com` as the link.

scottmcm: Oh!  I see; thank you.  Fixed.

scottmcm: And I checked the rendered <https://github.com/rust-lang/rustc-dev-guide/blob/library-stabilization/src/stabilization_guide.md> this time to confirm, which I should have done last time but didn't think of.

JohnTitor (approved):
LGTM, thanks!

PR #1328, opened 2022-03-12
fix winget install instructions installing wrong packages

### LINKED ISSUE #1324: Windows prerequisite instructions install incorrect Python package
I'm working on contributing to rust for the first time and was following the instructions here: https://rustc-dev-guide.rust-lang.org/building/prerequisites.html

This says to use "winget install python", but this will install "Introduction to Python Programming by GoLearningBus". I think the correct instruction should be "winget install Python.Python.3".

I assume this got messed up when the msstore source was added, and winget is now just picking an arbitrary python package. In light of that, it might be appropriate to change the cmake instruction to "winget install Kitware.CMake" as well.

skippy10110:
update winget install instructions to ensure proper packages are installed (-e for --exact, and full package names to ensure arbitrary packages from the msstore source aren't installed)

fixes #1324

### DIFF
--- a/src/building/prerequisites.md
+++ b/src/building/prerequisites.md
@@ -31,8 +31,8 @@ on Windows.
 Run the following in a terminal:
 
 ```powershell
-winget install python
-winget install cmake
+winget install -e Python.Python.3
+winget install -e Kitware.CMake
 ```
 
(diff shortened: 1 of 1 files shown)


wesleywiser (approved):
LGTM!

PR #1329, opened 2022-03-21
update section for type system constants

lcnr:
after reverting a part of https://github.com/rust-lang/rust/pull/87280 in https://github.com/rust-lang/rust/pull/92805 this was out of date.

Also generally extended that section to be a bit more helpful hopefully.

### DIFF
--- a/src/constants.md
+++ b/src/constants.md
@@ -16,6 +16,32 @@ or a generic parameter, e.g. `[u8; N]`, converting a constant to its [`ty::Const
 returns an unevaluated constant. Even fully concrete constants which do not depend on
 generic parameters are not evaluated right away.
 
+Anonymous constants are typechecked separately from their containing item, e.g.
+```rust
+fn foo<const N: usize>() -> [u8; N + 1] {
+    [0; N + 1]
+}
+```
+is treated as
+```rust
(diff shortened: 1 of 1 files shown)


src/constants.md
```diff
@@ -16,6 +16,32 @@ or a generic parameter, e.g. `[u8; N]`, converting a constant to its [`ty::Const
...
+}
+```
+
+### Unifying constants
+
+For the compiler, `ANON_CONST_1` and `ANON_CONST_2` are completely different, so
+we have to somehow look into unevaluated constants to check whether they should
+unify.
+
+For this we use [InferCtxt::try_unify_abstract_consts](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_infer/infer/struct.InferCtxt.html#method.try_unify_abstract_consts). This builds a custom AST for the two inputs from their THIR. Thish is then used for
```
lcnr, line 40:
```suggestion
For this we use [InferCtxt::try_unify_abstract_consts](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_infer/infer/struct.InferCtxt.html#method.try_unify_abstract_consts).
This builds a custom AST for the two inputs from their THIR. This is then used for
```

PR #1332, opened 2022-04-01
Add example how lints can be feature gated

xFrednet:
Just a simple addition, like the title says.

I hope that everyone reading this is having a great day. Keep up the good work :upside_down_face:

### DIFF
--- a/src/diagnostics.md
+++ b/src/diagnostics.md
@@ -594,6 +594,20 @@ declare_lint! {
 This makes the `ANONYMOUS_PARAMETERS` lint allow-by-default in the 2015 edition
 but warn-by-default in the 2018 edition.
 
+### Feature-gated lints
+
+Lints belonging to a feature should only be usable if the feature is enabled in the
+crate. To support this, lint declarations can contain a feature gate like so:
+
+```rust,ignore
+declare_lint! {
+    pub SOME_LINT_NAME,
+    Warn,
+    "a new and useful, but feature gated lint",
+    @feature_gate = sym::feature_name;
+}
+```
+
 ### Future-incompatible lints
 
 The use of the term `future-incompatible` within the compiler has a slightly


src/diagnostics.md
```diff
@@ -594,6 +594,20 @@ declare_lint! {
...
+
+Lints belonging to a feature should only be usable if the feature is enabled in the
+crate. To support this, lint declarations can contain a feature gate like this:
+
+```rust,ignore
+declare_lint! {
+    pub SOME_LINT_NAME,
+    Warn,
+    "a new and useful, but feature gated lint",
+    @feature_gate = rustc_span::sym::feature_name;
```
camelid, line 607: `sym` is already imported.

```suggestion
    @feature_gate = sym::feature_name;
```

src/diagnostics.md
```diff
@@ -594,6 +594,20 @@ declare_lint! {
 This makes the `ANONYMOUS_PARAMETERS` lint allow-by-default in the 2015 edition
 but warn-by-default in the 2018 edition.
 
+### Feature-gated lints
+
+Lints belonging to a feature should only be usable if the feature is enabled in the
+crate. To support this, lint declarations can contain a feature gate like this:
```
camelid, line 600:
```suggestion
crate. To support this, lint declarations can contain a feature gate like so:
```

camelid (approved):
Thanks for this PR! Just two suggestions and then LGTM.

One other thing: I think sometimes lints belonging to a feature are feature-gated but in the past they were not. Is the compiler/lang team's policy to always feature-gate lints belonging to a feature now? We don't really have to decide now since lints can always be un-feature-gated later if necessary, but I wanted to bring it up.

src/diagnostics.md
xFrednet: `this` is one of the words I use way to often, nice improvement :upside_down_face:

xFrednet:
Thank you for the review, I've added your suggestions and squashed the commits.

> Is the compiler/lang team's policy to always feature-gate lints belonging to a feature now? We don't really have to decide now since lints can always be un-feature-gated later if necessary, but I wanted to bring it up.

I don't know of any policies in that regard, it most likely depends on the lint. For instance, I'm currently working on the implementation of the `expect` attribute that emits a lint under specific circumstances. There a feature gate should be used IMO, as the lint can and should only be emitted if the feature is enabled. With other lints, it will most likely be a different story :upside_down_face:

PR #1333, opened 2022-04-04
sessiondiagnostic: translation

davidtwco:
Updates documentation on `#[derive(SessionDiagnostic)]` now that it can be used for translatable diagnostics.

~~Waiting on rust-lang/rust#95512 to be merged and for me to add links to the source after it has.~~

### DIFF
--- a/src/diagnostics/sessiondiagnostic.md
+++ b/src/diagnostics/sessiondiagnostic.md
@@ -1,27 +1,150 @@
-# Creating Errors With SessionDiagnostic
+# Creating translatable errors using `SessionDiagnostic`
+The `SessionDiagnostic` derive macro is the recommended way to create
+diagnostics. Diagnostics created with the derive macro can be translated into
+different languages and each have a slug that uniquely identifies the
+diagnostic.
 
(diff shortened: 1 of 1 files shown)


src/diagnostics/sessiondiagnostic.md
```diff
@@ -1,97 +1,207 @@
...
+`typeck-field-already-declared.label` (which has the message "field already
+declared"). If there is more than one sub-diagnostic of the same type, then
+these attributes can also take a value that is the attribute name to look for
+(e.g. `previous-decl-label` in our example).
+
+`#[help]` and `#[note]` can also be applied to the struct itself, in which case
+they work exactly like when applied to fields except the sub-diagnostic won't
+have a `Span`.
+
+Any attribute can also applied to an `Option<Span>` and will only emit a
```
JohnTitor, line 102:
```suggestion
Any attribute can also be applied to an `Option<Span>` and will only emit a
```

JohnTitor (approved):
Thanks for writing up! LGTM with a nit.

src/diagnostics/sessiondiagnostic.md
davidtwco: Fixed!

PR #1335, opened 2022-04-07
r-a: Use `python3 x.py` instead of `./x.py`

camelid:
This circumvents an issue on newer OSes (including now macOS Monterery):
rust-lang/rust#71818

Thanks to jyn514 for suggesting this fix.


JohnTitor (approved):
Seems a better one, thanks!

PR #1337, opened 2022-04-13
Minor type fix

eholk:
The `'b` in `let y` was an unbound region. Presumably this was meant to be `'c`.


JohnTitor (approved):
Good point, thanks!

PR #1341, opened 2022-04-15
Fix documentation regarding suggestion for expanding proc macros

fee1-dead:
This has been fixed by https://github.com/rust-lang/rust-analyzer/pull/11956, but I am not sure if this should be suggested for everyone - including those only working on the library - but with the rustc wrapper the performance impact is minimal.

### DIFF
--- a/src/building/suggested.md
+++ b/src/building/suggested.md
@@ -38,8 +38,9 @@ you can write: <!-- date: 2022-04 --><!-- the date comment is for the edition be
     ],
     "rust-analyzer.procMacro.enable": true,
     "rust-analyzer.cargo.buildScripts.enable": true,
+    "rust-analyzer.cargo.buildScripts.useRustcWrapper": true,
     "rust-analyzer.cargo.buildScripts.overrideCommand": [
-        "cargo",
+        "./build/$TARGET_TRIPLE/stage0/bin/cargo",
         "check",
         "-p",
         "rustc_driver",
@@ -56,10 +57,10 @@ you can write: <!-- date: 2022-04 --><!-- the date comment is for the edition be
 in your `.vscode/settings.json` file. This will ask `rust-analyzer` to use
 `./x.py check` to check the sources, and the stage 0 rustfmt to format them.
 
-> NOTE: Make sure to replace `TARGET_TRIPLE` in the `rust-analyzer.rustfmt.overrideCommand`
-> setting with the appropriate target triple for your machine. An example of such
-> a triple is `x86_64-unknown-linux-gnu`. An easy way to check your target triple
-> is to run `rustc -vV` and checking the `host` value of its output.
+> NOTE: Make sure to replace `TARGET_TRIPLE` with the appropriate target triple
+> for your machine. An example of such a triple is `x86_64-unknown-linux-gnu`.
+> An easy way to check your target triple is to run `rustc -vV` and checking the
+> `host` value of its output.
 
 If you have enough free disk space and you would like to be able to run `x.py` commands while
 rust-analyzer runs in the background, you can also add `--build-dir build-rust-analyzer` to the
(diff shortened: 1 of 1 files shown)


spastorino:
I'd guess we may want someone from t-libs to weigh in, if they are not interested in this suggestion maybe we can explicitly suggest one path for t-libs contributors and this one for t-compiler contributors.

JohnTitor:
> but I am not sure if this should be suggested for everyone - including those only working on the library

I think we could also add a note about this for those who want to contribute to the libraries rather than compiler.

spastorino:
Should we merge this or what is it waiting on?

JohnTitor:
This has a merge conflict and I forgot to ping, @fee1-dead Could you resolve it?

spastorino:
From what I've seen in the source code of R-A, the current settings on the dev-guide are correct. We could just probably remove `"rust-analyzer.cargo.buildScripts.enable": true` and `"rust-analyzer.procMacro.enable": true` because those are true by default.

JohnTitor:
Sorry, I made another conflict by #1380, could you also order like below?

```json
{
    "rust-analyzer.checkOnSave.overrideCommand": [
        "python3",
        "x.py",
        "check",
        "--json-output"
    ],
    "rust-analyzer.rustfmt.overrideCommand": [
        "./build/$TARGET_TRIPLE/stage0/bin/rustfmt",
        "--edition=2021"
    ],
    "rust-analyzer.procMacro.enable": true,
    "rust-analyzer.cargo.buildScripts.enable": true,
    "rust-analyzer.cargo.buildScripts.overrideCommand": [
        "./build/$TARGET_TRIPLE/stage0/bin/cargo",
        "check",
        "-p",
        "rustc_driver",
        "--message-format=json"
    ],
    "rust-analyzer.cargo.useRustcWrapperForBuildScripts": true,
    "rust-analyzer.rustc.source": "./Cargo.toml"
}
```

spastorino:
@JohnTitor from what I see on R-A source code https://github.com/rust-lang/rust-analyzer/blob/master/crates/rust-analyzer/src/config.rs, `rust-analyzer.procMacro.enable` is true by default, `rust-analyzer.cargo.buildScripts.enable` also true by default and `rust-analyzer.cargo.useRustcWrapperForBuildScripts` is now `rust-analyzer.cargo.buildScripts.useRustcWrapper` but seems to be a rename in place, anyway it's true by default. So I'd remove those 3 from the suggestions.

fee1-dead:
> So I'd remove those 3 from the suggestions.

Those configuration are there so we can suggest to people in libs to disable them later and would make the instructions clearer.

src/building/suggested.md
```diff
@@ -70,6 +71,17 @@ If you're running `coc.nvim`, you can use `:CocLocalConfig` to create a
 `editor.formatOnSave: true,` with
 `"coc.preferences.formatOnSaveFiletypes": ["rust"],`.
 
+> NOTE: The configuration once suggested disabling build scripts, but after
+> the version `2022-04-14` this was fixed by introducing an option for customizing
+> the command rust-analyzer uses to run build scripts. Please ensure that `rust-analyzer`
+> is up to date before enabling these options. If you do not wish to enable
+> `rust-analyzer.cargo.useRustcWrapperForBuildScripts`, the
```
RalfJung, line 78: Above the option is called `rust-analyzer.cargo.buildScripts.useRustcWrapper`, but here it is called by a different name?

src/building/suggested.md
```diff
@@ -38,8 +38,9 @@ you can write: <!-- date: 2022-04 --><!-- the date comment is for the edition be
     ],
     "rust-analyzer.procMacro.enable": true,
     "rust-analyzer.cargo.buildScripts.enable": true,
+    "rust-analyzer.cargo.buildScripts.useRustcWrapper": true,
```
RalfJung, line 41: This seems to be set by default, why set it explicitly here?
(Same for the `enable` options.)

fee1-dead: Answered above

src/building/suggested.md
flodiebold: It got renamed, `rust-analyzer.cargo.buildScripts.useRustcWrapper` is the new name and `rust-analyzer.cargo.useRustcWrapperForBuildScripts` is the old name that would still work.

flodiebold: I think this sentence should be left out anyway.

src/building/suggested.md
RalfJung: I think it'd be better to not have these lines.
- If people have this explicitly globally disabled, we should respect that.
- For the libs suggestion, we can just have another code block saying "also add these things".

fee1-dead:
closing, as #1384 is better

PR #1330, opened 2022-03-30
Update examples with 1.61.0-nightly (latest version)

ujjawal4046:
Specifically this handles 2 changes in upstream : 

1. Remove Config::stderr - https://github.com/rust-lang/rust/commit/5eeff3f0733c83be9e0406874e92adf24a5cf2b6
2. Add Config::crate_check_cfg  - https://github.com/rust-lang/rust/commit/3a73ca587bb8a8fb52d6045fbe31d50d5a56ff19

PR #1345, opened 2022-05-01
Update rustc-driver related examples

JohnTitor:
Updates rustc-driver related examples to the latest nightly (I forgot to update the references on #1330 :P).
cc #1344

### DIFF
--- a/examples/rustc-driver-interacting-with-the-ast.rs
+++ b/examples/rustc-driver-interacting-with-the-ast.rs
@@ -1,9 +1,9 @@
 #![feature(rustc_private)]
 
 // NOTE: For the example to compile, you will need to first run the following:
-//     rustup component add rustc-dev llvm-tools-preview
+//   rustup component add rustc-dev llvm-tools-preview
 
-// version: 1.61.0-nightly (68369a041 2022-02-22)
+// version: 1.62.0-nightly (7c4b47696 2022-04-30)
 
 extern crate rustc_ast_pretty;
 extern crate rustc_error_codes;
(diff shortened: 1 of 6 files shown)


examples/rustc-driver-interacting-with-the-ast.rs
```diff
@@ -76,7 +77,7 @@ fn main() {
                                     let hir_id = expr.hir_id; // hir_id identifies the string "Hello, world!"
                                     let def_id = tcx.hir().local_def_id(item.hir_id()); // def_id identifies the main function
                                     let ty = tcx.typeck(def_id).node_type(hir_id);
-                                    println!("{:?}: {:?}", expr, ty); // prints expr(HirId { owner: DefIndex(3), local_id: 4 }: "Hello, world!"): &'static str
+                                    println!("{:?}: {:?}", expr, ty); // prints "`rustc_hir::Expr`: `rustc_middle::ty::Ty`"
```
jyn514, line 80: are you sure? that seems very strange for it to print, normally it gives details about the item itself

JohnTitor: It actually prints:
```
fn main() { let message = "Hello, world!"; println!("{}", message); }
Expr { hir_id: HirId { owner: DefId(0:3 ~ rust_out[a9df]::main), local_id: 2 }, kind: Lit(Spanned { node: Str("Hello, world!", Cooked), span: <main.rs>:1:27: 1:42 (#0) }), span: <main.rs>:1:27: 1:42 (#0) }: &str
```

Well, I remember I truncated the output because it's a bit long but... yeah it's quite different.
Any suggestions here? I feel it's verbose and would like to shorten if possible. Though I don't mind pasting it as-is if you think it's better :)

tshepang: I would rid of the comment

JohnTitor: Removed!

PR #1346, opened 2022-05-05
mention `WithOptConstParam`

lcnr:
Should probably go more in-depth here, but that isn't something I have the time for rn :sweat_smile: this should probably be merged with the explanation in https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.WithOptConstParam.html


JohnTitor:
I'd like to get some eyes from @rust-lang/project-const-generics, could someone review?

oli-obk:
Yea, this is minimal, but good that we at least mention it now.

We can expand on it later

JohnTitor (approved):
Thanks both for writing and reviewing! Rubber-stamping in favor of oli-obk's approval.

PR #1350, opened 2022-05-09
Add docs for logging of queries.

ehuss:
Support for automatic logging of all queries was recently added in https://github.com/rust-lang/rust/pull/96697.

Perhaps @oli-obk could review?

### DIFF
--- a/src/tracing.md
+++ b/src/tracing.md
@@ -60,6 +60,25 @@ still get a short message per ignored `do_mir_borrowck`, but none of the things 
 calls. This helps you in looking through the calls that are happening and helps you adjust
 your regex if you mistyped it.
 
+## Query level filters
+
+Every [query](query.md) is automatically tagged with a logging span so that
+you can display all log messages during the execution of the query. For
+example, if you want to log everything during type checking:
+
+```
+RUSTC_LOG=[typeck]
+```
+
+The query key is included as a tracing field which means that you can filter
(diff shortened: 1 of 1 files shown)


src/tracing.md
```diff
@@ -60,6 +60,25 @@ still get a short message per ignored `do_mir_borrowck`, but none of the things
...
+RUSTC_LOG=[typeck]
+```
+
+The query key is included as a tracing field which means that you can filter
+on the debug display of the key. For example, the `typeck` query has a `DefId`
+of what is being checked. You can use a regex to match on that `DefId` to log
+type checking for a specific function:
+
+```
+RUSTC_LOG=[typeck{key=.*name_of_item.*}]
```
oli-obk, line 79: The argument name (`key`) is not the same in all queries, maybe point to https://github.com/rust-lang/rust/blob/87fd70c107b23fe08336a12a20a0f5e85561d499/compiler/rustc_middle/src/query/mod.rs#L18 so that people can browse the list of queries and see their arguments?

oli-obk:
thanks for doing this! I should start documenting all the little things I do that help debugging instead of just shouting it out on twitter

src/tracing.md
ehuss: good point! I pushed an update to try to make that clearer.

oli-obk (approved):

PR #1352, opened 2022-05-11
Fix configuration names for vscode/r-a

WaffleLapkin:
See https://github.com/rust-lang/rust-analyzer/pull/12010 and https://github.com/rust-lang/rust-analyzer/pull/12213

r-a should automatically patch the old configuration, but we might just as well suggest the new one :)


JohnTitor (approved):
Makes sense, thanks!

PR #1354, opened 2022-05-20
Add note about patching dependencies and warnings.

ehuss:
I have occasionally seen people get tripped up by the issue where if you use `patch` to a local path dependency, the build starts failing due to `deny-warnings`. This adds a note about this possible issue.


JohnTitor (approved):
Looks good, thanks!

PR #1356, opened 2022-05-21
Exclude `tomlee.co` from link-checking

JohnTitor:
https://tomlee.co/2014/04/a-more-detailed-tour-of-the-rust-compiler/ returns 404 on CI even though it's not actually 404.
So, that doesn't return any useful result, let's ignore it.
Also, formatted the exclude list for readability.


camelid (approved):
Good idea!

PR #1360, opened 2022-05-28
Triage some date references related to traits

JohnTitor:
Address renaming of the types team, previously wg-traits.
r? @rust-lang/types 

cc #1344


compiler-errors (approved):
I'm pretty sure candidate assembly hasn't changed since 2021. Thank you for linking to the Zulip stream.

PR #1365, opened 2022-06-07
Make build scripts and proc macros work with the suggested rust-analyzer config

jonas-schievink:
For some reason, using `procMacro.ignored` does not seem to work (many functions were missing from analysis because `#[tracing::instrument]` apparently expanded to nothing), so this config makes them work properly instead.


jyn514 (approved):

JohnTitor (approved):
Looks good, thanks!

PR #1369, opened 2022-06-18
Add an "is" and rearange "We next" to "Next, we"

EdwinRy:
I think this PR could slightly improve the readability of the paragraph.


JohnTitor (approved):
Thanks!

PR #1371, opened 2022-06-21
small improves

tshepang:
Also, RUST_BACKTRACE=full is now default

### DIFF
--- a/src/diagnostics.md
+++ b/src/diagnostics.md
@@ -280,13 +280,13 @@ There are two main ways to find where a given error is emitted:
   will treat the first error being emitted as an Internal Compiler Error, which
   allows you to get a
   stack trace at the point the error has been emitted. Change the `1` to
-  something else if you wish to trigger on a later error. One limitation
-  with this approach is that some calls get elided from the stack trace because
-  they get inlined in the compiled `rustc`.
-  Another limitation is the same problem we faced with
-  the prior approach, where the _construction_ of the error is far away from
-  where it is _emitted_. In some cases, we buffer multiple errors in order to
-  emit them in order.
+  something else if you wish to trigger on a later error.
+
+  There are limitations with this approach:
+  - Some calls get elided from the stack trace because they get inlined in the compiled `rustc`.
+  - The _construction_ of the error is far away from where it is _emitted_,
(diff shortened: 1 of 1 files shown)


src/diagnostics.md
```diff
@@ -273,18 +273,19 @@ There are two main ways to find where a given error is emitted:
...
   will treat the first error being emitted as an Internal Compiler Error, which
-  allows you to use the environment variable `RUST_BACKTRACE=full` to get a
+  allows you to get a
   stack trace at the point the error has been emitted. Change the `1` to
-  something else if you whish to trigger on a later error. Some limitations
+  something else if you whish to trigger on a later error. One limitation
   with this approach is that some calls get elided from the stack trace because
-  they get inlined in the compiled `rustc`, and the same problem we faced with
+  they get inlined in the compiled `rustc`.
+  Another limitation is the same problem we faced with
```
JohnTitor, line 286: IMO: I feel something like the below is slightly better:
```
This approach has some limitations. First, ... Second, ...
```

How about?

src/diagnostics.md
```diff
@@ -273,18 +273,19 @@ There are two main ways to find where a given error is emitted:
...
   constructed behind a relatively deep call-stack. Even then, it is a good way
   to get your bearings.
-- Invoking `rustc` with the nightly-only flag `-Z treat-err-as-bug=1`, which
+- Invoking `rustc` with the nightly-only flag `-Z treat-err-as-bug=1`
   will treat the first error being emitted as an Internal Compiler Error, which
-  allows you to use the environment variable `RUST_BACKTRACE=full` to get a
+  allows you to get a
   stack trace at the point the error has been emitted. Change the `1` to
-  something else if you whish to trigger on a later error. Some limitations
+  something else if you whish to trigger on a later error. One limitation
```
JohnTitor, line 283: Pre-existing but:

```suggestion
  something else if you wish to trigger on a later error. One limitation
```

src/diagnostics.md
tshepang: I restructured it a bit, so the paragraph is less heavy

JohnTitor (approved):
Looks great, thanks!

Issue #1132, opened 2021-05-26
One sentence per line?


RalfJung:
The docs currently use a fixed-width style for formatting the source, so even small changes (adding a single word somewhere) can lead to large diffs. This also amplifies conflicts.

It might be worth to consider using one-sentence-per-line or one-logical-unit-per-line (allowing line-breaks at reasonable places even inside sentences). I have made good experience with that formatting. The change doesn't have to happen all at once; when existing text is changed one could encourage also changing it to the new format.

Cc @jyn514

rylev:
I'd personally prefer to each paragraph on its own line. Most editors can visually line break (making reading the raw markdown easy).

RalfJung:
That makes diffs quite bad though -- diff is line-based, which aligns well with one-sentence-per-line IMO.

rylev:
That's true though I think GitHub's diff can handle intra-line diffs a bit better than plain `diff`, but this doesn't make local diffing nicer. I'm fine with moving to a per line scheme, but is their tooling that will do this automatically. It would be a huge when to not have to think about this at all when writing (similar to how rustfmt can be run on save in many editors).

tshepang:
@RalfJung this looks like can be closed, given #1241

RalfJung:
https://github.com/rust-lang/rustc-dev-guide/issues/1241 was closed as a duplicate of this one. If we also close this one there's none left to track this. ;)

jyn514:
> I'm fine with moving to a per line scheme, but is their tooling that will do this automatically. It would be a huge when to not have to think about this at all when writing (similar to how rustfmt can be run on save in many editors).

this is my opinion as well - unless we have some way to enforce it, we'll use an enormous amount of energy to switch over the existing docs only to have new PRs not follow the format (and either have another round of back and forth to fix it, or it just won't get fixed because the reviewer won't think about it). Tooling would greatly reduce the effort for both the initial switch and continued maintenance.

FWIW I would expect this to be fairly simple to write ourselves - the hard part is distinguishing line endings from abbreviations (`e.g.`), but maybe there's some existing dictionary we can use?

tshepang:
am working on the tool... it is so far not so clever, but I could already start using to migrate things gradually

Nilstrieb:
[...]

PR #1372, opened 2022-06-21
not obvious what Ex is, so rather get rid

tshepang:


src/SUMMARY.md
```diff
@@ -76,8 +76,8 @@
 - [Prologue](./part-3-intro.md)
 - [Command-line arguments](./cli.md)
 - [The Rustc Driver and Interface](./rustc-driver.md)
-    - [Ex: Type checking through `rustc_interface`](./rustc-driver-interacting-with-the-ast.md)
-    - [Ex: Getting diagnostics through `rustc_interface`](./rustc-driver-getting-diagnostics.md)
+    - [Type checking through `rustc_interface`](./rustc-driver-interacting-with-the-ast.md)
+    - [Getting diagnostics through `rustc_interface`](./rustc-driver-getting-diagnostics.md)
```
JohnTitor, line 80: According to https://rustc-dev-guide.rust-lang.org/rustc-driver-interacting-with-the-ast.html, it means "example", could you replace them with "Example" instead?

JohnTitor:
Hmm, the line length checker isn't much helpful here, maybe we should rethink https://github.com/rust-lang/rustc-dev-guide/issues/1132.

For now, could you exclude SUMMARY.md from the check on https://github.com/rust-lang/rustc-dev-guide/blob/master/ci/check_line_lengths.sh?

JohnTitor:
I'm not sure if it really improves readability here. and SUMMARY.md is not for reading generally. But yeah, the current looks fine and we could discuss later on the above issue, merging for now. Thanks!

JohnTitor (approved):

PR #1374, opened 2022-06-26
humorust: Forbid pineapple on pizza

Alcaro:
The PR wasn't merged, but neither was the rustc-ty punfest, so I think it belongs here.


JohnTitor (approved):
Thanks!

PR #1380, opened 2022-07-01
leave formatOnSave to the user

RalfJung:
`editor.formatOnSave` is not rust-lang/rust specific, so I think we should leave that decision to the user. Also many test files in the rustc repo are not formatted so this will make diffs unnecessarily big.

I also grouped the enabling of proc macros and build scripts together since it kind of feels like a related thing. ;)


JohnTitor (approved):
Makes sense, thanks!

PR #1382, opened 2022-07-02
Fix path to hir_id_validator.rs

danobi:
It doesn't look like the old path ever existed.

### DIFF
--- a/src/lowering.md
+++ b/src/lowering.md
@@ -18,7 +18,7 @@ of such structures include but are not limited to
     * Converted to a virtual `existential type` declaration
 
 Lowering needs to uphold several invariants in order to not trigger the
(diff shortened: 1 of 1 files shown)


tshepang (approved):

src/lowering.md
```diff
@@ -18,7 +18,7 @@ of such structures include but are not limited to
     * Converted to a virtual `existential type` declaration
 
 Lowering needs to uphold several invariants in order to not trigger the
-sanity checks in `compiler/rustc_middle/src/hir/map/hir_id_validator.rs`:
+sanity checks in `compiler/rustc_passes/src/hir_id_validator.rs`:
```
JohnTitor, line 21: We could turn it into a link to find it as a linkcheck failure.

PR #1393, opened 2022-07-15
remove outdated info on debugging

### LINKED ISSUE #1391: Debugging support chapter is somewhat outdated
The chapter at https://rustc-dev-guide.rust-lang.org/debugging-support-in-rustc.html has several things that I noticed were out-of-date. Some observations:

* The GDB fork hasn't been a thing for about 4 years now. I think that link and the mention that there is a fork could be removed.
* The LLDB fork also hasn't been a thing for about as long. It is not built in CI, so all the notes about macOS and such should be removed. I believe none of it was upstreamed, so I think all of this information about LLDB is old enough to not be valuable, and I would suggest removing it.
* I would remove the sections about shipping GDB and LLDB. I do not believe there are any plans to do this for the foreseeable future.

tshepang:
Closes #1391

Also, do some small fixes/improvements while at it.

### DIFF
--- a/src/debugging-support-in-rustc.md
+++ b/src/debugging-support-in-rustc.md
@@ -3,11 +3,13 @@
 <!-- toc -->
 
 This document explains the state of debugging tools support in the Rust compiler (rustc).
-The document gives an overview of debugging tools like GDB, LLDB etc. and infrastructure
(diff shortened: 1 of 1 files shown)


JohnTitor (approved):
LGTM reading https://github.com/rust-lang/rustc-dev-guide/issues/1391

PR #1394, opened 2022-07-16
make date-check more lightweight

tshepang:


JohnTitor (approved):
Makes some sense to me.

src/contributing.md
```diff
@@ -440,15 +440,8 @@ Just a few things to keep in mind:
...
-
-  - Additionally, include a machine-readable comment of the form `<!-- date:
-    2022-04 -->` (if the current month is April 2022). We have an automated
-    tool that uses these (in `ci/date-check`).
-
-    So, for the month of April 2022, the comment would look like: `As of <!--
-    date: 2022-04 --> April 2022`. Make sure to put the comment *between* `as of`
-    and `April 2022`; see [PR #1066][rdg#1066] for the rationale.
+    We have CI action (in `"~/.github/workflows/date-check.yml"`)
+    that generates an issue if any of these are over 6 months old.
```
camelid, line 444:
```suggestion
    We have a CI action (in `~/.github/workflows/date-check.yml`)
    that generates a monthly issue with any of these that are over 6 months old.
```

src/contributing.md
```diff
@@ -440,15 +440,8 @@ Just a few things to keep in mind:
   - The date the comment was added, e.g. instead of writing _"Currently, ..."_
     or _"As of now, ..."_, consider writing
     _"As of January 2021, ..."_.
-    Try to format the date as `<MONTH> <YEAR>` to ease search.
```
camelid, line 443: I'd rather keep this line (or something like it) since the format does matter.

spastorino:
Hmmm, I'm neutral to this change but consider that after this is merged, the text is more likely to be changed or removed to something else without being noticed.

ci/date-check/src/main.rs
```diff
@@ -36,29 +37,20 @@ impl fmt::Display for Date {
...
-        r"(?x) # insignificant whitespace mode
-        <!--\s*
-        [dD]ate:\s*
-        (?P<y>\d{4}) # year
-        -
-        (?P<m>\d{2}) # month
-        \s*-->",
-    )
-    .unwrap()
+    Regex::new(r"[aA]s of (\w+) (\d{4})").unwrap()
```
camelid, line 40: This shouldn't silently fail if newlines are used instead.

```suggestion
    Regex::new(r"[aA]s\s+of\s+(\w+)\s+(\d{4})").unwrap()
```

ci/date-check/src/main.rs
```diff
@@ -183,20 +175,18 @@ mod tests {
...
     }
 
     #[test]
     fn test_collect_dates_from_file() {
-        let text = "Test1\n<!-- date: 2021-01 -->\nTest2\nFoo<!-- date: 2021-02 \
-                    -->\nTest3\nTest4\nFoo<!-- date: 2021-03 -->Bar\n<!-- date: 2021-04 \
-                    -->\nTest5\nTest6\nTest7\n<!-- date: \n\n2021-05 -->\nTest8
+        let text = "Test1\nAs of Jan 2021\nTest2\nAs of Feb 2021 \
+                    \nTest3\nTest4\nAs of march 2021Bar\nas of apr 2021 \
+                    \nTest5\nTest6\nTest7\n\n\nas of may 2021\nTest8
```
camelid, line 189: Can you add tests with newlines interspersed?

camelid (changes requested):
Left a few comments above. Overall this looks good. Nice idea!

camelid:
> Hmmm, I'm neutral to this change but consider that after this is merged, the text is more likely to be changed or removed to something else without being noticed.

Hmm, that is a good point. It looking more natural does make it less obvious that it's read by a tool...

tshepang:
I found it tedious to write the date twice when updating things. I also made it more clear what date formats are accepted (as suggested by @camelid), so hopefully it's less hidden if the formats are processed by a tool. Am also thinking that as a WG, we are all aware of this tool (at least the more active of us), and all PRs that violate the format can be fixed, without even burdening the submitter to comply, since we could just add a fixup commit to their PR.

camelid:
> If we don't want to repeat the date and still be explicit that we are checking it, what about ...
> 
> When you don't want the date to show up in the text: `<!-- check-date: July 2022 -->` When you want the date to show up in the text: `<!-- check-date: -->July 2022`

I like this idea. Or even shorter: `<!-- track: July 2022 -->` and `<!-- track: -->July 2022`. Maybe that wouldn't be clear enough though?

tshepang:
I named it `date-check` in a branch am still working on, which matches name of tool

tshepang:
This was rebased, but the only real change is the latest commit https://github.com/rust-lang/rustc-dev-guide/pull/1394/commits/e82c2457b42c93f92b235d9461a7ba09ed92bd76.

ci/date-check/src/main.rs
```diff
@@ -35,50 +36,56 @@ impl fmt::Display for Date {
...
-        (?P<y>\d{4}) # year
-        -
-        (?P<m>\d{2}) # month
-        \s*-->",
-    )
-    .unwrap()
+fn make_date_regex() -> Vec<Regex> {
+    let patterns = [
+        r"<!--\s+date-check:\s+(\w+)\s+(\d+{4})\s+-->",
+        r"<!--\s+date-check\s+-->\s+(\w+)\s+(\d+{4})",
```
camelid, line 42: I don't think the `+` after `\d` is correct (it wasn't there before either):

```suggestion
        r"<!--\s+date-check:\s+(\w+)\s+(\d{4})\s+-->",
        r"<!--\s+date-check\s+-->\s+(\w+)\s+(\d{4})",
```

ci/date-check/src/main.rs
```diff
@@ -35,50 +36,56 @@ impl fmt::Display for Date {
...
+    ];
+    let set = RegexSet::new(&patterns).unwrap();
+    set.patterns()
+        .iter()
+        .map(|pattern| Regex::new(pattern).unwrap())
+        .collect()
```
camelid, line 48: What does this usage of `RegexSet` do? It seems like has the same result as just returning `vec![Regex::new(r1), Regex::new(r2)]`.

ci/date-check/src/main.rs
```diff
@@ -35,50 +36,56 @@ impl fmt::Display for Date {
...
+                    .chars()
+                    .filter(|c| *c == '\n')
+                    .count();
+                end_of_last_cap = byte_range.end;
+                (line, date)
+            })
+            .collect();
+        output.extend(results);
+    }
+    output
```
camelid, line 80: Please sort the output by line number. Now that two passes are being made the output will be disordered.

ci/date-check/src/main.rs
camelid: Actually, could you combine these regexes into one to avoid all this complexity with making multiple passes? It might be a little more permissive wrt accepted input but I think it'd be better overall.

src/contributing.md
```diff
@@ -437,17 +437,28 @@ Just a few things to keep in mind:
...
+    consider adding the date, in one of the following formats:
+    - Jan 2021
+    - January 2021
+    - jan 2021
+    - january 2021
 
-  - Additionally, include a machine-readable comment of the form `<!-- date:
-    2022-04 -->` (if the current month is April 2022). We have an automated
-    tool that uses these (in `ci/date-check`).
+    There is a CI action (in "~/.github/workflows/date-check.yml")
```
camelid, line 447:
```suggestion
    There is a CI action (in `~/.github/workflows/date-check.yml`)
```

src/contributing.md
```diff
@@ -437,17 +437,28 @@ Just a few things to keep in mind:
...
+        <!-- date-check -->
+
+    Example:
+
+        As of <!-- date-check --> Jul 2022, the foo did the bar.
+
+    For cases where the date should not be part of the visible rendered output,
+    use the following instead:
+
+        <!-- date-check: Jul 2022 -->
```
camelid, line 461: nit: could you use fenced code blocks and mark it as markdown?

camelid (commented):
I think this is very close. My only main concern is the complexity of having two regexes. I think it should be possible to have just one (slightly more permissive) regex. The rest of the comments are just stylistic nitpicks.

ci/date-check/src/main.rs
tshepang: a confused mess is what it is

ci/date-check/src/main.rs
tshepang: is this alright

ci/date-check/src/main.rs
tshepang: I did this with an earlier iteration, but it moved complexity elsewhere, where I had 4 regex groups (instead of just 2). Is that preferable, or is there a better alternative?

tshepang: ok, it's not so bad 2a5b586c24d4bed7b78227d5a5481bbb6b47e5ff

tshepang:
@camelid much wonderful review, thanks much... things are better as a result

ci/date-check/src/main.rs
```diff
@@ -37,13 +38,11 @@ impl fmt::Display for Date {
...
-        <!--\s*
-        [dD]ate:\s*
-        (?P<y>\d{4}) # year
-        -
-        (?P<m>\d{2}) # month
-        \s*-->",
+        r"(?x)
+        (?:<!--\s+date-check:\s+(\D+)\s+(\d{4})\s+-->)
+        |
+        (?:<!--\s+date-check\s+-->\s+(\D+)\s+(\d{4})\b)
```
camelid, line 44: A couple tweaks:

- Added whitespace and comments to make it easier to read
- Switched back to `?P` (named capture groups, should improve the code further down)
- Switched some instances of `\s+` to `\s*` to match the old behavior

```suggestion
        r"(?x) # insignificant whitespace mode
        (<!--\s*
          date-check:\s*
          (?P<m1>\D+)\s+
          (?P<y1>\d{4})\s*-->
        )
        |
        (<!--\s*
          date-check\s*-->\s+
          (?P<m2>\D+)\s+
          (?P<y2>\d{4})\b
        )
```

ci/date-check/src/main.rs
```diff
@@ -52,15 +51,19 @@ fn collect_dates_from_file(date_regex: &Regex, text: &str) -> Vec<(usize, Date)>
...
-                cap.get(0).unwrap().range(),
-                Date {
-                    year: cap["y"].parse().unwrap(),
-                    month: cap["m"].parse().unwrap(),
-                },
-            )
+        .captures_iter(text)
+        .filter_map(|cap| {
+            if let (Some(month), Some(year), None, None) | (None, None, Some(month), Some(year)) =
+                (cap.get(1), cap.get(2), cap.get(3), cap.get(4))
```
camelid, line 57: This will have to be updated to use named capture groups.

camelid (commented):
Ok, I left some suggestions to simplify the regex. After that, this looks great! Thanks for implementing this and writing so many tests.

camelid (approved):
Thank you again! I really like this new version of date-check.

PR #1395, opened 2022-07-16
Rename debugging_opts to unstable_opts, use link

5225225:


5225225:
Would be neat if we could do this automatically in some form, both link to the docs by writing a path (like how rustdoc does it, kinda) and check that the links are valid. That would be Complex but maybe a Fun project. Probably not worth it it though!

tshepang (approved):

tshepang:
@5225225 is that a project you are considering, somewhat

5225225:
Oh no, I've just volunteered myself to write that. Oh dear.

Maybe if I get through all my rustc PRs and have not much else to do, I can see how easy it would be to write that :)

PR #1402, opened 2022-07-18
revamp doc-build chapter

tshepang:
it's a tiny chapter, so made it lightweight by doing bullet points instead of section headers

### DIFF
--- a/src/building/compiler-documenting.md
+++ b/src/building/compiler-documenting.md
@@ -19,9 +19,8 @@ like the standard library (std) or the compiler (rustc).
   ./x.py doc --stage 1
   ```
 
-  First,
-  the compiler gets built to make sure rustdoc compiles,
-  then that newly-built rustdoc is used to document the components.
+  This ensures that (current) rustdoc gets built,
+  then that is used to document the components.
 
 - Much like running individual tests or building specific components,
   you can build just the documentation you want:
--- a/src/building/compiler-documenting.md
+++ b/src/building/compiler-documenting.md
@@ -36,7 +36,7 @@ A more conventional way is using `/x.py doc`.
 - Document internal rustc items
 
   Compiler documentation is not built by default.
-  To enable it, modify "config.toml":
+  To enable it, modify `config.toml`:
 
   ```toml
   [build]
--- a/src/building/compiler-documenting.md
+++ b/src/building/compiler-documenting.md
@@ -1,13 +1,13 @@
 # Building documentation
 
 This chapter describes how to build documentation of toolchain components,
-either in whole or individually.
+like the standard library (std) or the compiler (rustc).
 
 - Document everything
 
   This uses `rustdoc` from the beta toolchain,
   so will produce (slightly) different output to stage 1 rustdoc,
-  as `rustdoc` is under active development:
+  as rustdoc is under active development:
 
   ```bash
   ./x.py doc
@@ -20,8 +20,8 @@ either in whole or individually.
   ```
 
   First,
-  the compiler and rustdoc get built to make sure everything is okay,
-  and then it documents the files.
+  the compiler gets built to make sure rustdoc compiles,
+  then that newly-built rustdoc is used to document the components.
 
 - Much like running individual tests or building specific components,
   you can build just the documentation you want:
--- a/src/building/compiler-documenting.md
+++ b/src/building/compiler-documenting.md
@@ -1,8 +1,7 @@
 # Building documentation
 
-You can run `rustdoc` directly on a file to make sure the HTML is correct,
-which runs quickly.
-A more conventional way is using `/x.py doc`.
+This chapter describes how to build documentation of toolchain components,
+either in whole or individually.
 
 - Document everything
 
--- a/src/building/compiler-documenting.md
+++ b/src/building/compiler-documenting.md
@@ -35,7 +35,7 @@ like the standard library (std) or the compiler (rustc).
 - Document internal rustc items
 
   Compiler documentation is not built by default.
-  To enable it, modify `config.toml`:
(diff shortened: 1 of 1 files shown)


src/building/compiler-documenting.md
```diff
@@ -1,55 +1,51 @@
...
+  ./x.py doc compiler library
+  ```
 
-Much like individual tests or building certain components you can build only
- the documentation you want.
+- Document internal rustc items
 
-## Document internal rustc items
+  Compiler documentation is not built by default.
+  To enable it, modify "config.toml":
```
JohnTitor, line 39: I'd prefer backticks if it's a file name.

```suggestion
  To enable it, modify `config.toml`:
```

src/building/compiler-documenting.md
```diff
@@ -1,55 +1,51 @@
 # Building documentation
 
-You might want to build documentation of the various components
-available like the standard library. There’s two ways to go about this.
-You can run rustdoc directly on the file to make sure the HTML is
-correct, which is fast. Alternatively, you can build the documentation
-as part of the build process through `x.py`. Both are viable methods
-since documentation is more about the content.
+You can run `rustdoc` directly on a file to make sure the HTML is correct,
+which runs quickly.
```
JohnTitor, line 4: This seems a little too concise to me. I'd like to talk that this section is to document things on rust-lang/rust like std or submodule docs at first.

tshepang: should I remove this mention... I only kept it because I found it there, and I feel it's not worth a mention

JohnTitor: > and I feel it's not worth a mention

Hmm, why? I think describing what this page talks about makes some sense as not all readers will read through the entire chapter.

tshepang: I meant only the part about `rustdoc` being able to be run on a single file

JohnTitor: I mean you removed the "You might want to build documentation of the various components available like the standard library." sentence but the introduction now sounds a little too concise.

tshepang: my latest commit is also concise, but it explains what the chapter is about... whatyouthink

jyn514: I agree with @JohnTitor, "toolchain components" seems like jargon that won't be familiar to most readers. I think "the various components available, like the standard library or compiler," would be more clear.

src/building/compiler-documenting.md
```diff
@@ -1,55 +1,50 @@
...
-```
+  ```bash
+  ./x.py doc --stage 1
+  ```
 
-First the compiler and rustdoc get built to make sure everything is okay
-and then it documents the files.
+  First,
+  the compiler and rustdoc get built to make sure everything is okay,
+  and then it documents the files.
```
jyn514, line 24: hmm, this sentence is saying a lot of complicated things in very few words.
- rustdoc doesn't check that programs are valid, so you have to run rustc directly to make sure the program compiles
- rustdoc uses the built metadata to document the compiler (it's very unclear in this sentence whether "it" refers to rustdoc or bootstrap).

unless it's talking about stage1, in which case it's just completely wrong - the reason you have to build rustdoc is so you can run it, because it might be different from the beta rustdoc.

src/building/compiler-documenting.md
```diff
@@ -1,55 +1,50 @@
...
+  ./x.py doc compiler library
+  ```
 
-Much like individual tests or building certain components you can build only
- the documentation you want.
+- Document internal rustc items
 
-## Document internal rustc items
+  Compiler documentation is not built by default.
+  To enable it, modify `config.toml`:
```
jyn514, line 38: This was changed a while back to only control the defaults, not be a hard-off switch. https://github.com/rust-lang/rust/pull/90683
```suggestion
  To create it by default with `x.py doc`, modify `config.toml`:
```

src/building/compiler-documenting.md
jyn514: (this is also one of the few cases where `--keep-stage 0` is sound - if you've only modified documentation, it can't affect the build artifacts, so it's ok to only rebuild the documentation and not rustdoc itself.)

src/building/compiler-documenting.md
tshepang: What does this mean... what's the difference?

src/building/compiler-documenting.md
tshepang: not sure how much detail to provide, but please have a look at the small change I made

src/building/compiler-documenting.md
jyn514: Before if you ran `x.py doc compiler ` it would say "skipping because compiler-docs is disabled". Now it builds the docs in that case, but not if you only run `x.py doc`.

src/building/compiler-documenting.md
```diff
@@ -20,8 +20,8 @@ either in whole or individually.
   ```
 
   First,
-  the compiler and rustdoc get built to make sure everything is okay,
-  and then it documents the files.
+  the compiler gets built to make sure rustdoc compiles,
```
jyn514, line 23: this is super nitpicky and you can ignore it if you like, but I still feel like "make sure rustdoc compiles" is ambiguous. We're not *checking* that it compiles, we're actually compiling it. I would be less worried about this if it weren't for the fact that `check` is a cargo/x.py subcommand 😅

Maybe it makes more sense to just say "first rustdoc is built," ?

jyn514 (approved):
This looks a lot better than before, thanks :)

src/building/compiler-documenting.md
tshepang: glad you are "nitpicky"... was not too happy with it, so thanks for this

PR #1405, opened 2022-07-22
Document how to build a cross-compiler

bstrie:


src/building/how-to-build-and-run.md
```diff
@@ -223,6 +220,52 @@ fall back to using `cargo` from the installed `nightly`, `beta`, or `stable` too
...
+Targets also may require additional configuration, for example `wasm32-wasi` also requires LLD.
+As a result, building support for `wasm32-wasi` requires `config.toml` to contain the following:
+
+```toml
+[rust]
+lld = true
+
+[target.wasm32-wasi]
+wasi-root = "/path/to/wasi/sysroot"
+```
```
jyn514, line 253: this specifically seems like it should go in https://doc.rust-lang.org/rustc/platform-support.html like @ehuss suggested, and we can link there from the guide. everything else looks good :)

jyn514 (changes requested):

src/building/how-to-build-and-run.md
bstrie: I'm happy to add a new page there for wasm32-wasi, however, I'm not qualified to fill out all the fields in the platform docs template (I don't know who the maintainers are, I don't know the platform requirements...). Is it okay if I just leave those fields blank and only fill out the "Building" section?

bstrie: Actually, for now I can just have this document link to any of the other platform pages that already exist. I can work on getting a wasm32-wasi page added separately.

jyn514: > Is it okay if I just leave those fields blank and only fill out the "Building" section?

Yes, I think that's fine :) incremental progress is better than no progress.

bstrie:
Feedback addressed, ready for review.

src/building/how-to-build-and-run.md
```diff
@@ -1,10 +1,7 @@
 # How to Build and Run the Compiler
 
 The compiler is built using a tool called `x.py`. You will need to
-have Python installed to run it. But before we get to that, if you're going to
-be hacking on `rustc`, you'll want to tweak the configuration of the compiler.
-The default configuration is oriented towards running the compiler as a user,
-not a developer.
+have Python installed to run it.
```
jyn514, line 4: Was it intentional to remove this bit?

jyn514 (approved):
Lgtm

JohnTitor:
Unrelated to this PR but @jyn514, you're familiar with the building and bootstrapping rustc parts and help review PRs related to them, so I think it's better for you to have write access to this repo (I cannot believe you're _inactive_!). Do you mind if we re-add you as a WG member?

jyn514:
yeah, I'm happy for you to readd me :) but I still think we should give access to all team members to approve PRs, it shouldn't be limited to the dev-guide WG

src/building/how-to-build-and-run.md
bstrie: Yes, although let me know if it's too cheeky to sneak it in here rather than making an entirely separate PR for it. The lines that were removed were directly contradictory to the rest of the document; you don't want to tweak configuration *before* you install Python, you want to install Python so that you can run `./x.py setup` so that an appropriate config can be generated for you (and there is no "default" configuration, the setup tool prompts you to indicate what you're doing, which also encompasses developer rather than user usecases).

jyn514: fair enough, LGTM

PR #1412, opened 2022-07-29
Add instructions to fix build errors in std after adding a new target

Urgau:
This PR add the instructions to fix the build errors that may happen after adding a new target to compiler and adding support in std.

cc @jyn514 @Mark-Simulacrum 
Fixes https://github.com/rust-lang/rust/issues/99901

### DIFF
--- a/src/building/new-target.md
+++ b/src/building/new-target.md
@@ -76,6 +76,22 @@ will then add a corresponding file for your new target containing a
 
 Look for existing targets to use as examples.
 
+After adding your target to the `rustc_target` crate you may want to add
+`core`, `std`, ... with support for your new target. In that case you will
+probably need access to some `target_*` cfg. Unfortunately when building with
+stage0 (the beta compiler), you'll get an error that the target cfg is
+unexpected because stage0 doesn't know about the new target specification and
+we pass `--check-cfg` in order to tell it to check.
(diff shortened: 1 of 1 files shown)


src/building/new-target.md
```diff
@@ -76,6 +76,22 @@ will then add a corresponding file for your new target containing a
 
 Look for existing targets to use as examples.
 
+After adding your target to the `rustc_target` crate you may want to add
+`core`, `std`, ... with support for your new target, in that case you will
+probably need access to some `target_*` cfg, unfortunately when building with
+stage0 (the beta compiler) some errors indicating that the value of the target
+cfg is unexpected, this is due to the fact that stage0 doesn't know about the
+new target specification.
```
jyn514, line 84: this is all good info, it would just be nice to have it be more than one long sentence haha
```suggestion
After adding your target to the `rustc_target` crate you may want to add
`core`, `std`, ... with support for your new target. In that case you will
probably need access to some `target_*` cfg. Unfortunately, when building with
stage0 (the beta compiler), you'll get an error that the target cfg is unexpected because stage0 doesn't know about the new target specification and we pass `--check-cfg` in order to tell it to check.
```

jyn514 (approved):

Issue #1409, opened 2022-07-25
The "Debugging support in the Rust compiler" chapter seems relevant for users rather than contributors


JohnTitor:
_Originally posted by @rylev in https://github.com/rust-lang/rustc-dev-guide/pull/1406#pullrequestreview-1049224734_:

> This also seems correct though some of the information provided seems more relevant for *users* of Rust rather than those working on the Rust compiler.

Should we move this chapter somewhere or clarify how it's involved in hacking rustc?

chenyukang:
Sounds great!

For deugging rustc we already has:
https://rustc-dev-guide.rust-lang.org/compiler-debugging.html#configuring-the-compiler

I want to debug rustc with GDB, and can not find enough information on it.
So I spent some time to make it work.
Issues comes from an old GDB version, https://github.com/rust-lang/rust/issues/60705#issuecomment-761602642.

I will try to make a PR to complement it.

jieyouxu:
Triage: clarified in #1417.

PR #1417, opened 2022-08-06
Add gdb tips for symbol-mangling-version

chenyukang:
From https://github.com/rust-lang/rustc-dev-guide/issues/1409#issuecomment-1207225605

### DIFF
--- a/src/compiler-debugging.md
+++ b/src/compiler-debugging.md
@@ -16,9 +16,29 @@ set `debug = true` in your config.toml.
 
 Setting `debug = true` turns on many different debug options (e.g., `debug-assertions`,
 `debug-logging`, etc.) which can be individually tweaked if you want to, but many people
-simply set `debug = true`. Check out the comments in config.toml.example for more info.
+simply set `debug = true`. 
 
-You will need to rebuild the compiler once you've changed any configuration options.
+If you want to use GDB to debug rustc, please set `config.toml` with options:
+
+```toml
+[rust]
+debug = true
+debuginfo-level = 2 
+# WANRING: this will cost a lot of disk space (about 35GB) and compiling time.
+# Without debuginfo-level, you may also track the execution path, but lost
+# the symbol information for debugging.
+```
+
+Default configuration will enable `symbol-mangling-version` v0, you need to install a GDB 
+with version higher than 10.2, otherwise you need to disable new symbol-mangling-version.
+
+```toml
+[rust]
+new-symbol-mangling = false
+```
+
+Check out the comments in config.toml.example for more info, you will need to rebuild the 
+compiler once you've changed any configuration options.
 
 ## `-Z` flags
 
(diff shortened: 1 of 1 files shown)


src/compiler-debugging.md
```diff
@@ -18,6 +18,22 @@ Setting `debug = true` turns on many different debug options (e.g., `debug-asser
 `debug-logging`, etc.) which can be individually tweaked if you want to, but many people
 simply set `debug = true`. Check out the comments in config.toml.example for more info.
 
+If you want to use GDB to debug rustc, please set `config.toml` with options:
+
+```toml
+[rust]
+debug = true
+debuginfo-level = 2
```
jyn514, line 26: This should come with a big warning I think, I've found it to increase compile times and disk sizes a lot.

jyn514 (approved):
👍 the change makes sense to me

src/compiler-debugging.md
chenyukang: Yes, it will increase disk size cost.
I will add a warning tip.
Do you know what the difference that's we don't set this option?
I may need to have a try.

jyn514: I haven't compared before and after, but I remember rustc_driver.so being over 3 GB (!!) with this enabled.

chenyukang: emm, if we don't add `debuginfo-level = 2`, we can also follow the execution path with command `next`, `step`, but we can not print out the values.

With debuginfo we can print out those symbol info such as: 

```
(gdb) p callee_expr
$2 = (*mut rustc_hir::hir::Expr) 0x7fffe81512c8
(gdb) p callee_expr.kind
$3 = rustc_hir::hir::ExprKind::Path(rustc_hir::hir::QPath::Resolved(core::option::Option<&rustc_hir::hir::Ty>::None, 0x7fffe80513e8))
```

And, yes, those debug info will cost much more disk size and also the compiling speed.
For my case:

Disk size: 9G => 45G 
Time cost: 29 Minutes => 50 Minutes.

I will add more details in this PR. Anyway, even it costs so much, I think it's a helpful way to debug `rustc`, especially useful for beginners such as me.

chenyukang: @jyn514
Added more details for the option.

src/compiler-debugging.md
```diff
@@ -16,9 +16,29 @@ set `debug = true` in your config.toml.
...
+simply set `debug = true`. 
 
-You will need to rebuild the compiler once you've changed any configuration options.
+If you want to use GDB to debug rustc, please set `config.toml` with options:
+
+```toml
+[rust]
+debug = true
+debuginfo-level = 2 
+# WANRING: this will cost a lot of disk space (about 35GB) and compiling time.
```
jyn514, line 27:
```suggestion
# WARNING: this will use a lot of disk space (about 35GB) and compile time.
```

jyn514 (approved):
Thanks!

PR #1428, opened 2022-08-14
make date-check more easy to use

tshepang:
One could not run it like the following, for the would be a panic:

    cargo run --manifest-path ci/date-check/Cargo.toml .

### DIFF
--- a/ci/date-check/src/main.rs
+++ b/ci/date-check/src/main.rs
@@ -3,6 +3,7 @@ use std::{
     convert::TryInto as _,
     env, fmt, fs,
     path::{Path, PathBuf},
+    process,
     str::FromStr,
 };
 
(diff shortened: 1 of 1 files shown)


jyn514 (changes requested):
Rather than removing this functionality altogether, can you make `root_dir` using CARGO_MANIFEST_PATH or something like that? Or use `unwrap_or` instead of unwrap?

tshepang:
thanks for the suggestion... didn't notice I made some things less pretty

ci/date-check/src/main.rs
```diff
@@ -124,9 +124,7 @@ fn filter_dates(
 }
 
 fn main() {
-    let root_dir = env::args()
-        .nth(1)
-        .expect("expect root Markdown directory as CLI argument");
+    let root_dir = env::args().nth(1).unwrap_or(".".into());
```
camelid, line 127: I'd rather not make this change because then people doing `cargo run` in date-check's directory will get weird results and not realize why.

camelid (changes requested):

jyn514 (approved):

camelid (approved):
Thanks!

PR #1431, opened 2022-08-14
fix lifetime name

tshepang:
Also
- remove stale reference
- I don't see how ty chapter explains anything here


JohnTitor (approved):
Makes sense!

PR #1436, opened 2022-08-18
diagnostics: fix outdated use of string slugs

davidtwco:
These were all replaced by uses of typed identifiers but some documentation wasn't updated.


compiler-errors (approved):

compiler-errors:
r=me with conflicts rebased and fixed

PR #1448, opened 2022-08-27
Updates prereqs to build LLVM from source for C++17.

red1bluelost:
With the recent upgrade to C++17 in LLVM, this increases the minimum requirements for building LLVM from source. Matches those at https://llvm.org/docs/GettingStarted.html#host-c-toolchain-both-compiler-and-standard-library

### DIFF
--- a/src/building/prerequisites.md
+++ b/src/building/prerequisites.md
@@ -12,7 +12,7 @@ Before building the compiler, you need the following things installed:
 
 If building LLVM from source (the default), you'll need additional tools:
 
-* `g++` 5.1 or later, `clang++` 3.5 or later, or MSVC 2017 or later.
+* `g++` 7.1 or later, `clang++` 5.0 or later, or MSVC 2019 or later.
 * `ninja`, or GNU `make` 3.81 or later (ninja is recommended, especially on Windows)
 * `cmake` 3.13.4 or later
 
--- a/src/building/prerequisites.md
+++ b/src/building/prerequisites.md
@@ -12,7 +12,8 @@ Before building the compiler, you need the following things installed:
 
 If building LLVM from source (the default), you'll need additional tools:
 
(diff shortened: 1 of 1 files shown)


tshepang:
do these minimums apply to older llvm versions

red1bluelost:
Good point! LLVM upgraded on August 6 which was very recent. Rust's fork of LLVM is still using C++14 so the old compilers would still be viable.
Maybe this PR can just be blocked until Rust updates the fork to the C++17 LLVM.

tshepang:
the text can be changed to link to pages of the supported llvm versions, which can be older than the fork

red1bluelost:
I edited the text point to the recent LLVM 14 release documentation and added a date check comment. Does this work?

src/building/prerequisites.md
```diff
@@ -12,7 +12,8 @@ Before building the compiler, you need the following things installed:
 
 If building LLVM from source (the default), you'll need additional tools:
 
-* `g++` 5.1 or later, `clang++` 3.5 or later, or MSVC 2017 or later.
+* `g++`, `clang++`, or MSVC with version listed on <!-- date-check: Aug 2022 -->
```
tshepang, line 15:
```suggestion
* `g++`, `clang++`, or MSVC with versions listed on <!-- date-check: Aug 2022 -->
```

tshepang:
it does, thanks

tshepang (approved):

tshepang:
> I edited the text point to the recent LLVM 14 release documentation and added a date check comment. Does this work?

I assume you meant LLVM 13

red1bluelost:
Yes, oops my comment was wrong. The docs match regardless but I put 13 since it counts as one of the "last few" releases.

tshepang:
it looks like the oldest one supported by nightly https://github.com/rust-lang/rust/pull/100460

PR #1450, opened 2022-08-28
Update for removal of RLS

ehuss:
RLS has been removed as of https://github.com/rust-lang/rust/pull/100863. This updates the guide to remove information related to it.

### DIFF
--- a/src/contributing.md
+++ b/src/contributing.md
@@ -249,37 +249,37 @@ subtrees) actually needs to use `git subtree`.
 
 ### External Dependencies (submodules)
 
-Currently building Rust will also build the following external projects:
+Building Rust will also use external git repositories tracked using [git
+submodules]. The complete list may be found in the [`.gitmodules`] file. Some
+of these projects are required (like `stdarch` for the standard library) and
+some of them are optional (like [Miri]).
 
-* [miri](https://github.com/rust-lang/miri)
-* [rls](https://github.com/rust-lang/rls/)
+Usage of submodules is discussed more in the [Using Git
+chapter](git.md#git-submodules).
 
-We allow breakage of these tools in the nightly channel. Maintainers of these
-projects will be notified of the breakages and should fix them as soon as
-possible.
-
-After the external is fixed, one could add the changes with
-
-```sh
-git add path/to/submodule
-```
-
-outside the submodule.
-
-In order to prepare your tool-fixing PR, you can run the build locally by doing
-`./x.py build src/tools/TOOL`. If you will be editing the sources
-there, you may wish to set `submodules = false` in the `config.toml`
-to prevent `x.py` from resetting to the original branch.
+Some of the submodules are allowed to be in a "broken" state where they either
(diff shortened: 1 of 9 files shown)


src/rustc-driver.md
```diff
@@ -7,7 +7,7 @@ using the interface defined in the [`rustc_interface`] crate.
 The `rustc_interface` crate provides external users with an (unstable) API
 for running code at particular times during the compilation process, allowing
 third parties to effectively use `rustc`'s internals as a library for
-analyzing a crate or emulating the compiler in-process (e.g. the RLS or rustdoc).
+analyzing a crate or emulating the compiler in-process (e.g. rustdoc).
```
eddyb, line 10: Heh, this should mention clippy and miri, just like all the other places that already list all the custom drivers (this chapter might need it more than most, since it looks like what I would link for "custom driver", though surprisingly it doesn't itself seem to use that term).

src/git.md
```diff
@@ -434,6 +434,5 @@ exist and that they correspond to some sort of embedded subrepository dependency
 that Git can nicely and fairly conveniently handle for us.
 
 [Git submodules]: https://git-scm.com/book/en/v2/Git-Tools-Submodules
-[`rust-toolstate`]: https://rust-lang-nursery.github.io/rust-toolstate/
```
tshepang, line 437: how did you notice this is unused... do you have a tool

src/contributing.md
```diff
@@ -322,69 +322,21 @@ Here are those same steps in detail:
...
-
-This says that when we updated the submodule, the version number in our
-`src/tools/rls/Cargo.toml` changed. The new version is different from
-the version in `Cargo.lock`, so the build can no longer continue.
-
-To resolve this, we need to update `Cargo.lock`. Luckily, cargo provides a
-command to do this easily.
-
-```
-$ cargo update -p rls
```
tshepang, line 383: this is a rather large chunk to omit... is it mostly outdated

src/git.md
ehuss: I just noticed while searching for rust-toolstate. I would like to have a tool to detect those automatically. Unfortunately I don't think pulldown-cmark provides quite enough information for that.

src/contributing.md
ehuss: Correct. The `[patch]` tables are gone.

BTW, I was noticing that there seems to be a lot of detailed information about git in this "introduction" chapter. Almost nobody needs to know about how to update a submodule. I feel like more of this could be pushed off into the git chapter (or maybe even a subchapter of that). I held back from making larger changes.

src/rustc-driver.md
tshepang: @eddyb could that be a separate issue, so it doesn't block this PR

JohnTitor (approved):
As @tshepang said, I think the review comments could be follow-ups and this is ready to merge as-is. Please beat me if I'm missing something!

PR #1453, opened 2022-08-30
Add reference for updating Windows PATH and fix typo

Samyak2:
Yes, the reference is a page on Java's documentation but that's the most complete and most official looking reference I could find.


tshepang:
can you qualify it a bit, maybe "this is Java-related, but explains how to edit PATH"

tshepang (approved):

PR #1456, opened 2022-09-03
Link to the correct page in "about this guide"

jyn514:
Previously, "Building and debugging `rustc`" linked to the getting started page, which isn't correct. If people want that page, it already appears in the sidebar, and is the next sequential page.

### DIFF
--- a/src/about-this-guide.md
+++ b/src/about-this-guide.md
@@ -5,7 +5,7 @@ as well as to help new contributors get involved in rustc development.
 
 There are seven parts to this guide:
 
-1. [Building and debugging `rustc`][p1]:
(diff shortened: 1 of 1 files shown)


tshepang:
this one should perhaps link to 2 separate pages, for we also have a separate debugging chapter

tshepang:
if so, this should then be named `Building rustc`

jyn514:
Ok, I've updated this to only say "Building rustc". I don't think we should nest links like that on the main page, there's already a *lot* of content and nesting links makes it more noisy.

tshepang (approved):

PR #1462, opened 2022-09-11
Fix path to `rustfmt` in Suggested Workflows

BGR360:
Was pairing with a new contributor the other day and we realized the path to `rustfmt` was wrong.

I think it also might be the case that users now have to manually build stage0 `rustfmt` by doing `./x.py build --stage 0 src/tools/rustfmt`. It doesn't seem to get built when doing `./x.py build` on a compiler config.

### DIFF
--- a/src/building/suggested.md
+++ b/src/building/suggested.md
@@ -33,7 +33,7 @@ you can write: <!-- date-check: apr 2022 --><!-- the date comment is for the edi
         "--json-output"
     ],
     "rust-analyzer.rustfmt.overrideCommand": [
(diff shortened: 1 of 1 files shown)


ehuss:
I think the original is correct. There's just a bug that rustfmt doesn't get downloaded until you run a command that needs it (like `x.py fmt` or `x.py test tidy`). This is tracked in https://github.com/rust-lang/rust/issues/101306.

BGR360:
Ah gotcha thanks for clarifying.

PR #1469, opened 2022-09-19
Fix some typos

JohnTitor:

### DIFF
--- a/src/opaque-types-impl-trait-inference.md
+++ b/src/opaque-types-impl-trait-inference.md
@@ -147,7 +147,7 @@ flowchart TD
 
 ### Relating an opaque type to another type
 
-There is one central place where an opaqe type gets its hidden type constrained,
+There is one central place where an opaque type gets its hidden type constrained,
 and that is the `handle_opaque_type` function.
 Amusingly it takes two types, so you can pass any two types,
 but one of them should be an opaque type.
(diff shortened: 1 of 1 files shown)


src/opaque-types-impl-trait-inference.md
```diff
@@ -216,15 +216,15 @@ and then handle it correctly.
...
+but ignores obviously dead code (e.g. after a panic).
 The borrow checker is also the source of truth when it comes to hidden types,
 as it is the only one who can properly figure out what lifetimes on the hidden type correspond
 to which lifetimes on the opaque type declaration.
 
 ## Backwards compatibility hacks
 
 `impl Trait` in return position has various quirks that were not part
-of any RFCs and are likely accidental stabilizations.
+of any RFCs and are likely accidental stabilization.
```
tshepang, line 227:
```suggestion
of any RFCs and are likely an accidental stabilization.
```
is this what was meant

JohnTitor: I'm not sure but does it become a "countable noun" in this context?

tshepang: hm, I think it would instead have to be `... and is likely an accidental stabilization.`

JohnTitor: Well, I'm talking about if it's a countable noun or an uncountable one and I think it depends on the context, but I'm not sure if this is the case.

tshepang: I don't know... it just didn't read well, or was not obvious, hence the suggested edit.

PR #1470, opened 2022-09-20
Refine GHA config

JohnTitor:
I recommend reviewing commit by commit. Regarding mdbook changes, it seems fine on my local.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,10 +49,9 @@ jobs:
 
       - name: Install latest nightly Rust toolchain
         if: steps.mdbook-cache.outputs.cache-hit != 'true'
-        uses: actions-rs/toolchain@v1
-        with:
-            toolchain: nightly
-            override: true
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -49,10 +49,9 @@ jobs:
 
       - name: Install latest nightly Rust toolchain
         if: steps.mdbook-cache.outputs.cache-hit != 'true'
-        uses: actions-rs/toolchain@v1
-        with:
-            toolchain: nightly
-            override: true
+        run: |
+          rustup update nightly
+          rustup override set nightly
```
tshepang, line 54: would doing `rustup default nightly` instead of these 2 commands result in a stale toolchain

tshepang: also, looking at CI log, it appears the override is not needed

tshepang (approved):

tshepang:
please be on standby if we need to revert this

PR #1471, opened 2022-09-25
don't refer to the compile-time interpreter as "Miri"

RalfJung:
This leads to a lot of confusion when people think the compile-time interpreter does the same things as [Miri-the-tool](https://github.com/rust-lang/miri). Also given that the folders in the compiler source are called `interpret`, not `miri`, I think it's better to just call this "the (compile-time) (MIR) interpreter", not "Miri".

r? @oli-obk


oli-obk (approved):

src/interpret.md
```diff
@@ -1,14 +1,13 @@
-# Miri
+# Interpreter
```
JohnTitor, line 1: We could move this under `src/mir`, placing it on the top-level dir seems a bit ambiguous to me.

JohnTitor (commented):
Could you also add `miri.html` to https://github.com/rust-lang/rustc-dev-guide/blob/1c79085f41b566c45a43522067b07f8a5b1381e6/book.toml#L43-L45 not to break the old link?

src/interpret.md
RalfJung: In the TOC it is listed below const-eval so maybe we should make that a folder?

JohnTitor (approved):

JohnTitor:
Thanks!

PR #1474, opened 2022-10-03
fix typo and make paragraph consistent

Joshument:
Sorry if this being really minor is a bad thing, but I noticed it and thought I might as well suggest the edit


JohnTitor (approved):
Thanks!

PR #1476, opened 2022-10-05
Update running tests with the new flags

pietroalbini:
https://github.com/rust-lang/rust/pull/102193 changed the flags you need to pass to run `remote-test-server`, and this PR updates the documentation to reflect that.


JohnTitor (approved):
Thanks!

PR #1477, opened 2022-10-05
diagnostic structs: derive on enum

davidtwco:
rust-lang/rust#102189 added support for enums to the diagnostic derive, so this documentation needs updating.


JohnTitor (approved):
Thanks!

PR #1480, opened 2022-10-08
Use llvm subdomain for compiler-explorer link

dkm:
Using `llvm.godbolt.org` subdomain should automatically select the correct LLVM-IR language.

### DIFF
--- a/src/backend/debugging.md
+++ b/src/backend/debugging.md
@@ -224,7 +224,7 @@ really helpful for this.
 create a minimal working example with Godbolt. Go to
 [llvm.godbolt.org](https://llvm.godbolt.org).
 
-2. Choose `LLVM-IR` as programming language if not already selected.
+2. Choose `LLVM-IR` as programming language.
 
 3. Use `llc` to compile the IR to a particular target as is:
     - There are some useful flags: `-mattr` enables target features, `-march=`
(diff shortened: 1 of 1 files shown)


src/backend/debugging.md
```diff
@@ -222,9 +222,9 @@ really helpful for this.
 
 1. Once you have some LLVM IR for the problematic code (see above), you can
 create a minimal working example with Godbolt. Go to
-[gcc.godbolt.org](https://gcc.godbolt.org).
+[llvm.godbolt.org](https://llvm.godbolt.org).
 
-2. Choose `LLVM-IR` as programming language.
+2. Choose `LLVM-IR` as programming language if not already selected.
```
tshepang, line 227: I would just remove this line

dkm: ok :)

tshepang: I meant the whole line, since `llvm.godbolt.org` makes it obsolete

dkm: Oh. I initially wondered also. You'll get LLVM-IR by default, but if you change the language, close the tab, reopen it, the site will pick your last state and give you whatever you had when closing the tab, hence the initial change `if not already selected`.

tshepang (approved):

PR #1481, opened 2022-10-11
Add missing prerequisite for some Linux distros

scarvalhojr:
Fedora and Ubuntu users have reported missing `libstdc++-static` when attempting to build the compiler.


scarvalhojr:
For reference: https://github.com/rust-lang/rust/issues/97138

JohnTitor (approved):
Thanks!

PR #1483, opened 2022-10-14
Update outdated queries section

Nilstrieb:
- queries always need a description
- `QueryDescription` isn't just for descriptions (it's actually not for descriptions at all now) and always implemented by the macro
- queries are not grouped into categories

cc @cjgillot

### DIFF
--- a/src/query.md
+++ b/src/query.md
@@ -175,8 +175,8 @@ Defining a query takes place in two steps:
 To declare the query name and arguments, you simply add an entry to
 the big macro invocation in [`compiler/rustc_middle/src/query/mod.rs`][query-mod].
 Then you need to add a documentation comment to it with some _internal_ description.
-Then, provide the `desc` attribute which contains a short description of the query.
-This description is shown to the user in query cycles.
+Then, provide the `desc` attribute which contains a _user-facing_ description of the query.
+The `desc` attribute is shown to the user in query cycles.
 
 This looks something like:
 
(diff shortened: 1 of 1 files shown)


src/query.md
```diff
@@ -292,8 +276,6 @@ rustc_queries! {
 }
```
jyn514, line 276: this still looks outdated, and a bit redundant with the first section - can we remove it and add `desc` at the top?

src/query.md
```diff
@@ -180,20 +180,14 @@ the big macro invocation in
 
 ```rust,ignore
 rustc_queries! {
-    Other {
-        /// Records the type of every item.
-        query type_of(key: DefId) -> Ty<'tcx> {
-            cache { key.is_local() }
-        }
+    /// Records the type of every item.
```
jyn514, line 183: can we mention that you should add a doc-comment, or else it shames you about it in the generated docs lmao
https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/context/struct.TyCtxt.html#method.trigger_delay_span_bug

jyn514 (commented):
Thanks! Have a couple comments, but appreciate you updating this

src/query.md
Nilstrieb: given that `desc` is required this makes a lot of sense

src/query.md
Nilstrieb: We should probably also move the shaming from rustdoc to shaming you with a compiler instead at some point

jyn514: https://github.com/rust-lang/rust/issues/104008

src/query.md
```diff
@@ -260,40 +259,13 @@ impl<'tcx> QueryConfig for type_of<'tcx> {
...
-There is an additional trait that you may wish to implement called
-[`self::config::QueryDescription`][QueryDescription]. This trait is
-used during cycle errors to give a "human readable" name for the query,
-so that we can summarize what was happening when the cycle occurred.
-Implementing this trait is optional if the query key is `DefId`, but
-if you *don't* implement it, you get a pretty generic error ("processing `foo`...").
-You can put new impls into the `config` module. They look something like this:
+There is an additional trait with more methods called
+[`self::config::QueryDescription`][QueryDescription]. This trait contains a few
+extra methods which are used by the query system.
```
jyn514, line 264: I think this is a bad place to document the query internals. These change semifrequently (you mentioned wanting to make this part of QueryConfig) and they aren't needed to understand how to use queries. I'd rather spend this space explaining local vs extern providers or something like that. Query internals should be documented in either query_impl or query_system IMO.

Nilstrieb: Sounds good, I'll delete it. As for local vs extern providers I would first need to fully understand it myself before I could write about it so I guess that's follow up work :D

jyn514: Ok, opened https://github.com/rust-lang/rust/issues/104011 for documenting internals.

jyn514 (approved):

PR #1485, opened 2022-10-18
add tool to "enforce" semantic line breaks

tshepang:
See #1132

Run it so:
```
cd ci/semantic-line-breaks
cargo run --- ../../src
```

Diff is large, so maybe should be automated.


JohnTitor:
I don't have an objection to enforcing sembr but the implementation does look fragile, e.g. there's no tokenizer. Have you considered using an existing tool like https://github.com/bobheadxi/readable?

tshepang:
I have not... thanks for the suggestion, will have a look.

m-spitfire:
I have tried using `readable` to have a GitHub action that runs check, and format all the existing code, but the tool is immature, it gave errors for some files saying `Cannot take length of multi-line node '[object Object]'` and for others [the aforementioned](https://github.com/rust-lang/rustc-dev-guide/issues/1132#issuecomment-1225713221) issue is there, it does the wrong thing when it encounters `e.g.`. I am not sure what should be done here, make `readable` better or create another tool?

tshepang:
This tool of mine handles both issues without choking. I intend to get back to making it more fancy... I keep postponing.

Kobzol:

```
git fetch origin --all
git checkout <pr-branch>
git push --force-with-lease
```

tshepang:
I don't know how to do tokenization, but I attached a sample chapter to show output from the tool

PR #1486, opened 2022-10-19
Document multiple alternative suggestions on diagnostic structs

Xiretza:
Added by rust-lang/rust#103209.


Xiretza:
Not sure if that works here, but:

@rustbot label -blocked

JohnTitor (approved):
Thanks!

PR #1487, opened 2022-10-19
Update rust-analyzer suggestions

Veykril:
This setting will cause r-a to invoke the build script command only once in the root of the project instead of once per workspace which causes problems when the `linkedProjects` settings is used to target other workspaces in the repo.

In theory the same setting could be applied to `checkOnSave`, but for that to work `x.py` needs to fix up the paths of diagnostics to be relative to the project root instead of the workspace roots the diagnostics are emitted from.

This depends on https://github.com/rust-lang/rust-analyzer/pull/13128 which will have the next stable r-a release on coming monday / nightly in ~2 hours


JohnTitor (approved):
Thanks for the detailed explanation! I'm going to merge this the next Monday then.

Veykril:
Pushed another config thats necessary as well now, implemented in https://github.com/rust-lang/rust-analyzer/pull/13466 and will also be available on the next stable.

PR #1488, opened 2022-10-20
miri is no longer a submodule but a subtree.

luqmana:
As of https://github.com/rust-lang/rust/pull/102028 miri is now subtree.

### DIFF
--- a/src/contributing.md
+++ b/src/contributing.md
@@ -183,6 +183,7 @@ As a developer to this repository, you don't have to treat the following externa
 differently from other crates that are directly in this repo:
 
 * [Clippy](https://github.com/rust-lang/rust-clippy)
+* [Miri](https://github.com/rust-lang/miri)
 * [rustfmt](https://github.com/rust-lang/rustfmt)
 * [rust-analyzer](https://github.com/rust-lang/rust-analyzer)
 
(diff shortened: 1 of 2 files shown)


src/git.md
```diff
@@ -176,6 +176,8 @@ There is a workaround in [the issue][#77620-workaround].
 [#77620]: https://github.com/rust-lang/rust/issues/77620
 [#77620-workaround]: https://github.com/rust-lang/rust/issues/77620#issuecomment-705228229
 
+(Note that as of <!-- date-check --> Sept 2022 `miri` is a subtree and not a submodule.)
```
JohnTitor, line 179: I don't think we have to do the date check as the state isn't likely to be changed in the near future.

src/contributing.md
```diff
@@ -322,6 +322,12 @@ Here are those same steps in detail:
 
 #### Updating submodules
 
+(As of <!-- date-check --> Sept 2022 `miri` is no longer a submodule but a subtree. The following
+instructions are still broadly correct for updating submodules in general. For a more specific
+example you can also see the steps for [upgrading llvm][upgrading-llvm].)
+
+[upgrading-llvm]: https://rustc-dev-guide.rust-lang.org/backend/updating-llvm.html
```
JohnTitor, line 329: We could just redirect to the "Git submodules" section on git.md and remove this to avoid duplication.

luqmana: I just removed that section altogether. We already link to the the "Git submodules" section above in the "External Dependencies (submodules)" section.

src/git.md
luqmana: done

JohnTitor (approved):
Thanks!

PR #1494, opened 2022-10-24
Update `traits/resolution.md`

compiler-errors:
Co-authored by: @lcnr and @spastorino

### DIFF
--- a/src/traits/resolution.md
+++ b/src/traits/resolution.md
@@ -122,7 +122,7 @@ the obligation contains unbound inference variables.
 
 The subroutines that decide whether a particular impl/where-clause/etc applies
 to a particular obligation are collectively referred to as the process of
-_matching_. As of <!-- date-check --> Oct 2022, For `impl` candidates,
+_matching_. For `impl` candidates <!-- date-check: Oct 2022 -->,
 this amounts to unifying the impl header (the `Self` type and the trait arguments)
 while ignoring  nested obligations. If matching succeeds then we add it
 to a set of candidates. There are other rules when assembling candidates for
--- a/src/traits/resolution.md
+++ b/src/traits/resolution.md
@@ -132,7 +132,9 @@ Once this first pass is done, we can examine the set of candidates. If
 it is a singleton set, then we are done: this is the only impl in
 scope that could possibly apply. Otherwise, we can **winnow** down the set
 of candidates by using where clauses and other conditions. Winnowing uses
-`evaluate_candidate` to check whether the nested obligations may apply. If this still leaves more than 1 candidate, we use ` fn candidate_should_be_dropped_in_favor_of` to prefer some candidates over others. 
+`evaluate_candidate` to check whether the nested obligations may apply. 
+If this still leaves more than 1 candidate, we use ` fn candidate_should_be_dropped_in_favor_of` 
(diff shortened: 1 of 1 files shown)


src/traits/resolution.md
```diff
@@ -111,6 +111,8 @@ may lead to other errors downstream).
 
 ### Candidate assembly
 
+**TODO**: Talk about _why_ we have different candidates, and why it needs to happen in ap robe.
```
compiler-errors, line 114:
```suggestion
**TODO**: Talk about _why_ we have different candidates, and why it needs to happen in a probe.
```

src/traits/resolution.md
```diff
@@ -281,38 +236,16 @@ result of selection would be an error.
...
-identify which impl applies, but we do not (yet) need to decide how to select
-any nested obligations. Nonetheless, we *do* currently do a complete resolution,
-and that is because it can sometimes inform the results of type inference.
+One interesting twist has to do with nested obligations. In general, in codegen,
+we only to figure out which candidate applies, we do not care about nested obligations, as these are already assumed to be true. Nonetheless, we *do* currently do fulfill all of them. That is because it can sometimes inform the results of type inference.
 That is, we do not have the full substitutions in terms of the type variables
 of the impl available to us, so we must run trait selection to figure
 everything out.
-
-**TODO**: is this still talking about trans?
```
compiler-errors, line 301: We removed this following section because it's not yet implemented, is kinda an implementation detail, and the example is not exactly where the consideration above ("because it can sometimes inform the results of type inference") actually matters, which is with projection types specifically.

src/traits/resolution.md
```diff
@@ -281,38 +236,16 @@ result of selection would be an error.
...
-### Selection during translation
+### Selection during codegen
 
 As mentioned above, during type checking, we do not store the results of trait
-selection. At trans time, we repeat the trait selection to choose a particular
-impl for each method call. In this second selection, we do not consider any
+selection. At codegen time, we repeat the trait selection to choose a particular
+impl for each method call. This is done using `fn codegen_select_candidate`. In this second selection, we do not consider any
 where-clauses to be in scope because we know that each resolution will resolve
 to a particular impl.
```
compiler-errors, line 245:
```suggestion
impl for each method call. This is done using `fn codegen_select_candidate`. 
In this second selection, we do not consider any where-clauses to be in scope 
because we know that each resolution will resolve to a particular impl.
```

spastorino (approved):

src/traits/resolution.md
```diff
@@ -120,68 +122,23 @@ the obligation contains unbound inference variables.
...
-
-**TODO**: what does "unifying the `Self` types" mean? The `Self` of the
-obligation with that of an impl?
-
-The basic idea for candidate assembly is to do a first pass in which
-we identify all possible candidates. During this pass, all that we do
-is try and unify the type parameters. (In particular, we ignore any
-nested where clauses.) Presuming that this unification succeeds, the
-impl is added as a candidate.
+_matching_. As of <!-- date-check --> Oct 2022, For `impl` candidates,
```
spastorino, line 125: This can also be written like ...

```suggestion
_matching_. For `impl` candidates <!-- date-check: Oct 2022 -->,
```

PR #1506, opened 2022-11-05
Add -Ztrack-diagnostics information

mejrs:
Followup to https://github.com/rust-lang/rust/pull/103217

### DIFF
--- a/src/compiler-debugging.md
+++ b/src/compiler-debugging.md
@@ -185,6 +185,40 @@ stack backtrace:
 
 Cool, now I have a backtrace for the error!
 
+## Getting the the error creation location
+
+`-Z track-diagnostics` can help figure out where errors are emitted. It uses `#[track_caller]`
+for this and prints its location alongside the error:
+
+```bash
+$ RUST_BACKTRACE=1 rustc +stage1 error.rs -Z track-diagnostics
+error[E0277]: cannot add `()` to `{integer}`
+ --> src\error.rs:2:7
+  |
+2 |     1 + ();
+  |       ^ no implementation for `{integer} + ()`
+-Ztrack-diagnostics: created at compiler/rustc_trait_selection/src/traits/error_reporting/mod.rs:638:39
+  |
+  = help: the trait `Add<()>` is not implemented for `{integer}`
+  = help: the following other types implement trait `Add<Rhs>`:
+            <&'a f32 as Add<f32>>
+            <&'a f64 as Add<f64>>
+            <&'a i128 as Add<i128>>
+            <&'a i16 as Add<i16>>
+            <&'a i32 as Add<i32>>
+            <&'a i64 as Add<i64>>
+            <&'a i8 as Add<i8>>
+            <&'a isize as Add<isize>>
+          and 48 others
+
+For more information about this error, try `rustc --explain E0277`.
+```
+
+This is similar but different to `-Z treat-err-as-bug`:
+- it will print the locations for all errors emitted
+- it does not require a compiler built with debug symbols
+- you don't have to read through a big stack trace.
+
 ## Getting logging output
 
 The compiler uses the [`tracing`] crate for logging.
--- a/src/compiler-debugging.md
+++ b/src/compiler-debugging.md
@@ -128,7 +128,7 @@ fn main() {
 }
 ```
 
-```bash
+```
 $ rustc +stage1 error.rs
 error[E0277]: cannot add `()` to `{integer}`
  --> error.rs:2:7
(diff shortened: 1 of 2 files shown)


src/compiler-debugging.md
```diff
@@ -185,6 +185,40 @@ stack backtrace:
 
 Cool, now I have a backtrace for the error!
 
+## Getting the the error creation location
+
+`-Z track-diagnostics` can help figure out where errors are emitted. It uses `#[track_caller]`
+for this and prints its location alongside the error:
+
+```bash
```
tshepang, line 193:
```suggestion
```
```
I would not add this since what follows is not shell code, just console output... would indeed be nice to have colors matching rustc error messages

File: src/diagnostics.md [hunk 2]
```diff
@@ -287,6 +287,8 @@ There are two main ways to find where a given error is emitted:
   - The _construction_ of the error is far away from where it is _emitted_,
     a problem similar to the one we faced with the `grep` approach.
     In some cases, we buffer multiple errors in order to emit them in order.
+- Invoking `rustc` with the nightly-only flag `-Z track-diagnostics` will print error creation
```
tshepang, line 290:
```suggestion
- Invoking `rustc` with the nightly-only flag, `-Z track-diagnostics`, will print error creation
```

src/compiler-debugging.md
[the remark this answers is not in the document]
mejrs: I went with consistency with the treat-as-err example above, which also uses bash syntax to set env variables

src/diagnostics.md
mejrs: I'm not sure that makes sense, but I'm not a native speaker. I went for consistency with treat-bug-as-err, which does not have commas in this position.

src/compiler-debugging.md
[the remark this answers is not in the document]
tshepang: you may fix that as well if you like

src/diagnostics.md
tshepang: You can, alternatively, do this (which avoids the commas)...
> - Invoking `rustc` with `-Z track-diagnostics` will print error creation

mejrs: Thanks, that is much nicer

src/compiler-debugging.md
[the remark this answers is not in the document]
mejrs: Done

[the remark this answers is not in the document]
tshepang: I would not separate commands and their outputs as you did, especially since it does not help with highlighting at all. I can make the changes for you, if you feel like this is too much back-n-forth.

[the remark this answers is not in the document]
mejrs: Sorry for the delay - I  somewhat forgot about this Pr. I think the current separation is fine. Feel free to change it if you disagree.

src/compiler-debugging.md
```diff
@@ -130,6 +130,9 @@ fn main() {
 
 ```bash
 $ rustc +stage1 error.rs
+```
+
+```text
```
tshepang, line 135:
```suggestion
```
$ rustc +stage1 error.rs
```

File: src/compiler-debugging.md [hunk 4]
```diff
@@ -145,6 +148,9 @@ Now, where does the error above come from?
 
 ```bash
 $ RUST_BACKTRACE=1 rustc +stage1 error.rs -Z treat-err-as-bug
+```
+
+```text
```
tshepang, line 153:
```suggestion
```
$ RUST_BACKTRACE=1 rustc +stage1 error.rs -Z treat-err-as-bug
```

File: src/compiler-debugging.md [hunk 5]
```diff
@@ -185,6 +191,43 @@ stack backtrace:
...
+## Getting the the error creation location
+
+`-Z track-diagnostics` can help figure out where errors are emitted. It uses `#[track_caller]`
+for this and prints its location alongside the error:
+
+```bash
+$ RUST_BACKTRACE=1 rustc +stage1 error.rs -Z track-diagnostics
+```
+
+```text
```
tshepang, line 203:
```suggestion
```
$ RUST_BACKTRACE=1 rustc +stage1 error.rs -Z track-diagnostics
```

tshepang:
github ui failed me here, and does not seem to offer a way to fix it... @mejrs will you please fix my mess (replace the missing backticks on the code blocks). Please rebase as well.

mejrs:
Done. I'm not sure I understood what you wanted though...

tshepang (approved):

PR #1509, opened 2022-11-08
Consistent ordered list indexing

RCoder01:
I noticed the weird formatting of the ordered lists at https://rustc-dev-guide.rust-lang.org/macro-expansion.html. I believe this is caused by the 0-based numbering of these lists. Because it seemed like all the other ordered lists in this book were 1-based, I assumed this was an anomaly and changed all the 0-based lists to start from 1.

I apologize If it was intentional to have these start from 0, and I will close this PR if that is the case.

### DIFF
--- a/src/bug-fix-procedure.md
+++ b/src/bug-fix-procedure.md
@@ -53,9 +53,9 @@ Please see [the RFC][rfc 1122] for full details!
 The procedure for making a breaking change is as follows (each of these steps is
 described in more detail below):
 
-0. Do a **crater run** to assess the impact of the change.
-1. Make a **special tracking issue** dedicated to the change.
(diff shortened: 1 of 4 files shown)


tshepang:
In case you are not aware, markdown does automatic number incrementing, where the following...
```
1. first
1. second
1. third
```
becomes...

1. first
1. second
1. third

So, I would just take advantage of the feature, to avoid manually tracking such things, especially if they list changes in future.

I can approve the PR as-is if you prefer though.

RCoder01:
My bigger issue was with 0-indexes messing up markdown's automatic formatting, but I figured I'd fix the other weird numbering issues while I was at it. Not a problem either way, as long as the end result looks fine.

tshepang (approved):

PR #1512, opened 2022-11-22
Update path for `try_mark_green` implementation

ghassanachi:
Updates the location of the `try_mark_green` algorithm from `compiler/rustc_middle/src/dep_graph/graph.rs` to [`compiler/rustc_middle/src/dep_graph/graph.rs`](https://github.com/rust-lang/rust/blob/stable/compiler/rustc_query_system/src/dep_graph/graph.rs#L574) .

### DIFF
--- a/src/queries/incremental-compilation-in-detail.md
+++ b/src/queries/incremental-compilation-in-detail.md
@@ -177,7 +177,7 @@ fn try_mark_green(tcx, current_node) -> bool {
 }
 
 // Note: The actual implementation can be found in
-//       compiler/rustc_middle/src/dep_graph/graph.rs
+//       compiler/rustc_query_system/src/dep_graph/graph.rs
 ```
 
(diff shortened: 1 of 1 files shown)


tshepang (approved):

src/queries/incremental-compilation-in-detail.md
```diff
@@ -177,7 +177,7 @@ fn try_mark_green(tcx, current_node) -> bool {
 }
 
 // Note: The actual implementation can be found in
-//       compiler/rustc_middle/src/dep_graph/graph.rs
+//       compiler/rustc_query_system/src/dep_graph/graph.rs
```
JohnTitor, line 180: Shouldn't we move it outside of codeblock and add a link so that linkchecker can catch it?

tshepang: @JohnTitor https://github.com/rust-lang/rustc-dev-guide/pull/1517

PR #1518, opened 2022-11-28
clarify subtree tool policy

RalfJung:
Cc https://github.com/rust-lang/rust/issues/104962

### DIFF
--- a/src/contributing.md
+++ b/src/contributing.md
@@ -191,8 +191,10 @@ differently from other crates that are directly in this repo:
 
 In contrast to `submodule` dependencies
 (see below for those), the `subtree` dependencies are just regular files and directories which can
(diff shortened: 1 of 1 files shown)


JohnTitor (approved):
Looks good!

flip1995 (approved):
This policy sounds good to me. So :+1: from the Clippy side.

calebcartwright (approved):
Agreed. :+1: on this from rustfmt perspective

camelid:
It looks like this is ready to merge. Is that correct, or are we waiting on approval from another team?

JohnTitor:
I also think this is ready to merge as it has some approvals from some subtree maintainers. Let's merge, and follow-up if there are any concerns/comments.

PR #1521, opened 2022-12-03
chore: Update `actions/github-script` to v6

JohnTitor:
v5 requires to migrate from `github.issues` to `github.rest.issues`, and v6 updates Node.js image to v16.
Ref. https://github.com/actions/github-script#breaking-changes

This also removes `github_token` input as `GITHUB_TOKEN` will be used by default: https://github.com/actions/github-script#using-a-separate-github-token

### DIFF
--- a/.github/workflows/date-check.yml
+++ b/.github/workflows/date-check.yml
@@ -27,9 +27,8 @@ jobs:
           cargo run -- ../../src/ > ../../date-check-output.txt
 
       - name: Open issue
-        uses: actions/github-script@v3
+        uses: actions/github-script@v6
         with:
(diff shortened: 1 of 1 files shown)


camelid:
Do we know if there were breaking changes in V4? In their Breaking Changes section, it only seems to talk about V5 and V6.

JohnTitor:
According to https://github.com/actions/github-script/releases/tag/v4.0.0 and https://github.com/actions/github-script/compare/v3.1.1...v4.0.0, they seem to have added a change around `require` import, but I think it doesn't affect us as we only import `fs`.

camelid (approved):
Great, thanks for checking!

PR #1523, opened 2022-12-03
Remove duplicate paragraph

anirudh24seven:
Removing outdated paragraph that was improved in #1429


JohnTitor (approved):
Good catch, thanks!

PR #1525, opened 2022-12-05
Fix small inaccuracy in monomorph page

nameEqualsJared:
Just fixing a small inaccuracy. Pretty sure we meant to say heap here :)


camelid (approved):
Good catch, thanks!

PR #1532, opened 2022-12-17
Add back the `canonicalization` chapter.

### LINKED ISSUE #1527: "Canonical queries" section doesn't explain canonical queries
https://rustc-dev-guide.rust-lang.org/traits/canonical-queries.html doesn't mention the word "canonical" once. It looks like the docs got deleted in https://github.com/rust-lang/rustc-dev-guide/pull/679. @crlf0710 do you have time to add these back? Canonicalization is important and still happens in rustc, not chalk.

https://github.com/rust-lang/rust/blob/e940f845be13ff37d4ba28df5f40d74e5b0895a0/compiler/rustc_trait_selection/src/traits/mod.rs#L103-L114

crlf0710:
Fixes #1527

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -123,6 +123,7 @@
         - [Lowering to logic](./traits/lowering-to-logic.md)
         - [Goals and clauses](./traits/goals-and-clauses.md)
         - [Canonical queries](./traits/canonical-queries.md)
+        - [Canonicalization](./traits/canonicalization.md)
     - [Next-gen trait solving](./solve/trait-solving.md)
         - [Invariants of the type system](./solve/invariants.md)
         - [The solver](./solve/the-solver.md)
--- a/src/traits/canonicalization.md
+++ b/src/traits/canonicalization.md
@@ -1,0 +1,256 @@
+# Canonicalization
+
+Canonicalization is the process of **isolating** an inference value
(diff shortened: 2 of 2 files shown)


spastorino:
cc @lcnr

jyn514 (commented):
Thanks! This looks good but it seems strange to add it as a new page instead of replacing `canonical-queries` (which today doesn't talk about canonical queries at all). Also, putting it under "chalk-based trait solving" seems strange because it doesn't involve chalk at all. Maybe you can rename the current page to "trait queries" and move both to the higher "trait solving section"?

That said, `canonical-queries` today seems to have a lot of overlap with `lowering-to-logic` so maybe it makes sense to just remove it altogether? 🤔

spastorino:
@lcnr thoughts about this?.

lcnr:
I don't have an opinion here, it is probably useful to add this. there's some overlap with https://rustc-dev-guide.rust-lang.org/solve/canonicalization.html 🤷 

don't feel like i have the capacity to look at this in more detail in the near future, so i am in favor of merging this as is.

crlf0710:
@spastorino I don't think i'm able to apply some in-depth editing to the text, but i added the `FIXME` note as suggested.

spastorino (approved):

Issue #1542, opened 2023-01-05
rust-analyzer config not working with bootstrap


Nilstrieb:
I was talking to @jyn514 about what config to use to work on `src/bootstrap` and we couldn't figure out why it doesn't work for me.

We tried this config:
```json
{
    "rust-analyzer.checkOnSave.overrideCommand": [
        "x",
        "check",
        "--json-output",
        "--build-dir",
        "build-rust-analyzer"
    ],
    "rust-analyzer.rustfmt.overrideCommand": [
        "./build/x86_64-unknown-linux-gnu/stage0/bin/rustfmt",
        "--edition=2021"
    ],
    "rust-analyzer.procMacro.server": "./build/x86_64-unknown-linux-gnu/stage0/libexec/rust-analyzer-proc-macro-srv",
    "rust-analyzer.procMacro.enable": true,
    "rust-analyzer.cargo.buildScripts.enable": true,
    "rust-analyzer.cargo.buildScripts.invocationLocation": "root",
    "rust-analyzer.cargo.buildScripts.invocationStrategy": "once",
    "rust-analyzer.cargo.buildScripts.overrideCommand": [
        "x",
        "check",
        "--json-output",
    ],
    "rust-analyzer.cargo.sysroot": "./build/x86_64-unknown-linux-gnu/stage0-sysroot",
    "rust-analyzer.rustc.source": "./Cargo.toml",
    "rust-analyzer.linkedProjects": ["Cargo.toml", "src/bootstrap/Cargo.toml"],
}
```

This config works for working on bootstrap but breaks when I work on the compiler:
```
[ERROR rust_analyzer::main_loop] flycheck 1: File with cargo diagnostic not found in VFS: file not found: /home/nilsh/projects/rust/src/bootstrap/compiler/rustc_builtin_macros/src/format.rs
```

jyn514:
@Nilstrieb are you still running into this problem?

Nilstrieb:
I'm not! Magic! I will report back if it starts being a problem again.

Issue #1516, opened 2022-11-25
Fix RA recommended configuration


spastorino:
https://rustc-dev-guide.rust-lang.org/building/suggested.html#configuring-rust-analyzer-for-rustc

it mentions `stage0-sysroot`, more discussion https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/empty.20build.2F.24HOST.20directory

Is not clear what's the right suggestion

cc @jyn514

jyn514:
This config is no longer maintained in the guide itself. The version in the rust-lang/rust repo has been fixed.

PR #1545, opened 2023-01-06
Add Neovim configuration information

RossSmyth:
The JSON provided for VSCode works with coc and nlsp-settings. This is verified by
https://github.com/fannheyward/coc-rust-analyzer#configurations for coc, and feeding the JSON in to this schema
https://github.com/tamago324/nlsp-settings.nvim/blob/main/schemas/_generated/rust_analyzer.json for validating nlsp-settings.

The Lua translation is straight-forward and is how all rust-analyzer settings must be used with native nvim lsp.

Note that I just moved and my main computer that has nvim setup on it is not unpacked so I did not test these. But I believe they should work, and I think I've actually done something like this in the past.

### DIFF
--- a/src/building/suggested.md
+++ b/src/building/suggested.md
@@ -27,6 +27,44 @@ you to create a `.vscode/settings.json` file which will configure Visual Studio 
 This will ask `rust-analyzer` to use `./x.py check` to check the sources, and the
 stage 0 rustfmt to format them.
 
+
+For Neovim users there are several options for configuring for rustc. You can use the native LSP server and make your own logic for changing the rust-analyzer configuration to the above. You can also use [nlsp-settings](https://github.com/tamago324/nlsp-settings.nvim), which allows for project-local configuration files. This plugin allows for the above JSON to be directly put in to a file located at `rust/.nlsp-settings/rust_analyzer.json`. If you use [coc-rust-analyzer](https://github.com/fannheyward/coc-rust-analyzer) you can also use the above JSON, but placed in `rust/.vim/coc-settings.json`.
+
+Below is the Lua needed to configure the native Neovim LSP the same as the above VSCode configuration
+
+```lua
+{
+    ["rust-analyzer"] = {
+        checkOnSave = {
+            overrideCommand = {
+                {"python3", "x.py", "check", "--json-output"}
+            }
+        },
+        rustfmt = {
+            overrideCommand = {
+                {"./build/host/stage0/bin/rustfmt", "--edition=2021"}
+            }
+        },
+        procMacro = {
+            server = "./build/host/stage0/libexec/rust-analyzer-proc-macro-srv",
+            enable = true
+        },
+        cargo = {
+            buildScripts = {
+                enable = true,
+                invocationLocation = "root",
+                invocationStrategy = "once",
+                overrideCommand = {"python3", "x.py", "check", "--json-output"}
+            },
+            sysroot = "./build/host/stage0-sysroot"
+        },
+        rustc = {
+            source = "./Cargo.toml"
+        }
+    }
+}
+```
+
 If you have enough free disk space and you would like to be able to run `x.py` commands while
 rust-analyzer runs in the background, you can also add `--build-dir build-rust-analyzer` to the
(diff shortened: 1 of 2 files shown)


RossSmyth:
May want to block merging, at least the Lua portion, until #1516 and/or #1542 are resolved

RossSmyth:
I did test the nlsp-settings config and it accepted the config.

It still has some problems, but not with the config. Mainly rust-lang/rust-analyzer#8330 and some other issue I don't know. Making RA work with rustc is a struggle. But I have autocomplete, code actions, and jumping to definitions which is better than anything I got with VSCode lol.

RossSmyth:
Actually it wasn't working, but it is because the JSON in the book is invalid. It has a trailing comma which is invalid in JSON. So it was causing nvim to send a null dict. But I fixed it and also made the steps clear as it was confusing to me (first open a buffer, then update RA settings)

JohnTitor:
This is a pre-existing problem, but the explanation of VSCode and Neovim is mixed up and difficult to understand to me. Could you break down them as a sub-section of the "Configuring rust-analyzer for rustc"?

src/building/suggested.md
```diff
@@ -49,22 +49,62 @@ you can write: <!-- date-check: nov 2022 --><!-- the date comment is for the edi
...
+                invocationStrategy = "once",
+                overrideCommand = {"python3", "x.py", "check", "--json-output"}
+            },
+            sysroot = "./build/host/stage0-sysroot"
+        },
+        rustc = {
+            source = "./Cargo.toml"
+        }
+    }
+}
```
spastorino, line 101: I'd avoid adding all this again because we already have a json file in the rust repo.

RossSmyth: Is that in there? That is the Lua config, not the JSON config. We could also put the Lua config in the Rust repo, but there's not really a good way for `x config`  to automatically configure it.

RossSmyth:
> This is a pre-existing problem, but the explanation of VSCode and Neovim is mixed up and difficult to understand to me. Could you break down them as a sub-section of the "Configuring rust-analyzer for rustc"?

Sounds good

src/building/suggested.md
RossSmyth: Could just delete it and leave only one "suggested" way to configure it. This is for people who don't want to use an extension that can read the JSON file.

RossSmyth: Could just delete it and leave only one "suggested" way to configure it. This is for people who don't want to use an extension that can read the JSON file.

spastorino: Yep, I got the intention of adding this. I'd try to suggest some docs about how to convert from the json to Lua config. I'm worried that having this duplicated information would lead us to have places outdated. From what I've been seeing RA config is changing frequenly so my fear is that things are not in sync and outdated or either in the Lua one or in the json one.

cc @jyn514

RossSmyth: Yeah I have the same worry. Especially since they aren't living in the same document anymore. I'd like to make it easier for people who don't want to use plugins, but also those people probably know how to translate the JSON in to Lua? But I don't really know.

RossSmyth: Sounds good, I'll do that when I get home.

RossSmyth:
Also I cannot build these docs on my computer (Windows). 
```powershell
PS C:\Users\Ross\Documents\rustc-dev-guide> mdbook build
2023-02-09 16:37:13 [INFO] (mdbook::book): Book building has started
2023-02-09 16:37:14 [INFO] (mdbook::book): Running the html backend
2023-02-09 16:37:16 [INFO] (mdbook::book): Running the linkcheck backend
2023-02-09 16:37:16 [INFO] (mdbook::renderer): Invoking the "linkcheck" renderer
2023-02-09 16:37:16 [ERROR] (mdbook::utils): Error: Rendering failed
2023-02-09 16:37:16 [ERROR] (mdbook::utils):    Caused By: Unable to start the backend
2023-02-09 16:37:16 [ERROR] (mdbook::utils):    Caused By: %1 is not a valid Win32 application. (os error 193)
```

jyn514 (approved):
This is great, thank you!

PR #1551, opened 2023-01-10
add section for the new trait solver

lcnr:
still not done but a hopefully useful first step.


src/solve/canonicalization.md
```diff
@@ -0,0 +1,8 @@
+# Canonicalization
+
+While the exact approach to canonicalization for this solver will differ slightly
+wrt to lifetimes, please visit [the relevant chalk chapter][chalk] for now.
+
+As of the 10th January canonicalization is not yet fully implemented for the solver.
```
BoxyUwU, line 6:
```suggestion
As of the 10th January 2023 canonicalization is not yet fully implemented for the solver.
```

src/solve/trait-solving.md
```diff
@@ -0,0 +1,99 @@
...
+Before we dive into the new solver lets first take the time to go through all of our requirements
+on the trait system. We can then use these to guide our design later on.
+
+TODO: elaborate on these rules and get more precise about their meaning.
+Also add issues where each of these rules have been broken in the past
+(or still are).
+
+### 1. The trait solver has to be *sound*
+
+This means that we must never return *success* for goals for which no `impl` exists. That would
```
BoxyUwU, line 53: This is not really accurate since the solver will return success for goals that can be proven by param env candidates that dont hold if you were to try and prove them as a caller :thinking: I get what is trying to be communicated here though and I don't know what a better way of wording it would be lol

src/solve/coinduction.md
```diff
@@ -0,0 +1,201 @@
...
+While our implementation can not check for coninduction by trying to construct an infinite
+tree as that would take infinite ressources, it still makes sense to think of coinduction
+from this perspective.
+
+As we cannot check for infinite trees, we instead search for patterns for which we know that
+they would result in an infinite proof tree. The currently pattern we detect are (canonical)
+cycles. If `T: Send` relies on `T: Send` then it's pretty clear that this will just go on forever.
+
+With cycles we have to be careful with caching. Due to canonicalization of regions and inference
+variables we also have to rerun queries until the provisional result returned when hitting the cycle
```
BoxyUwU, line 61: I don't think you mentioned "provisional result" yet at this point so i expect nobody will know what this means. It also might be worth linking to https://hackmd.io/M5vJyLMGSTaeY0n96wqEWQ somewhere or inlining it since it steps through the "rerun queries until the provisional result" stuff.

src/solve/trait-solving.md
lcnr: `param_env` candidates mean that the caller has to provide an impl so it feels fairly accurate :thinking: I guess just adding a second sentence about that would help

src/solve/canonicalization.md
```diff
@@ -0,0 +1,8 @@
+# Canonicalization
+
+While the exact approach to canonicalization for this solver will differ slightly
+wrt to lifetimes, please visit [the relevant chalk chapter][chalk] for now.
+
+As of the 10th January 2023 canonicalization is not yet fully implemented for the solver.
```
tshepang, line 6:
```suggestion
As of 10<!-- date-check --> January 2023, canonicalization is not yet fully implemented for the solver.
```
We have a tool that checks these hidden comments to ensure that we get notified if things get stale (currently set at 6 months), and it produces [this example report](https://github.com/rust-lang/rustc-dev-guide/issues/1539).

src/solve/trait-solving.md
```diff
@@ -0,0 +1,102 @@
...
+simply be unsound by assuming a trait is implemented even though it is not. When using predicates
+from the `where`-bounds, the `impl` whill be proved by the user of the item.
```
spastorino, line 55:
```suggestion
from the `where`-bounds, the `impl` will be proved by the user of the item.
```

src/solve/coinduction.md
```diff
@@ -0,0 +1,250 @@
...
+
+## Why is coinduction correct
+
+When checking whether some trait goals holds, we're asking "does there exist an `impl`
+which satisfies this bound". Even if are infinite chains of nested goals, we still have a
+unique `impl` which should be used.
+
+## How to implement coinduction
+
+While our implementation can not check for coninduction by trying to construct an infinite
```
spastorino, line 52:
```suggestion
While our implementation can not check for coinduction by trying to construct an infinite
```

src/solve/coinduction.md
```diff
@@ -0,0 +1,250 @@
...
+-  `wf(<Ready<T> as Try>::Residual)` :tada: **cycle**
+
+### Issues when extending coinduction to more goals
+
+There are some additional issues to keep in mind when extending coinduction.
+The issues here are not relevant for the current solver.
+
+#### Implied super trait bounds
+
+Our trait system currectly treats super traits, e.g. `trait Trait: SuperTrait`,
```
spastorino, line 166:
```suggestion
Our trait system currently treats super traits, e.g. `trait Trait: SuperTrait`,
```

src/solve/trait-solving.md
```diff
@@ -0,0 +1,102 @@
...
+
+### 2. If type checker solves generic goal concrete instantiations of that goal have the same result
+
+Pretty much: If we successfully typecheck a generic function concrete instantiations
+of that function should also typeck. We should not get errors post-monomorphization.
+We can however get overflow.
```
spastorino, line 61: Should we provide an example here?

src/solve/trait-solving.md
```diff
@@ -0,0 +1,102 @@
...
+
+If a trait goal holds with an empty environment, there is a unique `impl`,
+either user-defined or builtin, which is used to prove that goal.
+
+This is necessary for codegen to select a unique method.
+An exception here are *marker traits* which are allowed to overlap.
+
+### 4. Normalization in empty environments results in a unique type
+
+Normalization for alias types/consts has a unique result. Otherwise we can easily implement
```
spastorino, line 73: Also it would be maybe a good idea to explain further. Like use an example of what would happen if you were to allow multiple results

src/solve/canonicalization.md
```diff
@@ -0,0 +1,9 @@
+# Canonicalization
+
+While the exact approach to canonicalization for this solver will differ slightly
+wrt to lifetimes, please visit [the relevant chalk chapter][chalk] for now.
+
+As of 10<!-- date-check --> January 2023, canonicalization is not yet fully implemented
```
spastorino, line 6: You can place the date inside the comment if it bothers the writing style :)

src/solve/coinduction.md
```diff
@@ -0,0 +1,250 @@
...
+trait Foo {}
+struct Wrapper<T>(T);
+
+impl<T> Foo for Wrapper<Wrapper<T>>
+where
+    Wrapper<T>: Foo
+{} 
+```
+Proving `Wrapper<?0>: Foo` uses the impl `impl<T> Foo for Wrapper<Wrapper<T>>` which constrains
+`?0` to `Vec<?1>` and then requires `Wrapper<?1>: Foo`. Due to canonicalization this would be
```
spastorino, line 73:
```suggestion
`?0` to `Wrapper<?1>` and then requires `Wrapper<?1>: Foo`. Due to canonicalization this would be
```

src/solve/coinduction.md
```diff
@@ -0,0 +1,250 @@
...
+        match self {
+            List::Nil => List::Nil,
+            List::Succ(head, tail) => List::Succ(head.clone(), tail.clone()),
+        }
+    }
+}
+```
+
+We are using `tail.clone()` in this impl. For this we have to prove `Box<List<T>>: Clone`
+which requires `List<T>: Clone` but that relies on the currently impl which we are currently
```
spastorino, line 115:
```suggestion
which requires `List<T>: Clone` but that relies on the impl which we are currently
```

src/solve/coinduction.md
spastorino, line 166:
```suggestion
Our trait system currently treats super traits, e.g. `trait Trait: SuperTrait`,
```

src/solve/coinduction.md
```diff
@@ -0,0 +1,250 @@
...
+The issues here are not relevant for the current solver.
+
+#### Implied super trait bounds
+
+Our trait system currectly treats super traits, e.g. `trait Trait: SuperTrait`,
+by 1) requiring that `SuperTrait` has to hold for all types which implement `Trait`,
+and 2) assuming `SuperTrait` holds if `Trait` holds.
+
+Relying on 2) while proving 1) is unsound. This can only be observed in case of
+coinductive cycles. Without a cycles, whenever we rely on 2) we must have also
```
spastorino, line 171:
```suggestion
coinductive cycles. Without cycles, whenever we rely on 2) we must have also
```

spastorino:
@lcnr I've left some minor comments. I'm fine to merge this after you address these nits, and we can keep improving it as we go.

src/solve/trait-solving.md
lcnr: getting overflow in general is easy by having recursion with growing generic arguments, e.g.
```rust
fn foo<T: FnMut()>(mut f: T) {
    foo(&mut f)
}

fn main() {
    foo(|| ());
}
```
having this result in trait solver overflow instead of monomorphization collector overflow requires predicates whose required depth grows faster than the monomorphization depth. Don't have a good example for this right now.

src/solve/canonicalization.md
lcnr: not sure what exactly you mean here '^^ feel free to edit this section if you want :grin: i think it's a bit ugly rn but don't care too much.

lcnr:
:+1: should have dealt with all your nits, thanks

src/solve/canonicalization.md
tshepang: I would be something like...
```
<!-- date-check: jan 2023 -->
As of 10 January 2023,  canonicalization is not yet fully implemented
```

compiler-errors (approved):
Let's merge this and iterate in-tree :smiley:

PR #1553, opened 2023-01-11
Update incremental-compilation-in-detail.md

gftea:
complete the explanation of full DAG example

### DIFF
--- a/src/queries/incremental-compilation-in-detail.md
+++ b/src/queries/incremental-compilation-in-detail.md
@@ -63,7 +63,7 @@ have in the cache and, transitively, the result of `type_check_item(foo)`
 might have changed too. We thus re-run `type_check_item(foo)`, which in
 turn will re-run `type_of(bar)`, which will yield an up-to-date result
 because it reads the up-to-date version of `Hir(bar)`. Also, we re-run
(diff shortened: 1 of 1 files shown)


JohnTitor:
I'm not familiar with this, maybe @michaelwoerister is a good reviewer here?

src/queries/incremental-compilation-in-detail.md
```diff
@@ -62,7 +62,8 @@ the result of `type_of(bar)` might yield a different result than what we
 have in the cache and, transitively, the result of `type_check_item(foo)`
 might have changed too. We thus re-run `type_check_item(foo)`, which in
 turn will re-run `type_of(bar)`, which will yield an up-to-date result
-because it reads the up-to-date version of `Hir(bar)`.
+because it reads the up-to-date version of `Hir(bar)`. Also, we re-run
+`type_check_item(bar)` because result of `type_of(bar)` has changed.
```
michaelwoerister, line 66:
```suggestion
`type_check_item(bar)` because result of `type_of(bar)` might have changed.
```

We don't know if `type_of(bar)` has really changed (it might have the same value, even though one of its inputs is different). We only know that it might have changed because of the changed input.

michaelwoerister:
Thanks, @gftea! Looks good to me.

PR #1555, opened 2023-01-13
bootsrapping stages overview list

mightyiam:


mightyiam:
@tshepang @jyn514 and @JohnTitor :heart:

src/building/bootstrapping.md
```diff
@@ -18,6 +18,13 @@ rustc, then uses it to compile the new compiler.
 
 ## Stages of bootstrapping
 
+Overview:
+
+- Stage 0: the pre-compiled compiler
+- Stage 1: from current code, by an earlier compiler
+- Stage 2: the truly current compiler
+- Stage 3: the same-result test
```
ehuss, line 26: (drive-by comment) I don't think stage 3 works today, and I suspect it won't be fixed anytime soon (there doesn't appear to be much demand for it).  I might suggest just removing it from the docs altogether.

mightyiam: Remove from docs because broken and no interest? Should it be removed from implementation and then docs?

ehuss: I don't think there is anything to remove from the implementation. Rustbuild just handles "stage N will fetch things from stage N-1".  There's just a particular bug when N is too high. I don't know if it would be worth adding an explicit check in rustbuild that prevents using something that is known to be broken. There might be some unusual use cases that could be affected. @jyn514 could maybe comment on that.

jyn514: the issue @ehuss is talking about is https://github.com/rust-lang/rust/issues/90244. I would like to fix that eventually; stage 3 *is* useful if you're trying to verify reproducible builds (https://github.com/rust-lang/rust/issues/90244#issuecomment-1134983794). Seems fine to link to the issue, but I don't want to remove the docs altogether.

mightyiam:
Looking forward for a review. Feel free to merge as far as I'm concerned.

spastorino:
@jyn514 can you take a look at this?

mightyiam:
Please?

tshepang (approved):

Issue #1491, opened 2022-10-21
cleanup the remaining mentions of `rustc_typeck`


lcnr:
`rustc_typeck` has been split into `rustc_hir_analysis` and `rustc_hir_typeck`, we should update the dev guide for this. partially done in #1490

PR #1559, opened 2023-01-15
rustc_typeck exists no more

tshepang:
part of #1491

cc @lcnr

### DIFF
--- a/src/type-checking.md
+++ b/src/type-checking.md
@@ -1,8 +1,12 @@
 # Type checking
 
-The [`hir_analysis`] crate contains the source for "type collection" as well as a bunch of related functionality. Checking the bodies of functions is implemented in the [`hir_typeck`] crate. (It draws heavily on the [type inference] and [trait solving].)
+The [`hir_analysis`] crate contains the source for "type collection" as well
+as a bunch of related functionality.
+Checking the bodies of functions is implemented in the [`hir_typeck`] crate.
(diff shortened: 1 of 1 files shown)


JohnTitor (approved):
LGTM!

src/type-checking.md
```diff
@@ -1,10 +1,10 @@
 # Type checking
 
-The [`rustc_typeck`][typeck] crate contains the source for "type collection"
+The [`hir_analysis`] crate contains the source for "type collection"
 and "type checking", as well as a few other bits of related functionality. (It
 draws heavily on the [type inference] and [trait solving].)
```
lcnr, line 5: this isn't completely right as we've moved type check to `rustc_hir_typeck`.

maybe something like
```suggestion
The [`hir_analysis`] crate contains the source for "type collection" as well as a bunch of related functionality. Checking the bodies of functions is implemented in the [`hir_typeck`] crate. (It draws heavily on the [type inference] and [trait solving].)
```
```

tshepang: which of these 2 crates "draws heavily..."

JohnTitor:
Should we also correct the description of https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir_analysis/index.html? It has some description for type checking.

src/type-checking.md
lcnr: both do :grin:

tshepang:
is CI stuck

PR #1562, opened 2023-01-18
Update rust-analyzer config `checkOnSave` -> `check`

lukas-code:
This setting was renamed in https://github.com/rust-lang/rust-analyzer/pull/13799.

Turns out that was the reason my analyzer hasn't been working recently.

### DIFF
--- a/src/building/suggested.md
+++ b/src/building/suggested.md
@@ -27,7 +27,7 @@ you can write: <!-- date-check: nov 2022 --><!-- the date comment is for the edi
 
 ```JSON
 {
-    "rust-analyzer.checkOnSave.overrideCommand": [
+    "rust-analyzer.check.overrideCommand": [
         "python3",
         "x.py",
         "check",


flodiebold:
The old config keys should still work though, right @Veykril ?

Veykril:
Ye we should still be accepting the old key here 🤔

lukas-code:
This might be a rust-analyzer bug then. I just double checked with r-a default settings and the old config doesn't work, but with this patch it works.

r-a version: 0.4.1371-standalone

Veykril:
I'll look into it.

Veykril:
Found the problem https://github.com/rust-lang/rust-analyzer/pull/13980, we should still update the dev-guide to be up to date, but ideally let this PR sit for a bit, since I assume some people that work on rustc might be using the rust-analyzer shipped as a stable rustup component which doesn't have the updated configs yet.

clubby789:
The suggested config has been moved into bootstrap [here](https://github.com/rust-lang/rust/blob/master/src/etc/vscode_settings.json).

lukas-code:
I'll close this per https://github.com/rust-lang/rustc-dev-guide/pull/1562#issuecomment-1424270676. The config in the rust repo should still be updated eventually.

PR #1565, opened 2023-01-25
Update hir.md

pourplusquoi:
Fix method name and the link, as it's outdated.


tshepang:
is this right... should it not be `parent_id`

tshepang:
oh, I see it's a more convenient call

tshepang (approved):

PR #1570, opened 2023-01-28
Add section on comparing types

Nilstrieb:
For rust-lang/rust#107393

There are a few questions remaining that I would like some help one :)

### DIFF
--- a/src/ty.md
+++ b/src/ty.md
@@ -148,17 +148,35 @@ These methods all return a `Ty<'tcx>` – note that the lifetime you get back is
 arena that this `tcx` has access to. Types are always canonicalized and interned (so we never
 allocate exactly the same type twice).
 
-> N.B.
-> Because types are interned, it is possible to compare them for equality efficiently using `==`
-> – however, this is almost never what you want to do unless you happen to be hashing and looking
-> for duplicates. This is because often in Rust there are multiple ways to represent the same type,
-> particularly once inference is involved. If you are going to be testing for type equality, you
-> probably need to start looking into the inference code to do it right.
-
 You can also find various common types in the `tcx` itself by accessing its fields:
 `tcx.types.bool`, `tcx.types.char`, etc. (See [`CommonTypes`] for more.)
 
 [`CommonTypes`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/context/struct.CommonTypes.html
+
+<!-- N.B: This section is linked from the type comparison internal lint. -->
+## Comparing types
+
+Because types are interned, it is possible to compare them for equality efficiently using `==`
+– however, this is almost never what you want to do unless you happen to be hashing and looking
+for duplicates. This is because often in Rust there are multiple ways to represent the same type,
+particularly once inference is involved.
+
+For example, the type `{integer}` (an integer inference variable, the type of an integer
+literal like `0`) and `u8` should often be treated as equal when testing whether they
+can be assigned to each other (which is a common operation in diagnostics code).
+`==` on them will return `false` though, since they are different types.
+
+The simplest way to compare two types correctly requires an inference context (`infcx`).
+If you have one, you can use `infcx.can_eq(ty1, ty2)` to check whether the types can be made equal,
+so whether they can be assigned to each other.
+When working with an inference context, you have to be careful to ensure that potential inference
+variables inside the types actually belong to that inference context. If you are in a function
+that has access to an inference context already, this should be the case.
+
+Another consideration is normalization. Two types may actually be the same, but one is behind an associated type.
+To compare them correctly, you have to normalize the types first. <!-- FIXME: When do you have to worry about this? When not? -->
+
+<!-- What to do when you don't have an inference context available already? Just create one and hope all goes well? -->
 
 ## `ty::TyKind` Variants
 
--- a/src/ty.md
+++ b/src/ty.md
@@ -167,7 +167,8 @@ equal when testing whether they can be assigned to each other (which is a common
 diagnostics code). `==` on them will return `false` though, since they are different types.
 
 The simplest way to compare two types correctly requires an inference context (`infcx`).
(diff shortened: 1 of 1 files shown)


src/ty.md
```diff
@@ -148,18 +148,36 @@ These methods all return a `Ty<'tcx>` – note that the lifetime you get back is
...
+for duplicates. This is because often in Rust there are multiple ways to represent the same type,
+particularly once inference is involved.
+
+For example, the type `{integer}` (an integer inference variable, the type of an integer
+literal like `0`) and `u8` should often be treated as equal when testing whether they
+can be assigned to each other (which is a common operation in diagnostics code).
+`==` on them will return `false` though, since they are different types.
+
+The simplest way to compare two types correctly requires an inference context (`infcx`).
+If you have one, you can use `infcx.can_eq(ty1, ty2)` to check whether the types can be made equal,
```
compiler-errors, line 170: this needs a param-env

src/ty.md
```diff
@@ -148,18 +148,36 @@ These methods all return a `Ty<'tcx>` – note that the lifetime you get back is
...
+
+Because types are interned, it is possible to compare them for equality efficiently using `==`
+– however, this is almost never what you want to do unless you happen to be hashing and looking
+for duplicates. This is because often in Rust there are multiple ways to represent the same type,
+particularly once inference is involved.
+
+For example, the type `{integer}` (an integer inference variable, the type of an integer
+literal like `0`) and `u8` should often be treated as equal when testing whether they
+can be assigned to each other (which is a common operation in diagnostics code).
+`==` on them will return `false` though, since they are different types.
```
compiler-errors, line 167: perhaps mention that one is `ty::UInt(..)` and one's `ty::Infer(ty::IntVar(..))`

src/ty.md
```diff
@@ -148,18 +148,36 @@ These methods all return a `Ty<'tcx>` – note that the lifetime you get back is
...
+
+The simplest way to compare two types correctly requires an inference context (`infcx`).
+If you have one, you can use `infcx.can_eq(ty1, ty2)` to check whether the types can be made equal,
+so whether they can be assigned to each other.
```
compiler-errors, line 171:
```suggestion
If you have one, you can use `infcx.can_eq(ty1, ty2)` to check whether the types can be made equal.
This is typically what you want to check during diagnostics, which is concerned with questions such as whether two types can be assigned to each other.
```

src/ty.md
```diff
@@ -148,18 +148,36 @@ These methods all return a `Ty<'tcx>` – note that the lifetime you get back is
...
+When working with an inference context, you have to be careful to ensure that potential inference
```
compiler-errors, line 172: Maybe put this on another paragraph?

src/ty.md
```diff
@@ -148,18 +148,36 @@ These methods all return a `Ty<'tcx>` – note that the lifetime you get back is
...
+variables inside the types actually belong to that inference context. If you are in a function
```
compiler-errors, line 173: Could you mention specifically that this is the case during HIR type-check and MIR borrow-check?

src/ty.md
```diff
@@ -148,18 +148,36 @@ These methods all return a `Ty<'tcx>` – note that the lifetime you get back is
...
+that has access to an inference context already, this should be the case.
+
+Another consideration is normalization. Two types may actually be the same, but one is behind an associated type.
+To compare them correctly, you have to normalize the types first. <!-- FIXME: When do you have to worry about this? When not? -->
```
compiler-errors, line 177: This is a concern primarily during HIR typeck, or when you get a type back from a TyCtxt query, since it's unnormalized there. Possibly worth mentioning that diagnostics can use `normalize_erasing_regions` after typeck (`FnCtxt::normalize` or `ObligationCtxt::normalize` should be used before).

src/ty.md
```diff
@@ -148,18 +148,36 @@ These methods all return a `Ty<'tcx>` – note that the lifetime you get back is
...
+
+<!-- What to do when you don't have an inference context available already? Just create one and hope all goes well? -->
```
compiler-errors, line 179: Well, so this is the thing. I think it's actually fine to compare types with equality during things like late lints, since all the inference variables are resolved and all the lifetimes are erased. Similarly for post-mono code. 

Diagnostics in other places *can* create an `InferCtxt`, but it's only really meaningful if you're also substituting an `EarlyBinder` with fresh inference variables (e.g. asking the question of "can `Wrapper<T>` for any substitution of `T = _` be unified with `Wrapper<i32>`), so that should be kept in mind..

Otherwise, diagnostics code needs to try its best to funnel any `InferCtxt` that the types came from, since as you noted before, letting infer vars leak from one infcx to another causes issues.

src/ty.md
compiler-errors: ^

src/ty.md
```diff
@@ -148,18 +148,61 @@ These methods all return a `Ty<'tcx>` – note that the lifetime you get back is
...
+that has access to an inference context already, this should be the case. Specifically, this is the
+case during HIR type checking or MIR borrow checking.
+
+Another consideration is normalization. Two types may actually be the same, but one is behind an
+associated type. To compare them correctly, you have to normalize the types first. This is
+primarily a concern during HIR type checking and with all types from a `TyCtxt` query
+(for example from `tcx.type_of()`).
+
+When a `FnCtxt` or an `ObligationCtxt` is available during type checking, `.normalize(ty)`
+can be used on them to normalize the type. After type checking, diagnostics code can use
```
compiler-errors, line 186:
```suggestion
should be used on them to normalize the type. After type checking, diagnostics code can use
```

compiler-errors (approved):
r=me other than one word nit and the fact that `InferCtxt::can_eq` is still missing a param-env.

src/ty.md
Nilstrieb: 💀

tshepang (approved):

PR #1574, opened 2023-02-01
Update rustfmt path

zephaniahong:
As per https://github.com/rust-lang/rust/pull/107297#discussion_r1092764740, the change broke the rust-analyzer config. Hence, changing the docs to match the new path

### DIFF
--- a/src/building/suggested.md
+++ b/src/building/suggested.md
@@ -34,7 +34,7 @@ you can write: <!-- date-check: nov 2022 --><!-- the date comment is for the edi
         "--json-output"
     ],
     "rust-analyzer.rustfmt.overrideCommand": [
-        "./build/host/stage0/bin/rustfmt",
+        "./build/host/rustfmt/bin/rustfmt",
         "--edition=2021"
     ],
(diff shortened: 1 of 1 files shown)


albertlarsan68 (approved):

JohnTitor:
At least @jyn514 wants to revert the change if possible, it seems. I'd hold off until their comment gets a response if so.

apiraino:
In addition to that, I think a breaking change of this bit of rust-analyzer configuration should probably need ample visibility in advance.

jyn514:
> At least @jyn514 wants to revert the change if possible, it seems. I'd hold off until their comment gets a response if so.

It turns out this is non-trivial to do since it means we'd also have to revert https://github.com/rust-lang/rust/pull/105609.

https://github.com/rust-lang/rust/issues/107547#issuecomment-1412980809, I'm ok with merging this PR as-is. Ideally we would have some way to test bootstrap ... cc https://github.com/rust-lang/rust/issues/102563

jyn514 (approved):

PR #1550, opened 2023-01-10
Add a citation file

JohnTitor:
Fixes #1534
This places a CFF file following https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files to add citation metadata.
Some data is taken from https://github.com/rust-lang/rustc-dev-guide/blob/master/book.toml.
The release date is taken from https://github.com/rust-lang/rustc-dev-guide/commit/27ec0dc6d6e5afb7a4848ed00b457d24ec12b8f3.

cc @rust-lang/wg-rustc-dev-guide I'm not familiar with citations at all, any feedback is greatly welcome!

Signed-off-by: Yuki Okushi <jtitor@2k36.org>

PR #1576, opened 2023-02-02
a better name

tshepang:


JohnTitor:
Why don't you suggest the change on https://github.com/rust-lang/rustc-dev-guide/pull/1550? This would bring a merge conflict and it's needless work for me if you suggest on the above PR :(

tshepang:
Oh, was not aware the other PR was changing same... forgot/missed it.

JohnTitor:
I see! Thanks for suggesting anyway <3

PR #1577, opened 2023-02-02
Fix broken links

JohnTitor:
I'd recommend reviewing commit by commit.


src/mir/visitor.md
```diff
@@ -37,10 +37,10 @@ code that will execute whenever a `foo` is found. If you want to
 recursively walk the contents of the `foo`, you then invoke the
 `super_foo` method. (NB. You never want to override `super_foo`.)
 
-A very simple example of a visitor can be found in [`LocalUseCounter`].
-By implementing `visit_local` method, this visitor counts how many times each local is used.
+A very simple example of a visitor can be found in [`LocalUseVisitor`].
+By implementing `visit_local` method, this visitor counts how many times each local is mutably used.
 
-[`LocalUseCounter`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_mir_transform/simplify_try/struct.LocalUseCounter.html
+[`LocalUseVisitor`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_mir_transform/const_debuginfo/struct.LocalUseVisitor.html
```
tshepang, line 43: why the choice of this over the other type

tshepang:
just one question, otherwise looks all good

sidenote: would it be alright to split these kinds into separate PRs, for it can take a while to review the whole thing at once

src/mir/visitor.md
JohnTitor: See https://github.com/rust-lang/rust/pull/107256, which is mentioned in the commit message of this change, the struct was removed to simplify, we have to find another example to show how `visit_local` is used.

JohnTitor:
> sidenote: would it be alright to split these kinds into separate PRs in future, for it can take a while to review the whole thing at once

To make reviews easier, I've split the commits with the culprit PRs' links, and there are only 6 commits with 10 or so diff lines. I don't think it's supposed to take hours to review...

tshepang (approved):

tshepang:
I was reviewing commit-by-commit (as suggested), but only realized later that links are added to the commit messages... such are rather easy to miss with GitHub UI. Thanks for making it easy.

JohnTitor:
I see, I'm going to mention it on OP explicitly next time then! I'd like to try splitting if the context is complex and needs some time/eyes to review 👍

Issue #1581, opened 2023-02-03
Update the rustc-driver examples to play with the latest nightly


Hiroki6:
I tried to compile the examples in `rustc 1.69.0-nightly (f3126500f 2023-02-02)`

```
$ rustup -V
rustup 1.25.2 (17db695f1 2023-02-01)
info: This is the version for the rustup toolchain manager, not the rustc compiler.
info: The currently active `rustc` version is `rustc 1.69.0-nightly (f3126500f 2023-02-02)`

$ rustc +nightly rustc-driver-interacting-with-the-ast.rs
$ rustc +nightly rustc-driver-getting-diagnostics.rs
$ rustc +nightly rustc-driver-example.rs
```

All of them raised errors
```
error[E0560]: struct `rustc_interface::Config` has no field named `input_path`
  --> rustc-driver-example.rs:49:9
   |
49 |         input_path: None,  // Option<PathBuf>
   |         ^^^^^^^^^^ `rustc_interface::Config` does not have this field
   |
   = note: available fields are: `opts`, `crate_cfg`, `crate_check_cfg`, `input`, `output_dir` ... and 8 others

error[E0599]: `rustc_interface::queries::QueryResult<'_, rustc_ast::ast::Crate>` is not an iterator
    --> rustc-driver-example.rs:74:50
     |
74   |             let parse = queries.parse().unwrap().take();
     |                                                  ^^^^ `rustc_interface::queries::QueryResult<'_, rustc_ast::ast::Crate>` is not an iterator

error[E0599]: `rustc_interface::queries::QueryResult<'_, rustc_interface::passes::QueryContext<'_>>` is not an iterator
    --> rustc-driver-example.rs:77:44
     |
77   |             queries.global_ctxt().unwrap().take().enter(|tcx| {
     |                                            ^^^^ `rustc_interface::queries::QueryResult<'_, rustc_interface::passes::QueryContext<'_>>` is not an iterator

error: aborting due to 3 previous errors
```

- `input_path` was removed from Config by the commit https://github.com/rust-lang/rust/pull/106810/commits/f5c601492ee520d2ee4f6c133f7f4dfa7b0c13e2.
- `global_ctxt` now returns `QueryResult` which doesn't have the `take` method by the commit https://github.com/rust-lang/rust/pull/105603/commits/58782a8842bdd74e3304092251a1d06e5b6b550b.

JohnTitor:
[...]

Issue #1556, opened 2023-01-13
fail to compile the examples


gftea:
I tried to compile the examples below

```bash
rustup +nightly component add rust-src rustc-dev llvm-tools-preview
rustc +nightly rustc-driver-interacting-with-the-ast.rs 
```

I got below errors

```bash
error[E0433]: failed to resolve: could not find `DiagnosticOutput` in `rustc_session`
  --> rustc-driver-interacting-with-the-ast.rs:46:43
   |
46 |         diagnostic_output: rustc_session::DiagnosticOutput::Default,
   |                                           ^^^^^^^^^^^^^^^^ could not find `DiagnosticOutput` in `rustc_session`

error[E0560]: struct `rustc_interface::Config` has no field named `diagnostic_output`
  --> rustc-driver-interacting-with-the-ast.rs:46:9
   |
46 |         diagnostic_output: rustc_session::DiagnosticOutput::Default,
   |         ^^^^^^^^^^^^^^^^^ `rustc_interface::Config` does not have this field
   |
   = note: available fields are: `opts`, `crate_cfg`, `crate_check_cfg`, `input`, `input_path` ... and 9 others

error: aborting due to 2 previous errors

```

JohnTitor:
As noted in https://rustc-dev-guide.rust-lang.org/rustc-driver-interacting-with-the-ast.html#getting-the-type-of-an-expression:
> To get the type of an expression, use the global_ctxt to get a TyCtxt. The following was tested with nightly-2022-06-05 (see [here](https://github.com/rust-lang/rustc-dev-guide/blob/master/examples/rustc-driver-interacting-with-the-ast.rs) for the complete example):

A PR is welcome to update the nightly version :)

gftea:
`DiagnosticOutput`  is removed from `Config` by commit "641f8249f99b407af7e5376b098323926eab1696", 2 months ago

PR submitted

PR #1373, opened 2022-06-21
make clear that other versions can work

tshepang:

PR #1582, opened 2023-02-04
Highlight the nightly version needed to compile the examples

JohnTitor:
Readers didn't notice the examples need a specific nightly version. Indeed other versions _can_ work, but it introduces some confusion like https://github.com/rust-lang/rustc-dev-guide/issues/1556 and #1581. This emphasizes the nightly version to use.
A partial revert of https://github.com/rust-lang/rustc-dev-guide/pull/1373.

### DIFF
--- a/examples/rustc-driver-example.rs
+++ b/examples/rustc-driver-example.rs
@@ -3,6 +3,7 @@
 // NOTE: For the example to compile, you will need to first run the following:
 //   rustup component add rustc-dev llvm-tools-preview
 
+// You need the below nightly version to compile:
 // version: rustc 1.68.0-nightly (935dc0721 2022-12-19)
 
 extern crate rustc_error_codes;
--- a/examples/rustc-driver-getting-diagnostics.rs
+++ b/examples/rustc-driver-getting-diagnostics.rs
@@ -3,6 +3,7 @@
 // NOTE: For the example to compile, you will need to first run the following:
 //   rustup component add rustc-dev llvm-tools-preview
 
+// You need the below nightly version to compile:
 // version: rustc 1.68.0-nightly (935dc0721 2022-12-19)
 
 extern crate rustc_error_codes;
--- a/examples/rustc-driver-interacting-with-the-ast.rs
+++ b/examples/rustc-driver-interacting-with-the-ast.rs
@@ -3,6 +3,7 @@
 // NOTE: For the example to compile, you will need to first run the following:
 //   rustup component add rustc-dev llvm-tools-preview
 
+// You need the below nightly version to compile:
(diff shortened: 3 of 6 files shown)


tshepang:
To me, reports like that just indicate that we should update the code so that it builds again. I would not call it confusion. I also think we should use CI to preempt such.

One concern with this change is it may make someone install some old nightly without a need.

JohnTitor:
> To me, reports like that just indicate that we should update the code so that it builds again (with latest nightly)

While I agree it should be up-to-date always, the internal API can be changed at any time and it may require updating so often. Our maintenance hands are limited, you can see untriaged date checks and a lot of open issues as proof of it. So pinning the nightly version and suggesting it should make some sense, we could make sure the examples work and readers can know how it works at least.

> One concern with this change is it may make someone install some old nightly without a need.

What's the problem with it at all? rustup can help it and users can install it easily. Also, the examples already need to install additional components like rustc-dev, so I don't believe it makes much difference.

JohnTitor:
To begin with, the purpose of the example is to let the reader intuitively know how it works. Forcing readers to report every time it breaks would hinder that.
What's more, date-checks already encourage us to update regularly.

tshepang:
> > To me, reports like that just indicate that we should update the code so that it builds again (with latest nightly)
> 
> While I agree it should be up-to-date always, the internal API can be changed at any time and it may require updating so often. Our maintenance hands are limited, you can see untriaged date checks and a lot of open issues as proof of it. So pinning the nightly version and suggesting it should make some sense, we could make sure the examples work and readers can know how it works at least.

The old text already makes it clear which nightly is known to work, so that should cover it while we have not got to update the code.

> > One concern with this change is it may make someone install some old nightly without a need.
> 
> What's the problem with it at all? rustup can help it and users can install it easily. Also, the examples already need to install additional components like rustc-dev, so I don't believe it makes much difference.

It is waste, especially given how large toolchain downloads are. The new text is also misleading... it makes it look like no other nightly works.

JohnTitor:
> The old text already makes it clear which nightly is known to work, so that should cover it while we have not got to update the code.

The current results in opening the issues linked above, I wouldn't call it clear in the face of such a situation.

> It is waste, especially given how large toolchain downloads are.

That is a problem with the toolchain itself, not with running the example on the old nightly. Repeating myself but it already increases the size by using rustc-dev/llvm-tools components.

> The new text is also misleading... it makes it look like no other nightly works.

It _can_, but how does the reader know the other versions? And how can we help readers by providing that information?
Again, the primary goal here is to get the reader to actually compile and run the example. There seems to be no problem in showing only one nightly version to accomplish this.

JohnTitor:
The PR here is meaningless according to @tshepang, closing.

PR #1588, opened 2023-02-07
Replace settings.json with x.py setup note

clubby789:
https://github.com/rust-lang/rust/pull/107757 moves the recommended `settings.json` setup into `x.py setup`, so update the dev guide accordingly.

cc @jyn514

### DIFF
--- a/src/building/suggested.md
+++ b/src/building/suggested.md
@@ -22,48 +22,17 @@ You can also install the hook as a step of running `./x.py setup`!
 `rust-analyzer` can help you check and format your code whenever you save
 a file. By default, `rust-analyzer` runs the `cargo check` and `rustfmt`
 commands, but you can override these commands to use more adapted versions
-of these tools when hacking on `rustc`. For example, for Visual Studio Code,
-you can write: <!-- date-check: nov 2022 --><!-- the date comment is for the edition below -->
-
-```JSON
-{
-    "rust-analyzer.checkOnSave.overrideCommand": [
-        "python3",
-        "x.py",
-        "check",
-        "--json-output"
-    ],
-    "rust-analyzer.rustfmt.overrideCommand": [
-        "./build/host/rustfmt/bin/rustfmt",
-        "--edition=2021"
-    ],
-    "rust-analyzer.procMacro.server": "./build/host/stage0/libexec/rust-analyzer-proc-macro-srv",
-    "rust-analyzer.procMacro.enable": true,
-    "rust-analyzer.cargo.buildScripts.enable": true,
-    "rust-analyzer.cargo.buildScripts.invocationLocation": "root",
-    "rust-analyzer.cargo.buildScripts.invocationStrategy": "once",
-    "rust-analyzer.cargo.buildScripts.overrideCommand": [
-        "python3",
-        "x.py",
-        "check",
-        "--json-output"
-    ],
-    "rust-analyzer.cargo.sysroot": "./build/host/stage0-sysroot",
-    "rust-analyzer.rustc.source": "./Cargo.toml",
-}
-```
-
-in your `.vscode/settings.json` file. This will ask `rust-analyzer` to use
-`./x.py check` to check the sources, and the stage 0 rustfmt to format them.
+of these tools when hacking on `rustc`. For example, `x.py setup` will prompt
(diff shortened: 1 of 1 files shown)


src/building/suggested.md
```diff
@@ -22,47 +22,17 @@ You can also install the hook as a step of running `./x.py setup`!
...
-        "--json-output"
-    ],
-    "rust-analyzer.cargo.sysroot": "./build/host/stage0-sysroot",
-    "rust-analyzer.rustc.source": "./Cargo.toml",
-}
-```
-
-in your `.vscode/settings.json` file. This will ask `rust-analyzer` to use
-`./x.py check` to check the sources, and the stage 0 rustfmt to format them.
+of these tools when hacking on `rustc`. If using Visual Studio Code, `x.py setup`
```
jyn514, line 25: I don't think this is true? It will do this unconditionally even if you're using vim, no?

(I would like it to support all sorts of configs, including for vim, but I don't want to block your change on it.)

jyn514 (commented):
This looks great, thanks! I would like to wait until the rust-lang/rust PR merges before merging this but otherwise LGTM :)

jyn514 (approved):

src/building/suggested.md
clubby789: Would
> For example, `x.py setup` will prompt you to create a `.vscode/settings.json` file which will configure Visual Studio code.

be better here?

jyn514: yes, that sounds great :)

src/building/suggested.md
```diff
@@ -22,47 +22,17 @@ You can also install the hook as a step of running `./x.py setup`!
...
+stage 0 rustfmt to format them.
 
 If you have enough free disk space and you would like to be able to run `x.py` commands while
 rust-analyzer runs in the background, you can also add `--build-dir build-rust-analyzer` to the
 `overrideCommand` to avoid x.py locking.
 
 If you're running `coc.nvim`, you can use `:CocLocalConfig` to create a
-`.vim/coc-settings.json` and enter the same settings as above, but replacing
-`editor.formatOnSave: true,` with
+`.vim/coc-settings.json` and copy the settings from [this file](https://github.com/rust-lang/rust/blob/master/src/etc/vscode_settings.json), but replacing `editor.formatOnSave: true,` with
```
spastorino, line 35: there's `no editor.formatOnSave: true` in https://github.com/rust-lang/rust/blob/master/src/etc/vscode_settings.json

clubby789: Looks like this was removed from the dev guide in f423a3ad9571a6dbebc7e65849ef5c08672cf295 but this note was never updated, I'll fix this now

RalfJung:
I think this is unfortunate; having the file content available on a website made it much easier to update the file when changes are required (which happens fairly regularly, 1-2 times a year): when things behave strangely, I would just go to https://rustc-dev-guide.rust-lang.org/building/suggested.html#configuring-rust-analyzer-for-rustc and see what is different from my config.

Doing `./x.py setup` instead doesn't work, it asks me a question instead of just printing the desired vscode settings:
```
Welcome to the Rust project! What do you want to do with x.py?
a) library: Contribute to the standard library
b) compiler: Contribute to the compiler itself
c) codegen: Contribute to the compiler, and also modify LLVM or codegen
d) tools: Contribute to tools which depend on the compiler, but do not modify it directly (e.g. rustdoc, clippy, miri)
e) user: Install Rust from source
Please choose one (a/b/c/d/e): 
```
The correct answer for me is "a, b, c, d". :joy: Also my config.toml is fine, I don't want it to be changed please.

How can I get the recommended vscode settings without having the tool change my config.toml?

jyn514:
@RalfJung you can run `cp src/etc/vscode_settings.json .VSCode/settings.json`. I'm happy to make a note of that in the docs, are you interested in making a PR with that change? ;)

Separately, `setup` is doing enough things now that I think allowing things like `setup hook` and `setup vscode` makes sense so you don't have to go through the whole flow.

RalfJung:
> you can run cp src/etc/vscode_settings.json .VSCode/settings.json

No that won't work, my settings.json contains a bunch of custom stuff as well.

Maybe the wiki page could just link to https://github.com/rust-lang/rust/blob/master/src/etc/vscode_settings.json ?

PR #1594, opened 2023-02-11
use actual names

tshepang:


JohnTitor (approved):
LGTM!

spastorino:
I don't like this change that much to be honest. First of all because in the rest of the summary we refer to things using their names in english and not to the crates names.
Maybe if we were to improve things we could do something like:

`The Rustc Driver (rustc_driver) and Interface (rustc_interface)`

But if so, I'd do the same to every section. Maybe that turns being noisy and in that case, we can have a paragraph on every section saying that this thing lives in rustc_something.

cc @jyn514

tshepang:
It doesn't feel like it follows that the English name of a crate is just the upper-cased version of it, minus the `rustc_` prefix. It would not apply cleanly, especially since a lot of the names are generic english/technical names (driver and interface). That's why I prefer using actual crate names, and that way we need not concoct a translation system (from crate to english).

A similar thing that happened is someone using the word _structure_ when talking about _struct_ keyword.

tshepang:
I found one exception, which is `borrowck`, though maybe it's not much of an exception, since _borrow checker_ is (I think) a unique thing to Rust. That is, we can use _borrow checker_ to refer to the crate `borrowck` (though that can also apply to `polunius`).

PR #1595, opened 2023-02-13
new solver: write canonicalization chapter

lcnr:
still have to explain canonicalization in-depth. Before that I want to check whether the current way this is written is even helpful.

cc @rust-lang/initiative-trait-system-refactor

### DIFF
--- a/src/solve/canonicalization.md
+++ b/src/solve/canonicalization.md
@@ -76,7 +76,7 @@ TODO: link to code once the PR lands and elaborate
     considering universes
 - generic parameters in the input get treated as placeholders in the root universe
 - all regions in the input get all mapped to existentially bound vars and we "uniquify" them.
-    `T: Trait<'a, 'a>` gets canonicalized to `exists<'0, '1> T: Trait<'0, '1>`. We do not care
+    `&'a (): Trait<'a>` gets canonicalized to `exists<'0, '1> &'0 (): Trait<'1>`. We do not care
     about their universes and simply put all regions into the highest universe of the input.
 - once we collected all canonical vars we compress their universes, see comment in `finalize`.
 - in the output everything in a universe of the caller gets put into the root universe and only
--- a/src/solve/canonicalization.md
+++ b/src/solve/canonicalization.md
@@ -30,8 +30,8 @@ We now call the canonical query with the canonical goal.
 
 ### Instantiating the canonical goal inside of the query
 
-To actually try to prove the canonical goal we start by instantiating existential variables with
-inference variables and universal variables without inference variables and placeholders again.
+To actually try to prove the canonical goal we start by instantiating the bound variables with
+inference variables and placeholders again.
 
 This happens inside of the query in a completely separate `InferCtxt`. Inside of the query we
 now have a goal `u32: Trait<?0>`. We also remember which value we've used to instantiate the bound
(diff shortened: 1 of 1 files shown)


src/solve/canonicalization.md
```diff
@@ -1,10 +1,74 @@
...
+We remember the original values of all bound variables in the original context. Here this would
+map `T` back to `?x`. These original values are used later on when dealing with the query
+response.
+
+We now call the canonical query with the canonical goal.
+
+### Instantiating the canonical goal inside of the query
+
+To actually try to prove the canonical goal we start by instantiating existential variables with
+inference variables and universal variables without inference variables and placeholders again.
```
lcnr, line 34: that's wrong... It would have to be "existentially bound variables" but that's a lot :sweat_smile: idk what's a good wording here

BoxyUwU, line 34:
```suggestion
To actually try to prove the canonical goal we start by instantiating existential variables with
inference variables and universal variables with placeholders
```
?

src/solve/canonicalization.md
```diff
@@ -1,10 +1,74 @@
...
+variables in the canonical goal, which maps `T` to `?0`.
+
+We now compute the goal `u32: Trait<?0>` and figure out that this holds, but we've constrained
+`?0` to `Vec<?1>`. We finally convert this result to something useful to the caller.
+
+### Canonicalizing the query response
+
+The caller has to know both the result of the query and any inference constraints from inside
+of the query. The result is already context independent, so we can return that as is, we do have
+to canonicalize the inference constraints though.
```
BoxyUwU, line 47: The distinction between "result of the query" and "any inference constraints" is unclear to me and made this confusing to read. "the result is context independent but we do have to canonicalize part of the result even though we just said it was context independent"

src/solve/canonicalization.md
```diff
@@ -1,10 +1,74 @@
...
+It now equates the original value of `T` (`?x`) with the value for `T` in the
+response (`Vec<?z>`), which correctly constrains `?x` to `Vec<?z>`.
+
+## `ExternalConstraints`
+
+Computing a trait goal may not only constrain inference variables, it can also add region
+obligations, e.g. given a goal `(): AOutlivesB<'a, 'b>` we would like to return the fact that
+`'a: 'b` has to hold.
+
+This is done by not only returning the canonical `var_values` from the query but also extracting
```
BoxyUwU, line 68: I dont think `var_values` has been mentioned yet, should be explicitly clear somewhere that the previously mentioned "mapping from `T` to `exists<U> Vec<U>`" is `var_values`

BoxyUwU: Would it make sense to avoid bringing up the exact struct field names here and just talk about it in a high level way like the rest of this document? Maybe leaving the exact impl details to the "how exactly does canonicalization work"?

spastorino:
@lcnr I'm not sure if you wanted to merge this or leave it as a WIP. But I think it's fine if we merge and we iterate over it with this chapter living on master.

lcnr:
alright, ready for merge then :grin:

src/solve/canonicalization.md
```diff
@@ -1,10 +1,84 @@
...
 
-[chalk]: https://rust-lang.github.io/chalk/book/canonical_queries/canonicalization.html#canonicalization
\ No newline at end of file
+Let's first go over the way canonical queries work and then dive into the specifics of
+how canonicalization works.
+
+## A walkthrough of canonical queries
+
+We're going to use the goal `u32: Trait<?x>` as an example and assume that this only holds
+for `u32: Trait<Vec<?z>>` where `?z` is unconstrained.
```
BoxyUwU, line 17: Would it maybe make sense to say "this only holds for `for<T> u32: Trait<Vec<T>>`" or something to make it clear that its not that there is only a trait impl _specifically_ for `?z` but that it holds for any type argument to `Vec`. Or maybe even just directly say "assumes there is only an `impl<T> Trait<Vec<T>> for u32 {`

BoxyUwU (approved):

src/solve/canonicalization.md
lcnr: done

PR #1605, opened 2023-02-17
lower-case "Compiler" in headings, for consistency (and looks)

tshepang:


camelid (commented):
Hmm, if we're going to do this, I think we should also lowercase the other titles (e.g. "Suggested Workflows" or "Sanitizers").

But I actually wonder if we should just capitalize all the titles since title case is standard for chapter names?

tshepang:
I wanted to keep it small and just start with being consistent with where "compiler" is found. I also prefer non-title case (just aesthetics), and I think there was a PR that moved headings away from title case.

src/building/bootstrapping.md
```diff
@@ -90,7 +90,7 @@ because one must first build the new compiler with an older compiler
 and then use that to build the new compiler with itself.
 For development, you usually only want the `stage1` compiler,
 which you can build with `./x.py build library`.
-See [Building the Compiler](./how-to-build-and-run.html#building-the-compiler).
+See [Building the compiler](./how-to-build-and-run.html#building-the-compiler).
```
camelid, line 93: Could you lowercase the header at the link destination too?

camelid (approved):
Thanks! One small change and then this is good to go.

PR #1606, opened 2023-02-17
rustc_codegen_ssa feels permanent enough not to need date-check

tshepang:


src/backend/backend-agnostic.md
```diff
@@ -2,11 +2,12 @@
 
 <!-- toc -->
 
-As of <!-- date-check --> Aug 2022, `rustc_codegen_ssa` provides an
-abstract interface for all backends to implement, to allow other codegen
-backends (e.g. [Cranelift]).
+[`rustc_codegen_ssa`]
+provides an abstract interface for all backends to implement,
+to allow other codegen backends (e.g. [Cranelift]).
```
camelid, line 7: While we're at it, what do you think about changing the wording to this?

```suggestion
[`rustc_codegen_ssa`]
provides an abstract interface for all backends to implement,
enabling the creation of new backends (e.g. [Cranelift]).
```

src/backend/backend-agnostic.md
```diff
@@ -2,11 +2,12 @@
...
 
 [Cranelift]: https://github.com/bytecodealliance/wasmtime/tree/HEAD/cranelift
```
camelid, line 9: What do you think about changing this to link to Rust's cranelift backend rather than the wasmtime repo? I think in its existing state it might be a bit confusing.

src/backend/backend-agnostic.md
```diff
@@ -2,11 +2,12 @@
...
+[`rustc_codegen_ssa`]: https://doc.rust-lang.org/stable/nightly-rustc/rustc_codegen_ssa/index.html
```
camelid, line 10: Might be better to link to the latest docs.

```suggestion
[`rustc_codegen_ssa`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_codegen_ssa/index.html
```

camelid (approved):
Thanks! This looks good; I just have a couple of small suggestions while we're at it.

src/backend/backend-agnostic.md
tshepang: how about "to allow code sharing (e.g. with [Cranelift])"

tshepang: Actually that doesn't sound right. I think we should just remove the part starting with "enabling", or just mention all 3 backends?

src/backend/backend-agnostic.md
tshepang: I thought that would be a better link target as well. I do it.

tshepang: see if this is resolved with https://github.com/rust-lang/rustc-dev-guide/pull/1606/commits/dfb468a52fb606b6647e7240756f2c399654f179

src/backend/backend-agnostic.md
tshepang: see if this is resolved with https://github.com/rust-lang/rustc-dev-guide/pull/1606/commits/dfb468a52fb606b6647e7240756f2c399654f179

camelid (approved):
Thanks again!

PR #1610, opened 2023-02-17
Bump dependencies to fix CI

camelid:
See https://github.com/rust-lang/rustc-dev-guide/actions/runs/4207763764/jobs/7302956059.


JohnTitor (approved):
Thanks 👍

PR #1615, opened 2023-02-18
bootstrap now creates a usable toolchain when building stage0 std

the8472:
https://github.com/rust-lang/rust/issues/101691


jyn514 (approved):
❤️ 

@jdno could we also change the scripts in the cloud dev machines to link this toolchain as well?

PR #1639, opened 2023-03-09
Update labels overview

ehuss:
This updates the overview of the different kinds of labels to add some missing entries, fix some incorrect ones, and to rearrange it into a table which should hopefully make it a little easier to read.

Since GitHub's markdown preview does not allow style tags, this is a screenshot of what it looks like rendered with mdbook:

image


JohnTitor (approved):
Great, thank you!

PR #1646, opened 2023-03-16
Add chapter on fuzzing

### LINKED ISSUE #1645: Guide for fuzzer-users
There's quite a lot of documentation on how to build, run, and debug the compiler. That's good! However, there's a notable subsection of compiler contributors that needs special direction:

People who bring in a fuzzer and make rustc crab rave until the next ICE Age.

And as Rust's profile increases, more of them show up, as more and more people use the Rust compiler in their research projects, and fuzzing happens to have a lot of interesting computer science research attached to it. And they open issues. Lots of issues. Our own @matthiaskrgr has opened _over 500_ issues. And over 400 of them have been fixed!

Almost-inevitably the first few issues include at least one "total whiff": it's code that produces an ICE, sure. However, it was generated with something like a specific compilation flag combined with internally unstable features used to implement the core library, to produce some ICE that we probably will not bother to fix, despite the local enthusiasm for dropping ICE on GitHub. Usually after a bit of feedback they dial in on issues that are a bit "higher value", but having to have that back-and-forth with each person feels like it's going to involve a lot of repetition. If they're mostly-automating bug-finding, it seems reasonable to automate (or at least cache in documentation) some of our responses. This documentation thus may cover more "how to file an issue that is useful" but still.

In particular, for such documentation, we need to capture the differences in how best to go about. "fuzzing the compiler" and "fuzzing the nightly compiler" and "fuzzing the internal features of the compiler". We don't want people to never fuzz things that require internal features, because something like "the MIR validator breaks on trying to validate some unusual MIR" is really useful to find out, even though that's an internal feature. But if the code requires misusing magic attributes that change layout or use an internal intrinsic outside its parameters, we probably will shrug that off. Likewise there's "fuzzing a nightly feature" vs. "fuzzing a nightly feature that had its prototype skeleton impl with 100 `todo!()`s landed yesternight".

Pinging Matthias because perhaps he may have some thoughts on what he would say to his clone and/or to himself after traveling back in time.

langston-barrett:
Fixes #1645.

cc @workingjubilee and @matthiaskrgr

### DIFF
--- a/src/fuzzing.md
+++ b/src/fuzzing.md
@@ -22,11 +22,12 @@ project, please read this guide before reporting fuzzer-generated bugs!
 - Include all of the information requested in the bug report template
 - Search for existing reports with the same message and query stack
 - Format the test case with `rustfmt`, if it maintains the bug
+- Indicate that the bug was found by fuzzing
 
 *Please don't:*
 
 - Report lots of bugs that use internal features, including but not limited to
-  `custom_mir`, `lang_items`, `no_std`, and `rustc_attrs`.
+  `custom_mir`, `lang_items`, `no_core`, and `rustc_attrs`.
 - Seed your fuzzer with inputs that are known to crash rustc (details below).
 
 ### Discussion
@@ -34,7 +35,30 @@ project, please read this guide before reporting fuzzer-generated bugs!
 If you're not sure whether or not an ICE is a duplicate of one that's already
 been reported, please go ahead and report it and link to issues you think might
 be related. In general, ICEs on the same line but with different *query stacks*
-are usually distinct bugs.
+are usually distinct bugs. For example, [#109020][#109202] and [#109129][#109129]
+had similar error messages:
+
+```
+error: internal compiler error: compiler/rustc_middle/src/ty/normalize_erasing_regions.rs:195:90: Failed to normalize <[closure@src/main.rs:36:25: 36:28] as std::ops::FnOnce<(Emplacable<()>,)>>::Output, maybe try to call `try_normalize_erasing_regions` instead
+```
+```
+error: internal compiler error: compiler/rustc_middle/src/ty/normalize_erasing_regions.rs:195:90: Failed to normalize <() as Project>::Assoc, maybe try to call `try_normalize_erasing_regions` instead
+```
+but different query stacks:
+```
+query stack during panic:
+#0 [fn_abi_of_instance] computing call ABI of `<[closure@src/main.rs:36:25: 36:28] as core::ops::function::FnOnce<(Emplacable<()>,)>>::call_once - shim(vtable)`
+end of query stack
+```
+```
+query stack during panic:
+#0 [check_mod_attrs] checking attributes in top-level module
+#1 [analysis] running analysis passes on this crate
+end of query stack
+```
+
+[#109020]: https://github.com/rust-lang/rust/issues/109020
+[#109129]: https://github.com/rust-lang/rust/issues/109129
 
 ## Building a corpus
 
@@ -56,19 +80,19 @@ To build a corpus, you may want to use:
 
 Here are a few things you can do to help the Rust project after filing an ICE.
 
-- Add the minimal test case to [Glacier][glacier]
 - [Bisect][bisect] the bug to figure out when it was introduced 
 - Fix unrelated problems with the test case (things like syntax errors or
   borrow-checking errors)
 - Minimize the test case (see below)
+- Add the minimal test case to [Glacier][glacier]
 
 [bisect]: https://github.com/rust-lang/cargo-bisect-rustc/blob/master/TUTORIAL.md
 
 ## Minimization
 
-It can be helpful to *minimize* the fuzzer-generated input. When minimizing, be
-careful to preserve the original error, and avoid introducing distracting
-problems such as syntax, type-checking, or borrow-checking errors.
+It is helpful to carefully *minimize* the fuzzer-generated input. When
+minimizing, be careful to preserve the original error, and avoid introducing
+distracting problems such as syntax, type-checking, or borrow-checking errors.
 
 There are some tools that can help with minimization. If you're not sure how
 to avoid introducing syntax, type-, and borrow-checking errors while using
@@ -86,10 +110,13 @@ these tools, post both the complete and minimized test cases. Generally,
 When fuzzing rustc, you may want to avoid generating code, since this is mostly
 done by LLVM. Try `--emit=mir` instead.
 
-A variety of compiler flags can uncover different issues.
+A variety of compiler flags can uncover different issues. `-Zmir-opt=4` will
+turn on MIR optimization passes that are not run by default, potentially
+uncovering interesting bugs.
 
 If you're fuzzing a compiler you built, you may want to build it with `-C
-target-cpu=native` to squeeze out a few more executions per second.
+target-cpu=native` or even PGO/BOLT to squeeze out a few more executions per
+second.
 
 ## Existing projects
 
--- a/src/fuzzing.md
+++ b/src/fuzzing.md
@@ -26,9 +26,10 @@ project, please read this guide before reporting fuzzer-generated bugs!
 
 *Please don't:*
 
-- Report lots of bugs that use internal features, including but not limited to
-  `custom_mir`, `lang_items`, `no_core`, and `rustc_attrs`.
-- Seed your fuzzer with inputs that are known to crash rustc (details below).
+- Don't report lots of bugs that use internal features, including but not
+  limited to `custom_mir`, `lang_items`, `no_core`, and `rustc_attrs`.
+- Don't seed your fuzzer with inputs that are known to crash rustc (details
+  below).
 
 ### Discussion
 
@@ -107,16 +108,17 @@ these tools, post both the complete and minimized test cases. Generally,
 
 ## Effective fuzzing
 
-When fuzzing rustc, you may want to avoid generating code, since this is mostly
-done by LLVM. Try `--emit=mir` instead.
+When fuzzing rustc, you may want to avoid generating machine code, since this
+is mostly done by LLVM. Try `--emit=mir` instead.
 
 A variety of compiler flags can uncover different issues. `-Zmir-opt-level=4`
 will turn on MIR optimization passes that are not run by default, potentially
-uncovering interesting bugs.
+uncovering interesting bugs. `-Zvalidate-mir` can help uncover such bugs.
 
 If you're fuzzing a compiler you built, you may want to build it with `-C
 target-cpu=native` or even PGO/BOLT to squeeze out a few more executions per
-second.
+second. Of course, it's best to try multiple build configurations and see
+what actually results in superior throughput.
 
 ## Existing projects
 
--- a/src/fuzzing.md
+++ b/src/fuzzing.md
@@ -1,0 +1,105 @@
+# Fuzzing
+
+<!-- date-check: Mar 2023 -->
+
+For the purposes of this guide, *fuzzing* is any testing methodology that
+involves compiling a wide variety of programs in an attempt to uncover bugs in
+rustc. Fuzzing is often used to find internal compiler errors (ICEs). Fuzzing
+can be beneficial, because it can find bugs before users run into them and
+provide small, self-contained programs that make the bug easier to track down.
+However, some common mistakes can reduce the helpfulness of fuzzing and end up
+making contributors' lives harder. To maximize your positive impact on the Rust
+project, please read this guide before reporting fuzzer-generated bugs!
+
+## Guidelines
+
+### In a nutshell
+
+*Please do:*
+
+- Ensure the bug is still present on the latest nightly rustc
+- Include a reasonably minimal, standalone example along with any bug report
+- Include all of the information requested in the bug report template
+- Search for existing reports with the same message and query stack
+- Format the test case with `rustfmt`, if it maintains the bug
+
+*Please don't:*
+
+- Report lots of bugs that use internal features, including but not limited to
+  `custom_mir`, `lang_items`, `no_std`, and `rustc_attrs`.
+- Seed your fuzzer with inputs that are known to crash rustc (details below).
+
+### Discussion
+
+If you're not sure whether or not an ICE is a duplicate of one that's already
+been reported, please go ahead and report it and link to issues you think might
+be related. In general, ICEs on the same line but with different *query stacks*
+are usually distinct bugs.
+
+## Building a corpus
+
+When building a corpus, be sure to avoid collecting tests that are already
+known to crash rustc. A fuzzer that is seeded with such tests is more likely to
+generate bugs with the same root cause, wasting everyone's time. The simplest
+way to avoid this is to loop over each file in the corpus, see if it causes an
+ICE, and remove it if so.
+
+To build a corpus, you may want to use:
+
+- The rustc/rust-analyzer/clippy test suites (or even source code) --- though avoid
+  tests that are already known to cause failures, which often begin with comments
+  like `// failure-status: 101` or `// known-bug: #NNN`.
+- The already-fixed ICEs in [Glacier][glacier] --- though avoid the unfixed
+  ones in `ices/`!
+
+## Extra credit
+
+Here are a few things you can do to help the Rust project after filing an ICE.
+
+- Add the minimal test case to [Glacier][glacier]
+- [Bisect][bisect] the bug to figure out when it was introduced 
+- Fix unrelated problems with the test case (things like syntax errors or
+  borrow-checking errors)
+- Minimize the test case (see below)
+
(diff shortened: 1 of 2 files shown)


src/fuzzing.md
```diff
@@ -0,0 +1,105 @@
...
+- Ensure the bug is still present on the latest nightly rustc
+- Include a reasonably minimal, standalone example along with any bug report
+- Include all of the information requested in the bug report template
+- Search for existing reports with the same message and query stack
+- Format the test case with `rustfmt`, if it maintains the bug
+
+*Please don't:*
+
+- Report lots of bugs that use internal features, including but not limited to
+  `custom_mir`, `lang_items`, `no_std`, and `rustc_attrs`.
```
compiler-errors, line 29: `no_core` -- `no_std` bugs are valid and should be reported

src/fuzzing.md
```diff
@@ -0,0 +1,105 @@
...
+
+### In a nutshell
+
+*Please do:*
+
+- Ensure the bug is still present on the latest nightly rustc
+- Include a reasonably minimal, standalone example along with any bug report
+- Include all of the information requested in the bug report template
+- Search for existing reports with the same message and query stack
+- Format the test case with `rustfmt`, if it maintains the bug
```
compiler-errors, line 24: I would also ask to indicate that you found your bug via fuzzing

src/fuzzing.md
```diff
@@ -0,0 +1,105 @@
...
+- Report lots of bugs that use internal features, including but not limited to
+  `custom_mir`, `lang_items`, `no_std`, and `rustc_attrs`.
+- Seed your fuzzer with inputs that are known to crash rustc (details below).
+
+### Discussion
+
+If you're not sure whether or not an ICE is a duplicate of one that's already
+been reported, please go ahead and report it and link to issues you think might
+be related. In general, ICEs on the same line but with different *query stacks*
+are usually distinct bugs.
```
compiler-errors, line 37: Maybe a few examples, such as:

```
panicked at 'index out of bounds: the len is 1 but the index is 1', [...] ena-0.14.1/src/snapshot_vec.rs:199:10
```

There are other common ones that maybe @matthiaskrgr remembers.

src/fuzzing.md
```diff
@@ -0,0 +1,105 @@
...
+  like `// failure-status: 101` or `// known-bug: #NNN`.
+- The already-fixed ICEs in [Glacier][glacier] --- though avoid the unfixed
+  ones in `ices/`!
+
+## Extra credit
+
+Here are a few things you can do to help the Rust project after filing an ICE.
+
+- Add the minimal test case to [Glacier][glacier]
+- [Bisect][bisect] the bug to figure out when it was introduced
```
compiler-errors, line 60: :sparkles: 

If this list is arbitrarily ordered, I would prefer if we reorder this in terms of value to the ICE solver / compiler person like me :smile_cat: :
1. Bisecting
2. Minimiztion and fixing unrelated problems are *kinda* the same thing, though if you want to make some distinction, maybe correct me.
3. Glacier (these are only super useful for ICEs we expect to be there for a long time)

src/fuzzing.md
```diff
@@ -0,0 +1,105 @@
...
+  borrow-checking errors)
+- Minimize the test case (see below)
+
+[bisect]: https://github.com/rust-lang/cargo-bisect-rustc/blob/master/TUTORIAL.md
+
+## Minimization
+
+It can be helpful to *minimize* the fuzzer-generated input. When minimizing, be
+careful to preserve the original error, and avoid introducing distracting
+problems such as syntax, type-checking, or borrow-checking errors.
```
workingjubilee, line 71: I would rephrase this slightly, because if the issue is an ICE, it _is_ helpful to minimize the fuzzer-generated input, period... if done properly.

src/fuzzing.md
langston-barrett: Haha, whoops! Fixed.

src/fuzzing.md
langston-barrett: > Minimization and fixing unrelated problems feel like the same the same thing, though if you are trying to distinguish them here for some reason, feel free correct me :P

My only thought here is that it's possible to *shink* the test case in terms of bytes, LoC, etc. without reducing the number of unrelated errors (indeed, while increasing them).

src/fuzzing.md
langston-barrett: I rephrased: "It is helpful to carefully minimize"

src/fuzzing.md
```diff
@@ -0,0 +1,105 @@
...
+`halfempty` is not, but is generally a high-quality tool.
+
+[halfempty]: https://github.com/googleprojectzero/halfempty
+[picireny]: https://github.com/renatahodovan/picireny
+[treereduce]: https://github.com/langston-barrett/treereduce
+
+## Effective fuzzing
+
+When fuzzing rustc, you may want to avoid generating code, since this is mostly
+done by LLVM. Try `--emit=mir` instead.
```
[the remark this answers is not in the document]
langston-barrett: Added `-Zmir-opt`, thanks! I'm not sure if the extra detail about `--emit=llvm-{ir,bc}` is crucial, let me know if you think otherwise.

src/fuzzing.md
langston-barrett: Added an example!

src/fuzzing.md
langston-barrett: Whoops, just forgot that!

langston-barrett:
Github really doesn't want me to re-request review from two people right now... but all reviews are appreciated ofc!

src/fuzzing.md
matthiaskrgr: A couple more common ICEs
````
      7     "error_reason": "thread 'rustc' panicked at 'Box<dyn Any>', /rustc/1203e0866e6c3659775efcb8aecad21dc13ef38b/compiler/rustc_errors/src/lib.rs:995:33",
     13     "error_reason": "thread 'rustc' panicked at 'Box<dyn Any>', /home/matthias/vcs/github/rust_debug_assertions/compiler/rustc_errors/src/lib.rs:1644:9",
     14     "error_reason": "thread 'rustc' panicked at 'forcing query with already existing `DepNode`",
     20     "error_reason": "error: internal compiler error: compiler/rustc_infer/src/infer/lexical_region_resolve/mod.rs:203:17: cannot relate region: ReErased",
     37     "error_reason": "error: internal compiler error: no errors encountered even though `delay_span_bug` issued",
     80     "error_reason": "thread 'rustc' panicked at 'assertion failed: `(left == right)`",
     91     "error_reason": "error: internal compiler error: unexpected panic",
````

File: src/fuzzing.md [hunk 7]
```diff
@@ -0,0 +1,132 @@
...
+provide small, self-contained programs that make the bug easier to track down.
+However, some common mistakes can reduce the helpfulness of fuzzing and end up
+making contributors' lives harder. To maximize your positive impact on the Rust
+project, please read this guide before reporting fuzzer-generated bugs!
+
+## Guidelines
+
+### In a nutshell
+
+*Please do:*
```
matthiaskrgr, line 18: could also mention debug assertions here.

src/fuzzing.md
```diff
@@ -0,0 +1,132 @@
...
+- Include all of the information requested in the bug report template
+- Search for existing reports with the same message and query stack
+- Format the test case with `rustfmt`, if it maintains the bug
+- Indicate that the bug was found by fuzzing
+
+*Please don't:*
+
+- Report lots of bugs that use internal features, including but not limited to
+  `custom_mir`, `lang_items`, `no_core`, and `rustc_attrs`.
+- Seed your fuzzer with inputs that are known to crash rustc (details below).
```
matthiaskrgr, line 31: I would rephrase this a bit:

Please don't:

Don't report lot of bugs that..

Don't seed your fuzzer wth inputs...

src/fuzzing.md
```diff
@@ -0,0 +1,132 @@
...
+[`treereduce-rust`][treereduce] and [picireny][picireny] are syntax-aware.
+`halfempty` is not, but is generally a high-quality tool.
+
+[halfempty]: https://github.com/googleprojectzero/halfempty
+[picireny]: https://github.com/renatahodovan/picireny
+[treereduce]: https://github.com/langston-barrett/treereduce
+
+## Effective fuzzing
+
+When fuzzing rustc, you may want to avoid generating code, since this is mostly
```
matthiaskrgr, line 110: "generating code" -> "generating machine code" since fuzzing is all about generating RUST code?

src/fuzzing.md
```diff
@@ -0,0 +1,132 @@
...
+done by LLVM. Try `--emit=mir` instead.
+
+A variety of compiler flags can uncover different issues. `-Zmir-opt-level=4`
+will turn on MIR optimization passes that are not run by default, potentially
+uncovering interesting bugs.
```
matthiaskrgr, line 115: -Zvalidate-mir is also noteworthy here.

src/fuzzing.md
```diff
@@ -0,0 +1,132 @@
...
+
+If you're fuzzing a compiler you built, you may want to build it with `-C
+target-cpu=native` or even PGO/BOLT to squeeze out a few more executions per
+second.
+
```
matthiaskrgr, line 120: I'm note sure if PGOing will work that well.

Usually when profiling, you try to optimize performance for codepaths that you know are very common but when fuzzing, you try to run as many rare/undicovered codepaths as possible.

src/fuzzing.md
langston-barrett: I don't know about this - can you say more?

src/fuzzing.md
langston-barrett: For my own edification, what happens when MIR validation fails? Does rustc emit an ICE, or some other kind of message?

src/fuzzing.md
langston-barrett: Ultimately, such questions are entirely specific to the environment/context/fuzzer, and need to be decided by looking at the executions per second. I'll say as much here.

src/fuzzing.md
matthiaskrgr: When bootstrapping rust, you can enable debug assertions in your config.toml which then enables debug assertions in rustc https://github.com/rust-lang/rust/blob/master/config.example.toml#L438

(bugs that might still be useful to know, but sometimes are not that bad or discovering them automatically has a noticeable performance hit that we don't want to deal with all the time etc..)

https://github.com/rust-lang/rust/labels/requires-debug-assertions

src/fuzzing.md
matthiaskrgr: I think it's a bit like debug-assertions but for the mir, some extra checks that make sure the mir generated at compile time is well formed and does not contain patterns that would be know to be obviously wrong.
Sometimes mir optimizations can break code in a way that does not produce an ICE right away and Zvalidate-mir can catch that I think.
https://github.com/rust-lang/rust/issues/101243

src/fuzzing.md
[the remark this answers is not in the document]
workingjubilee: It's not crucial at all, I just decided to yammer a bit!

src/fuzzing.md
workingjubilee: Per @Nilstrieb:
Every MIR pass transforms the MIR.
After all passes, the MIR is validated.
`-Zvalidate-mir` validates after every pass, instead of once.
This catches transient violations of MIR's well-formedness.

Yes, it will probably induce an ICE or other entertaining error.

src/fuzzing.md
langston-barrett: Great, I covered that a bit later on. Thanks for the info!

src/fuzzing.md
```diff
@@ -0,0 +1,134 @@
...
+### In a nutshell
+
+*Please do:*
+
+- Ensure the bug is still present on the latest nightly rustc
+- Include a reasonably minimal, standalone example along with any bug report
+- Include all of the information requested in the bug report template
+- Search for existing reports with the same message and query stack
+- Format the test case with `rustfmt`, if it maintains the bug
+- Indicate that the bug was found by fuzzing
```
Nilstrieb, line 25: I would like to add something like "remove distractions". Often, fuzzed snippets are full of weird code and one of (or even multiple together) these weirdnesses is triggering the ICE. While it's not always easy without understanding the issue, at least trying to remove as many weirdnesses as possible for a few minutes can be great in making the report more obvious.
This includes things like malformed attributes, unresolved names, invalid types etc.

Nilstrieb: I see you already mentioned something like this below, but I really like the "distractions" wording, Generally, a little less minimal but less "weird" example is easier to understand.

langston-barrett: This is discussed a bit later on, under "extra credit" and also under "minimization".

My inclination is to avoid putting *too* many things in the "Please do" list to make sure that people don't see it and think either (1) "That's so much work! I'm not going to do all that, I'm just going to report it however I feel" or (2) "That's so much work! I'm going to fuzz something else.". I think we want a short list of guidelines that include everything at the "just right" intersection of *important* and *easy*.

I put bisection (even though it's important), minimization, and adding to Glacier under "extra credit", because they seem a bit harder. I put `rustfmt` in this list because it's so easy. Removing spurrious other errors also seems a bit difficult, or at least something I wouldn't necessarily want to do for dozens of bugs.

All of that being said! I've never fixed a bug in rustc, so I don't have a good picture of what's most important.

langston-barrett: > I see you already mentioned something like this below, but I really like the "distractions" wording, Generally, a little less minimal but less "weird" example is easier to understand.

I clarified this point in the "extra credit" section and used this wording!

langston-barrett:
It occurs to me that there's a lot in this chapter that applies to ICEs in general (though much of it applies especially to fuzzed ICEs), such as requesting minimization, avoiding "distracting" errors, searching for duplicates, bisection, etc. I wonder if it might be helpful to have a separate chapter to hold some of this information. Such a chapter could then be linked to from this one, and maybe also in the ICE issue template.

src/fuzzing.md
```diff
@@ -0,0 +1,149 @@
...
+It is helpful to carefully *minimize* the fuzzer-generated input. When
+minimizing, be careful to preserve the original error, and avoid introducing
+distracting problems such as syntax, type-checking, or borrow-checking errors.
+
+There are some tools that can help with minimization. If you're not sure how
+to avoid introducing syntax, type-, and borrow-checking errors while using
+these tools, post both the complete and minimized test cases. Generally,
+*syntax-aware* tools give the best results in the least amount of time.
+[`treereduce-rust`][treereduce] and [picireny][picireny] are syntax-aware.
+`halfempty` is not, but is generally a high-quality tool.
```
langston-barrett, line 103: Using these tools can be tricky, especially if you want to avoid introducing distracting errors. I think I'll write a little tool that uses `treereduce` under the hood but is set up by default to shrink ICEs without introducing errors.

src/fuzzing.md
```diff
@@ -0,0 +1,149 @@
...
+
+- [Bisect][bisect] the bug to figure out when it was introduced 
+- Fix "distractions": problems with the test case that don't contribute to
+  triggering the ICE, such as syntax errors or borrow-checking errors
+- Minimize the test case (see below)
+- Add the minimal test case to [Glacier][glacier]
+
+[bisect]: https://github.com/rust-lang/cargo-bisect-rustc/blob/master/TUTORIAL.md
+
+## Minimization
```
langston-barrett, line 92: I just found two Zulip threads about minimization:

- https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/minimization/near/342432199
- https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/Script.20for.20minimizing.20huge.20minimal.20examples/near/266972944

One of them mentions [`rust-reduce`](https://github.com/jethrogb/rust-reduce), I will give that tool a try. If successful, I'll add it to the list.

workingjubilee:
> It occurs to me that there's a lot in this chapter that applies to ICEs in general (though much of it applies especially to fuzzed ICEs), such as requesting minimization, avoiding "distracting" errors, searching for duplicates, bisection, etc. I wonder if it might be helpful to have a separate chapter to hold some of this information. Such a chapter could then be linked to from this one, and maybe also in the ICE issue template.

Hmm. I think this guide is getting pretty close to "good as-is", and I'd be fine to land this as one chapter and then do a split in a followup that reorganizes things slightly and cleanly divides "minimization as relevant to fuzzing" out.

langston-barrett:
> Hmm. I think this guide is getting pretty close to "good as-is", and I'd be fine to land this as one chapter and then do a split in a followup that reorganizes things slightly and cleanly divides "minimization as relevant to fuzzing" out.

Agreed! Someone on Zulip also mentioned minimization as a potential chapter topic.

src/fuzzing.md
langston-barrett: (Might not do this before merging, but I will try to get to it at some point!)

compiler-errors (approved):
I'm happy to merge this now, then.

PR #1651, opened 2023-03-24
Add locale_resources

nathanhammond:
After adding this field it works in: `rustc 1.70.0-nightly (a266f1199 2023-03-22)`


jyn514 (approved):
thanks!

PR #1657, opened 2023-04-01
Update some links and information for cargo-bisect-rustc

ehuss:
This updates some links for cargo-bisect-rustc which has a new documentation guide.


JohnTitor (approved):
Thanks!

PR #1679, opened 2023-04-22
Fix links in how-to-build-and-run.md

briankung:
Four links were broken, either pointing to the wrong url or the wrong extension (or both). This fixes the links.


JohnTitor (approved):
Thank you!

PR #1681, opened 2023-04-29
Define more lint terms.

ehuss:
This adds some explicit definitions of various terms used in the lint chapters *before* they are used. It can be confusing when the text uses terminology that is not defined (or is not defined until much later).

This uses some text taken from a comment by @Nilstrieb.

### DIFF
--- a/src/diagnostics.md
+++ b/src/diagnostics.md
@@ -582,7 +582,7 @@ More information about lint registration can be found in the [LintStore]
 chapter.
 
 [deprecated plugin system]: https://doc.rust-lang.org/nightly/unstable-book/language-features/plugin.html
-[LintStore]: lintstore.md
+[LintStore]: diagnostics/lintstore.md
 
 ### Declaring a lint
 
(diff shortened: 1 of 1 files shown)


src/diagnostics.md
```diff
@@ -507,6 +507,83 @@ module.
...
+- *Tool* lints are lints with a path prefix like `clippy::` or `rustdoc::`.
+- *Internal* lints are the `rustc::` scoped tool lints that only run on the
+  rustc source tree itself and are defined in the compiler source like a
+  regular built-in lint.
+
+More information about lint registration can be found in the [LintStore]
+chapter.
+
+[deprecated plugin system]: https://doc.rust-lang.org/nightly/unstable-book/language-features/plugin.html
+[LintStore]: lintstore.md
```
tshepang, line 585: I thinks this is it
```suggestion
[LintStore]: diagnostics/lintstore.md
```

tshepang (approved):

PR #1686, opened 2023-05-10
Add note about shell completion scripts

clubby789:
Add a note about the new shell completion scripts that will land in https://github.com/rust-lang/rust/pull/111388. cc @jyn514

### DIFF
--- a/src/building/suggested.md
+++ b/src/building/suggested.md
@@ -339,3 +339,13 @@ pkgs.mkShell {
   RUST_BOOTSTRAP_CONFIG = config;
 }
 ```
+
+## Shell Completions
+
+If you use Bash, Fish or PowerShell, you can find automatically-generated shell completion scripts for `x.py` in [`src/etc/completions`](https://github.com/rust-lang/rust/tree/master/src/etc/completions).
+Zsh support will also be included once issues with [`clap_complete`](https://crates.io/crates/clap_complete) have been resolved.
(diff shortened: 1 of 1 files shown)


src/building/suggested.md
```diff
@@ -339,3 +339,8 @@ pkgs.mkShell {
   RUST_BOOTSTRAP_CONFIG = config;
 }
 ```
+
+## Shell Completions
+
+If you use Bash, Fish or PowerShell, you can find automatically-generated shell completion scripts for `x.py` in [`src/etc/completions`](https://github.com/rust-lang/rust/tree/master/src/etc/completions).
```
jyn514, line 345: Can we also document how to use the scripts? is it as simple as `source src/etc/completions/bash.sh`?

clubby789: I'll add some quick suggestions

clubby789:
I think CI broke as part of Github outage here

ChrisDenton:
```
Checking line lengths in all source files <= 100 chars...
Offending files and lines:
	src/building/suggested.md:348 : You can use `source ./src/etc/completions/x.py.<extension>` to load completions for your shell of choice,
1 offending lines found.
```

clubby789:
Ah thank you, should be good now

jyn514 (approved):
amazing, thank you!

PR #1688, opened 2023-05-16
Update track_caller reference link

Zyranivia:
This PR updates the (probably) outdated https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-track_caller-attribute to https://doc.rust-lang.org/reference/attributes/codegen.html#the-track_caller-attribute


JohnTitor (approved):
Thanks!

PR #1705, opened 2023-06-27
Fix some links

ehuss:
This fixes some broken links that I noticed.


src/diagnostics/translation.md
```diff
@@ -87,9 +87,9 @@ information that needs to be provided by the code to do so.
 
 ### Compile-time validation and typed identifiers
 rustc's Fluent resources for the default locale (`en-US`) are in the
-[`compiler/rustc_error_messages/locales/en-US`] directory. Currently, each crate
-which defines translatable diagnostics has its own Fluent resource, such as
-`parser.ftl` or `typeck.ftl`.
+[`compiler/rustc_error_messages/messages.ftl`] file. Currently, each crate
+which defines translatable diagnostics has its own Fluent resource in
+a file named `messages.ftl`.
```
ehuss, line 92: I'm not actually up to speed on the Fluent stuff, so this may not be worded the best. It was converted from a directory to a file in https://github.com/rust-lang/rust/pull/103042, and then renamed again in https://github.com/rust-lang/rust/pull/108682.

tshepang: This would be correct if the first sentence was removed. I would also add an example or two:
> ... such as `compiler/rustc_borrowck/messages.ftl` and `compiler/rustc_parse/messages.ftl`

tshepang:
just one comment, but it's otherwise all good

src/diagnostics/translation.md
ehuss: Sure, I went ahead and pushed a change. Skimming over the chapter, it looks like there are quite a few things that are out-of-date. I don't know if there is someone involved in translation who could be pinged to update it.

tshepang (approved):

src/diagnostics/translation.md
tshepang: I noticed, and intend to update it soon.

tshepang:
@ehuss what is your workflow like when doing a change such as this... do you do it in one session, or do you collect these over time?

est31:
Can't speak for @ehuss but when I do changes like these, I often start with one thing that disturbs me. I decide to make a PR for a one line change, and then look at the places around it and see there is similar problems there.

ehuss:
> @ehuss what is your workflow like when doing a change such as this... do you do it in one session, or do you collect these over time?

I just ran into some broken links, so I decided to fix them, and then decided while I was there I was just going to fix the rest. I used `mdbook-linkcheck` to find them. I had to edit the script in this repo to run it manually against all pages (removing `FLAGS`, and setting `USE_TOKEN`).

PR #1711, opened 2023-06-30
update some of the diagnostic translations info

tshepang:
I did not touch parts about `Emitter` trait... perhaps someone that knows what movement went there could do a follow-up PR.


JohnTitor (approved):
Thanks! Let's update it even if it is partial.

PR #1721, opened 2023-07-05
improve explanation of placing tests in separate file

tshepang:


src/tests/intro.md
```diff
@@ -43,20 +43,19 @@ The standard library relies very heavily on documentation tests to cover its fun
...
 ```rust,ignore
 #[cfg(test)]
 mod tests;
 ```
 
-If it wasn't done this way, and the tests were placed in the same file as the source,
-then changing or adding a test would cause the crate you are working on to be recompiled.
-If you were working on something like `core`,
-then that would require recompiling the entire standard library, and the entirety of `rustc`.
+If it wasn't done this way,
```
JohnTitor, line 56:
```suggestion
If it wasn't done this way,
```

JohnTitor (approved):

PR #1732, opened 2023-07-11
Explain more in depth what early and late bound generic parameters are

BoxyUwU:
I moved all the text about the `ty::` representation of bound vars and generic parameters out into the subsection about the `ty` module since I found it confusing to be put along side the explanation of what it means for a generic parameter to be early or late bound. To me there is a large difference between the two concepts and they do not deserve to be put together. (Ideally we would rewrite the bound-vars-and-params page and rename `RegionKind::EarlyBound/LateBound` imo)

This explanation of what early/late bound parameters are came from the deep dive that t-types did earlier this week, rust-lang/types-team#94. Thanks to @jyn514 also for looking over the document to give feedback before editing it to put in here.

cc @rust-lang/types


src/early-late-bound.md
```diff
@@ -1,107 +1,199 @@
...
 
+If you are looking for information on the `RegionKind` variants `ReLateBound` and `ReEarlyBound`
+you should look at the section on [bound vars and params](./bound-vars-and-params.md). This section
+discusses what makes generic parameters on functions and closures late/early bound. Not the general
+concept of bound vars and generic parameters which `RegionKind` has named somewhat confusingly
+with this topic.
+
+## What does it mean for parameters to be early or late bound
+
+All function definitions conceptually have a zst (this is represented by `TyKind::FnDef` in rustc).
```
compiler-errors, line 17:
```suggestion
All function definitions conceptually have a ZST (this is represented by `TyKind::FnDef` in rustc).
```

src/bound-vars-and-params.md
```diff
@@ -0,0 +1,61 @@
...
+[`ParamTy`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.ParamTy.html
+[`EarlyBoundRegion`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.EarlyBoundRegion.html
+[`Generics`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.Generics.html
+
+## Late-bound parameters
+
+Late-bound parameters in `rustc` are handled quite differently (they are also
+specialized to lifetimes since, right now, only late-bound lifetimes are
+supported, though with GATs that has to change). We indicate their potential
+presence by a [`Binder`] type. The [`Binder`] doesn't know how many variables
```
compiler-errors, line 35:
```suggestion
Late-bound parameters in `rustc` are handled differently. We indicate their
presence by a [`Binder`] type. The [`Binder`] doesn't know how many variables
```

src/bound-vars-and-params.md
```diff
@@ -0,0 +1,61 @@
...
+there are at that binding level. This can only be determined by walking the
+type itself and collecting them. So a type like `for<'a, 'b> ('a, 'b)` would be
+`for (^0.a, ^0.b)`. Here, we just write `for` because we don't know the names
```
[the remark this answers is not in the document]
BoxyUwU: yes but not in this PR because I don't really want to touch this too much until we sort out the naming of RegionKind's variants

src/bound-vars-and-params.md
```diff
@@ -0,0 +1,61 @@
...
+```rust,ignore
+trait Foo<T> {
+  type Bar<U> = (Self, T, U);
+}
+```
+
+Here, the type `(Self, T, U)` would be `($0, $1, $2)`, where `$N` means a
+[`ParamTy`] with the index of `N`.
+
+In rustc, the [`Generics`] structure carries this information. So the
```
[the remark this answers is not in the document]
BoxyUwU: probably but also i dont really want to in this PR

src/early-late-bound.md
```diff
@@ -1,107 +1,199 @@
...
+If you are looking for information on the `RegionKind` variants `ReLateBound` and `ReEarlyBound`
+you should look at the section on [bound vars and params](./bound-vars-and-params.md). This section
+discusses what makes generic parameters on functions and closures late/early bound. Not the general
+concept of bound vars and generic parameters which `RegionKind` has named somewhat confusingly
+with this topic.
+
+## What does it mean for parameters to be early or late bound
+
+All function definitions conceptually have a ZST (this is represented by `TyKind::FnDef` in rustc).
+The only generics on this zst are the early bound parameters of the function definition. e.g.
```
compiler-errors, line 18:
```suggestion
The only generics on this ZST are the early bound parameters of the function definition. e.g.
```

compiler-errors (approved):

PR #1733, opened 2023-07-12
"Rust reference" -> "Rust book" in about-this-guide.md

ismailarilik:
Shouldn't it be named as "Rust book"?


ChrisDenton:
I believe it's meant to link to the [Rust Reference](https://doc.rust-lang.org/nightly/reference/).

JohnTitor:
Oh seems it was fixed already. Thanks for the PR anyway!

ismailarilik:
It seems that it still goes to the book.

PR #1736, opened 2023-07-13
fix(how-to-build-and-run): fix a typo ("fromer" -> "former")

ismailarilik:
I think it is a typo, isn't it? 🤔

### DIFF
--- a/src/building/how-to-build-and-run.md
+++ b/src/building/how-to-build-and-run.md
@@ -130,7 +130,7 @@ of a checkout. It also looks up the appropriate version of `python` to use.
 You can install it with `cargo install --path src/tools/x`.
 
 To clarify that this is another global installed binary util, which is
-similar to the fromer `x` declared in section [What is `x.py`](#what-is-xpy) but
+similar to the former `x` declared in section [What is `x.py`](#what-is-xpy) but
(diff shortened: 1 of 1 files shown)


src/building/how-to-build-and-run.md
```diff
@@ -130,7 +130,7 @@ of a checkout. It also looks up the appropriate version of `python` to use.
 You can install it with `cargo install --path src/tools/x`.
 
 To clarify that this is another global installed binary util, which is
-similar to the fromer `x` declared in section [What is `x.py`](#what-is-xpy) but
+similar to the former `x` declared in section [What is `x.py`](#what-is-xpy) but
```
tshepang, line 133:
```suggestion
similar to the one declared in section [What is `x.py`](#what-is-xpy) but
```
reads better I think

ismailarilik: Yeah, it is. But it might be another PR not to cause any confusion; this PR just fixes a typo.

tshepang: PRs often do change direction, and that's ok... I will keep the typo commit so as not to lose history of what happened.

tshepang (approved):

PR #1753, opened 2023-07-25
feat(part-5-intro): make "Part 5" obvious

ismailarilik:


tshepang:
I don't feel like it's worth it, unless this was done for sake of consistency, which itself is lacking in these high-level sections. I would rather remove the "Part X" from the other headings instead.

ismailarilik:
> I don't feel like it's worth it, unless this was done for sake of consistency, which itself is lacking in these high-level sections. I would rather remove the "Part X" from the other headings instead.

There are some references to part numbers, like this one:

image

And this:

image

So I think it would be more consistent to adding part numbers to the first and fifth parts. Also the section [How to build and run the compiler](https://rustc-dev-guide.rust-lang.org/building/how-to-build-and-run.html#how-to-build-and-run-the-compiler) might be "Part 0".

But you know better, right? (Since I am very new.) =)

So for consistency, there are two ways:

1. Removing part info from headers. The reference in the second picture above should also be adjusted not to have part number.
2. Adding part info to the headers which don't have it. This doesn't require any other change.

There is also one concern: In any case do we need to change file names (add or remove part numbers)?

Please guide me. =)

tshepang:
Part numbers were decided without my involvement, so maybe someone that remembers justification would comment. As for me, I would get rid of them altogether.

Maybe let's wait for a week or so, and if no one comments, we can decide together.

Whatever happens, I would not change the file names, unless we add redirects (see book.toml for where this is done).

ismailarilik:
> Part numbers were decided without my involvement, so maybe someone that remembers justification would comment. As for me, I would get rid of them altogether.
> 
> Maybe let's wait for a week or so, and if no one comments, we can decide together.
> 
> Whatever happens, I would not change the file names, unless we add redirects (see book.toml for where this is done).

It's been almost two weeks so what do you think? I am OK with the current state.

tshepang:
am in favor of removing part info from headers

ismailarilik:
Hmm, I am just wondering why. They are parts so can have part numbers. The URL also have part numbers for all parts except the first one. Isn't it self-descriptive?

tshepang:
Parts feel to me like organisational things, not things that should be exposed in headings. That is, am happy if with go with option 2, where we remove "Part X" in headings, even if we leave the file names unchanged.

ismailarilik:
@tshepang how is it now?

tshepang (approved):

src/part-2-intro.md
```diff
@@ -7,7 +7,7 @@ end-to-end process of compilation _and_ readers interested in learning about a
 specific system they wish to contribute to. If anything is unclear, feel free
 to file an issue on the [rustc-dev-guide
 repo](https://github.com/rust-lang/rustc-dev-guide/issues) or contact the compiler
-team, as detailed in [this chapter](./compiler-team.md).
```
tshepang: I would not have reverted this

PR #1758, opened 2023-07-27
fix(serialization.md): fix invalid `Lazy<[T]>` to valid `LazyArray<T>`

ismailarilik:
I am not totally sure that it should be like this but I am pretty sure because I couldn't find a type for `Lazy<[T]>` [here](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_metadata/rmeta/) so I opened a PR instead of an issue. Feel free to close this if it was a mistake (and sorry about that).


JohnTitor:
Oh, it seems this has a merge conflict, could you check it?

tshepang:
could you remove the merge commit

ismailarilik:
Sorry, I don't have time for it nowadays. Can't GitHub do this for us?

tshepang:
I think not. Let me do another PR.

PR #1764, opened 2023-08-03
Add RPITIT documentation

spastorino:
This is more or less the document we've created for the types team deep dive as in but I don't think it needs more work.

r? @compiler-errors


compiler-errors (approved):

compiler-errors:
CI needs fixing, r=me after

Issue #1704, opened 2023-06-27
`warning` attribute missing in diagnostics derive macro reference


nicklimmm:
Previously, the `warn_` attribute is used to add warning diagnostics. But it was changed to `warning` in [this PR](https://github.com/rust-lang/rust/pull/100959).

Some information is also outdated in both `Diagnostic` and `Subdiagnostic` derives, e.g. no `warning` reference in `Subdiagnostic`, the attribute is also applicable on structs.

Diagnostic derive documentation from nightly rustc (source of truth):
- [rustc `#[derive(Diagnostic)`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_macros/derive.Diagnostic.html)
- [rustc `#[derive(LintDiagnostic)]`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_macros/derive.LintDiagnostic.html)
- [rustc `#[derive(Subdiagnostic)]`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_macros/derive.Subdiagnostic.html)

What we have in the guide instead:
- [`#[derive(Diagnostic)]`dev guide reference](https://rustc-dev-guide.rust-lang.org/diagnostics/diagnostic-structs.html#reference)
- [`#[derive(Subdiagnostic)]`dev guide reference](https://rustc-dev-guide.rust-lang.org/diagnostics/diagnostic-structs.html#reference-1)

Updating the `warning` references in the guide can help new contributors to support the translation effort (issue tracker https://github.com/rust-lang/rust/issues/100717)

nicklimmm:
If this looks good, I'd like to update the docs accordingly

PR #1765, opened 2023-08-07
Update from `#[warn_]` to `#[warning]` diagnostic attributes

nicklimmm:
Context: https://github.com/rust-lang/rustc-dev-guide/issues/1704

### DIFF
--- a/src/diagnostics/diagnostic-structs.md
+++ b/src/diagnostics/diagnostic-structs.md
@@ -70,7 +70,7 @@ Diagnostics are more than just their primary message, they often include
 labels, notes, help messages and suggestions, all of which can also be
 specified on a `Diagnostic`.
 
-`#[label]`, `#[help]` and `#[note]` can all be applied to fields which have the
+`#[label]`, `#[help]`, `#[warning]` and `#[note]` can all be applied to fields which have the
 type `Span`. Applying any of these attributes will create the corresponding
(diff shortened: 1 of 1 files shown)


tshepang (approved):

src/diagnostics/diagnostic-structs.md
```diff
@@ -180,8 +180,8 @@ following attributes:
   - Value is a path to an item in `rustc_errors::fluent` for the note's
     message.
     - Defaults to equivalent of `.label`.
-- `#[warn_]` or `#[warn_(slug)]` (_Optional_)
-  - _Applied to `Span` fields._
+- `#[warning]` or `#[warning(slug)]` (_Optional_)
+  - _Applied to struct or `Span`/`()` fields._
```
tshepang, line 184: isn't `Span` a struct

nicklimmm: Technically yes, but here it emphasizes placing it on a struct field of type `Span`.

In case this is ambiguous, it should mean like this:
- struct, or
- `Span`/`()` fields

tshepang: would `(), Span, or other struct types` be accurate

tshepang: ahh, much better

PR #1771, opened 2023-08-22
Fix broken MD link format

chenyukang:


src/generics.md
```diff
@@ -127,16 +127,15 @@ You may have a couple of followup questions…
...
-another list of types.
+use [`instantiate`][instantiate] to replace a `GenericArgsRef` with  another list of types.
 
 [Here is an example of actually using `instantiate` in the compiler][instantiatex].
 The exact details are not too important, but in this piece of code, we happen to be
 converting from the `rustc_hir::Ty` to a real `ty::Ty`. You can see that we first get some args
 (`args`).  Then we call `type_of` to get a type and call `ty.instantiate(tcx, args)` to get a new
 version of `ty` with the args made.
 
+[instantiate]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/generic_args/struct.EarlyBinder.html#method.instantiate
```
tshepang, line 138: this works too, and avoids the repetition
```suggestion
use [`instantiate`] to replace a `GenericArgsRef` with  another list of types.

[Here is an example of actually using `instantiate` in the compiler][instantiatex].
The exact details are not too important, but in this piece of code, we happen to be
converting from the `rustc_hir::Ty` to a real `ty::Ty`. You can see that we first get some args
(`args`).  Then we call `type_of` to get a type and call `ty.instantiate(tcx, args)` to get a new
version of `ty` with the args made.

[`instantiate`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/generic_args/struct.EarlyBinder.html#method.instantiate
```

tshepang (approved):

Issue #1699, opened 2023-06-18
link to std-dev-guide from the landing page


jyn514:
i.e add a new top-level "how do i contribute to std?" section to https://rustc-dev-guide.rust-lang.org/getting-started.html

cc https://discord.com/channels/273534239310479360/592856094527848449/1120138588336689274

zahash:
Hi, can i do this?

zahash:
https://github.com/rust-lang/rustc-dev-guide/pull/1783

zahash:
can this issue be closed now?

PR #1783, opened 2023-08-29
link std-dev-guide from landing page (#1699)

zahash:


src/getting-started.md
```diff
@@ -171,6 +171,10 @@ incredibly helpful:
 
 See ["How to build and run the compiler"](./building/how-to-build-and-run.md).
 
+## How do i contribute to std?
```
tshepang, line 174: I would change the heading, to match the others
```suggestion
## Contributing to std (standard library)
```
I would also move it to before "Contributing code to other Rust projects", because std is more central than those other projects.

zahash: done.

tshepang (approved):

PR #1787, opened 2023-09-04
Add section about building an optimized version of `rustc`

Kobzol:
This PR adds a new section to the dev guide, which describes how can one build as optimized version of `rustc` as possible. A possible use-case is for building an optimized version of `rustc` for a Linux distribution.

Related issue: https://github.com/rust-lang/rust/issues/115251

### DIFF
--- a/src/building/optimized-build.md
+++ b/src/building/optimized-build.md
@@ -22,7 +22,7 @@ lto = "thin"
 
 > Note that LTO for `rustc` is currently supported and tested only for
 > the `x86_64-unknown-linux-gnu` target. Other targets *may* work, but no guarantees are provided.
-> Notably, LTO optimized `rustc` currently produces [miscompilations] on Windows.
+> Notably, LTO-optimized `rustc` currently produces [miscompilations] on Windows.
 
 [miscompilations]: https://github.com/rust-lang/rust/issues/109114
 
@@ -63,7 +63,7 @@ compile `rustc` for a specific instruction set architecture, you can set the `ta
 option in `RUSTFLAGS`:
 
 ```bash
-$ RUSTFLAGS="-C target_cpu=x86-64-v3" x.py build ...
+RUSTFLAGS="-C target_cpu=x86-64-v3" ./x build ...
 ```
 
 If you also want to compile LLVM for a specific instruction set, you can set `llvm` flags
@@ -78,9 +78,9 @@ cflags = "-march=x86-64-v3"
 ## Profile-guided optimization
 
 Applying profile-guided optimizations (or more generally, feedback-directed optimizations) can
-produce a large increase to `rustc` performance, by up to 25%. However, these techniques are not
-simply enabled by a configuration option, but rather they require a complex build workflow that
-compiles `rustc` multiple times and profiles it on selected benchmarks.
+produce a large increase to `rustc` performance, by up to 15% ([1], [2]). However, these techniques
+are not simply enabled by a configuration option, but rather they require a complex build workflow
+that compiles `rustc` multiple times and profiles it on selected benchmarks.
 
 There is a tool called `opt-dist` that is used to optimize `rustc` with [PGO] (profile-guided
 optimizations) and [BOLT] (a post-link binary optimizer) for builds distributed to end users. You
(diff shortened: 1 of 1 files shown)


Kobzol:
CC @cuviper, as you're the only distro maintainer I know :laughing: Would something like this be useful to you, in order to build optimized `rustc` binaries for Linux distros? I'd be glad for any feedback.

src/building/optimized-build.md
```diff
@@ -0,0 +1,131 @@
...
+in `config.toml`:
+
+```toml
+[rust]
+lto = "thin"
+```
+
+> Note that LTO for `rustc` is currently supported and tested only for
+> the `x86_64-unknown-linux-gnu` target. Other targets *may* work, but no guarantees are provided.
+> Notably, LTO optimized `rustc` currently produces [miscompilations] on Windows.
```
tshepang, line 25:
```suggestion
> Notably, LTO-optimized `rustc` currently produces [miscompilations] on Windows.
```

src/building/optimized-build.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Instruction set
+
+By default, `rustc` is compiled for a generic (and conservative) instruction set architecture
+(depending on the selected target), to make it support as many CPUs as possible. If you want to
+compile `rustc` for a specific instruction set architecture, you can set the `target_cpu` compiler
+option in `RUSTFLAGS`:
+
+```bash
+$ RUSTFLAGS="-C target_cpu=x86-64-v3" x.py build ...
```
tshepang, line 66: removes extraneous character, and use recommended x command
```suggestion
RUSTFLAGS="-C target_cpu=x86-64-v3" ./x build ...
```

src/building/optimized-build.md
```diff
@@ -0,0 +1,131 @@
...
+```toml
+[llvm]
+cxxflags = "-march=x86-64-v3"
+cflags = "-march=x86-64-v3"
+```
+
+## Profile-guided optimization
+
+Applying profile-guided optimizations (or more generally, feedback-directed optimizations) can
+produce a large increase to `rustc` performance, by up to 25%. However, these techniques are not
```
tshepang, line 81: can we have a link to some benchmark

src/building/optimized-build.md
```diff
@@ -0,0 +1,131 @@
...
+    ```
+   Note that the default Linux environment expects several hardcoded paths to exist:
+    - `/checkout` should contain a checkout of the Rust compiler repository that will be compiled.
+    - `/rustroot` should contain the compiled LLVM toolchain (containing BOLT).
+    - A Python 3 interpreter should be available under the `python3` binary.
+    - `/tmp/rustc-perf` should contain a downloaded checkout of the Rust benchmark suite.
+
+You can modify `LinuxEnvironment` (or implement your own) to override these paths.
+
+[`Environment`]: https://github.com/rust-lang/rust/blob/65e468f9c259749c210b1ae8972bfe14781f72f1/src/tools/opt-dist/src/environment/mod.rs#L8-L7
```
tshepang, line 129:
```suggestion
[`Environment`]: https://github.com/rust-lang/rust/blob/65e468f9c259749c210b1ae8972bfe14781f72f1/src/tools/opt-dist/src/environment/mod.rs#L8-L70
```

src/building/optimized-build.md
Kobzol: PGO on Linux was enabled in parts (rustc/LLVM), so I don't really have a single benchmark link that would such big performance wins at once. This number comes from my past experience, plus from some attempts to disable PGO (so I saw the opposite regressions). I have changed the amount to `up to 15%` and added a link to PGO for Windows: https://github.com/rust-lang/rust/pull/96978, which enabled PGO for `rustc` and LLVM at once, and also to https://blog.rust-lang.org/inside-rust/2020/11/11/exploring-pgo-for-the-rust-compiler.html#final-numbers-and-a-benchmarking-plot-twist.

tshepang (approved):

cuviper:
> Would something like this be useful to you, in order to build optimized `rustc` binaries for Linux distros?

Some of this I already knew, but I'm not applying all these techniques yet. It's good to have it written down, in any case!

PR #1788, opened 2023-09-04
Set max line length in `.editorconfig` to 100

Kobzol:
It is enforced by CI, but it wasn't set in the config. Noticed in https://github.com/rust-lang/rustc-dev-guide/actions/runs/6077069991/job/16486130135.


JohnTitor (approved):
Thanks!

PR #1802, opened 2023-10-04
update new trait solver docs

lcnr:
wanted to do this for a while, cc @BoxyUwU @compiler-errors


src/solve/invariants.md
```diff
@@ -0,0 +1,140 @@
...
+mean we assume a trait is implemented even though it is not, which is very likely to result in
+actual unsoundness. When using `where`-bounds to prove a goal, the `impl` will be provided by the
+user of the item.
+
+This invariant only holds if we check region constraints. As we do not check region constraints
+during implicit negative overlap check in coherence, this invariant is broken there. As this check
+relies on *completeness* of the trait solver, it is not able to use the current region constraints
+check - `InferCtxt::resolve_regions` - as its handling of type outlives goals is incomplete.
+
+### Normalization of semantically equal aliases empty environments results in a unique type ✅
```
compiler-errors, line 48:
```suggestion
### Normalization of semantically equal aliases in empty environments results in a unique type ✅
```

src/solve/invariants.md
```diff
@@ -0,0 +1,140 @@
...
+It is however very difficult to imagine a sound type system without this invariant, so
+the issue is that the invariant is broken, not that we incorrectly rely on it.
+
+### Generic goals and their instantiations have the same result ✅
+
+Pretty much: If we successfully typecheck a generic function concrete instantiations
+of that function should also typeck. We should not get errors post-monomorphization.
+We can however get overflow errors at that point.
+
+TODO: example for overflow error post-monomorpghization
```
compiler-errors, line 72:
```suggestion
TODO: example for overflow error post-monomorphization
```

src/solve/invariants.md
```diff
@@ -0,0 +1,140 @@
...
+a few of them do not hold in Rust right now. This is either a fundamental to its design or
+caused by bugs and something that may change in the future.
+
+It is important to know about the things you can assume while working on - and with - the
+type system, so here's an incomplete and inofficial list of invariants of
+the core type system:
+
+- ✅: this invariant mostly holds, with some weird exceptions, you can rely on it outside
+of these cases
+- ❌: this invariant does not hold, either due to bugs or by design, you must not rely on
```
compiler-errors, line 16: I wonder if we should distinguish "by design" or "due to bugs".

compiler-errors (approved):

src/solve/invariants.md
spastorino: I think it would be nice to distinguish yeah, maybe we can do that as a followup.

spastorino:
I know this has some FIXMEs and TODOs and the comment from @compiler-errors error that could be addressed but this is a huge improvement over what we currently have so I'm merging this and we can continue improving things in followups.

PR #1808, opened 2023-10-17
Start a chapter about the evolving const effect system

oli-obk:
r? @fee1-dead @compiler-errors


compiler-errors (approved):

src/effects.md
```diff
@@ -0,0 +1,60 @@
+# Effects and effect checking
+
+Note: all of this describes the implementation of the unstable `effects` and
+`const_trait` features. None of this implementation is usable or visible from
```
fee1-dead, line 4: Should this be `const_trait_impl`?

src/effects.md
```diff
@@ -0,0 +1,60 @@
...
+need to know whether there is a `const_trait` impl for the type and trait being
+used (or whether the `const fn` is used at runtime, then any type implementing the
+trait is ok, just like with other bounds).
+
+We perform these checks via a const generic boolean that gets attached to all
+`const fn` and `const trait`. The following sections will explain the desugarings
+and the way we perform the checks at call sites.
+
+The const generic boolean is inverted to the meaning of `const`. In the compiler
+it is called `host`, because it enables "host APIs" like `static` items, network
```
fee1-dead, line 21: We might be renaming that (there were concerns on an earlier PR) but don't think that blocks this PR

src/effects.md
```diff
@@ -0,0 +1,60 @@
...
+a generic parameter, it means "maybe const" (meaning we're in a const fn or const
+trait).
+
+## `const fn`
+
+All `const fn` have a `#[rustc_host] const host: bool` generic parameter that is
+hidden from users. Any `~const Trait` bounds in the generics list or `where` bounds
+of a `const fn` get converted to `Trait<host> + Trait<true>` bounds. The `Trait<true>`
+exists so that associated types of the generic param can be used from projections
+like `<T as Trait>::Assoc`, because there are no `<T as ~Trait>` projections for now.
```
fee1-dead, line 33:
```suggestion
like `<T as Trait>::Assoc`, because there are no `<T as ~const Trait>` projections for now.
```

src/effects.md
```diff
@@ -0,0 +1,60 @@
...
+they have all the regular generic parameters of the trait. Any `~const Trait` super-trait
+bounds get desugared to `Trait<host> + Trait<true>` in order to allow using associated
+types and consts of the super traits in the trait declaration. This is necessary, because
+`<Self as SuperTrait>::Assoc` is always `<Self as SuperTrait<true>>::Assoc` as there is
+no `<Self as ~const SuperTrait>` syntax.
+
+## `typeck` performing method and function call checks.
+
+When generic parameters are instantiated for any items, the `host` generic parameter
+is always instantiated as an inference variable. These variables fall back to `true` at
```
fee1-dead, line 48: We should probably mention that it is a special kind of const infer var just so that the fallback works

src/effects.md
```diff
@@ -0,0 +1,60 @@
...
+the end of typeck to ensure that `let _ = some_fn_item_name;` will keep compiling.
+
+All actually used (in function calls, casts, or anywhere else) function items, will
+have the `enforce_context_effects` method invoked.
+It trivially returns if the function being called has no `host` generic parameter.
+
+TODO: how do we bail out if a non-const function is called in a const context?
```
fee1-dead, line 55: AFAIK that is still in the Mir check_consts pass

src/effects.md
compiler-errors: It should probably reflect what it's called now, and can be renamed later when it gets renamed in the compiler.

src/effects.md
compiler-errors: Yeah, should link or mention `fallback_effects`

fee1-dead (approved):
LGTM!

fee1-dead:
Also make sure to open a PR from a branch on your fork instead 😅

PR #1811, opened 2023-10-22
Include `glibc.static` in `shell.nix`

eopb:
This is now required to link [tests/ui/process/nofile-limit.rs](https://github.com/rust-lang/rust/blob/724ba7fe9084e607c79e7656ac173eb617d173f5/tests/ui/process/nofile-limit.rs)


eopb:
Oops, this breaks more tests than it fixes

PR #1813, opened 2023-10-23
remove change-id from sample build configuration

onur-ozkan:
Since the change-id is frequently updated alongside breaking changes in the build configuration, keeping the change-id in sample configurations synced with the latest change-id becomes challenging. To address this, we are removing the change-id from the sample configurations so that users can directly view the logs and decide whether to use it or not.

ref https://github.com/rust-lang/rust/pull/115898#issuecomment-1775909050


JohnTitor (approved):
Sounds sensible, thanks!

PR #1821, opened 2023-11-04
Add explanations on how to run rustc_codegen_gcc tests

GuillaumeGomez:
As mentioned in https://github.com/rust-lang/rust/pull/117313, adding documentation about how to run cg_gcc tests would be very helpful.

cc @RalfJung

### DIFF
--- a/src/tests/running.md
+++ b/src/tests/running.md
@@ -320,3 +320,36 @@ communicate with the server to coordinate running tests (see
 [remote-test-server]: https://github.com/rust-lang/rust/tree/master/src/tools/remote-test-server
 [src/bootstrap/test.rs]: https://github.com/rust-lang/rust/tree/master/src/bootstrap/test.rs
 
+## Running rustc_codegen_gcc tests
+
+First thing to know is that it only supports linux x86_64 at the moment. We will
+extend its support later on.
+
+You need to update `codegen-backends` value in your `config.toml` file in the
+`[rust]` section and add "gcc" in the array:
+
+```toml
+codegen-backends = ["llvm", "gcc"]
+```
+
+Then you need to install libgccjit 12. For example with `apt`:
+
+```bash
+$ apt install libgccjit-12-dev
+```
+
+Now you can run the following command:
+
+```bash
+$ ./x.py test compiler/rustc_codegen_gcc/
+```
+
+If it cannot find the `.so` library (if you installed it with `apt` for example), you
+need to pass the library file path with `LIBRARY_PATH`:
+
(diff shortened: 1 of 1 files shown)


src/tests/running.md
```diff
@@ -320,3 +320,36 @@ communicate with the server to coordinate running tests (see
...
+
+```bash
+$ apt install libgccjit-12-dev
+```
+
+Don't forget to check if the `.so` is available in your test. With `apt` it's not
+and you need to do:
+
+```bash
+$ ln -s /usr/lib/gcc/x86_64-linux-gnu/12/libgccjit.so /usr/lib/x86_64-linux-gnu/libgccjit.so
```
RalfJung, line 345: This looks like it needs to be run as root? Sorry, but I don't think it is acceptable to ask people to do system-wide modifications for this. I certainly would refuse to ever mutate anything inside `/usr` by hand. `/usr` is reserved for `apt` and `dpkg` to manage, I don't want to cause system instability by messing with that.

GuillaumeGomez: Well, better giving people a solution rather than letting them look around, no? I have no clue why it is packaged like this and I don't want to require people to play with environment variables in order to be able to find the location of the `.so`...

RalfJung: Asking people to change an environment variable (which is a local action only affecting that single `x.py` call, it's entirely ephemeral) is much nicer than asking them to *permanently mutate the system directory with root privileges* (which potentially affects the entire system forever).

Sorry, but I do have pretty strong opinions on not messing with system state. I've screwed up my system that way. IMO it's bad advice to ask people to do that.

GuillaumeGomez: I'm fine with not giving any direction at all. Feel free to close the PR.

RalfJung: I was hoping there's a way to run the tests that's not so painful...

GuillaumeGomez: Maybe it's only specific to `apt` and other package managers are not doing whatever that is. That would be interesting to see.

RalfJung: I found a work-around. People just have to run 
```
LIBRARY_PATH=/usr/lib/gcc/x86_64-linux-gnu/12/ ./x.py test compiler/rustc_codegen_gcc/
```
So let's just tell people to do that.

GuillaumeGomez: Thanks! Much better indeed.

GuillaumeGomez:
Updated! Thanks for the improvements @RalfJung.

PR #1826, opened 2023-11-21
Emphasize that the experts map is not up to date

nouritsu:
From [Zulip Conversation](https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/Looking.20for.20Mentorship.20.28Diagnostics.29)
Adds greater emphasis on the fact that the expert's map is outdated.  
Also cleans up accidental double spaces and formats document.

### DIFF
--- a/src/getting-started.md
+++ b/src/getting-started.md
@@ -48,10 +48,10 @@ guide :)
 ### Experts
 
 Not all `t-compiler` members are experts on all parts of `rustc`; it's a pretty
-large project.  To find out who has expertise on different parts of the
+large project. To find out who has expertise on different parts of the
 compiler, [consult this "experts map"][map].
 
-It's not perfectly complete, though, so please also feel free to ask questions
+The experts map incomplete and not actively maintained, so please also feel free to ask questions
 even if you can't figure out who to ping.
 
 Another way to find experts for a given part of the compiler is to see who has made recent commits.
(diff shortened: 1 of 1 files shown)


ChrisDenton:
Hmm... should we simply remove the map? It seems less than useful if it isn't kept up to date.

nouritsu:
I vote for yes, for the following reasons -
- the map's sole purpose is to connect contributors to experts, its redundant if the list of experts is not up to date
- the community helps one another out even if they are not an expert in a certain topic, both on zulip and on discord
- experts already offer to mentor contributors in the github issues section for rust

workingjubilee:
Yeah, we can simply delete the refs to the experts map.

nouritsu:
I'll push to my repository so this pr is updated

tshepang (approved):

tshepang:
instead of hiding that it exists, we should instead delete it first, and only then remove the reference

nouritsu:
I second that.

spastorino:
I think we should probably delete the expert file and this reference altogether but I'd rather have @davidtwco or @wesleywiser weighing in here.

The idea of documenting somewhere or having a way to see who could we consult on certain areas of the compiler is a good one but pretty hard to maintain and keep up to date.
David, Wesley, any thought about this?.

davidtwco:
> I think we should probably delete the expert file and this reference altogether but I'd rather have @davidtwco or @wesleywiser weighing in here.
> 
> The idea of documenting somewhere or having a way to see who could we consult on certain areas of the compiler is a good one but pretty hard to maintain and keep up to date. David, Wesley, any thought about this?.

I think we should delete the map for now - we should solve this problem, like Santiago says, it's a worthwhile thing to have a solution for - but the out-of-date map is more confusing than helpful.

spastorino:
Opened a PR to remove the experts map https://github.com/rust-lang/compiler-team/pull/698.
If we land that we should update the dev guide again to remove occurrences of it too.

PR #1831, opened 2023-12-05
Replace experts map with reviewers from triagebot

spastorino:
Now that we have removed experts map https://github.com/rust-lang/compiler-team/pull/698, fix the guide accordingly.

cc @davidtwco


JohnTitor (approved):
👍

PR #1835, opened 2023-12-11
Add a chapter on editions.

ehuss:
This adds a chapter to explain how editions work in rustc and the things to consider for different kinds of changes.

I don't know how best to place this in the chapter listing. It doesn't fit any of the "phases" of the compiler, since it covers multiple phases. I decided to start a new "General Guides" section, but I don't know if that really makes sense.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -184,6 +184,10 @@
 - [Sanitizers Support](./sanitizers.md)
 - [Debugging support in the Rust compiler](./debugging-support-in-rustc.md)
 
+# General Guides
+
+- [Editions](guides/editions.md)
+
 ---
 
 [Appendix A: Background topics](./appendix/background.md)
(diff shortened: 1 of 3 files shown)


spastorino (approved):

src/SUMMARY.md
```diff
@@ -184,6 +184,10 @@
 - [Sanitizers Support](./sanitizers.md)
 - [Debugging support in the Rust compiler](./debugging-support-in-rustc.md)
 
+# General Guides
+
+- [Editions](guides/editions.md)
```
JohnTitor, line 189: I guess we could place this under the "# Contributing to Rust" section like the stability attr guide.

ehuss: Sure! Posted https://github.com/rust-lang/rustc-dev-guide/pull/1838.

PR #1840, opened 2023-12-18
bootstrapping: Clarify that stage0 std code is not executed to produce stage1 compiler

Enselic:
The wording

> When compiling `rustc`, the stage0 compiler uses the freshly compiled
> `std`.

can be misread as

> code from `std` compiled by stage0 `rustc` is executed to produce the
> stage1 `rustc`.

Clarify that that is not the case. (I personally misread it that way which has had me confused for a long time.)


JohnTitor (approved):
Thanks!

PR #1841, opened 2023-12-18
Suggest `gcc_multi` to make `mir_opts` run on nixos

Nadrieril:
`./x.py test mir_opts` requires a cross-architecture toolchain. `gcc_multi` seems to have everything needed for it to run on NixOS. I hope there was no particular reason to pin `gcc9` in particular; `gcc_multi` uses the latest available `gcc`.


JohnTitor (approved):
Sounds reasonable 👍

PR #1842, opened 2023-12-23
Add a description of `unpretty=hir` to the HIR docs

mox692:
I believe rust playground use `unpretty=hir` option by default, so I think it would also be nice to add a mention of this option.


JohnTitor (approved):
Makes sense, thanks!

PR #1844, opened 2023-12-30
Prominently mention `profiler = true` on the coverage page

Zalathar:
Enabling the profiler runtime is an essential part of being able to properly work on the coverage instrumentation code.

There's already a mention of it on this page, but it's made in passing and is easy to miss. This patch adds a much more prominent section containing recommended `config.toml` settings, including `profiler = true`.


Zalathar:
Back when I started looking at the coverage code, this was quite a puzzle to me, so hopefully this can help others avoid the same problems.

tshepang (approved):

PR #1849, opened 2024-01-12
Add some explanations for frequently used rustbot commands

chenyukang:


TaKO8Ki (approved):

src/walkthrough.md
```diff
@@ -186,10 +186,17 @@ request that reviewer by leaving a comment on the thread with `r?
...
+some requested changes. When you finished iterating on the changes, you can mark the PR as 
+`S-waiting-on-review` again by leaving a comment with `@rustbot ready`, this will remove the 
+`S-waiting-on-author` label and add the `S-waiting-on-review` label.
+
+Feel free to ask questions or discuss things you don't understand or disagree with. However,
 recognize that the PR won't be merged unless someone on the Rust team approves
-it.
+it. If a reviewer leave a comment like `r=me after fixing ...`, that means they approve the PR and 
+you can merge it with comment with `@bors r=reviwer-github-id`(e.g. `@bors r=eddyb`) to merge it 
+after fixing lefted trivial issues.
```
JohnTitor, line 199: Worth mentioning that r=someone requires permission and the delegate cmd gives it to you.

src/contributing.md
fmease, line 1: In `rustbot.md`, there's a similar section mentioning `@rustbot label -S-waiting-on-author +S-waiting-on-review`. Should also be adjusted I guess.

chenyukang: yeah, I noticed that part, but I think that section is introducing `@rustbot label`, which is a more general way.

I added more explanation about shortcuts after that.

src/walkthrough.md
```diff
@@ -186,10 +186,18 @@ request that reviewer by leaving a comment on the thread with `r?
...
+`S-waiting-on-review` again by leaving a comment with `@rustbot ready`, this will remove the 
+`S-waiting-on-author` label and add the `S-waiting-on-review` label.
+
+Feel free to ask questions or discuss things you don't understand or disagree with. However,
 recognize that the PR won't be merged unless someone on the Rust team approves
-it.
+it. If a reviewer leave a comment like `r=me after fixing ...`, that means they approve the PR and 
+you can merge it with comment with `@bors r=reviwer-github-id`(e.g. `@bors r=eddyb`) to merge it 
+after fixing lefted trivial issues. `r=someone` requires permission and the reviwer use
+delegate command gives it to you in this way.
```
JohnTitor, line 200: Re-thinking this and the below might sound better, how about?

```suggestion
after fixing trivial issues. Note that `r=someone` requires permission and bors could say something like
"🔑 Insufficient privileges..." when commenting `r=someone`. In that case, you have to ask the reviewer
to revisit your PR.
```

chenyukang: looks great.

PR #1855, opened 2024-01-19
Add a perf. bot command and a link to its help page

Kobzol:
Suggested [here](https://github.com/rust-lang/rust/pull/120093#issuecomment-1899238160).

### DIFF
--- a/src/tests/perf.md
+++ b/src/tests/perf.md
@@ -46,5 +46,15 @@ suite against it.
 After the performance tests finish, the bot will post a comment on the PR with
 a summary and a link to a full report.
 
-More details are available in the [perf collector
+If you want to do a perf run for an already built artifact (e.g. for a previous try
+build that wasn't benchmarked yet), you can run this instead:
+
+`@rust-timer build <commit-sha>`
+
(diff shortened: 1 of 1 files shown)


src/tests/perf.md
```diff
@@ -46,5 +46,15 @@ suite against it.
...
 
-More details are available in the [perf collector
+If you want to do a perf. run for an already built artifact (e.g. for a previous try
+build that wasn't benchmarked yet), you can run this instead:
+
+`@rust-timer build <commit-sha>`
+
+You cannot benchmark the same artifact twice though.
+
+More information about the available perf. bot commands can be found
```
WaffleLapkin, line 56: Text above does not have dots after "perf", so
```suggestion
If you want to do a perf run for an already built artifact (e.g. for a previous try
build that wasn't benchmarked yet), you can run this instead:

`@rust-timer build <commit-sha>`

You cannot benchmark the same artifact twice though.

More information about the available perf bot commands can be found
```

WaffleLapkin (approved):
lgtm, modulo nitpick

Kobzol:
Removed dots.

Issue #1812, opened 2023-10-22
`shell.nix` fails to link `tests/ui/process/nofile-limit.rs` on NixOS


eopb:
The [recommended `shell.nix`](https://github.com/rust-lang/rustc-dev-guide/blob/b0ee9ec8fa59a6c7620165e061f4747202377a62/src/building/suggested.md#using-nix-shell) fails to link [`tests/ui/process/nofile-limit.rs`](https://github.com/rust-lang/rust/blob/724ba7fe9084e607c79e7656ac173eb617d173f5/tests/ui/process/nofile-limit.rs).

<details>

<summary><b>Output</b></summary>

```
rust on  master [?] via 🐍 v3.9.17 via 🦀 v1.75.0-nightly via ❄  impure (rustc) took 2m24s
❯ ./x test tests/ui/process/nofile-limit.rs
Building bootstrap
    Finished dev [unoptimized] target(s) in 0.04s
Updating submodule src/tools/cargo
Submodule 'src/tools/cargo' (https://github.com/rust-lang/cargo.git) registered for path 'src/tools/cargo'
Cloning into '/home/ethan/Code/rust-lang/rust/src/tools/cargo'...
fatal: unable to access 'https://github.com/rust-lang/cargo.git/': Recv failure: Connection reset by peer
fatal: clone of 'https://github.com/rust-lang/cargo.git' into submodule path '/home/ethan/Code/rust-lang/rust/src/tools/cargo' failed
Failed to clone 'src/tools/cargo'. Retry scheduled
Cloning into '/home/ethan/Code/rust-lang/rust/src/tools/cargo'...
remote: Enumerating objects: 2122, done.
remote: Counting objects: 100% (2122/2122), done.
remote: Compressing objects: 100% (1474/1474), done.
remote: Total 2122 (delta 560), reused 1307 (delta 431), pack-reused 0
Receiving objects: 100% (2122/2122), 2.82 MiB | 2.85 MiB/s, done.
Resolving deltas: 100% (560/560), done.
remote: Total 0 (delta 0), reused 0 (delta 0), pack-reused 0
remote: Enumerating objects: 361, done.
remote: Counting objects: 100% (361/361), done.
remote: Compressing objects: 100% (143/143), done.
remote: Total 189 (delta 159), reused 70 (delta 43), pack-reused 0
Receiving objects: 100% (189/189), 79.51 KiB | 3.31 MiB/s, done.
[...]

PR #1858, opened 2024-01-21
make shell.nix better

Nilstrieb:
This makes it less opinionated.

cc @Nadrieril @fee1-dead i know you're using nixos

### DIFF
--- a/src/building/suggested.md
+++ b/src/building/suggested.md
@@ -273,7 +273,10 @@ If you're using nix, you can use the following nix-shell to work on Rust:
 pkgs.mkShell {
   name = "rustc";
   nativeBuildInputs = with pkgs; [
-    gcc_multi binutils cmake ninja openssl pkg-config python39 git curl cacert patchelf nix
+    binutils cmake ninja pkg-config python3 git curl cacert patchelf nix
+  ];
+  buildInputs = [
+    openssl glibc.out glibc.static
   ];
   # Avoid creating text files for ICEs.
   RUSTC_ICE = "0";
--- a/src/building/suggested.md
+++ b/src/building/suggested.md
@@ -275,7 +275,7 @@ pkgs.mkShell {
   nativeBuildInputs = with pkgs; [
     binutils cmake ninja pkg-config python3 git curl cacert patchelf nix
   ];
-  buildInputs = [
+  buildInputs = with pkgs; [
     openssl glibc.out glibc.static
   ];
   # Avoid creating text files for ICEs.
--- a/src/building/suggested.md
+++ b/src/building/suggested.md
@@ -283,8 +283,10 @@ pkgs.mkShell {
 }
 ```
 
-Note that when using nix on a not-NixOS distribution, it may be necessary to set **`patch-binaries-for-nix = true` in `config.toml`**.
-Bootstrap tries to detect whether it's running in nix and enable patching automatically, but this detection can have false negatives.
+Note that when using nix on a not-NixOS distribution, it may be necessary to set
+**`patch-binaries-for-nix = true` in `config.toml`**.
+Bootstrap tries to detect whether it's running in nix and enable patching automatically,
+but this detection can have false negatives.
 
 You can also use your nix shell to manage `config.toml`:
 
--- a/src/building/suggested.md
+++ b/src/building/suggested.md
@@ -280,6 +280,23 @@ pkgs.mkShell {
 }
 ```
 
+Note that when using nix on a not-NixOS distribution, it may be necessary to set **`patch-binaries-for-nix = true` in `config.toml`**.
+Bootstrap tries to detect whether it's running in nix and enable patching automatically, but this detection can have false negatives.
+
+You can also use your nix shell to manage `config.toml`:
+
+```nix
+let
+  config = pkgs.writeText "rustc-config" ''
+    # Your config.toml content goes here
+  ''
+pkgs.mkShell {
+  /* ... */
+  # This environment varaible tells bootstrap where our config.toml is.
+  RUST_BOOTSTRAP_CONFIG = config;
+}
+```
+
 ## Shell Completions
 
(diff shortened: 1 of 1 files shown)


5225225:
One thing that would be useful is somehow getting `x` in the PATH and having it work.

Just doing 

```nix
x_application = pkgs.writeShellApplication {
  name = "x";
  text = ''exec ${./x} "$@"'';
};
```
and putting that in `nativeBuildInputs` doesn't work, since the script looks up stuff relative to itself, which is fair enough.

```nix
x_application = pkgs.writeShellApplication {
  name = "x";
  text = ''
      cd "${builtins.toString ./.}"
      ./x "$@"
  '';
};
```
seems to work fine? I don't know nix or bootstrap terribly well so there's probably a cleaner way of doing this.

Nadrieril:
Oh RUSTC_ICE is the setting I didn't know I needed x) So many files turning up for no apparent reason!

I agree with most of the changes except that I really liked the RUSTC_BOOTSTRAP_CONFIG example. I wouldn't have known it was an option but now I can share my config.toml between machines. You could perhaps edit the comment instead, to have it say "this is an example config.toml, edit at your convenience"?

rowan-sl:
In order to fix #1812 `nativeBuildInputs` can be changed to include `glibc.out glibc.static` (*in that order*)
 - additionally, `gcc_multi` can be removed

 for some reason `glibc.out` doesn't create the issues with `stdlib.h` that `glibc` did - [see this discussion](https://discourse.nixos.org/t/how-to-get-this-basic-c-build-to-work-in-a-nix-shell/12262/3)

Nadrieril:
I added `gcc_multi` because it was required to run the `mir_opts` tests (see [this discussion](https://rust-lang.zulipchat.com/#narrow/stream/182449-t-compiler.2Fhelp/topic/Running.20mir-opt.20on.20nixos))

eopb:
I generally agree with this change but I think it would be a shame to get rid of the recommendation of `patch-binaries-for-nix = true`.

That line is needed for nix shells on non-NixOS systems because [bootstrap doesn't have perfect nix detection](https://github.com/eopb/rust/blob/b8c207435c85955d0c0806240d5491f5ccd1def5/src/bootstrap/bootstrap.py#L651). I added it in https://github.com/rust-lang/rustc-dev-guide/pull/1774.

If we are going to get rid of the `config.toml` part of this `shell.nix` we should probably recommend `patch-binaries-for-nix` separately on the same page.

fee1-dead:
Thanks for the ping. I actually don't use a shell.nix for hacking the rust compiler, since I thought it would be more hassle free if I installed these required things globally.

It might be helpful to showcase that using nix to manage config.toml is possible, but leave a comment saying that if anyone wants to change the behavior to use the `./config.toml` instead they can remove it.

Nilstrieb:
> In order to fix #1812 `nativeBuildInputs` can be changed to include `glibc.out glibc.static` (_in that order_)
> 
>     * ~additionally, `gcc_multi` can be removed~ Looks like it was used in a different part than the part of the test suite i ran
> 
> 
> for some reason `glibc.out` doesn't create the issues with `stdlib.h` that `glibc` did - [see this discussion](https://discourse.nixos.org/t/how-to-get-this-basic-c-build-to-work-in-a-nix-shell/12262/3)

lol, i tried this out with awesome results: compiling rustc with this causes a rustc that just hangs (when cargo is invoking it to figure out cfgs etc). no problem, I try to `perf top` to figure out what's going on.

`perf: Segmentation fault`

rowan-sl:
> > In order to fix #1812 `nativeBuildInputs` can be changed to include `glibc.out glibc.static` (_in that order_)
> > ```
> > * ~additionally, `gcc_multi` can be removed~ Looks like it was used in a different part than the part of the test suite i ran
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > for some reason `glibc.out` doesn't create the issues with `stdlib.h` that `glibc` did - [see this discussion](https://discourse.nixos.org/t/how-to-get-this-basic-c-build-to-work-in-a-nix-shell/12262/3)
> 
> lol, i tried this out with awesome results: compiling rustc with this causes a rustc that just hangs (when cargo is invoking it to figure out cfgs etc). no problem, I try to `perf top` to figure out what's going on.
> 
> `perf: Segmentation fault`

it has this problem when using the fix I gave? (or did you mean when `gcc_multi` is removed?)

Nilstrieb:
ok my perf is always broken somehow (???). I did remove `gcc_multi` because it's useless now (as mir opt tests now don't require C anymore) but the change I made is adding `glibc.out glibc.static` to `nativeBuildInputs` (shouldn't it be `buildInputs` too?) which makes programs (build scripts, rustc) hang... will debug it further and report back
@Nemo157 mentioned that it works though

Nilstrieb:
moving it to `buildInputs` fixed the hangs. i do not want to know where they come from. but i will put it there, that seems more appropriate anyways.

Nadrieril:
Can you try running `./x.py test tests/mir-opts --bless` just to be sure? I remember `gcc_multi` being required for entirely silly reasons (bless is important because that runs more tests than without it 😅)

Nilstrieb:
saethlin very recently made this not required anymore, being able to bless without a C toolchain.

Nadrieril:
Ah I see, amazing then :D

src/building/suggested.md
```diff
@@ -270,67 +270,34 @@ If you're using nix, you can use the following nix-shell to work on Rust:
...
-    in
-    pkgs.writeText "rustc-ripgreprc" "--ignore-file=${ignoreFile}";
-in
 pkgs.mkShell {
   name = "rustc";
   nativeBuildInputs = with pkgs; [
-    gcc_multi binutils cmake ninja openssl pkgconfig python39 git curl cacert patchelf nix psutils
+    binutils cmake ninja pkg-config python3 git curl cacert patchelf nix
+  ];
+  buildInputs = [
```
5225225, line 278: this needs `with pkgs;`

Nilstrieb:
This should be fine to merge now. We could also mention this in `x setup` when it detects nix (which it already) can, but that's something for later.

Nadrieril (approved):

Issue #1861, opened 2024-01-27
Clarify what is a UI test


Arthur-Milchior:
>  These are used in ui tests to test whether the expected set of paths exist in the dependency graph.

is found in https://rustc-dev-guide.rust-lang.org/incrcomp-debugging.html

Either UI means User Interface, and I would suggest you clarify how user interfaces are related to this compiler bug.

Or UI means something that is entirely different, and that seems harder to find by searching online for "UI test" or "UI rust", as most answer are related to user interface.

ChrisDenton:
This is documented in the testing section. E.g. https://rustc-dev-guide.rust-lang.org/tests/ui.html#introduction

The one line summary in https://rustc-dev-guide.rust-lang.org/tests/compiletest.html states it like so:

> [ui](https://rustc-dev-guide.rust-lang.org/tests/ui.html) — tests that check the stdout/stderr from the compilation and/or running the resulting executable

Arthur-Milchior:
Thank you. I fear in the meantime I forgot about it. 
May I suggest just adding a link on `ui` to this page?

ChrisDenton:
Yes, adding a link would make sense. Would you be willing to make a pull request?

Arthur-Milchior:
Already done https://github.com/rust-lang/rustc-dev-guide/pull/1862

PR #1862, opened 2024-01-27
Clarify debugging graph dependency

Arthur-Milchior:
I found the explanation very confusing. I tried to take the same test and give more details, so that the next reader could understand this part of the test even if they are not that used to macros.

I also added a link to the example file mentioned

Closes https://github.com/rust-lang/rustc-dev-guide/issues/1861


JohnTitor (approved):
Thanks!

PR #1868, opened 2024-01-28
Add links for arena and interning.

Arthur-Milchior:
I suspect that people who don't already know about arenas may be interested in the wikipedia page. It may not be the best material on this topic, but I could not find anything that is specific to the arena used by rustc, and this is still a good first approach.

I also copied the link about interning from overview to memory, as the reader may have forgotten that interning was already discussed previously.  Admittedly, the wikipedia page is called string interning, but it mentions "Objects other than strings can be interned.", and still explains the idea in a useful manner.

I suspect it may also be useful to link to
https://en.wikipedia.org/wiki/Hash_consing , as it seems the standard way to ensure that each tree (in this case, the AST of a type) is unique in memory. Except that I don't know whether it's actually what Rust uses.

### DIFF
--- a/src/memory.md
+++ b/src/memory.md
@@ -6,7 +6,7 @@ it will take a lot of time and space to do so.
 
 One of the main way the compiler manages this is using [arena]s and [interning].
 
-[arena]: (https://en.wikipedia.org/w/index.php?title=Arena_allocation)
+[arena]: https://en.wikipedia.org/wiki/Region-based_memory_management
 [interning]: https://en.wikipedia.org/wiki/String_interning
 
 ## Arenas and  Interning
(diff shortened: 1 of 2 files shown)


src/memory.md
```diff
@@ -4,7 +4,10 @@ Rustc tries to be pretty careful how it manages memory. The compiler allocates
 _a lot_ of data structures throughout compilation, and if we are not careful,
 it will take a lot of time and space to do so.
 
-One of the main way the compiler manages this is using arenas and interning.
+One of the main way the compiler manages this is using [arena]s and [interning].
+
+[arena]: (https://en.wikipedia.org/w/index.php?title=Arena_allocation)
```
JohnTitor, line 9:
```suggestion
[arena]: https://en.wikipedia.org/wiki/Region-based_memory_management
```

JohnTitor (approved):

PR #1880, opened 2024-02-05
Update for upcoming markdown changes.

ehuss:
An upcoming version of mdbook will have some minor changes in how it interprets markdown to better conform with the standards. The changes are:

* Fixed the missing footnote definition for `[^core]`.
* Headings now allow arbitrary attributes inside curly braces at the end to set HTML attributes. There was one heading where this caused it to be misinterpreted as an attribute. Fixed by using backslashes to avoid the attribute parsing.
* The `llvm-and-covmap-versions` footnote wasn't following GitHub-flavor footnote syntax. I adjusted it so it works in both the old and new versions.

### DIFF
--- a/src/borrow_check/drop_check.md
+++ b/src/borrow_check/drop_check.md
@@ -152,3 +152,5 @@ One possible way these inconsistencies can be fixed is by MIR building
 to be more pessimistic, probably by making `Ty::needs_drop` weaker, or
 alternatively, changing `dropck_outlives` to be more precise, requiring
 fewer regions to be live.
+
+[^core]: This is the core assumption of [#110288](https://github.com/rust-lang/rust/issues/110288) and [RFC 3417](https://github.com/rust-lang/rfcs/pull/3417).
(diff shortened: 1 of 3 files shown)


src/borrow_check/drop_check.md
```diff
@@ -146,7 +146,7 @@ a type is contained in a larger local. **This is inconsistent, and
 should be fixed: an example [for
 arrays](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8b5f5f005a03971b22edb1c20c5e6cbe)
 and [for
-`PhantomData`](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=44c6e2b1fae826329fd54c347603b6c8).**[^core]
```
spastorino, line 149: Found what was supposed to be core.

`[^core]: This is the core assumption of [#110288] and [RFC 3417].`

Can you add that?

spastorino (changes requested):

src/borrow_check/drop_check.md
ehuss: Sure, it should be updated now.

JohnTitor (approved):
LGTM once @spastorino checked the follow-up, thanks!

spastorino (approved):

PR #1888, opened 2024-02-14
fix update llvm lld build command

gnana-ganesh-tw:


gnana-ganesh-tw:
This was fixed in [this PR](https://github.com/rust-lang/rustc-dev-guide/pull/1887)

PR #1908, opened 2024-02-26
Use different type in an example

Arthur-Milchior:
Sentences such as «without the argument u32» were ambiguous, as there were two distincts u32. Having a single one, the one in the monomorphization of the type, remove the ambiguity.

### DIFF
--- a/src/ty.md
+++ b/src/ty.md
@@ -281,7 +281,7 @@ modules choose to import a larger or smaller set of names explicitly.
 Let's consider the example of a type like `MyStruct<u32>`, where `MyStruct` is defined like so:
 
 ```rust,ignore
-struct MyStruct<T> { x: u32, y: T }
+struct MyStruct<T> { x: str, y: T }
 ```
 
 The type `MyStruct<u32>` would be an instance of `TyKind::Adt`:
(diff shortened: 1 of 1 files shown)


src/ty.md
```diff
@@ -281,7 +281,7 @@ modules choose to import a larger or smaller set of names explicitly.
 Let's consider the example of a type like `MyStruct<u32>`, where `MyStruct` is defined like so:
 
 ```rust,ignore
-struct MyStruct<T> { x: u32, y: T }
+struct MyStruct<T> { x: str, y: T }
```
fmease, line 284: This is not valid. Unsized types like `str` cannot be used here, they may only appear as the type of the very last field.

Please choose a different type.

fmease (approved):
Your explanation makes sense and I agree. I request one change though.

src/ty.md
Nilstrieb, line 284:
```suggestion
struct MyStruct<T> { x: u8, y: T }
```
let's choose this type

Nilstrieb:
@fmease you could have just applied a suggestion yourself^^

fmease:
Yeah, I could've :) but I wanted to give the author the chance to update it themself :P

PR #1921, opened 2024-03-02
Mention forking before cloning the Rust repo

mmicu:


src/building/how-to-build-and-run.md
```diff
@@ -13,7 +13,9 @@ and a bunch of tools (e.g. `rustdoc`, the bootstrapping infrastructure, etc).
 
 [repo]: https://github.com/rust-lang/rust
 
-The very first step to work on `rustc` is to clone the repository:
+The very first step to work on `rustc` is to clone the repository.
+However, you might consider to fork it first.
```
JohnTitor, line 17: This chapter is "How to build and run the compiler" and I'm not sure how it's useful, while I totally agree if some wants to _contribute_.

jieyouxu:
Triage: I'm inclined to agree with the assessment that it's strange to mention forking on the build and run the compiler page (but would make sense for the contributing page). Feel free to send a PR that adds the forking mention to the contributing page if such remark doesn't already exist. Thanks for the PR anyways!

PR #1926, opened 2024-03-04
Update parallel-rustc.md

Tbkhi:
Minor updates to syntax and improvements for readability.


Tbkhi:
CC: @spastorino

src/parallel-rustc.md
```diff
@@ -1,31 +1,35 @@
...
-these parallelization functions. **To try out the current parallel compiler**, 
-one can install rustc from source code with `parallel-compiler = true` in 
-the `config.toml`.
+As of <!-- date-check --> August 2022, the only stage of the compiler that is
+parallel is [code generation stage][codegen] (codegen). Some parts of the
+compiler have parallel implementations, such as query evaluation, type check
+and [monomorphization][monomorphization], but the general version of the
+compiler does not include parallelization functions. **To try out the current
+parallel compiler**, install `rustc` from source code with `parallel-compiler =
+true` in the `Config.toml`.
```
jieyouxu, line 9: cc @SparrowLii I feel like some of this is quite outdated, right?

SparrowLii: Most of the processes in the rust compiler have been parallelized.

The codegen part is executed concurrently by default.  Users can control the number of tasks with the `-C codegen-units` option.

The compilation parts after HIR lowering to codegen, such as type checking, borrow checking and mir optimization have been parallelized in the nightly version. Currently, they are executed in serial by default, and parallelization is manually enabled by the user using the `-Z threads = n` option.

Other parts such as lex parsing, HIR lowering, and macro expansion are still executed in serial mode.

SparrowLii (commented):
Thanks for the PR! 

A lot of content in this article needs to be updated, my opinion is we merge this PR first, after the [#132282](https://github.com/rust-lang/rust/pull/132282) merged I will carry on the complete update work (should be next monday).

src/parallel-rustc.md
SparrowLii, line 9: Most of the rust compilers are now parallelized.

The codegen part is executed concurrently by default. You can use the `-C codegen-units=n` option to control the number of concurrent tasks.

The parts after HIR lowering to codegen such as type checking, borrowing checking, and mir optimization are parallelized in the nightly version. Currently, they are executed in serial by default, and parallelization is manually enabled by the user using the `-Z threads = n` option.

Other parts, such as lexical parsing, HIR lowering, and macro expansion, are still executed in serial mode.

jieyouxu:
Yeah, let's merge this first, we can update in a follow-up.

src/parallel-rustc.md
jieyouxu: I'll slightly adjust this to reflect your wording and then back link to the issue to update the docs for parallel front-end.

jieyouxu:
Thanks for the PR!

PR #1935, opened 2024-03-11
Update name-resolution.md

Tbkhi:
Add links makes minor edits.


Noratrieb (approved):
thanks a lot, i've made some small edits (some of the words shouldn't be in a code block). especially the links are very helpful!

PR #1940, opened 2024-03-16
Correct an example

Arthur-Milchior:
The function, as written, is not valid. `a` is not bounded. Given the parameter and the return type, I assume we are considering the identity function for u32. In this case, the return value should be `b`. As I was touching this code, I considered also changing the function name, for the sake of the clarity.

### DIFF
--- a/src/early-late-bound-summary.md
+++ b/src/early-late-bound-summary.md
@@ -3,7 +3,7 @@
 This section discusses what it means for generic parameters to be early or late bound.
 
 ```rust
-fn foo<'a, T>(b: &'a u32) -> &'a u32 { a }
+fn id_u32<'a, T>(b: &'a u32) -> &'a u32 { b }
 //     ^^  ^early bound
 //     ^^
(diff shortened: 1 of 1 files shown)


src/early-late-bound-summary.md
```diff
@@ -3,7 +3,7 @@
 This section discusses what it means for generic parameters to be early or late bound.
 
 ```rust
-fn foo<'a, T>(b: &'a u32) -> &'a u32 { a }
+fn id_u32<'a, T>(b: &'a u32) -> &'a u32 { b }
 //     ^^  ^early bound
 //     ^^
 //     ^^late bound
```
jieyouxu, line 9: Could you also fix the alignment of the `^^`s?

Arthur-Milchior:
Very sorry for the delay @jieyouxu . Corrected. Thanks for the review.

Arthur-Milchior:
oops, actually it was already solved. Great.

PR #1951, opened 2024-03-30
Add quickstart for how to build and run the compiler

Nilstrieb:
The chapter is quite long, and a lot of the information is, while valuable, not very important for newcomers. I think it makes sense to have a condensed version for anyone just wanting to get started with only the most important information.

### DIFF
--- a/src/building/quickstart.md
+++ b/src/building/quickstart.md
@@ -1,0 +1,60 @@
+# Quickstart
+
+This is a quickstart guide about getting the compiler running. For more information in the individual steps,
+see the other pages in this chapter.
+
+First, clone the repository:
+
+```sh
+git clone https://github.com/rust-lang/rust.git
+cd rust
+```
+
+When building the compiler, we don't use `cargo` directly, instead we use a wrapper called "x".
+On Unix-like platforms, use `./x`. On Windows, use `.\x.ps1`. This guide will use `./x`.
+
+We need to create a configuration for the build. Use `./x setup` to create a good default.
+
+```sh
+./x setup
+```
+
+Then, we can build the compiler. Use `./x build` to build the compiler, standard library and a few tools.
+You can also `./x check` to just check it.
+All these commands can take specific components/paths as arguments, for example `./x check compiler` to just check the compiler.
+
+```sh
+./x build
+```
+
+> When doing a change to the compiler that does not affect the way it compiles the standard library
+(so for example, a change to an error message), use `--keep-stage-std 1` to avoid recompiling it.
+
+After building the compiler and standard library, you now have a working compiler toolchain.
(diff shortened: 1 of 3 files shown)


src/building/quickstart.md
```diff
@@ -0,0 +1,60 @@
...
+
+```sh
+rustc +stage testfile.rs
+```
+
+After doing a change, you can run the compiler test suite with `./x test`.
+
+`./x test` runs the full test suite, which is slow and rarely what you want.
+Usually, `./x test tests/ui` is what you want after a comiler change,
+testing all "UI" tests that invoke the compiler on a specific test file and check the output.
```
fee1-dead, line 50: It would probably make sense to link the chapter on ui tests here if they want to read more on that.

src/building/quickstart.md
```diff
@@ -0,0 +1,60 @@
...
+
+```sh
+./x test tests/ui
```
compiler-errors, line 53: `--bless`?

compiler-errors (approved):

src/building/quickstart.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+> `./x suggest` can also be helpful for suggesting which tests to run after a change.
+
+Congrats, you are now ready to make a change to the compiler! If you have more questions,
```
fee1-dead, line 58: as a next step, could direct them to suggested workflows to setup their editor properly.

src/building/quickstart.md
```diff
@@ -0,0 +1,60 @@
...
+You can use it with rustup by linking it.
+
+```sh
+rustup toolchain link stage1 build/host/stage1
+```
+
+Now you have a toolchain called `stage1` linked to your build. You can use it to test the compiler.
+
+```sh
+rustc +stage testfile.rs
```
fee1-dead, line 43:
```suggestion
rustc +stage1 testfile.rs
```

Nilstrieb:
I am not going to fix the length check CI failure, I'm going to wait for #1952 instead.

src/building/quickstart.md
```diff
@@ -0,0 +1,60 @@
...
+
+First, clone the repository:
+
+```sh
+git clone https://github.com/rust-lang/rust.git
+cd rust
+```
+
+When building the compiler, we don't use `cargo` directly, instead we use a wrapper called "x".
+On Unix-like platforms, use `./x`. On Windows, use `.\x.ps1`. This guide will use `./x`.
```
ChrisDenton, line 14: `./x` *should* work on Windows too, if run from powershell, because it'll autocomplete the extension.

Nilstrieb: oh, i will remove the sentence about windows and hope no one gets confused

PR #1952, opened 2024-03-30
Delete length check

Nilstrieb:
It's super annoying to be forced to use this bad convention, and apparently everyone agrees. The only reason no improvements have been done is because those were blocked on writing a better checker(#1485).

I strongly believe that no checker is better than a bad checker, so let's just delete it in the meantime. I kindly asked anyone who sees this to complain about overly long sentences in review in the future, I think we can make this turn out fine.

addresses https://github.com/rust-lang/rustc-dev-guide/issues/1132#issuecomment-2028241585

@tshepang fyi, but imo anyone can approve this


lcnr (approved):

lcnr:
agree with this. The linelength limit causes a significant annoyance because mdbook and github disagree on whether they should be rendered

PR #1953, opened 2024-04-01
Rewrite the `Parameter Environments` chapter

BoxyUwU:
The current section is confusing even to some t-types members, and also generally I have spoken to a large amount of people who did not understand how ParamEnv worked at all even after reading the current section on it.

Solution: rewrite it xd

Would quite like to find someone who currently doesnt understand `ParamEnv` and get them to read through this with me so I can figure out if it is actually more helpful and what problems crop up in trying to understand this :sparkles:

Also change the location of the chapter to be before most of the chapters on analysis as an understanding of what a `ParamEnv` is, is pretty useful in understanding trait solving and whatnot

cc @compiler-errors


compiler-errors:
I'll definitely review this tomorrow, please annoy me if I forget!

src/param_env.md
```diff
@@ -1,69 +1,159 @@
...
-functions/methods) it is often relevant to have more information about the
-`Self` or generic parameters. Trait bounds and similar information is encoded in
-the [`ParamEnv`][pe]. Often this is not enough information to obtain things like the
-type's `Layout`, but you can do all kinds of other checks on it (e.g. whether a
-type implements `Copy`) or you can evaluate an associated constant whose value
-does not depend on anything from the parameter environment.
+## What is it
 
-[pe]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.ParamEnv.html
+The type system relies on information in the environment in order for it to function correctly. This information is stored in the [`ParamEnv`][pe] type and it is important to use the correct `ParamEnv` when interacting with the type system.
```
compiler-errors, line 5: Should we be more clear abt what an "environment" means here before diving into the details?

src/param_env.md
```diff
@@ -1,69 +1,159 @@
...
 
-For example if you have a function
+The information represented by `ParamEnv` is a list of in scope where clauses, and a [`Reveal`][reveal]. A `ParamEnv` typically corresponds to a specific item's environment however it can also be created with arbitrary data that is not derived from a specific item. In most cases `ParamEnv`s are initially created via the [`param_env` query][query] which returns a `ParamEnv` derived from the provided item's where clauses.
```
compiler-errors: Should we explain what a `Reveal` is too?

compiler-errors: > A `ParamEnv` typically corresponds to a specific item's environment

Maybe reworded to:

> A `ParamEnv` is typically derived from the `where` clauses written on a specific item

(maybe also adding "plus some other implicit assumptions like `Sized` and `ConstArgHasType`")

compiler-errors: Should we also motivate the part:

> however it can also be created with arbitrary data that is not derived from a specific item

By explaining when this may be true, such as `compare_method_predicate_entailment`?

src/param_env.md
```diff
@@ -1,69 +1,159 @@
...
 ```rust
-fn foo<T: Copy>(t: T) { ... }
+// `foo` would have a `ParamEnv` of:
+// `[T: Sized, T: Trait, <T as Trait>::Assoc: Clone]`
+fn foo<T: Trait>()
+where
+    <T as Trait>::Assoc: Clone,
+{}
 ```
+If we were interacting with the type system from inside of `foo` we would use this `ParamEnv` everywhere that we interact with the type system. This would allow things such as normalization (TODO: write a chapter about normalization and link it), evaluating generic constants, and proving where clauses/goals, to rely on `T` being sized, implementing `Trait`, etc.
```
compiler-errors, line 18:
```suggestion
If we were interacting with the type system (for example, type-checking or implementing a lint) from inside of `foo` we would use this `ParamEnv` everywhere that we interact with the type system. This would allow things such as normalization (TODO: write a chapter about normalization and link it), evaluating generic constants, and proving where clauses/goals, to rely on `T` being sized, implementing `Trait`, etc.
```

src/param_env.md
```diff
@@ -1,69 +1,159 @@
...
 
-the parameter environment for that function is `[T: Copy]`. This means any
-evaluation within this function will, when accessing the type `T`, know about
-its `Copy` bound via the parameter environment.
+A more concrete example:
+```rust
+// `foo` would have a `ParamEnv` of:
+// `[T: Sized, T: Clone]`
+fn foo<T: Clone>(a: T) -> (T, T) {
```
fmease, line 24:
```suggestion
fn foo<T: Clone>(a: T) {
```

foo doesn't actually return a tuple

src/param_env.md
fmease, line 7: maybe “in-scope where clauses” or even “in-scope where-clauses” for legibility, up to preference

src/param_env.md
```diff
@@ -1,69 +1,159 @@
...
 
-You can get the parameter environment for a `def_id` using the
-[`param_env`][query] query. However, this `ParamEnv` can be too generic for
-your use case. Using the `ParamEnv` from the surrounding context can allow you
-to evaluate more things. For example, suppose we had something the following:
+Or alternatively an example that would not compile:
+```rust
+// `foo2` would have a `ParamEnv` of:
+// `[T: Sized]`
+fn foo2<T>(a: T) -> (T, T) {
```
fmease, line 40:
```suggestion
fn foo2<T>(a: T) -{
```

ditto

src/param_env.md
```diff
@@ -1,69 +1,159 @@
...
+    // When typechecking `foo2` we attempt to prove `T: Clone`.
+    // As we are type checking `foo2` we use `foo2`'s environment
+    // when trying to prove `T: Clone`.
+    //
+    // Trying to prove `T: Clone` with a `ParamEnv` of `[T: Sized]` will
+    // fail as there is nothing in the environment telling the trait solver
+    // that `T` implements `Clone`.
```
fmease, line 47: I feel like this should clarified. If we can't find anything in the `ParamEnv` there may still be other candidates like user-written impls. Might sound a bit nit-picky but I think this could fortify the reader's understanding. Of course, there's no such thing as `impl<T> Clone for T { … }` but we might want mention that somehow.

BoxyUwU: no no this is perfect thank you for pointing this out. I remember vaguely thinking of this when writing but it must have slipped my mind thanks <3 I definitely dont want to give the impression that the _only_ way of proving things is via the bounds in `ParamEnv`

compiler-errors (commented):
i was partly through leaving some reviews when fmease came thru, some of these may be not relevant anymore but im putting them up so they dont bitrot (e.g. if lcnr also reviews tomorrow while i am asleep haha)

src/param_env.md
```diff
@@ -1,69 +1,160 @@
...
+This workaround does not work in all cases as normalizing associated types requires a `ParamEnv` which introduces a bootstrapping problem. We need a normalized `ParamEnv` in order for normalization to give correct results, but we need to normalize to get that `ParamEnv`. Currently we normalize the `ParamEnv` once using the unnormalized param env and it tends to give okay results in practice even though there are some examples where this breaks ([example]).
+
+In the next-gen trait solver the requirement for all where clauses in the `ParamEnv` to be fully normalized is not present and so we do not normalize when constructing `ParamEnv`s.
+
+[example]: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e6933265ea3e84eaa47019465739992c
+
+---
+
+When needing a `ParamEnv` in the compiler there generally three options for obtaining one:
+- The correct env is already in scope simply use it (or pass it down the call stack to where you are)
```
compiler-errors, line 139: Maybe explain where you're likely to find it: in `FnCtxt`, in `LateLintContext`, anywhere else? Maybe `WfCheckingCtxt`?

src/param_env.md
```diff
@@ -1,69 +1,160 @@
...
+- Call `tcx.param_env(def_id)`
```
compiler-errors, line 140: (or `param_env_reveal_all_normalized` -- maybe worth mentioning and explaining why that may be desired)

Also are we forgetting `ParamEnv::reveal_all()`? :thinking: Useful for monomorphization usages

compiler-errors (commented):
(some other random comments when i was reading from the bottom up, will review the middle parts tomorrow :smile_cat:)

src/param_env.md
fmease: and some kinds of implied outlives-bounds IINM

compiler-errors: Ya at least the ones from structs

src/param_env.md
```diff
@@ -1,69 +1,160 @@
...
+}
+
+fn requires_impl<T: Clone + SuperSuperTrait>(a: T) {}
+```
+
+If we did not elaborate the env then the `requires_impl` call would fail to typecheck as we would not be able to prove `T: Clone` or `T: SuperSuperTrait`. In practice we elaborate the env which means that `bar`'s `ParamEnv` is actually:
+`[T: Sized, T: Copy, T: Clone, T: Trait, T: SuperTrait, T: SuperSuperTrait]`
+This allows us to prove `T: Clone` and `T: SuperSuperTrait` when type checking `bar`.
+
+The `Clone` trait has a `Sized` supertrait however we do not end up with two `T: Sized` bounds in the env (one for the supertrait and one for the implicitly added `T: Sized` bound). This is because the elaboration process (implemented via [`util::elaborate`][elaborate]) deduplicates the where clauses to avoid this.
```
jieyouxu, line 81: The links can be very helpful 👍

jieyouxu (commented):
As someone who is :ferrisClueless: about typeck and `ParamEnv`, this looks great to me and is a very clear explanation on what `ParamEnv` is / what it's used for. Pointing out the pitfalls is also very 😎.

src/param_env.md
BoxyUwU: Perhaps though I don't actually know how to do that. I'm kind of using environment here as a handwavey vibes based thing because I can't put into words what I mean concretely :sweat_smile:

src/param_env.md
BoxyUwU: Updated this section to talk about how to acquire `ParamEnv`s with `Reveal::All` and when this is desired.

src/param_env.md
BoxyUwU: I think the existing doc comments on `Reveal` do an adequate job of describing what the purpose of `Reveal` is. I added a note explicitly instructing people to click on the link if they want information.

src/param_env.md
```diff
@@ -1,69 +1,160 @@
...
+    // Trying to prove `T: Clone` with a `ParamEnv` of `[T: Sized]` will
+    // fail as there is nothing in the environment telling the trait solver
+    // that `T` implements `Clone` and there exists no user written impl
+    // that could apply.
+    requires_clone(a);
+}
+```
 
-[query]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_ty_utils/ty/fn.param_env.html
+It's very important to use the correct `ParamEnv` when interacting with the type system as otherwise it can lead to ICEs or things compiling when they shouldn't (or vice versa). See [#82159](https://github.com/rust-lang/rust/pull/82159) and [#82067](https://github.com/rust-lang/rust/pull/82067) as examples of PRs that changed rustc to use the correct param env to avoid ICE.
```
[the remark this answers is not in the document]
BoxyUwU: We do this to some degree in the 3rd subchapter which is hopefully enough. It is pretty difficult to give general guidance here yeah :sweat_smile: hopefully what we have now is adequate

compiler-errors (approved):
✨

BoxyUwU:
big thanks to all of you for the reviews it was invaluable <3

PR #1958, opened 2024-04-05
Fix some broken links under bootstrapping.

pmundkur:
Two links may have broken when `bootstrapping` moved to its own section; another broke when `cc-rs` moved docs to `docs.rs` from its github repo.


jieyouxu (approved):
Ah yeah, thanks for the fixes, I forgor to update the links when I moved bootstrap stuff into a subfolder.

PR #1960, opened 2024-04-08
compiletest: specify which special env var and which particular CI job

jieyouxu:
The description was irking me because it casually drops "a special env var" and "one CI job" but refuses to elaborate.

### DIFF
--- a/src/tests/headers.md
+++ b/src/tests/headers.md
@@ -168,8 +168,9 @@ The following header commands will check LLVM support:
   Note: The test will fail on CI if the component does not exist.
 * `needs-matching-clang` — ignores if the version of clang does not match the
   LLVM version of rustc.
-  These tests are always ignored unless a special environment variable is set
-  (which is only done in one CI job).
+  These tests are always ignored unless a special environment variable
+  `RUSTBUILD_FORCE_CLANG_BASED_TESTS` is set
+  (which is only done in one CI job `x86_64-gnu-debug`).
 
 See also [Debuginfo tests](compiletest.md#debuginfo-tests) for headers for
 ignoring debuggers.
--- a/src/tests/headers.md
+++ b/src/tests/headers.md
@@ -168,15 +168,16 @@ The following header commands will check LLVM support:
   Note: The test will fail on CI if the component does not exist.
 * `needs-matching-clang` — ignores if the version of clang does not match the
   LLVM version of rustc.
-  These tests are always ignored unless a special environment variable
-  `RUSTBUILD_FORCE_CLANG_BASED_TESTS` is set
-  (which is only done in one CI job `x86_64-gnu-debug`).
+  These tests are always ignored unless a special environment variable,
+  `RUSTBUILD_FORCE_CLANG_BASED_TESTS`, is set
+  (which is only done in one CI job [`x86_64-gnu-debug`]).
 
 See also [Debuginfo tests](compiletest.md#debuginfo-tests) for headers for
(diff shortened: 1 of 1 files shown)


tshepang:
how did you find that that test uses that env var

jieyouxu:
I'm not 100% sure this is right, but this is what I followed:

Working backwards from compiletest directive parsing `needs-matching-clang`:

https://github.com/rust-lang/rust/blob/75fd074338801fba74a8cf7f8c48c5c5be362d08/src/tools/compiletest/src/header/needs.rs#L102-L106

`run_clang_based_tests_with` corresponds to the compiletest flag `run-clang-based-tests-with`

https://github.com/rust-lang/rust/blob/ab3dba92db355b8d97db915a2dca161a117e959c/src/tools/compiletest/src/lib.rs#L233

which is set if

https://github.com/rust-lang/rust/blob/ab3dba92db355b8d97db915a2dca161a117e959c/src/bootstrap/src/core/build_steps/test.rs#L1895-L1898

which is set if

https://github.com/rust-lang/rust/blob/ab3dba92db355b8d97db915a2dca161a117e959c/src/bootstrap/src/utils/helpers.rs#L153C1-L169C2

which is set if env var `RUSTBUILD_FORCE_CLANG_BASED_TESTS` is set.

The only CI job that has this env var is `x86_64-gnu-debug`

https://github.com/rust-lang/rust/blob/ab3dba92db355b8d97db915a2dca161a117e959c/src/ci/docker/host-x86_64/x86_64-gnu-debug/Dockerfile#L32

If someone actually happens to know if this is wrong, please feel free to specify the right env var and CI job 😆

jieyouxu:
I'm also not actually sure if the description

> ignores if the version of clang does not match the LLVM version of rustc.

is accurate.

src/tests/headers.md
```diff
@@ -168,8 +168,9 @@ The following header commands will check LLVM support:
   Note: The test will fail on CI if the component does not exist.
 * `needs-matching-clang` — ignores if the version of clang does not match the
   LLVM version of rustc.
-  These tests are always ignored unless a special environment variable is set
-  (which is only done in one CI job).
+  These tests are always ignored unless a special environment variable
+  `RUSTBUILD_FORCE_CLANG_BASED_TESTS` is set
+  (which is only done in one CI job `x86_64-gnu-debug`).
```
tshepang, line 173: can you add link to this file

src/tests/headers.md
```diff
@@ -168,8 +168,9 @@ The following header commands will check LLVM support:
   Note: The test will fail on CI if the component does not exist.
 * `needs-matching-clang` — ignores if the version of clang does not match the
   LLVM version of rustc.
-  These tests are always ignored unless a special environment variable is set
-  (which is only done in one CI job).
+  These tests are always ignored unless a special environment variable
+  `RUSTBUILD_FORCE_CLANG_BASED_TESTS` is set
```
tshepang, line 172:
```suggestion
  These tests are always ignored unless a special environment variable,
  `RUSTBUILD_FORCE_CLANG_BASED_TESTS`, is set
```

src/tests/headers.md
jieyouxu: Added a link to the Dockerfile

src/tests/headers.md
jieyouxu: Added commas

tshepang (approved):

PR #1963, opened 2024-04-13
Add a note about checks with multiple revisions

tgross35:
I had to dig around a bit to figure this out, so update the docs to make it more clear.


jieyouxu (approved):
Thanks for the docs!

PR #1966, opened 2024-04-24
Add documentation for check-cfg by-default in UI tests

Urgau:
This PR adds some (very conservative) documentation for a new compiletest directive regarding check-cfg; as well as documentation regarding the use of the `FALSE` cfg as convention for an always-false config.

Related to https://github.com/rust-lang/rust/pull/124345 and shouldn't be merged before that PR.

### DIFF
--- a/src/tests/ui.md
+++ b/src/tests/ui.md
@@ -363,7 +363,7 @@ with the different outputs of the different revisions.
 
 > Note: cfg revisions also work inside the source code with `#[cfg]` attributes.
 > 
-> By-convention the `FALSE` cfg is used to have an always false config.
+> By convention, the `FALSE` cfg is used to have an always-false config.
 
 ## Controlling pass/fail expectations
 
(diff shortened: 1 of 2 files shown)


src/tests/ui.md
```diff
@@ -361,6 +361,9 @@ multiple `.stderr` files for the different outputs.
 In the example above, there would be a `.mir.stderr` and `.thir.stderr` file
 with the different outputs of the different revisions.
 
+> Note: cfg revisions also work inside the source code with `#[cfg]` attributes.
+> 
+> By-convention the `FALSE` cfg is used to have an always false config.
```
tshepang, line 366:
```suggestion
> By convention, the `FALSE` cfg is used to have an always-false config.
```

Urgau:
@jieyouxu https://github.com/rust-lang/rust/pull/124345 has been merged, this doc update is ready.

PR #1971, opened 2024-05-02
Update CI documentation to avoid mentioning `expand-yaml-anchors`

Kobzol:
Reflects the recent CI changes on `rust-lang/rust`. I will later also update the [Forge](https://forge.rust-lang.org/infra/docs/rustc-ci.html) (actually, I would like to copy the forge CI page here and then just link to rustc-dev-guide from the Forge).

### DIFF
--- a/src/tests/ci.md
+++ b/src/tests/ci.md
@@ -50,7 +50,21 @@ it is merged into the main branch.
 You can copy one of the definitions from the `auto` section to the `pr` or `try` sections.
 For example, the `x86_64-msvc` job is responsible for running the 64-bit MSVC tests.
 You can copy it to the `pr` section to cause it to be executed after a commit is pushed to your
-PR.
+PR, like this:
+
+```yaml
+pr:
+  ...
+  - image: x86_64-gnu-tools
+    <<: *job-linux-16c
+  # this item was copied from the `auto` section
+  # vvvvvvvvvvvvvvvvvv
+  - image: x86_64-msvc
+    env:
+      RUST_CONFIGURE_ARGS: --build=x86_64-pc-windows-msvc --enable-profiler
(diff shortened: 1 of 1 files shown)


src/tests/ci.md
```diff
@@ -40,35 +40,31 @@ For example, if a Windows build fails, but you don't have access to a Windows
...
+it is merged into the main branch.
 
-For example, the `x86_64-msvc-1` and `x86_64-msvc-2` jobs are responsible for
-running the 64-bit MSVC tests.
-You can copy those up to the `jobs.pr.strategy.matrix.include` section with
-the other jobs.
+You can copy one of the definitions from the `auto` section to the `pr` or `try` sections.
+For example, the `x86_64-msvc` job is responsible for running the 64-bit MSVC tests.
+You can copy it to the `pr` section to cause it to be executed after a commit is pushed to your
+PR.
```
jieyouxu, line 53: I think this should mention that you should not copy the YAML anchors in `env` if the job description has it.

jieyouxu: Also could you add an example for adding a job to PR CI (where the job description has a YAML anchor, such as one of the apple jobs)? I think that would be very helpful for people trying to edit PR CI to test stuff.

Kobzol: We should just remove these anchors altogether, I'll send a PR to `rust`.

Kobzol: I added an example how to add a job to PR CI. Hopefully this will soon be needed less once we get arbitrary try builds.

jieyouxu: Thanks for the improvements 💚

jieyouxu (approved):
Thanks for updating the docs 💚, I ran into this the other day and was slightly confused :3

PR #1978, opened 2024-05-14
use `&` to load completions for PowerShell

hycinth22:
in `src/building/suggested.md`
> You can use `source ./src/etc/completions/x.py.<extension>`
to load completions for your shell of choice,
or `source .\src\etc\completions\x.py.ps1` for PowerShell.

to my best knowledge, `source` command is available only in POSIX Shell, and PowerShell is not a POSIX Shell.
hence there is no `source` command in powershell, and you will get an error if you actually run `source .\src\etc\completions\x.py.ps1` for PowerShell.

My suggestion: use call operator(&) to load .ps1 script file for powershell.

There are some other options (e.g. dot sourcing)

### DIFF
--- a/src/building/suggested.md
+++ b/src/building/suggested.md
@@ -315,7 +315,7 @@ Zsh support will also be included once issues with [`clap_complete`](https://cra
 
 You can use `source ./src/etc/completions/x.py.<extension>`
 to load completions for your shell of choice,
-or `source .\src\etc\completions\x.py.ps1` for PowerShell.
+or `& .\src\etc\completions\x.py.ps1` for PowerShell.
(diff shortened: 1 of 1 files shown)


hycinth22:
Other choice we may have:

1. Execute the filename directly: `.\src\etc\completions\x.py.ps1`
This is equivalent to call operator, but its semantics are not clear enough

2. Invoke-Expression: `Invoke-Expression -Command ".\src\etc\completions\x.py.ps1"`
too long and no obvious benefit

3. Dot-Sourcing: `. .\src\etc\completions\x.py.ps1
all definitions in the script will be import to the current session. we dont need so much

hycinth22:
Is anyone going to review or discuss?

jieyouxu (approved):
Thanks for the fix!

PR #1984, opened 2024-05-21
Add Rust for Linux notification group entry

Kobzol:
Discussed [here](https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/Testing.20Rust.20for.20Linux.20in.20our.20CI).


camelid (approved):
Thanks!

PR #1985, opened 2024-05-27
Flesh out the "representing types" chapter

### LINKED ISSUE #1105: Update late bound parameters section from binder refactor
See https://rust-lang.zulipchat.com/#narrow/stream/144729-wg-traits/topic/binder.20refactor/near/233042116.

BoxyUwU:
Fixes #1105

I added a section about how we define generic parameters since explaining `ty::Generics` seemed like useful context for explaining `TyKind::Param` (and friends). I moved the section on early vs late bound parameters under this chapter since it seems like the natural place for this to live.

I moved the section on the `fold` module out of the chapter on representing types since it doesn't really have anything to do with how (or why) we structure `Ty`/`Const`/`Region` the way that we do.

As for the changes to the representing types chapter:
- Moved discussion of `TyKind::Adt`/`GenericArgs` into a single chapter as previously we were discussing these things split across three different files
- Split "bound vars and params" into a section solely about generic parameters and a separate section about `Binder`
- Discuss `EarlyBinder` and what we use it for and 
- Talk about placeholders ty/const/regions


src/what_is_ty_generics.md
```diff
@@ -0,0 +1,18 @@
...
+
+```rust,ignore
+trait Trait<T> {
+    fn foo<U>(&self);
+}
+```
+
+The `ty::Generics` used for `foo` would contain `[U]` and a parent of `Some(Trait)`. `Trait` would have a `ty::Generics` containing `[Self, T]` with a parent of `None`.
+
+The [`GenericParamDef`] struct is used to represent each individual generic parameter in a `ty::Generics` listing. The `GenericParamDef` struct contains information about the generic parameter, for example its name, defid, what kind of parameter it is (i.e. type, const, lifetime). It also contains a `u32` index representing what position the parameter is (starting from the outtermost parent).
```
jdonszelmann, line 13: what's this u32 actually useful for in practice? Also `outtermost` is with one t.

src/ty.md
```diff
@@ -10,7 +10,8 @@ The `ty` module defines how the Rust compiler represents types internally. It al
 When we talk about how rustc represents types,  we usually refer to a type called `Ty` . There are
 quite a few modules and types for `Ty` in the compiler ([Ty documentation][ty]).
 
-[ty]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/index.html
+[ty]: https://doc.rust-lang.org/nightly/nightly-rustc/ru
+]stc_middle/ty/index.html
```
jdonszelmann, line 14: broken link

src/ty_module/generic_arguments.md
```diff
@@ -0,0 +1,126 @@
+# ADTs and Generic Arguments
```
jdonszelmann, line 1: It might be nice to define/spell out ADT somewhere just for completeness

src/ty_module/generic_arguments.md
```diff
@@ -0,0 +1,126 @@
...
+// (1) represents the `MyStruct` part
+// (2) represents the `<u32>`, or "substitutions" / generic arguments
+```
+
+There are two parts:
+
+- The [`AdtDef`][adtdef] references the struct/enum/union but without the values for its type
+  parameters. In our example, this is the `MyStruct` part *without* the argument `u32`.
+  (Note that in the HIR, structs, enums and unions are represented differently, but in `ty::Ty`,
+  they are all represented using `TyKind::Adt`.)
```
jdonszelmann, line 27: oh, for the previous comment, you kinda do that here. Maybe say upfront: ADT = struct, enum, union

src/ty_module/generic_arguments.md
```diff
@@ -0,0 +1,126 @@
...
+```rust,ignore
+struct MyStruct<T>(T);
+type Foo = MyStruct<u32>
+```
+
+For the `MyStruct<U>` written in the `Foo` type alias, we would represent it in the following way:
+
+- There would be an `AdtDef` (and corresponding `DefId`) for `MyStruct`.
+- There would be a `GenericArgs` containing the list `[GenericArgKind::Type(Ty(u32))]`
+- This is one `TyKind::Adt` containing the `AdtDef` of `MyStruct` with the `GenericArgs` above.
```
jdonszelmann, line 126: I don't like the way this last sentence is phrased. Reread it twice, and it does make sense but I think it can be phrased better.

BoxyUwU: oh that should say `there`

src/ty_module/param_ty_const_regions.md
```diff
@@ -0,0 +1,99 @@
...
+Each `TyKind::Param` contains two things: the name of the parameter and an index.
+
+See the following concrete example of a usage of `TyKind::Param`:
+```rust,ignore
+struct Foo<T>(Vec<T>);
+```
+The `Vec<T>` type is represented as `TyKind::Adt(Vec, &[GenericArgKind::Type(Param("T", 0))])`.
+
+The name is somewhat self explanatory, it's the name of the type parameter. The index of the type parameter is an integer indicating
+its order in the list of generic parameters in scope (note: this includes parameters defined on items on outter scopes than the item the parameter is defined on). Consider the following examples:
```
jdonszelmann, line 34: outter

src/ty_module/binders.md
```diff
@@ -0,0 +1,52 @@
...
+
+We did not always explicitly track the set of bound vars introduced by each `Binder`, this caused a number of bugs (read: ICEs [#81193](https://github.com/rust-lang/rust/issues/81193), [#79949](https://github.com/rust-lang/rust/issues/79949), [#83017](https://github.com/rust-lang/rust/issues/83017)). By tracking these explicitly we can assert when constructing higher ranked where clauses/types that there are no escaping bound variables or variables from a different binder. See the following example of an invalid type inside of a binder:
+```
+Binder(
+    fn(&'^1_0 &'^1 T/#0),
+    &[BoundVariarbleKind::Region(...)],
+)
+```
+This would cause all kinds of issues as the region `'^1_0` refers to a binder at a higher level than the outtermost binder i.e. it is an escaping bound var. The `'^1` region (also writeable as `'^0_1`) is also ill formed as the binder it refers to does not introduce a second parameter. Modern day rustc will ICE when constructing this binder due to both of those regions, in the past we would have simply allowed this to work and then ran into issues in other parts of the codebase.
```
jdonszelmann, line 47: outter

src/ty_module/param_ty_const_regions.md
```diff
@@ -0,0 +1,99 @@
...
+  fn method<Z>() {
+    // inside here, X, Y and Z are all in scope
+    // X has index 0
+    // Y has index 1
+    // Z has index 2
+  }
+}
+```
+
+Concretely given the `ty::Generics` for the item the parameter is defined on, if the index is `10` then starting from the root `parent`, it will be the eleventh parameter to be introduced.
```
jdonszelmann, line 53: I think this example needs more context

src/ty_module/early_binder.md
```diff
@@ -0,0 +1,76 @@
...
+
+```rust,ignore
+fn foo<T, U>(a: T, _b: U) -> T { a }
+
+fn main() {
+    let c = foo::<i32, u128>(1, 2);
+}
+```
+
+When type checking `main` we cannot just naively look at the return type of `foo` and assign the type `T` to the variable `a`, after all the function `main` does not define any generic parameters, `T` is completely meaningless in this context. More generally whenever an item introduces (binds) generic parameters, when accessing types inside the item from outside, the generic parameters must be instantiated with values from the outer item.
```
jdonszelmann, line 13: do you mean `T` to the variable `c`?

BoxyUwU: yes :>

src/ty_module/instantiating_binders.md
```diff
@@ -0,0 +1,142 @@
...
+- The goal holds because we were successfully able to unify `T: Trait<'static>` with `T: Trait<'?0>`
+
+Instantiating binders with inference variables can be accomplished by using the [`instantiate_binder_with_fresh_vars`] method on [`InferCtxt`]. Binders should be instantiated with infer vars when we only care about one specific instantiation of the binder, if instead we wish to reason about all possible instantiations of the binder then placeholders should be used instead.
+
+## Instantiating with placeholders
+
+Placeholders are very similar to `Ty/ConstKind::Param`/`ReEarlyParam`, they represent some unknown type that is only equal to itself. `Ty`/`Const` and `Region` all have a `Placeholder` variant that is comprised of a `Universe` and a `BoundVar`. 
+
+The `Universe` tracks which binder the placeholder originated from, and the `BoundVar` tracks which parameter on said binder that this placeholder corresponds to. Equality of placeholders is determined solely by whether the universes are equal and the `BoundVar`s are equal. See the [chapter on Placeholders and Universes][ch_placeholders_universes] for more information.
```
jdonszelmann, line 39: I think some more links here might help

BoxyUwU:
cc @lcnr

lcnr (approved):
saying "generic parameters introduced via `for<'a>` syntax" feels wrong, you introduce bound variables, not generic parameters (except for late bound fn args, which are somewhere in the middle. I don't know of a better alternative to introducing these concepts, so I think it may actually be useful to slightly muddle these concepts.

I am quite happy with these changes and think they're an improvement. Feel free to merge after handling as many nits as you want. Don't want to block this from getting merged, given that it's already really useful and the nits don't change anything fundamental

src/ty_module/early_binder.md
```diff
@@ -0,0 +1,76 @@
...
+    let y = foo.x;
+}
+```
+
+In the compiler the `instantiate` call for this is done in [`FieldDef::ty`] ([src][field_def_ty_src]), at some point during type checking `bar` we will wind up calling `FieldDef::ty(x, &[u32, f32])` in order to obtain the type of `foo.x`.
+
+**Note on indices:** It is possible for the indices in `Param` to not match with what the `EarlyBinder` binds. For
+example, the index could be out of bounds or it could be the index of a lifetime when we were expecting a type.
+These sorts of errors would be caught earlier in the compiler when translating from a `rustc_hir::Ty` to a `ty::Ty`.
+If they occur later, that is a compiler bug.
```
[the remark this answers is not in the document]
BoxyUwU: I'm not entirely sure without going through every usage of `EarlyBinder` in the compiler. Nothing comes to mind though and if we have code that does so I'd be a bit :(. Ideally we could store a `DefId` in `EarlyBinder` to explicitly track the generics it binds like we do with `Binder` imo. The whole setup with opaque types' `generics_of` is sketchy enough that it might actually catch some bugs lol

src/ty_module/early_binder.md
```diff
@@ -0,0 +1,76 @@
...
+For example:
+```rust
+impl<T> Trait for Vec<T> {
+    fn foo(&self, b: Self) {}
+}
+```
+
+When constructing a `Ty` to represent the `b` parameter's type we need to get the type of `Self` on the impl that we are inside. This can be acquired by calling the [`type_of`] query with the `impl`'s `DefId`, however, this will return a `EarlyBinder<Ty>` as the impl block binds generic parameters that may have to be discharged if we are outside of the impl.
+
+The `EarlyBinder` type provides an [`instantiate_identity`] function for discharging the binder when you are "already inside of it". Conceptually this discharges the binder by instantiating it with placeholders in the root universe (we will talk about what this means in the next few chapters). In practice though it simply returns the inner value with no modification taking place.
```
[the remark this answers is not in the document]
BoxyUwU: Added a little sentence about this

PR #1986, opened 2024-05-29
Clarify a little bit in MIR chapter

camelid:
Mario Carneiro noticed some issues in the [reading club meeting](https://rust-lang.zulipchat.com/#narrow/stream/196385-t-compiler.2Fwg-rustc-dev-guide/topic/.5Breading.20club.5D.202024.2E05.2E02.20chapter.2039.20.28MIR.29).

cc @digama0

### DIFF
--- a/src/mir/index.md
+++ b/src/mir/index.md
@@ -61,7 +61,7 @@ show you the MIR for your program. Try putting this program into play
 (or [clicking on this link][sample-play]), and then clicking the "MIR"
 button on the top:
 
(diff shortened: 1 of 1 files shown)


camelid:
Not sure what to do about the disappearance of `// in scope 0` source-span comments. I mean, of course we should delete them if they're no longer shown, but do we know if it's intentional that they were removed?

camelid:
Ok, it looks like this is the change: rust-lang/rust@0a1fa411edbf6f9ad1205655d77cfda5a43aac06

Noratrieb (approved):

PR #1988, opened 2024-06-03
Fix dead links

hycinth22:
found some dead links when I read this book

### DIFF
--- a/src/memory.md
+++ b/src/memory.md
@@ -52,8 +52,8 @@ represented as a slice `&'tcx [tcx.types.i32, tcx.types.u32]`).
   defined and discussed in depth in the `AdtDef and DefId` section.
 - [`Predicate`] defines something the trait system has to prove (see `traits` module).
 
-[`GenericArgs`]: ./generic_arguments.html#GenericArgs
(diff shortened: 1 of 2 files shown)


src/memory.md
```diff
@@ -52,8 +52,8 @@ represented as a slice `&'tcx [tcx.types.i32, tcx.types.u32]`).
   defined and discussed in depth in the `AdtDef and DefId` section.
 - [`Predicate`] defines something the trait system has to prove (see `traits` module).
 
-[`GenericArgs`]: ./generic_arguments.html#GenericArgs
-[`TraitRef`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.TraitRef.html
+[`GenericArgs`]: ./ty_module/generic_arguments.html#the-genericargs-type
```
camelid, line 55: We usually use `.md` for self-links. I believe it's so these links are checked automatically in CI.

```suggestion
[`GenericArgs`]: ./ty_module/generic_arguments.md#the-genericargs-type
```

camelid (commented):
Looks good! Just one change.

src/memory.md
hycinth22: applied. thanks for the suggestion

PR #1989, opened 2024-06-07
Add docs for building Fuchsia locally and in CI

tmandry:
This also proposes "integration testing" as an umbrella term for Crater, Cargotest, and the integration CI builder.

I would like to expand the docs to include more details in the future, but this is a start. Concurrently with this PR I am working to land https://github.com/rust-lang/rust/pull/126105 and https://fxrev.dev/1061894, along with some simplifications to the Fuchsia build itself by Julia Ryan.


lqd:
> https://fuchsia-review.git.corp.google.com/c/fuchsia/+/1061894

fxrev.dev/1061894

tmandry:
Sorry.. trying to get that fixed and it makes me want to throw things! But I should remember to edit the links.

tmandry:
Added some additional details.

tmandry:
To do: Need to add some instructions about changing `rustc_version_string` after you change the compiler.

tmandry:
Mind giving this a skim @lqd? With your approval I think this can be merged.

src/tests/fuchsia.md
```diff
@@ -0,0 +1,162 @@
...
+initial build.
+
+## Fuchsia target support
+
+To learn more about Fuchsia target support, see the Fuchsia chapter in [the
+rustc book][platform-support].
+
+[regressions]: https://gist.github.com/tmandry/7103eba4bd6a6fb0c439b5a90ae355fa
+[build-toolchain]: https://fuchsia.dev/fuchsia-src/development/build/rust_toolchain
+[build-fuchsia.sh]: https://github.com/rust-lang/rust/pull/126105/files?file-filters%5B%5D=.sh&show-viewed-files=true
```
lqd, line 152: maybe link to the file within the repo?

```suggestion
[build-fuchsia.sh]: https://github.com/rust-lang/rust/blob/99f77a2eda555b50b518f74823ab636a20efb87f/src/ci/docker/host-x86_64/x86_64-gnu-integration/build-fuchsia.sh
```

src/tests/integration.md
```diff
@@ -0,0 +1,49 @@
...
+
+> Example: `./x test src/tools/cargotest`
+
+### Integration builders
+
+Integration jobs build large open-source Rust projects that are used as
+regression tests in CI.
+
+Fuchsia is currently the only project used in this way. See the [Fuchsia
+chapter](fuchsia.md) for more details.
```
lqd, line 28: We also have rust-for-linux

src/tests/fuchsia.md
```diff
@@ -0,0 +1,162 @@
...
+[Fuchsia](https://fuchsia.dev) is an open-source operating system with about 2
+million lines of Rust code.[^loc] It has caught a large number of [regressions]
+in the past and was subsequently included in CI.
+
+## Building Fuchsia in CI
+
+Fuchsia builds as part of the suite of bors tests that run before a pull request
+is merged.
+
+If you are worried that a pull request might break the Fuchsia builder and want to test it out before submitting it to the bors queue, simply add this line to your PR description:
```
lqd, line 12: (nit: maybe wrap this one like the other lines?)

src/tests/fuchsia.md
```diff
@@ -0,0 +1,162 @@
...
+}
+
+rustc_binary("example") {
+    crate_root = "src/bin.rs"
+    # ...existing keys here...
+    configs += [ ":everybody_loops" ]
+}
+```
+
+This will add the flag `-Zeverybody-loops` to rustc for the given target. You can
```
lqd, line 110: could we explicitly mention the target name here just to make it super clear? (I don't expect people to read a lot of GN docs, so it will be nice that we have an example here)

src/tests/fuchsia.md
```diff
@@ -0,0 +1,162 @@
...
+Going down one layer, `fx build` invokes `ninja`, which in turn eventually
+invokes `rustc`. All build actions are run inside the out directory, which is
+usually `out/default` inside the Fuchsia checkout.
+
+You can get ninja to print the actual command it invokes by forcing that command
+to fail, e.g. by adding a syntax error to one of the source files of the target.
+Once you have the command, you can run it from inside the output directory.
+
+After changing the toolchain itself, the build setting `rustc_version_string` in
+`out/default/args.gn` needs to be changed so that `fx build` or `ninja` will
```
lqd, line 130: Do you do this manually? Would it be worthwhile to show here how it looks like in practice, or e.g. could we show a command to do it (à la `fx set rustc_version_string` with `date +%s`)?

src/tests/fuchsia.md
```diff
@@ -0,0 +1,162 @@
...
+rebuild all the Rust targets. The contents of the string do not matter, as long
+as it changes from one build to the next. [build_fuchsia_from_rust_ci.sh] does
+this automatically by hashing the toolchain directory.
+
+The Fuchsia website has more detailed documentation of the [build system].
+
+#### Other tips and tricks
+
+You can comment out the `fx set` command after the initial run so it won't rerun
```
lqd, line 139: Is this is referring to the command to update `rustc_version_string`?

lqd (commented):
I left a few comments. 

I'm not sure how widespread the "integration" nomenclature is, within the project, but it seems fine and people can nitpick the definition.

While I have you (I'm sure I asked this before, sorry), and related to the previous point, do you mind if we rename the job to something that contains the word "fuchsia"? 

It wasn't clear to me whether integration also referred to your integration repo or to integration tests. (But I guess naming the jobs and builders is more of a t-infra decision)

tmandry:
> do you mind if we rename the job to something that contains the word "fuchsia"?

I don't mind. I think initially I shied away because it was possible we would build other projects in the same job, but we can always change it later. It makes sense to have the name reflect what is actually being built.

src/tests/fuchsia.md
tmandry: Yes, thanks – I wrote this before that PR was merged.

src/tests/integration.md
tmandry: Yes, that has changed since I wrote this :D

src/tests/fuchsia.md
tmandry: I normally do it manually in an editor. It's easier than trying to find the right `fx set` command from my history and modifying that.

src/tests/fuchsia.md
tmandry: No but clarified and added that too

src/tests/fuchsia.md
tmandry: Clarified.

lqd (approved):
LGTM.

We'll still keep improving e.g. the rust-for-linux contents, and update the fuchsia try job name when the renaming lands.

PR #1990, opened 2024-06-07
Mention `COMPILETEST_REQUIRE_ALL_LLVM_COMPONENTS` in `needs-llvm-components` documentation

erikdesjardins:
We should mention the env var that [controls](https://github.com/rust-lang/rust/blob/76e7a0849c07d73e4d9afde8036ee8c450127cc8/src/tools/compiletest/src/header.rs#L1593) this check for better discoverability, like we already do for `needs-matching-clang` below.

cc https://github.com/rust-lang/rust/pull/125949#issuecomment-2146631862


jieyouxu (approved):
Thanks!

PR #1996, opened 2024-06-15
tell about `STAGE0_MISSING_TARGETS` for new targets

onur-ozkan:
We just realized that the bootstrap target sanity check isn't documented anywhere, and it would be quite exhausting to learn this manually through a series of failing CI builds.

cc @Kobzol

### DIFF
--- a/src/building/new-target.md
+++ b/src/building/new-target.md
@@ -94,6 +94,17 @@ adding `NEW_TARGET_OS` as `target_os`:
 + (Some(Mode::Std), "target_os", Some(&["watchos", "NEW_TARGET_OS"])),
 ```
 
+To use this target in bootstrap, we need to explicitly add the target triple to the `STAGE0_MISSING_TARGETS`
+list in `src/bootstrap/src/core/sanity.rs`. This is necessary because the default compiler bootstrap uses does
+not recognize the new target we just added. Therefore, it should be added to `STAGE0_MISSING_TARGETS` so that the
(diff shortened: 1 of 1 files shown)


src/building/new-target.md
```diff
@@ -94,6 +94,17 @@ adding `NEW_TARGET_OS` as `target_os`:
 + (Some(Mode::Std), "target_os", Some(&["watchos", "NEW_TARGET_OS"])),
 ```
 
+To use this target in bootstrap, we need to explicitly add the target triple to the `STAGE0_MISSING_TARGETS`
+list in `bootstrap/src/core/sanity.rs`. This is necessary because the default compiler bootstrap uses does not
```
Kobzol, line 98:
```suggestion
list in `src/bootstrap/src/core/sanity.rs`. This is necessary because the default compiler bootstrap uses does not
```
Just to make it clearer.

Kobzol (commented):
The problem was more that without people knowing about this array, the CI wouldn't even fail when adding a new target :laughing:

Left one comment, otherwise LGTM.

src/building/new-target.md
onur-ozkan: Done https://github.com/rust-lang/rustc-dev-guide/compare/31d839376aaef45ff8a13722a4b7d128b640323a..a11ed34e997960b4a0ecdeba625ac1a8c5a36b08

Kobzol (approved):
Thanks!

PR #1997, opened 2024-06-18
Mention `RUSTC_ICE=0` to suppress ICE file

jieyouxu:


jieyouxu:
https://github.com/rust-lang/rust/pull/126615 should mostly make this not needed, but *someone* might still find it useful.

BoxyUwU (approved):

PR #1998, opened 2024-06-18
Fix note about compiletest header `rustfix-only-machine-applicable`

fmease:
`//@ rustfix-only-machine-applicable` actually only works in conjunction with `//@ run-rustfix`.


jieyouxu (approved):
Thanks!

PR #2000, opened 2024-06-19
Document hard-resetting submodules

jieyouxu:
Sometimes `src/llvm-project` just *breaks* and you can't run `git submodule update` successfully. I do not know why. This PR suggests to run `git submodule foreach git reset --hard` if you run into `src/llvm-project` submodule showing up in `git status` as `(new commits, modified content)` as a workaround.


Nilstrieb (approved):
submeowdule

PR #2002, opened 2024-06-24
Document how to run `run-make` tests on Windows

jieyouxu:
Apparently it is very convoluted to run `run-make` on Windows...

### DIFF
--- a/src/tests/running.md
+++ b/src/tests/running.md
@@ -213,6 +213,22 @@ This is much faster, but doesn't always work. For example, some tests
 include directives that specify specific compiler flags, or which rely
 on other crates, and they may not run the same without those options.
 
+## Running `run-make` tests
+
+### Windows
+
+Running the `run-make` test suite on Windows is a bit more involved. There are numerous
+prerequisites and environmental requirements:
+
+- Install msys2: <https://www.msys2.org/>
+- Specify `MSYS2_PATH_TYPE=inherit` in `msys2.ini` in the msys2 installation directory, run the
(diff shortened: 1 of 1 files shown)


src/tests/running.md
```diff
@@ -213,6 +213,23 @@ This is much faster, but doesn't always work. For example, some tests
...
 on other crates, and they may not run the same without those options.
 
+## Running `run-make` tests
+
+### Windows
+
+Running the `run-make` test suite on Windows is a bit more involved. There are numerous
+prerequisites and environmental requirements:
+
+- Install python on Windows: <https://www.python.org/downloads/>
```
ChrisDenton, line 223: Hm... do we need to specify python? That's already a prerequisite of bootstrap, no?

jieyouxu: Probably yeah. I'll remove it.

jieyouxu: Removed.

ChrisDenton:
Should we also make a note about types of run-make tests? E.g. you can run individual rust recipes without needing make. Or would that just complicate things more?

jieyouxu:
> Should we also make a note about types of run-make tests? E.g. you can run individual rust recipes without needing make. Or would that just complicate things more?

I would have these instructions for now while there are mixed Makefile/rmake.rs tests. Once we yeet all of the Makefile tests, we don't need this entire section anymore (modulo some external deps).

ChrisDenton (approved):
Fair enough!

PR #2005, opened 2024-06-30
Document `bootstrap` integration with `rustc-perf`

Kobzol:
This PR cleans up the section about profiling the compiler. It does the following things:
- Creates a new section that describes how to use `rustc-perf` for profiling.
- Documents the new `bootstrap` integration with `rustc-perf`, introduced in https://github.com/rust-lang/rust/pull/126318, https://github.com/rust-lang/rust/pull/127002 and https://github.com/rust-lang/rust/pull/128868.
- Removes example of `rustc-perf` usage from the `perf` section, since there is now a dedicated `rustc-perf` section. The new section mostly links to existing `rustc-perf` integration, I don't think that duplicating the docs is ideal (case in point: the existing `rustc-perf` example was using CLI options that are no longer used by `rustc-perf`).

### DIFF
--- a/src/profiling/with_rustc_perf.md
+++ b/src/profiling/with_rustc_perf.md
@@ -5,10 +5,7 @@ the Rust compiler. You can find instructions on how to use the suite in its [man
 
 However, using the suite manually can be a bit cumbersome. To make this easier for `rustc` contributors,
 the compiler build system (`bootstrap`) also provides built-in integration with the benchmarking suite,
-which will download and build the suite for you, build a local compiler toolchain and let you profile it using a
-simplified command-line interface.
-
-> This integration currently only supports profiling, benchmarking is not integrated yet.
+which will download and build the suite for you, build a local compiler toolchain and let you profile it using a simplified command-line interface.
 
 You can use the `./x perf -- <command> [options]` command to use this integration.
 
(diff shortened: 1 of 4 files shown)


src/profiling/with_rustc_perf.md
```diff
@@ -0,0 +1,35 @@
...
+
+The [Rust benchmark suite][rustc-perf] provides a comprehensive way of profiling and benchmarking
+the Rust compiler. You can find instructions on how to use the suite in its [manual][rustc-perf-readme].
+
+However, using the suite manually can be a bit cumbersome. To make this easier for `rustc` contributors,
+the compiler build system (`bootstrap`) also provides built-in integration with the benchmarking suite,
+which will download and build the suite for you, build a local compiler toolchain and let you profile or
+benchmark it using a simplified command-line interface.
+
+You can use the `./x perf -- <command> [options]` command to use this integration.
```
camelid, line 11: Can you give a complete example of a command that someone can run that roughly matches rustc-perf does? I.e. if I want to mimic what `rust-timer` does, what should I run?

camelid: The command also doesn't seem to be working for me. I tried `x perf -- eprintln --profiles Check`, but after it ran, all the files in `build/tmp/rustc-perf/results` were completely empty.

Kobzol: The compiler doesn't normally print anything to stderr. Usually you use eprintln by adding manual eprintln calls, and then postprocess the results (see rustc-perf documentation for more info).

Currently, this can't really do what rust-timer does, as benchmarking is not supported yet (only profiling).

camelid: Ah, perhaps you could add a clarification like this to the docs?

Kobzol: Pushed a new commit that tries to make this a bit more explicit.

spastorino:
I think this is ready to merge, @camelid leaving it to you to see if you're fine with the latest changes

jieyouxu (approved):
This LGTM, we can always fine-tune/improve the docs in the future.

PR #2006, opened 2024-07-01
Update new target check-cfg instructions

Urgau:
With https://github.com/rust-lang/rust/pull/127026 we no longer want to use the `EXTRA_CHECK_CFGS` list in bootstrap. New target specific cfgs should be set directly in the appropriate `[lints.rust.unexpected_cfgs.check-cfg]` lint table in one of the `library/{std,alloc,core}/Cargo.toml`.

### DIFF
--- a/src/building/new-target.md
+++ b/src/building/new-target.md
@@ -86,12 +86,18 @@ unexpected because stage0 doesn't know about the new target specification and
 we pass `--check-cfg` in order to tell it to check.
 
 To fix the errors you will need to manually add the unexpected value to the
-`EXTRA_CHECK_CFGS` list in `src/bootstrap/src/lib.rs`. Here is an example for
-adding `NEW_TARGET_OS` as `target_os`:
+different `Cargo.toml` in `library/{std,alloc,core}/Cargo.toml`. Here is an
+example for adding `NEW_TARGET_ARCH` as `target_arch`:
+
+*`library/std/Cargo.toml`*:
 ```diff
-- (Some(Mode::Std), "target_os", Some(&["watchos"])),
-+ // #[cfg(bootstrap)] NEW_TARGET_OS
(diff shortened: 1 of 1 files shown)


src/building/new-target.md
```diff
@@ -86,12 +86,18 @@ unexpected because stage0 doesn't know about the new target specification and
...
-`EXTRA_CHECK_CFGS` list in `src/bootstrap/src/lib.rs`. Here is an example for
-adding `NEW_TARGET_OS` as `target_os`:
+different `Cargo.toml` in `library/{std,alloc,core}/Cargo.toml`. Here is an
+example for adding `NEW_TARGET_ARCH` as `target_arch`:
+
+*`library/std/Cargo.toml`*:
 ```diff
-- (Some(Mode::Std), "target_os", Some(&["watchos"])),
-+ // #[cfg(bootstrap)] NEW_TARGET_OS
-+ (Some(Mode::Std), "target_os", Some(&["watchos", "NEW_TARGET_OS"])),
```
tshepang, line 94: should this be here

Urgau: I'm sorry, I don't understand the question.

The goal of this change is to no longer advertise the `EXTRA_CHECK_CFGS` list in bootstrap in favour of the individual `Cargo.toml` files.
Regarding the diff, I took the current std `Cargo.toml` as new base.

tshepang: I meant the line starting with `+ Some(`... I don't understand why it's there. Maybe it was meant to be a comment.

Urgau: Those three lines are from the previous diff. They are being removed with this PR.

*A diff in a diff can be difficult to read.*

tshepang (approved):

src/building/new-target.md
tshepang: hehe, I see... I need to rest

PR #2011, opened 2024-07-08
Fix typo: lists -> lints

camelid:
It's a bit of a tongue-twister it seems.

### DIFF
--- a/src/bug-fix-procedure.md
+++ b/src/bug-fix-procedure.md
@@ -276,7 +276,7 @@ FutureIncompatibleInfo {
 
 Remove this too.
 
-#### Add the lint to the list of removed lists.
+#### Add the lint to the list of removed lints.
 
 In `compiler/rustc_lint/src/lib.rs` there is a list of "renamed and removed lints".
 You can add this lint to the list:


src/bug-fix-procedure.md
```diff
@@ -276,7 +276,8 @@ FutureIncompatibleInfo {
 
 Remove this too.
 
-#### Add the lint to the list of removed lists.
+<a name="add-the-lint-to-the-list-of-removed-lists">
```
camelid, line 279: I added an anchor so that the rename doesn't break existing links.

fmease (approved):

src/bug-fix-procedure.md
fmease: Thanks, good idea! However, I'm not sure if it's actually needed, we break URL a lot for rustc-dev-guide

fmease: see ParamEnv PR and my HIR ty lowering PR which moved lowering to ast-lowering etc

camelid: > Thanks, good idea! However, I'm not sure if it's actually needed, we break URL a lot of times for rustc-dev-guide

Fair enough, I'll remove it. I was just thinking of a particular place in rustdoc that links here that I didn't want to break. But it's obvious enough what it's referring to anyway. :)

PR #2016, opened 2024-07-19
Update adding.md

10takla:
Add `@` to the test header `edition:2018` in the [example](https://rustc-dev-guide.rust-lang.org/tests/adding.html#step-1-add-a-test-file), since the [header](https://rustc-dev-guide.rust-lang.org/tests/headers.html) syntax requires it


jieyouxu (approved):
Thanks!

Issue #2018, opened 2024-07-23
Why is `book.toml` included in `.gitignore` ?


marxin:
I noticed that while I was trying to find a file `book.toml` in my VS Code instance, but it seems the editor ignores files mentioned in `.gitignores`. However, the file is tracked in git history and I'm curious why it's ignored?
CC: @mark-i-m

PR #2023, opened 2024-07-24
Remove properly tracked config file from .gitignore & add support for skipping of link-checking

marxin:
The file is part of the git history and is a configuration file.

Fixes: #2018


camelid:
I think the reason this is ignored is that we often want to change `book.toml` locally to turn off link-checking. I'm not sure it makes sense to remove it from the gitignore unless there's another way to disable link-checking locally.

marxin:
Sure, that makes sense and I've just added the support. What do you think about it now?

marxin:
@camelid May I please ping this?

marxin:
@camelid: A gentle ping.

ci/linkcheck.sh
```diff
@@ -3,10 +3,17 @@
...
 
+LINKCHECK_BINARY=mdbook-linkcheck2
+
 set_github_token() {
   jq '.config.output.linkcheck."http-headers"."github\\.com" = ["Authorization: Bearer $GITHUB_TOKEN"]'
 }
 
+if [ ! -z "$SKIP_LINKCHECK" ] ; then
+  echo "Skipping link check."
+  exec $LINKCHECK_BINARY -f ""
```
camelid, line 14: Shouldn't this be doing `exit 0` instead? I'm not sure why it's still running the linkcheck binary.

camelid (changes requested):
Thanks! This is a much better approach than our old approach of just commenting out the config. Just left a comment about something that seems wrong.

ci/linkcheck.sh
marxin: Yep, adjusted.

camelid (approved):

PR #2030, opened 2024-07-28
Save linkcheck cache always

marxin:
This PR always stores the linkcheck cache when being run as part of the scheduled CI job. Note the normal PRs ignore the cache and check only the modified files. I also extended the cache expiration time so that the Nighly job can face the congressional limit from Github.com (`"Server returned 429 Too Many Requests"`).

Hopefully, it's a viable solution.
@camelid

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,12 +35,13 @@ jobs:
             ~/.cargo/bin
           key: ${{ runner.os }}-${{ env.MDBOOK_VERSION }}--${{ env.MDBOOK_LINKCHECK2_VERSION }}--${{ env.MDBOOK_TOC_VERSION }}--${{ env.MDBOOK_MERMAID_VERSION }}
 
-      - name: Cache linkcheck
-        uses: actions/cache@v4
+      - name: Restore cached Linkcheck
+        if: github.event_name == 'schedule'
+        id: cache-linkcheck-restore
+        uses: actions/cache/restore@v4
         with:
-          path: |
-            ~/book/linkcheck
-          key: ${{ runner.os }}-${{ hashFiles('./book/linkcheck') }}
+          path: book/linkcheck/cache.json
+          key: linkcheck
 
       - name: Install latest nightly Rust toolchain
         if: steps.mdbook-cache.outputs.cache-hit != 'true'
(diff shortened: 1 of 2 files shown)


Kobzol:

```
git fetch origin --all
git checkout <pr-branch>
git push --force-with-lease
```

.github/workflows/ci.yml
```diff
@@ -35,12 +35,13 @@ jobs:
             ~/.cargo/bin
           key: ${{ runner.os }}-${{ env.MDBOOK_VERSION }}--${{ env.MDBOOK_LINKCHECK2_VERSION }}--${{ env.MDBOOK_TOC_VERSION }}--${{ env.MDBOOK_MERMAID_VERSION }}
 
-      - name: Cache linkcheck
-        uses: actions/cache@v4
```
camelid, line 39: Any idea what this code was doing before? AFAICT it was doing nothing, which is pretty strange. It seems like it cached an empty linkcheck directory, which is useless, and also never restored it...

camelid (commented):
I'm a little confused about what this change is accomplishing. Wouldn't making the cron job run every other day accomplish exactly the same thing?

.github/workflows/ci.yml
Kobzol: The way the [cache](https://github.com/actions/cache) action works is that you use it at the beginning of the workflow, and then you generate the data that it should cache. The action then does this:
1) When the action is executed: if there is some cache entry, restore it from the cache. If there is no cache entry, do nothing.
2) At the end of the workflow: if the given path contains something, save it to a cache entry.

It's a little counter-intuitive in that you use it at the beginning, and it then uses an implicit step at the end of the workflow that actually writes to the cache.

camelid: Thanks! I was wondering if it was something like that.

camelid:
> I'm a little confused about what this change is accomplishing. Wouldn't making the cron job run every other day accomplish exactly the same thing?

Never mind, I just realized that the difference is that the cache will allow us to check new links without rechecking old links. Not sure why I didn't think of that :sweat_smile:

.github/workflows/ci.yml
```diff
@@ -35,12 +35,13 @@ jobs:
...
+      - name: Restore cached Linkcheck
+        if: github.event_name == 'schedule'
+        id: cache-linkcheck-restore
+        uses: actions/cache/restore@v4
         with:
-          path: |
-            ~/book/linkcheck
-          key: ${{ runner.os }}-${{ hashFiles('./book/linkcheck') }}
+          path: book/linkcheck/cache.json
+          key: linkcheck
```
camelid, line 44: Does it make sense to include the mdbook-linkcheck2 version in the key? That way if we do an update of mdbook-linkcheck2 and the cache format changes, it doesn't cause the next cron job to spuriously fail.

marxin: Sure, good idea!

marxin:
> Not sure why I didn't think of that 😅

Yeah, the main goal of the cache expiration is to at least partially reuse some of the GH links from the cache as one can't check all of them unless the `"Server returned 429 Too Many Requests"` message is hit.

camelid (approved):
Thanks for doing this!

marxin:
Thanks for the fix and the CI fix you landed just after my change!

PR #2037, opened 2024-08-05
Correct rust code block in *Dataflow Analysis*

xFrednet:
While reading the guide, I noticed that the code example didn't compile. This change, simply corrects the code and adds syntax highlighting by marking the code block as rust code.

### DIFF
--- a/src/mir/dataflow.md
+++ b/src/mir/dataflow.md
@@ -127,16 +127,18 @@ value will be `true`, since our analysis is done as soon as we determine that
 `transmute` has been called. Our join operator will just be the boolean OR (`||`)
 operator. We use OR and not AND because of this case:
 
-```
-let x = if some_cond {
-    std::mem::transmute<i32, u32>(0_i32); // transmute was called!
-} else {
-    1_u32; // transmute was not called
-};
+```rust
+# unsafe fn example(some_cond: bool) {
+    let x = if some_cond {
+        std::mem::transmute::<i32, u32>(0_i32) // transmute was called!
+    } else {
+        1_u32 // transmute was not called
+    };
 
-// Has transmute been called by this point? We conservatively approximate that
-// as yes, and that is why we use the OR operator.
-println!("x: {}", x);
+    // Has transmute been called by this point? We conservatively approximate that
+    // as yes, and that is why we use the OR operator.
+    println!("x: {}", x);
+# }
 ```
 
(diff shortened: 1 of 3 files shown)


src/mir/dataflow.md
```diff
@@ -127,16 +127,18 @@ value will be `true`, since our analysis is done as soon as we determine that
...
+    let x = if some_cond {
+        std::mem::transmute::<i32, u32>(0_i32) // transmute was called!
+    } else {
+        1_u32 // transmute was not called
+    };
+
+    // Has transmute been called by this point? We conservatively approximate that
+    // as yes, and that is why we use the OR operator.
+    println!("x: {}", x);
+# }
```
camelid, line 141: Can you unindent the body of the function since we're hiding the function header anyway?

camelid:
There's also a link-checking error (not your fault; it's just because the file hasn't been changed and thus checked in a while). Do you mind fixing it? Thanks!

```
error: It looks like "mir/debugging.md" wasn't included in SUMMARY.md
    ┌─ mir/dataflow.md:217:1
    │
217 │ [Debugging MIR]. Start with `-Z dump-mir=F -Z dump-mir-dataflow`, where `F` is
    │ ^^^^^^^^^^^^^^^ It looks like "mir/debugging.md" wasn't included in SUMMARY.md

Error: One or more incorrect links
```

xFrednet:
> Do you mind fixing it? Thanks!

I've tried fixing it, but don't really understand the problem. The error mentions a file which is included in the summary. I've now update the summary to use the same title and the lint to point to the `.md` file, but the error still persists.

This could be https://github.com/Michael-F-Bryan/mdbook-linkcheck/issues/77 . Do you have an idea how to fix the error?

camelid (approved):
Thanks! It does look like a bug in linkcheck like you said, and since this change certainly doesn't cause the issue, we'll ignore it. It'd be good to get to the bottom of this though.

camelid:
Argh we can't merge until CI is fixed sigh.

camelid:
Will merge once someone else from @rust-lang/wg-rustc-dev-guide approves my CI change.

camelid:
Received the [go-ahead](https://rust-lang.zulipchat.com/#narrow/stream/196385-t-compiler.2Fwg-rustc-dev-guide/topic/linkcheck.20bugs/near/457137498) on Zulip.

PR #2039, opened 2024-08-08
Update LLVM docs

nikic:
The primary change here is to split the backport process into two. While upstream backports are accepted, they should be filed there first. Switch to direct backports after the upstream release is closed.

cc @rust-lang/wg-llvm

### DIFF
--- a/src/backend/updating-llvm.md
+++ b/src/backend/updating-llvm.md
@@ -3,32 +3,70 @@
 <!-- toc -->
 
 <!-- date-check: Jul 2023 -->
-There is no formal policy about when to update LLVM or what it can be updated to,
-but a few guidelines are applied:
+Rust supports building against multiple LLVM versions:
 
-* We try to always support the latest released version
-* We try to support the last few versions
-  (and the number changes over time)
-* We allow moving to arbitrary commits during development
-* We strongly prefer to upstream all patches to LLVM before including them in rustc
+* Tip-of-tree for the current LLVM development branch is usually supported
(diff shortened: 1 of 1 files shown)


src/backend/updating-llvm.md
```diff
@@ -3,49 +3,90 @@
...
-* We allow moving to arbitrary commits during development
-* We strongly prefer to upstream all patches to LLVM before including them in rustc
+* Tip-of-tree for the current LLVM development branch is usually supported
+  within a few days. PRs for such fixes are tagged with `llvm-main`.
+* The latest released major version is always supported.
+* The one or two preceding major versions are usually supported.
 
-## Why update LLVM?
+By default, Rust uses its own fork in the [rust-lang/llvm-project repository].
+This fork is based on a `release/NN.x` branch of the upstream project, rather
```
DianQK, line 14: Could you change to `release/$N.x` and move "where `$N` is the version of LLVM that's being released" to here?

DianQK (approved):
Thanks!
BTW, I've been thinking about using GitHub Actions or @bors to automate the process, but I haven't had the time to do it yet. :3

src/backend/updating-llvm.md
```diff
@@ -3,49 +3,90 @@
 <!-- toc -->
 
 <!-- date-check: Jul 2023 -->
```
spastorino, line 5: I guess while we are at this we should update this date to today's date.

cuviper (approved):

PR #2044, opened 2024-08-21
Replace direct http links to rustc-dev-guide.rust-lang.org

pmundkur:
Use relative links to the corresponding md source files; this helps local browsing of generated docs.

### DIFF
--- a/src/borrow_check/opaque-types-region-inference-restrictions.md
+++ b/src/borrow_check/opaque-types-region-inference-restrictions.md
@@ -158,7 +158,7 @@ See [#113971] for how we used to conflate the difference.
 
 [#113971]: https://github.com/rust-lang/rust/issues/113971
 [SCC]: https://en.wikipedia.org/wiki/Strongly_connected_component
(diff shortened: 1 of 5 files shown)


src/borrow_check/opaque-types-region-inference-restrictions.md
```diff
@@ -158,7 +158,7 @@ See [#113971] for how we used to conflate the difference.
 
 [#113971]: https://github.com/rust-lang/rust/issues/113971
 [SCC]: https://en.wikipedia.org/wiki/Strongly_connected_component
-[member constraints]: https://rustc-dev-guide.rust-lang.org/borrow_check/region_inference/member_constraints.html
+[member constraints]: region_inference/member_constraints.md
```
camelid, line 161: We usually prefix the local paths with `./` -- I believe for link-checking reasons, but I'm not fully sure. Do you mind doing that here and in the rest of the PR?

camelid: To be clear, paths that already have `../` can be left alone.

camelid (approved):
Thanks!

PR #2045, opened 2024-08-22
clarify the role of rustc_const_unstable

RalfJung:
This came up in https://github.com/rust-lang/rust/pull/129377.

@oli-obk I wonder if it would make sense to say that if a function is `#[unstable]` but a `const fn`, that implicitly makes it `#[rustc_const_unstable]` under the same feature gate?

### DIFF
--- a/src/stability.md
+++ b/src/stability.md
@@ -49,11 +49,11 @@ marks an item as stabilized. Note that stable functions may use unstable things 
 
 ## rustc_const_unstable
 
-The `#[rustc_const_unstable(feature = "foo", issue = "1234", reason = "lorem ipsum")]`
-has the same interface as the `unstable` attribute. It is used to mark
-`const fn` as having their constness be unstable. This allows you to make a
-function stable without stabilizing its constness or even just marking an existing
-stable function as `const fn` without instantly stabilizing the `const fn`ness.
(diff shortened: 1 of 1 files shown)


oli-obk:
I think the original idea was to force you to add a const_(un)stable attribute. We could also force that to have the same feature gate.

I think there was a reason we couldn't share the regular stability attribute on its own, but I don't remember why

RalfJung:
> I think the original idea was to force you to add a const_(un)stable attribute.

Yeah, we just didn't properly document that it seems, nor actually enforce it in rustc. This fixes the "document" part of that; I didn't find a good place to enforce this in rustc.

For actually changing the rules, let's discuss that in https://github.com/rust-lang/rust/issues/129815.

oli-obk:
Fairly certain we enforce it for the `rustc_stable` attribute

oli-obk (approved):

RalfJung:
> passes_missing_const_stab_err

Yes, indeed we do. But stdarch has tons of unstable `const fn` without any const stability. But I am not sure if it's worth fixing that, I think we want to change the way const stability checking works. But that's a topic for https://github.com/rust-lang/rust/issues/129815. :)

PR #2046, opened 2024-08-22
Explain the internal `#[rustc_*]` TEST attributes used for debugging and inside tests

fmease:
First of all, I'm not entirely sure if it's worth documenting this at all.

I guess it could be useful for newcomers who haven't heard of them yet.
Well, I guess the non-exhaustive table I added can be of use for seasoned rustc devs, too.
In fact, I added the table for selfish reasons as I often found myself grepping through `compiler/` and `tests/ui/` in search of specific attributes whose names I keep forgetting (their naming scheme is slightly inconsistent).

These TEST attributes serve a dual purpose. Namely as a means to "printf+"-debug the compiler as well as in UI tests to perform "white-box testing" (sorta). Hence *two* sections on this topic. I chose *compiler debugging* as the "main" article here (with the testing one referring back to the other one) but I'm open to switch it around. I know that originally the TEST attrs were introduced for use in UI tests. However, I think the "compiler debugging" aspect is more important.

### DIFF
--- a/src/compiler-debugging.md
+++ b/src/compiler-debugging.md
@@ -267,10 +267,34 @@ for debugging by dumping extra compiler-internal information. These are prefixed
 are gated behind the internal feature `rustc_attrs` (enabled via e.g. `#![feature(rustc_attrs)]`).
 
 For a complete and up to date list, see [`builtin_attrs`]. More specifically, the ones marked `TEST`.
+Here are some notable ones:
+
+| Attribute | Description |
+|----------------|-------------|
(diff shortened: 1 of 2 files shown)


jieyouxu (approved):
Thank you!

> I'm not entirely sure if it's worth documenting this at all.

I find these information *very* valuable, otherwise you'd either have to manage to stumble across this information on their own, or otherwise get passed down via review folklore (and has to be repeated every time someone is stuck on this). It's very hard to discover, realizing they even *exist* is a significant challenge in the first place. So thank you for taking the time and effort to document these!

PR #2048, opened 2024-08-23
Add docs for JS tests

notriddle:


src/rustdoc-internals/search.md
```diff
@@ -354,3 +354,102 @@ The unification filter ensures that:
...
+
+A `rustdoc-js` test has a `.rs` and `.js` file, with the same name.
+The `.rs` file specifies the hypothetical library crate to run
+the searches on (make sure you mark anything you need to find as `pub`).
+The `.js` file specifies the actual searches.
+The `rustdoc-js-std` tests are the same, but don't require an `.rs`
+file, since they use the standard library.
+
+The `.js` file is like a module (except the loader takes care of
+`exports` for you). It expects you to set these variables in the
```
lolbinarycat, line 372: the use of "expects" makes it sound like it is mandatory to set all of these (after all, `Option::expect` is used to throw an error if the receiver is `None`)

something like "It expects you to set *at least one of* these variables" or "the following variables have special meaning" would be more clear imo.

src/rustdoc-internals/search.md
```diff
@@ -354,3 +354,102 @@ The unification filter ensures that:
...
+* `// exact-check`: If search results appear that aren't part of the test case,
+  then fail. By default, the test case will tolerate the engine returning more
+  results than specified.
+* `// ignore-order`: By default, the entries in the test case must have a
+  matching order in the results. Setting this option allows you to test
+  filtering without also testing ranking, if you expect a test to be sensitive
+  to minor scoring changes.
+* `// should-fail`: Used to write negative tests.
+
+Standard library tests probably shouldn't specify any of these (we want the
```
lolbinarycat, line 393: except `ignore-order`, which is used a decent amount in `rustdoc-js-std`.

src/rustdoc-internals/search.md
```diff
@@ -354,3 +354,102 @@ The unification filter ensures that:
...
+libs team to be able to add new stuff without causing our tests to fail), but
+standalone tests will often want `// exact-check`.
+
+`FILTER_CRATE` can be left out (equivalent to searching "all crates"), but you
+have to specify `EXPECTED` or `PARSED`.
```
lolbinarycat, line 398: this should probably be moved to immediately after the table.

lolbinarycat (commented):
there's a few tiny wording and formatting things i think could be improved, but this is very helpful!

jieyouxu (approved):
LGTM, feel free to merge this unless you want someone from rustdoc to review.

jieyouxu:
(I think we don't have r? on this repo lol)

src/rustdoc.md
```diff
@@ -83,7 +83,8 @@ does is call the `main()` that's in this crate's `lib.rs`, though.)
 * All paths in this section are relative to `tests` in the rust-lang/rust repository.
 * Tests on search index generation are located in `rustdoc-js`, as a
   series of JavaScript files that encode queries on the standard library search
-  index and expected results.
+  index and expected results. The format is specified
```
camelid, line 86: This needs to be updated because it's talking only about what is now called rustdoc-js-std.

camelid (approved):
Looks good with one thing fixed.

src/rustdoc-internals/search.md
```diff
@@ -354,3 +354,101 @@ The unification filter ensures that:
 The bloom filter checks none of these things,
 and, on top of that, can have false positives.
 But it's fast and uses very little memory, so the bloom filter helps.
+
+## Testing the search engine
+
+While the DOM UI code itself is tested using `rustdoc-gui` tests, the
```
GuillaumeGomez, line 360: Phrasing seems a bit strange here. I would say instead:

```suggestion
While the generated UI is tested using `rustdoc-gui` tests, the
```

src/rustdoc-internals/search.md
```diff
@@ -354,3 +354,101 @@ The unification filter ensures that:
...
+| `EXPECTED`     | `[ResultsTable]\|ResultsTable` | List of tests to run, specifying what the hypothetical user types into the search box and sees in the tabs
+| `PARSED`       | `[ParsedQuery]\|ParsedQuery`   | List of parser tests to run, without running an actual search
+
+`FILTER_CRATE` can be left out (equivalent to searching "all crates"), but you
+have to specify `EXPECTED` or `PARSED`.
+
+Additionally, the following magic comments are supported.
+Put them on their own line, without indenting.
+
+* `// exact-check`: If search results appear that aren't part of the test case,
```
GuillaumeGomez, line 386: It also checks for the items sorting.

lolbinarycat: isn't sorting checked by default?

GuillaumeGomez: Yes but it's not mentioned anywhere before `// ignore-order` as far as I can see. So better to precise it.

lolbinarycat: it should be documented, but not under `exact-check`, as order is checked regardless of the presence of `exact-check`.

src/rustdoc-internals/search.md
```diff
@@ -354,3 +354,100 @@ The unification filter ensures that:
...
+| -------------- | ------------------------------ | -------------------------------------------------------------------------------------------------------------
+| `FILTER_CRATE` | `string`                       | Only include results from the given crate. In the GUI, this is the "Results in <kbd>crate</kbd>" drop-down menu.
+| `EXPECTED`     | `[ResultsTable]\|ResultsTable` | List of tests to run, specifying what the hypothetical user types into the search box and sees in the tabs
+| `PARSED`       | `[ParsedQuery]\|ParsedQuery`   | List of parser tests to run, without running an actual search
+
+`FILTER_CRATE` can be left out (equivalent to searching "all crates"), but you
+have to specify `EXPECTED` or `PARSED`.
+
+By default, the test fails if any of the expected results are missing,
+or if the results don't appear in the specified order.
```
GuillaumeGomez, line 384: This isn't true, it only checks the order by default:

```suggestion
By default, the test fails if the results don't appear in the specified order.
```

GuillaumeGomez: ... End of week. Sorry about that. ^^'

GuillaumeGomez (approved):

src/rustdoc-internals/search.md
GuillaumeGomez: I prefer the one you commented.

compiler-errors (approved):

PR #2051, opened 2024-08-30
Fix a link and create an implied internal link 

pmundkur:
Make an implied internal link explicit, and fix the link for `elaborate_drop` transform.


jieyouxu (approved):
Thanks!

PR #2066, opened 2024-09-17
Fix conditions lowering text for enums with no fields

spastorino:
It's really enum's with variants without fields but wrote the text a little bit similar to what it's said in the following section.


jieyouxu (approved):
Thanks

PR #2068, opened 2024-09-18
Remove chalk is owned by WG-traits

spastorino:
The WG is retired and chalk is being replaced by the new solver.


compiler-errors (approved):
Good catch

PR #2070, opened 2024-09-20
Emphasize how to run a single tool test

adwinwhite:
Currently `x` can run individual rustc tests by directly passing file paths to `./x test`.
But this doesn't work for tool tests.
This PR provides some guidance for people who are confused by this.


jieyouxu (approved):
Thanks!

PR #2076, opened 2024-09-24
Note lldb debuginfo requires `python310.dll` to be present in `PATH` on Windows

jieyouxu:
Not sure about running lldb debuginfo on the Unixes, but AFAIK this is true on Windows.

### DIFF
--- a/src/tests/compiletest.md
+++ b/src/tests/compiletest.md
@@ -254,6 +254,17 @@ test based on the debugger currently being used:
   NOTE: The "Rust" version of LLDB doesn't exist anymore, so this will always be ignored.
   This should probably be removed.
 
+> **Note on running lldb debuginfo tests locally**
+>
+> If you want to run lldb debuginfo tests locally, then currently on Windows it is required that:
+> 
+> - You have **exactly** Python 3.10 installed -- not a newer version, not an older version,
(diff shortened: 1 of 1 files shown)


Noratrieb (approved):
not sure if people will find this before it's too late but neat

src/tests/compiletest.md
```diff
@@ -254,6 +254,17 @@ test based on the debugger currently being used:
   NOTE: The "Rust" version of LLDB doesn't exist anymore, so this will always be ignored.
   This should probably be removed.
 
+> **Note on running lldb debuginfo tests locally**
+>
+> If you want to run lldb debuginfo tests locally, then currently on Windows it is required that:
+> 
+> - You have **exactly** Python 3.10 installed -- not a newer version, not an older version,
```
ChrisDenton, line 261: This is a bit misleading. You can have other versions installed but this could be taken as meaning that you have to uninstall like Python 3.12 or whatever.

jieyouxu: I'll revise this wording slightly.

jieyouxu: https://github.com/rust-lang/rustc-dev-guide/pull/2076

PR #2080, opened 2024-09-28
building/suggested: Add instructions for Emacs & Helix

mrkajetanp:
Add instructions for setting up rust-analyzer with Emacs and Helix.
Additionally, move some content applicable to all editors out of the vscode section into its own section for visibility.

### DIFF
--- a/src/building/suggested.md
+++ b/src/building/suggested.md
@@ -21,19 +21,18 @@ You can also install the hook as a step of running `./x setup`!
 
 ## Configuring `rust-analyzer` for `rustc`
 
-### Visual Studio Code
+### Project-local rust-analyzer setup
 
 `rust-analyzer` can help you check and format your code whenever you save
 a file. By default, `rust-analyzer` runs the `cargo check` and `rustfmt`
 commands, but you can override these commands to use more adapted versions
-of these tools when hacking on `rustc`. For example, `x setup vscode` will prompt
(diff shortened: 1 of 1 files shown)


src/building/suggested.md
```diff
@@ -42,6 +41,12 @@ If you have enough free disk space and you would like to be able to run `x` comm
 rust-analyzer runs in the background, you can also add `--build-dir build-rust-analyzer` to the
 `overrideCommand` to avoid x locking.
 
+### Visual Studio Code
+
+Running `x setup vscode` will prompt you to create a `.vscode/settings.json`
```
jieyouxu, line 46: Nit: `./x setup vscode` for cli usage.

jieyouxu (approved):
Thanks, one tiny nit, then looks good to me.

jieyouxu:
@mrkajetanp also note that if you're interested, you can adjust the `./x setup` code in boostrap to support creating the emacs and helix config locally, just like for vscode.

src/building/suggested.md
mrkajetanp: sure, done!

mrkajetanp:
Good idea for adjusting the `./x setup` code, I might play around with it!
At first glance it looks very vscode-specific as it is so might need some refactoring but still should be fine.

PR #2084, opened 2024-10-03
update `x install` documentation

onur-ozkan:
`x install` documentation was quite outdated, so it needed an update. Also added instructions on how to install in custom paths.

### DIFF
--- a/src/building/build-install-distribution-artifacts.md
+++ b/src/building/build-install-distribution-artifacts.md
@@ -7,23 +7,25 @@ You’ll want to run this command to do it:
 ./x dist
 ```
 
-# Install distribution artifacts
+# Install from source
 
-If you’ve built a distribution artifact you might want to install it and
-test that it works on your target system. You’ll want to run this command:
+You might want to prefer installing Rust (and tools configured in your configuration)
+by building from source. If so, you want to run this command:
 
 ```bash
 ./x install
 ```
 
-   Note: If you are testing out a modification to a compiler, you
-   might want to use it to compile some project.
-   Usually, you do not want to use `./x install` for testing.
-   Rather, you should create a toolchain as discussed in
-   [here][create-rustup-toolchain].
+   Note: If you are testing out a modification to a compiler, you might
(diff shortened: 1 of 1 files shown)


src/building/build-install-distribution-artifacts.md
```diff
@@ -7,23 +7,25 @@ You’ll want to run this command to do it:
 ./x dist
 ```
 
-# Install distribution artifacts
+# Install from source
 
-If you’ve built a distribution artifact you might want to install it and
-test that it works on your target system. You’ll want to run this command:
+You might want to prefer installing Rust(and tools configured in your configuration)
```
jieyouxu, line 12: Nit

```suggestion
You might want to prefer installing Rust (and tools configured in your configuration)
```

src/building/build-install-distribution-artifacts.md
```diff
@@ -7,23 +7,25 @@ You’ll want to run this command to do it:
...
 ./x install
 ```
 
-   Note: If you are testing out a modification to a compiler, you
-   might want to use it to compile some project.
-   Usually, you do not want to use `./x install` for testing.
-   Rather, you should create a toolchain as discussed in
-   [here][create-rustup-toolchain].
+   Note: If you are testing out a modification to a compiler, you might
+   want to build the compiler (with `./x build`) then create a toolchain as
```
jieyouxu, line 20: Nit: probably `./x build library --stage 1` just to be explicit here.

jieyouxu (approved):
Thank you! I have some minor nits, but LGTM otherwise.

src/building/build-install-distribution-artifacts.md
onur-ozkan: Default stage is already `1` and `build library` doesn't really make it clearer (as we say 'build the compiler'). I prefer to show the shortest and most straightforward commands whenever possible while relying on bootstrap defaults.

jieyouxu: Fair enough.

PR #2087, opened 2024-10-05
Fixed links to rust-analyzer configs for Emacs and Helix

YohDeadfall:
These link were not rendered properly due to the lost address parts. For `rust_analyzer_settings.json` it worked because its reference is located at the end of the document.


JohnTitor (approved):
Thanks!

PR #2092, opened 2024-10-07
Add redirects for integration-testing and headers

jieyouxu:
I forgor to add redirects for them from the old URLs to the new URLs.


book.toml
```diff
@@ -54,3 +54,5 @@ warning-policy = "error"
 "/diagnostics/sessiondiagnostic.html" = "diagnostic-structs.html"
 "/diagnostics/diagnostic-codes.html" = "error-codes.html"
 "/miri.html" = "const-eval/interpret.html"
+"/tests/integration.html" = "tests/ecosystem.html"
+"/tests/headers.html" = "tests/directives.html"
```
ehuss, line 58: These need to be relative to the source directory.

```suggestion
"/tests/integration.html" = "ecosystem.html"
"/tests/headers.html" = "directives.html"
```

ehuss (commented):
Thanks! ❤️

book.toml
jieyouxu: Wait wdym relative to the source directory? Because e.g. directives.html is still under the `tests/` folder no?

jieyouxu: I changed the right-side to be as suggested, though I can't claim I understand this redirect syntax exactly 🤷

jieyouxu: Ah okay thanks for the explanation.

DianQK (approved):
LGTM. Thanks!

PR #2093, opened 2024-10-08
Ask folks to use the Oxford comma in diagnostics

onkoe:
Hey there! In Rust, it seems that most user-facing types of content prefer unambiguous grammar. However, there currently isn't any instruction to use the Oxford comma when writing lists of items within diagnostics. 

I added that to avoid any ambiguity, such as that which I fixed in https://github.com/rust-lang/rust/pull/131038

Please let me know if any changes are necessary! :)

### DIFF
--- a/src/diagnostics.md
+++ b/src/diagnostics.md
@@ -156,6 +156,8 @@ use an error-level lint instead of a fixed error.
   compiler messages are an important learning tool.
 - When talking about the compiler, call it `the compiler`, not `Rust` or
   `rustc`.
+- Use the [Oxford comma](https://en.wikipedia.org/wiki/Serial_comma) when
+  writing lists of items.
 
 ### Lint naming
 


src/diagnostics.md
```diff
@@ -156,6 +156,8 @@ use an error-level lint instead of a fixed error.
   compiler messages are an important learning tool.
 - When talking about the compiler, call it `the compiler`, not `Rust` or
   `rustc`.
+- Use the [Oxford comma](https://en.wikipedia.org/wiki/Serial_comma) when
+  writing lists of items.
```
jieyouxu, line 160: Nit: I don't think we want to make this advise because like the wikipedia entry says

> The serial comma can serve to avoid ambiguity in specific contexts, though its employment may also generate ambiguity under certain circumstances.

I think it's context-dependent and is best left to the reviewer/author to adjust whether or not to deploy oxford commas.

tshepang: I have rarely seen it add ambiguity, and have far more seen it's lack add ambiguity

onkoe: It's a good call to think a bit before writing diagnostics. However, I intentionally qualified it with "when listing items". 

The ambiguities stemming from the extra comma are pretty rare. When they do appear, it's typically because the author was doing more than listing items.

With that in mind, do you think softer wording would be more effective? My goal was to avoid ambiguity in the diagnostic formatting rules.

jieyouxu: I don't have *too* strong of an opinion on this, so if tshepang is in favor then we can merge and see if this advice turns out to not always apply in practice (I suspect rarely).

PR #2094, opened 2024-10-09
Document compiletest directives `ignore-coverage-map` and `ignore-coverage-run`

Zalathar:
Follow-up to https://github.com/rust-lang/rust/pull/131400, which renamed these directives.

### DIFF
--- a/src/tests/directives.md
+++ b/src/tests/directives.md
@@ -145,6 +145,8 @@ Some examples of `X` in `ignore-X` or `only-X`:
 - When particular debugger versions are matched: `ignore-gdb-version`
 - Specific [compare modes]: `compare-mode-polonius`, `compare-mode-chalk`,
   `compare-mode-split-dwarf`, `compare-mode-split-dwarf-single`
+- The two different test modes used by coverage tests:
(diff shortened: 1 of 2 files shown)


src/tests/directives.md
```diff
@@ -145,6 +145,8 @@ Some examples of `X` in `ignore-X` or `only-X`:
 - When particular debugger versions are matched: `ignore-gdb-version`
 - Specific [compare modes]: `compare-mode-polonius`, `compare-mode-chalk`,
   `compare-mode-split-dwarf`, `compare-mode-split-dwarf-single`
+- The two different test modes used by coverage tests:
+  `ignore-coverage-map`, `ignore-coverage-run`
```
Zalathar, line 149: The style here is a bit inconsistent with most of the other list items, which omit the `ignore-` or `only-` prefix. I chose to use the full name here partly because it's more searchable, and partly because `only-coverage-map` etc are not in the permitted list of directives and therefore don't actually work.

jieyouxu: This is fine, thanks.

jieyouxu (approved):
Thanks!

PR #2098, opened 2024-10-13
update const stability docs

RalfJung:
This updates the docs for https://github.com/rust-lang/rust/pull/131349. Should only land once that PR landed.

### DIFF
--- a/src/stability.md
+++ b/src/stability.md
@@ -53,22 +53,38 @@ marks an item as stabilized. Note that stable functions may use unstable things 
 
 The `#[rustc_const_unstable(feature = "foo", issue = "1234", reason = "lorem
 ipsum")]` has the same interface as the `unstable` attribute. It is used to mark
-`const fn` as having their constness be unstable. Every `const fn` with
-stability attributes should carry either this attribute or
-`#[rustc_const_stable]` (see below).
+`const fn` as having their constness be unstable. This is only needed in rare cases:
+- If a `const fn` makes use of unstable language features or intrinsics.
+  (The compiler will tell you to add the attribute if you run into this.)
+- If a `const fn` is `#[stable]` but not yet intended to be const-stable.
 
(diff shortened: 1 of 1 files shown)


RalfJung:
@rust-lang/libs @rust-lang/libs-api FYI, and also would be good to get some feedback if you think this system makes sense or you would like to see some aspect of it work differently. :)

RalfJung:
https://github.com/rust-lang/rust/pull/131349 is merged now, so could someone merge this docs PR as well? :)

RalfJung:
Seems like this was merged with a rebase, not a clean merge? That makes it a lot harder to automatically track which of my branches have been merged already... :(

jieyouxu:
I think I just clicked one of the merge buttons, I'll talk with other dev-guide maintainers and see we can stick with *one* of the merging options, lol.

jieyouxu:
> Seems like this was merged with a rebase, not a clean merge? That makes it a lot harder to automatically track which of my branches have been merged already... :(

@RalfJung actually just to clarify, would a rebase and merge work? I may have accidentally clicked the squash and merge.

RalfJung:
Rebase and merge still means the commit as it exists in my branch does not end up in the repo, leaving me with no way to automatically determine that my branch was merged. So I have to delete all the old branches by hand, which is somewhat tedious. Lucky enough most rust-lang repos use a normal merge so I can have some scripts that detect all branches which have been merged (local and remote) and delete them. :)

I understand some people prefer a linear history, but I don't quite understand why... and in rust-lang we don't seem to have a default policy either way.

jieyouxu:
Right. So apparently in this repo the default "merge" button is disabled for some reason, I'll ask around why... This is just docs so I really don't think the rust-lang/rust no merge policy is relevant.

RalfJung:
rust-lang uses merge commits without rebase to merge PRs. The policy there is only about no merge commits *inside* a PR.

PR #2106, opened 2024-10-29
Describe why and how to use a separate build directory for rust-analyzer

Zalathar:
There is some existing advice about this in the dev guide, but it is terse and incomplete.

https://rust-lang.zulipchat.com/#narrow/channel/122651-general/topic/Setup.20multiple.20editors.20for.20work.20on.20standard.20lib.3F/near/479397923

### DIFF
--- a/src/building/suggested.md
+++ b/src/building/suggested.md
@@ -37,13 +37,34 @@ checking time. For example, if you are working on the compiler, you can override
 the command to `x check compiler --json-output` to only check the compiler part.
 You can run `x check --help --verbose` to see the available parts.
 
-If you have enough free disk space and you would like to be able to run `x`
-commands while rust-analyzer runs in the background, you can also add
-`--build-dir build-rust-analyzer` to the `overrideCommand` to avoid x locking.
-
 Running `./x setup editor` will prompt you to create a project-local LSP config
 file for one of the supported editors. You can also create the config file as a
 step of running `./x setup`.
+
+### Using a separate build directory for rust-analyzer
+
+By default, when rust-analyzer runs a check or format command, it will share
+the same build directory as manual command-line builds. This can be inconvenient
+for two reasons:
+- Each build will lock the build directory and force the other to wait, so it
+  becomes impossible to run command-line builds while rust-analyzer is running
+  commands in the background.
+- There is an increased risk of one of the builds deleting previously-built
+  artifacts due to conflicting compiler flags or other settings, forcing
+  additional rebuilds in some cases.
+
+To avoid these problems:
+- Add `--build-dir=build-rust-analyzer` to all of the custom `x` commands in
+  your editor's rust-analyzer configuration.
+  (Feel free to choose a different directory name if desired.)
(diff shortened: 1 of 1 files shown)


src/building/suggested.md
```diff
@@ -37,14 +37,35 @@ checking time. For example, if you are working on the compiler, you can override
...
+  your editor's rust-analyzer configuration.
+  (Feel free to choose a different directory name if desired.)
+- Modify the `rust-analyzer.procMacro.server` setting so that it points to the
+  copy of `rust-analyzer-proc-macro-srv` in that other build directory.
+
+Using separate build directories for command-line builds and rust-analyzer
+requires extra disk space, and also means that running `./x clean` on the
+command-line will not clean out the separate build directory. To clean the
+separate build directory, run `./x clean --build-dir=build-rust-analyzer`
+instead.
```
jieyouxu, line 67: Suggestion: instead of this, we can also just nest r-a's build directory under `build/`, i.e. `build/build-rust-analyzer`

Zalathar: That approach has always seemed a bit too cute to me, especially for a recommendation.

In particular:
- It becomes impossible to clean the main build directory *without* deleting the r-a build directory.
- Bootstrap doesn't actually know how to clean a nested build directory; it's just getting deleted as a byproduct of how the clean command is currently implemented.

jieyouxu: Fair enough.

jieyouxu (approved):
Thanks!

lnicola:
My comments on this from the r-a side:

 - if you prefer to run `./x.py check` from the command line, another option is to disable `checkOnSave`
 - a `check` started by r-a will block one started from the terminal, but the latter will be fast enough since it will have nothing to do besides checking for freshness; running two `check`s at once is wasteful but with CPU and disk space

Zalathar:
The issue is more that r-a running checks in the background will block *all* command-line bootstrap actions, including building the compiler or running tests, even if the check is relatively unimportant.

And once r-a has started its check, there's no way to stop it without stopping the server, which then implies several minutes of re-indexing when the server is restarted.

And aside from that, there are a few known cases where bootstrap will delete artifacts to avoid incompatibilities between different settings, which means that letting r-a touch the main build directory in the background is a really bad idea (even if everything is protected by locks).

lnicola:
> The issue is more that r-a running checks in the background will block all command-line bootstrap actions, including building the compiler or running tests, even if the check is relatively unimportant.

Yes, but will the equivalent of `cargo check & cargo test --no-run; wait` be faster than `cargo check; cargo test --no-run`? I wouldn't be sure.

> And once r-a has started its check, there's no way to stop it without stopping the server

In VS Code (and other clients which implement the extension) there is a "Cancel running flycheck" command that does exactly that. Clicking on the status bar item and Cancel might also work.

connortsui20:
https://github.com/rust-lang/rust/pull/132794 seems to have modified the behavior and path for Helix's build directory, so I think this needs to be updated slightly?

PR #2120, opened 2024-11-03
Add a couple of linkcheck exceptions:

marxin:
Addresses:
```

Server returned 500 Internal Server Error for https://github.com/rust-lang/rust/pulls?q=is%3Apr+label%3AS-inactive+is%3Aclosed
Server returned 401 Unauthorized for https://objects.githubusercontent.com/github-production-release-asset-2e65be/343907537/60cd5880-7b81-11eb-9502-68d5f473ebe2?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=releaseassetproduction%2F20241103%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20241103T085906Z&X-Amz-Expires=300&X-Amz-Signature=e532fd7d2bd271f12988982bc8e5a47a0f966b40452b55d444bbb74292ac1fb1&X-Amz-SignedHeaders=host&response-content-disposition=attachment%3B%20filename%3Drustc.generic.wpaProfile&response-content-type=application%2Foctet-stream
Server returned 500 Internal Server Error for https://github.com/rust-lang/rust/pulls?q=is%3Apr+label%3AS-inactive
```

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -42,9 +42,9 @@ exclude = [
     "marketplace\\.visualstudio\\.com",
     "objects\\.githubusercontent\\.com",
     # The bug listing URL works only if an user is logged in, otherwise one gets 404.
-    "github\\.com/issues\\?",
+    "github\\.com/issues\\?q=.*",
     # Similarly 500 is sometimes returned here.
-    "github\\.com/rust-lang/rust/pulls\\?",
+    "github\\.com/rust-lang/rust/pulls\\?q=.*",
     # 401 is returned here for unknown reason
     "github\\.com/wesleywiser/rustc-bootstrap-wpa-analysis",
     # Handle: connection closed before message completed
--- a/book.toml
+++ b/book.toml
@@ -43,8 +43,14 @@ exclude = [
     "objects\\.githubusercontent\\.com",
     # The bug listing URL works only if an user is logged in, otherwise one gets 404.
     "github\\.com/issues\\?",
+    # Similarly 500 is sometimes returned here.
+    "github\\.com/rust-lang/rust/pulls\\?",
+    # 401 is returned here for unknown reason
+    "github\\.com/wesleywiser/rustc-bootstrap-wpa-analysis",
     # Handle: connection closed before message completed
     "microsoft\\.com/en-us/research/publication/",
+    # 500 is returned for HEAD request
+    "code\\.visualstudio\\.com/docs/editor/tasks",
 ]
 cache-timeout = 86400
 warning-policy = "error"


book.toml
```diff
@@ -43,8 +43,14 @@ exclude = [
     "objects\\.githubusercontent\\.com",
     # The bug listing URL works only if an user is logged in, otherwise one gets 404.
     "github\\.com/issues\\?",
+    # Similarly 500 is sometimes returned here.
+    "github\\.com/rust-lang/rust/pulls\\?",
```
jieyouxu, line 47: I'm not sure we want to skip this, isn't PR number linking kinda important

jieyouxu (commented):
I'm not sure about the PR number exceptions, but I also am not sure about the linkcheck in general... so maybe cc @tshepang and @camelid?

book.toml
marxin: Yeah, it is. I want to skip only the 2 particular queries that bring the troubles:
```
https://github.com/rust-lang/rust/pulls?q=is%3Apr+label%3AS-inactive+is%3Aclosed
https://github.com/rust-lang/rust/pulls?q=is%3Apr+label%3AS-inactive
```

Does it make sense?

marxin:
> I'm not sure about the PR link exceptions, but I also am not sure about the linkcheck in general... so maybe cc @tshepang and @camelid?

Note these are the last missing pieces after which the full link check will become green again. Moreover, with `mdbook-linkcheck2`, we can also enable checking for modified files (with `-f` option) and should result in more stable links in the future.

book.toml
jieyouxu: Oh okay, that makes sense. But also at the same time, it's horrifying(?) that these endpoints can return internal server errors?

jieyouxu: @marxin out of curiosity, do you know if that happens frequently or did it occur during GitHub outages?

marxin: If I create a dummy book and include the link, then it's fine. But these links are verified in a batch with other GH links (in parallel) and maybe we hit some kind of DoS protection mechanism. Who knows :)

marxin: > @marxin out of curiosity, do you know if that happens frequently or did it occur during GitHub outages?

In my case, it happens every single time. Similarly, if you take a look at the nightly CI:
https://github.com/rust-lang/rustc-dev-guide/actions/runs/11653521916/job/32445996122

jieyouxu: I suspect they're actually just rate limits, but yeah anyway, thanks for the info.

marxin: The rate limits are properly reflected in the return code: `Server returned 429 Too Many Requests for https://github.com`, but these 500s are somehow different.

marxin:
May I ping this PR @camelid as the `mdbook-linkcheck2` PR has been merged?

book.toml
```diff
@@ -42,9 +42,15 @@ exclude = [
     "marketplace\\.visualstudio\\.com",
     "objects\\.githubusercontent\\.com",
     # The bug listing URL works only if an user is logged in, otherwise one gets 404.
-    "github\\.com/issues\\?",
+    "github\\.com/issues\\?q=.*",
+    # Similarly 500 is sometimes returned here.
+    "github\\.com/rust-lang/rust/pulls\\?q=.*",
```
jieyouxu, line 47: Question: is this still a problem? Occassionally github does have outages, but it seems surprising that basic queries like that can return 500.

book.toml
```diff
@@ -42,9 +42,15 @@ exclude = [
...
+    # 401 is returned here for unknown reason
+    "github\\.com/wesleywiser/rustc-bootstrap-wpa-analysis",
```
jieyouxu, line 49: Question: similarly, is this still a problem?

jieyouxu:
One more (general) question... does the linkchecker have any retry mechanism for 5xx errors?

book.toml
camelid: Either way, I don't think it really matters because this type of link should always be valid.

camelid (approved):

PR #2140, opened 2024-11-20
Make `Diag` a clickable link in Suggestion section

ohno418:
Since `Diag` struct  is the core struct in the Suggestions section, it would be better to make it a clickable link.
https://rustc-dev-guide.rust-lang.org/diagnostics.html#suggestions


jieyouxu (approved):
Thanks

Issue #2141, opened 2024-11-20
Actionables from 2024-11-20 reading-club meeting


jieyouxu:
- [x] remove "identifiers in the compiler" (individual chapters for each ast/ir should just go in detail if warranted)
- [x] move the stuff thats completely unrelated to source code repr to some new header such as "supporting infra" (and probably the closure expansion one to analysis since it basically just talks about type checking lol)
- [ ] #2143
- [x] dont inline all the definitions of inline asm ast/hir/mir/whatever into the dev guide just link to it
- [ ] do we talk about async anywhere lol

See https://rust-lang.zulipchat.com/#narrow/channel/196385-t-compiler.2Fwg-rustc-dev-guide/topic/.5Breading.20club.5D.202024.2E11.2E20.20chapter.2040.20to.2042/near/483542730

jieyouxu:
Two actionables were converted into separate issues.

PR #2142, opened 2024-11-21
Reorganize the "Source Code Representation" chapters

BoxyUwU:
This ticks off a number of boxes from #2141, specifically it:
- Creates a new chapter `Supporting Infrastructure` and moves a number of unrelated chapters out of the source code repr chapter
- Removes the chapter about IR IDs moving its information into the respective IR chapters
- Removes a number of type definitions from the inline asm chapter, instead opting to reference the rustc

The only remaining follow ups would be to flesh out the hir chapter some more, and talk about async stuff somewhere. It might be worth moving those to their own issues though

cc @jieyouxu


jieyouxu (approved):
Thanks! Very nice cleanup :+1:

PR #2149, opened 2024-11-28
Update `//@ proc-macro` aux build directive docs

jieyouxu:
For https://github.com/rust-lang/rust/pull/133540.

cc @ehuss

### DIFF
--- a/src/tests/compiletest.md
+++ b/src/tests/compiletest.md
@@ -572,6 +572,7 @@ There are multiple [directives](directives.md) to assist with that:
 - `aux-crate`
 - `aux-bin`
 - `aux-codegen-backend`
+- `proc-macro`
 
 `aux-build` will build a separate crate from the named source file. The source
 file should be in a directory called `auxiliary` beside the test file.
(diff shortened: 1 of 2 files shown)


src/tests/compiletest.md
```diff
@@ -618,45 +619,55 @@ for tests in `tests/ui-fulldeps`, since it requires the use of compiler crates.
...
+   to produce a dylib for the aux crate.
+3. The aux crate is made available to the test file via extern prelude with
+   `--extern <aux_crate_name>`. Note that since UI tests default to edition
+   2015, you still need to specify `extern <aux_crate_name>` unless the main
+   test file is using an edition that is 2018 or newer.
+4. The `proc_macro` crate is made available as an extern prelude module. Same
+   edition 2015 vs newer edition distinction for `extern proc_macro;` applies.
+
+For example, you might have a test `tests/ui/cat/meow.rs` and proc-macro
+auxiliary `tests/ui/cat/auxiliary/whiskers.rs`:
```
jieyouxu, line 639: Remark: purposefully did not talk about transitive aux builds because of cycle problems.

src/tests/compiletest.md
```diff
@@ -618,45 +619,55 @@ for tests in `tests/ui-fulldeps`, since it requires the use of compiler crates.
...
+same parent folder as the main test file. However, it also has four additional
+preset behavior compared to `aux-build` for the proc-macro test auxiliary:
+
+1. The aux test file is built with `--crate-type=proc-macro`.
+2. The aux test file is built without `-C prefer-dynamic`, i.e. it will not try
+   to produce a dylib for the aux crate.
+3. The aux crate is made available to the test file via extern prelude with
+   `--extern <aux_crate_name>`. Note that since UI tests default to edition
+   2015, you still need to specify `extern <aux_crate_name>` unless the main
+   test file is using an edition that is 2018 or newer.
```
ehuss, line 634: No need to change, but to clarify here, it is only required if you want to use the crate name in a `use` import. You can refer to it directly, like `fn main() { my_aux_thing::foo!(); }` without using `use` or `extern crate`.

src/tests/compiletest.md
```diff
@@ -618,45 +619,55 @@ for tests in `tests/ui-fulldeps`, since it requires the use of compiler crates.
...
+extern crate proc_macro;
+use proc_macro::*;
 
-fn main() {
-    my_proc_macro::foo!();
+#[proc_macro]
+pub fn identity(ts: TokenStream) -> TokenStream {
+    ts
 }
 ```
```
ehuss, line 670: Can we include a note about it not working with `build-aux-doc`? Maybe something like:

```suggestion

> **Note**: The `proc-macro` header currently does not work with the
> `build-aux-doc` header for rustdoc tests. In that case, you will need to use
> the `aux-build` header, and use `#![crate_type="proc_macro"]`, and 
> `//@ force-host` and `//@ no-prefer-dynamic` headers in the proc-macro.
```

Hopefully this will get fixed at some point, and we can remove this.

ehuss (commented):
Thanks for putting this together! ❤️ 

Can you also add it to the list in the table here? https://github.com/rust-lang/rustc-dev-guide/blob/220aaee4f3f1ff33b371cd1e815b47e7c0c830d1/src/tests/directives.md?plain=1#L55-L61

src/tests/compiletest.md
jieyouxu: Ah yes, good point

src/tests/compiletest.md
jieyouxu: Sure

src/tests/compiletest.md
jieyouxu: Added clarification.

src/tests/compiletest.md
jieyouxu: Included this note.

jieyouxu:
Addressed review feedback, lmw if there's anything else I'm missing.

jieyouxu:
I'm going to merge this first because the directive is already available, if there's any further changes we can send follow-ups.

ehuss:
Looks good, thanks!

PR #2152, opened 2024-12-02
Explain how to deal with exploded git submodules

jieyouxu:


BoxyUwU:
i hate when my submodules explode

PR #2155, opened 2024-12-05
extend closure constraints section

lcnr:
wrote this while looking into it myself, I think it's probably useful, but definitely not perfect :thinking:


jieyouxu (approved):
Something is better than nothing :3

PR #2157, opened 2024-12-06
squashing: recommend --keep-base when squashing without a conflict

RalfJung:


jieyouxu (approved):
Cool

jieyouxu:
(You can self-merge, we don't have the 1-approval requirement anymore if u have write access)

RalfJung:
I could but that doesn't mean I should. I'd rather have some double-check my edits and make sure there's not some context I am overlooking.

PR #2162, opened 2024-12-15
Rework the driver docs

bjorn3:
The removal of driver queries has changed the way rustc_interface can be used. I've also changed things a bit to fix things that were wrong in the docs even before the driver query removal. And finally this is now recommends using rustc_driver over rustc_interface like most tools are already doing. The rustc_driver api is a bit more stable and much easier to use when you don't need the flexibility of rustc_interface as is the case almost all the time. Rustdoc is one of the few tools that needs the flexibility of rustc_interface.

Blocked on https://github.com/rust-lang/rust/pull/134302 and testing the updated examples.

### DIFF
--- a/examples/rustc-driver-example.rs
+++ b/examples/rustc-driver-example.rs
@@ -1,35 +1,36 @@
 #![feature(rustc_private)]
 
+extern crate rustc_ast;
 extern crate rustc_ast_pretty;
+extern crate rustc_data_structures;
 extern crate rustc_driver;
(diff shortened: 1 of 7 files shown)


jieyouxu:
@bjorn3 let me know if you want to do further testing, or if this is ready (we can also improve example in follow-ups).

bjorn3:
Updated examples. Ready for review.

jieyouxu (approved):
Thanks!

PR #2163, opened 2024-12-16
Specify what a CGU is

connortsui20:
Adds (CGU) clarification

This might seems kind of silly but I actually tripped up on this because I'm not familiar with codegen units nor with the other material in this document so going from codegen unit <-> CGU was kind of hard :laughing:


jieyouxu (approved):
Thanks

PR #2164, opened 2024-12-17
Add suggestion for `--keep-stage 0`

ehuss:
This adds a suggestion for `--keep-stage 0` when iteratively working on the standard library (and wanting to have a full toolchain otherwise).


jieyouxu (approved):
Thanks

PR #2165, opened 2024-12-17
Fix trivial typo of "query-fied"

cbloodsworth:
"Query-field" should be "Query-fied" - three instances.

Nitpicky for sure, but threw me for a slight loop first time I read it.


tshepang (approved):

tshepang:
that is more a correction than a mere nitpick... thank you for the fix

PR #2181, opened 2024-12-30
Make linkcheck2 detection automatic

marxin:


ci/linkcheck.sh
```diff
@@ -9,8 +9,8 @@ set_github_token() {
   jq '.config.output.linkcheck."http-headers"."github\\.com" = ["Authorization: Bearer $GITHUB_TOKEN"]'
 }
 
-if [ -z "$ENABLE_LINKCHECK" ] ; then
-  echo "Skipping link check."
+if ! which $LINKCHECK_BINARY &>/dev/null ; then
+  echo "WARNING: Skipping link check. Consider running 'cargo install $LINKCHECK_BINARY'."
```
DianQK, line 13: It should be an error on CI.

marxin: Good point!

marxin:
CC: @camelid

Kobzol:

```
git fetch origin --all
git checkout <pr-branch>
git push --force-with-lease
```

Kobzol:
When I don't have the mdbook-linkcheck2 and try to run `mdbook build`, it fails with this error:
```
2025-01-06 09:55:26 [INFO] (mdbook::book): Book building has started
2025-01-06 09:55:26 [INFO] (mdbook::book): Running the html backend
2025-01-06 09:55:28 [INFO] (mdbook::book): Running the linkcheck backend
2025-01-06 09:55:28 [INFO] (mdbook::renderer): Invoking the "linkcheck" renderer
WARNING: Skipping link check. Consider running 'cargo install mdbook-linkcheck2'.
2025-01-06 09:55:28 [WARN] (mdbook::renderer): Error writing the RenderContext to the backend, Broken pipe (os error 32)
```
I assume that is not expectd?

camelid:
The RenderContext message is unrelated to mdbook-linkcheck2 and linkchecking in general AFAICT. I'm not sure why it's happening, but note that it's a warning not an error. The command still finishes successfully.

Kobzol:
Sorry, I didn't notice that, the "Error" in the warning message threw me off :)

camelid (commented):
I lean against doing this change for a couple of reasons:
- Binary detection can lead to weird inconsistencies where different environments lead to different behaviors. E.g. one person may have linkcheck2 installed for use in other projects but another doesn't, and it will be confusing if their build behaves differently.
- Running linkchecking wastes a bit of your time and network resources locally (especially for people who have bad internet). Since it's run in CI anyway, it seems unnecessary to also run it by default locally (for people who already have the binary).
- Nearly all of the time new links you're adding are going to be correct because you just copy and paste them. Links usually only break over time.
- It's not much of a hassle to add `ENABLE_LINKCHECK=1` locally if you want it.

marxin:
> I lean against doing this change for a couple of reasons:

Works for me!

PR #2182, opened 2024-12-30
Document josh subtrees and update subtree repository list

Kobzol:
Mention our usage of josh in `miri` and `rust-analyzer`, and add the cranelift backend to the list.

### DIFF
--- a/src/external-repos.md
+++ b/src/external-repos.md
@@ -12,8 +12,7 @@ changes; and use submodules for tools that are independent of the compiler.
 
 ## External Dependencies (subtrees)
 
-As a developer to this repository, you don't have to treat the following external projects
-differently from other crates that are directly in this repo:
+The following external projects are managed using some form of a `subtree`:
 
 * [clippy](https://github.com/rust-lang/rust-clippy)
 * [miri](https://github.com/rust-lang/miri)
@@ -30,7 +29,7 @@ implement a new tool feature or test, that should happen in one collective rustc
 
 `subtree` dependencies are currently managed by two distinct approaches:
 
-* Using `git subtree` directly
+* Using `git subtree`
   * `clippy` ([sync guide](https://doc.rust-lang.org/nightly/clippy/development/infrastructure/sync.html#performing-the-sync-from-rust-langrust-to-clippy))
   * `rustfmt`
   * `rustc_codegen_cranelift` ([sync script](https://github.com/rust-lang/rustc_codegen_cranelift/blob/113af154d459e41b3dc2c5d7d878e3d3a8f33c69/scripts/rustup.sh#L7))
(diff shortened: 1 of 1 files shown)


src/external-repos.md
```diff
@@ -29,6 +28,20 @@ to these tools should be filed against the tools directly in their respective
 upstream repositories. The exception is that when rustc changes are required to
 implement a new tool feature or test, that should happen in one collective rustc PR.
 
+`subtree` dependencies are currently managed by two distinct approaches:
+
+* Using `git subtree` directly
```
RalfJung, line 33:
```suggestion
* Using `git subtree`
```
There's no "indirect" usage so I find the additional qualifier confusing

src/external-repos.md
```diff
@@ -29,6 +28,20 @@ to these tools should be filed against the tools directly in their respective
...
+  * `clippy` ([sync guide](https://doc.rust-lang.org/nightly/clippy/development/infrastructure/sync.html#performing-the-sync-from-rust-langrust-to-clippy))
+  * `rustfmt`
+  * `rustc_codegen_cranelift` ([sync script](https://github.com/rust-lang/rustc_codegen_cranelift/blob/113af154d459e41b3dc2c5d7d878e3d3a8f33c69/scripts/rustup.sh#L7))
+* Using the [josh][josh] tool
+  * `miri` ([sync guide](https://github.com/rust-lang/miri/blob/master/CONTRIBUTING.md#advanced-topic-syncing-with-the-rustc-repo))
+  * `rust-analyzer` ([sync script](https://github.com/rust-lang/rust-analyzer/blob/2e13684be123eca7181aa48e043e185d8044a84a/xtask/src/release.rs#L147))
+
+The [josh][josh] tool is an alternative to git subtrees, which manages git history in a different way and scales better to larger repositories. The differences between `git subtree` and josh are described [here](https://hackmd.io/7pOuxnkdQDaL1Y1FQr65xg). It requires specific tooling to work with josh, you can check out the `miri` or `rust-analyzer` scripts linked above for inspiration.
```
RalfJung, line 41: > The differences between `git subtree` and josh are described [here](https://hackmd.io/7pOuxnkdQDaL1Y1FQr65xg).

This doesn't describe the difference, it's a migration guide.

src/external-repos.md
```diff
@@ -3,24 +3,23 @@
...
 
 As a general rule, use crates.io for libraries that could be useful for others in the ecosystem; use
 subtrees for tools that depend on compiler internals and need to be updated if there are breaking
 changes; and use submodules for tools that are independent of the compiler.
 
-## External Dependencies (subtree)
+## External Dependencies (subtrees)
 
 As a developer to this repository, you don't have to treat the following external projects
 differently from other crates that are directly in this repo:
```
RalfJung, line 16: That's not quite true, tools generally prefer PRs to their own repos when changes don't have to be done in lockstep with Rust compiler changes.

Kobzol: This was pre-existing, but I agree it was confusing, I reworded it.

src/external-repos.md
Kobzol: I wanted to distinguish "using subtrees via josh" and "using subtrees via git subtree", but the word was extraneous, I suppose. Removed it.

jieyouxu (approved):
Thanks!

src/external-repos.md
```diff
@@ -29,6 +27,20 @@ to these tools should be filed against the tools directly in their respective
 upstream repositories. The exception is that when rustc changes are required to
 implement a new tool feature or test, that should happen in one collective rustc PR.
 
+`subtree` dependencies are currently managed by two distinct approaches:
+
+* Using `git subtree`
```
RalfJung, line 32: portable-simd also uses this; it seems to be entirely missing from this list?

Kobzol: The list was incomplete before, I added a few entries, but since there was no global list before that, I didn't notice portable-simd. Good catch! Added it in https://github.com/rust-lang/rustc-dev-guide/pull/2214.

PR #2184, opened 2024-12-31
Use `id` attribute on `<a>` elements instead of deprecated `name` attribute

max-heller:
The `name` attribute on `<a>` elements [is deprecated and should be replaced by the `id` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#name).

`<a id="target">` is the pattern used by other mdbooks, e.g. the [Rust Book](https://github.com/rust-lang/book/blob/04d06dfe541607e6419f3d028c3f9b245f3be4d9/src/ch04-03-slices.md?plain=1#L237).

### DIFF
--- a/src/appendix/background.md
+++ b/src/appendix/background.md
@@ -4,7 +4,7 @@ This section covers a numbers of common compiler terms that arise in
 this guide. We try to give the general definition while providing some
 Rust-specific context.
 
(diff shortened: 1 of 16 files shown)


jieyouxu (approved):
Thanks

PR #2186, opened 2025-01-01
comment out FIXMEs to not display them on UI

onur-ozkan:
Prevents FIXMEs from appearing on the UI. Unless I am missing something, displaying these to readers doesn't make much sense. Some of them might make sense to include on the UI, but I am unsure which ones. If you think any should be displayed please mark them specifically.

This is how it currently looks:

image  

And this is how it looks after this patch:

image

### DIFF
--- a/src/solve/normalization.md
+++ b/src/solve/normalization.md
@@ -52,12 +52,14 @@ before assigning the resulting rigid type to an inference variable. This is used
 This has to be used whenever we match on the value of some type, both inside
 and outside of the trait solver.
 
+<!--
 FIXME: structure, maybe we should have an "alias handling" chapter instead as
 talking about normalization without explaining that doesn't make too much
 sense.
 
 FIXME: it is likely that this will subtly change again by mostly moving structural
 normalization into `NormalizesTo`.
+-->
 
 [structural_norm]: https://github.com/rust-lang/rust/blob/2627e9f3012a97d3136b3e11bf6bd0853c38a534/compiler/rustc_trait_selection/src/solve/alias_relate.rs#L140-L175
 [structural-relate]: https://github.com/rust-lang/rust/blob/a0569fa8f91b5271e92d2f73fd252de7d3d05b9c/compiler/rustc_trait_selection/src/solve/alias_relate.rs#L88-L107
@@ -72,11 +74,13 @@ possible. However, this only works for aliases referencing bound variables if th
 not ambiguous as we're unable to replace the alias with a corresponding inference
 variable without leaking universes.
 
+<!--
 FIXME: we previously had to also be careful about instantiating the new inference
 variable with another normalizeable alias. Due to our recent changes to generalization,
 this should not be the case anymore. Equating an inference variable with an alias
 now always uses `AliasRelate` to fully normalize the alias before instantiating the
 inference variable: [source][generalize-no-alias]
+-->
 
 [generalize-no-alias]: https://github.com/rust-lang/rust/blob/a0569fa8f91b5271e92d2f73fd252de7d3d05b9c/compiler/rustc_infer/src/infer/relate/generalize.rs#L353-L358
 
--- a/src/solve/normalization.md
+++ b/src/solve/normalization.md
@@ -1,5 +1,8 @@
 # Normalization in the new solver
 
+> **NOTE**: FIXME: The content of this chapter has some been changed quite
+significantly since it was written.
+
 With the new solver we've made some fairly significant changes to normalization when compared
 to the existing implementation.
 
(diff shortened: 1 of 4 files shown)


BoxyUwU:
In some cases the FIXME's are themselves information, some extra bit of nuance that the page doesn't cover that the reader should probably know exists even if we haven't actually talked about it yet. Or potentially it indicates they should take the page with a bit of salt as its not been updated in a while or is confusing as written

src/solve/invariants.md
```diff
@@ -1,6 +1,8 @@
 # Invariants of the type system
 
+<!--
 FIXME: This file talks about invariants of the type system as a whole, not only the solver
+-->
```
BoxyUwU, line 5: for example I think there is value in having this be user-visible, there's a potential "flaw" in the current document that is probably good for readers to be aware of and have explicitly called out

src/solve/opaque-types.md
```diff
@@ -60,8 +60,11 @@ Finally, we check whether the item bounds of the opaque hold for the expected ty
 [eq-prev]: https://github.com/rust-lang/rust/blob/384d26fc7e3bdd7687cc17b2662b091f6017ec2a/compiler/rustc_trait_selection/src/solve/normalizes_to/opaque_types.rs#L51-L59
 [insert-storage]: https://github.com/rust-lang/rust/blob/384d26fc7e3bdd7687cc17b2662b091f6017ec2a/compiler/rustc_trait_selection/src/solve/normalizes_to/opaque_types.rs#L68
 [item-bounds-ck]: https://github.com/rust-lang/rust/blob/384d26fc7e3bdd7687cc17b2662b091f6017ec2a/compiler/rustc_trait_selection/src/solve/normalizes_to/opaque_types.rs#L69-L74
+
+<!--
 [^1]: FIXME: this should ideally only result in a unique candidate given that we require the args to be placeholders and regions are always inference vars
 [^2]: FIXME: why do we check whether the expected type is rigid for this.
+-->
```
BoxyUwU, line 67: There are footnotes linking to this which should also be removed if we're going to remove this.

src/solve/opaque-types.md
```diff
@@ -86,9 +89,11 @@ opaque before the first defining use of the opaque.
 
 These are not supported and trying to define them right now should always error.
 
+<!--
 FIXME: Because looking up opaque types in the opaque type storage can now unify regions,
 we have to eagerly check that the opaque types does not reference placeholders. We otherwise
 end up leaking placeholders.
+-->
```
BoxyUwU, line 96: I think its fine to leave this FIXME here, its a crumb of information in a heading that otherwise has ~nothing

src/solve/opaque-types.md
```diff
@@ -99,8 +104,11 @@ The handling of member constraints does not change in the new solver. See the
 
 ## calling methods on opaque types
 
+<!--
 FIXME: We need to continue to support calling methods on still unconstrained
 opaque types in their defining scope. It's unclear how to best do this.
+-->
```
BoxyUwU, line 110: I think a FIXME is nice here since otherwise it seems like this header was just intentionally left blank with little explanation

src/solve/opaque-types.md
onur-ozkan: In that case they should stay as they were. I didn’t realize the links above.

src/solve/normalization.md
```diff
@@ -52,12 +52,14 @@ before assigning the resulting rigid type to an inference variable. This is used
...
 and outside of the trait solver.
 
+<!--
 FIXME: structure, maybe we should have an "alias handling" chapter instead as
 talking about normalization without explaining that doesn't make too much
 sense.
 
 FIXME: it is likely that this will subtly change again by mostly moving structural
 normalization into `NormalizesTo`.
+-->
```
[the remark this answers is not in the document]
onur-ozkan: Sure, [done](https://github.com/rust-lang/rustc-dev-guide/pull/2186/commits/db5d166d7e03509711d107db4f4009f69c1c9288) that.

image

onur-ozkan:
> In some cases the FIXME's are themselves information, some extra bit of nuance that the page doesn't cover that the reader should probably know exists even if we haven't actually talked about it yet. Or potentially it indicates they should take the page with a bit of salt as its not been updated in a while or is confusing as written

Yeah I am aware of that, I left some of them visible but I wasn't entirely sure about others, that's why I wanted for a help in the PR description.

src/tests/directives.md
```diff
@@ -2,7 +2,9 @@
 
 <!-- toc -->
 
-> **FIXME(jieyouxu)** completely revise this chapter.
+<!--
+FIXME(jieyouxu) completely revise this chapter.
```
jieyouxu, line 6: FTR, this was intentional and I think it's more accurate to display this. But maybe elaborate with:

- This chapter is quite outdated at various places.
- The organization is confusing.
- It's potentially better to have per-directive doc comments then backlink to them for details, instead of having them separately listed here.

jieyouxu:
For dev-guide specifically, it's not stable or user-facing, but more like internal contributor docs. IMO having contributor-facing FIXMEs are actually good because it conveys info about the page (esp. if it indicates the page is outdated or needs to be changed in some way).

onur-ozkan:
I honestly think that some FIXMEs shouldn't appear in the UI as they don't really make sense at all... Also, those pages often visited by newbies not just for internal use.

src/tests/directives.md
onur-ozkan: It even includes a specific username to be resolved from. Why should it appear on the UI? Maybe I don't understand, what value does it bring to that page?

jieyouxu:
> I honestly think that some FIXMEs shouldn't appear in the UI as they don't really make sense at all... Also, those pages often visited by newbies not just for internal use.

Makes sense, we should probably convert those FIXMEs to actual issues in this repo anyway.

src/tests/directives.md
jieyouxu: Makes sense, I opened #2187 to track this instead.

jieyouxu (approved):
Thanks!

PR #2191, opened 2025-01-03
Describe how to use rust-analyzer with `rmake.rs`

Zalathar:
This documents a local workaround that I found for the default lack of rust-analyzer integration for `rmake.rs` scripts.

(Actually, my original workaround was more complicated, but writing this PR prompted me to find the simpler version described here.)


jieyouxu:
Thanks! Ideally it'd be possible to have this by default, but it gets annoying if you change or add new tests so...

PR #2192, opened 2025-01-03
Split stuff out of representing types, and rewrite early/late bound chapter

BoxyUwU:
title. it's not the best overall, I think the binders chapter could do with some work in order for it to really "fit" in its new position but I didn't feel like rewriting even *more* of the moved chapters...


jieyouxu (approved):
Thanks, we can always continue to update/improve docs

PR #2199, opened 2025-01-07
Fix rib example

sagudev:
The way ribs were presented before one would assume `helper` would be available inside closure, but it's not:

```rust
let helper = || {
        helper(); // helper not found in this scope
};
```

that's because ribs for let are introduced at the end of let statement and not while in expresions.

related zulip thread: https://rust-lang.zulipchat.com/#narrow/channel/182449-t-compiler.2Fhelp/topic/Ribs.20in.20name.20resolution

It is possible that my understanding is wrong, so please correct me if I am wrong.

### DIFF
--- a/src/name-resolution.md
+++ b/src/name-resolution.md
@@ -120,9 +120,9 @@ even though they should be visible by ordinary scoping rules. An example:
 fn do_something<T: Default>(val: T) { // <- New rib in both types and values (1)
     // `val` is accessible, as is the helper function
     // `T` is accessible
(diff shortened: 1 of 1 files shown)


camelid (commented):
This seems reasonable, but it'd be good to get a review from someone more experienced with the resolver. cc @petrochenkov maybe? (feel free to "reassign" of course)

jieyouxu:
Ping @petrochenkov, if you have time, could you kindly double-check?

PR #2202, opened 2025-01-08
Add a CI workflow for performing rustc-pull automatically

Kobzol:
The CI workflow is configured to run weekly on Monday morning, and it can also be executed manually through `workflow_dispatch`. The workflow runs `rustc-pull`, and if it actually pulled something, it force pushes to the `rustc-pull` branch, which is a sticky branch that should only be used for rustc pull updates.

If there is an open PR created by the Github Actions bot, it will be updated by that push. If not, the workflow creates a new PR. This PR also opts this repo into the auto open/close PR functionality from triagebot, so that CI runs on pull requests opened by the GitHub Actions bot user.

The PR also adds explicit error messages if there is nothing to pull during `rustc-pull`.

[This](https://github.com/Kobzol/rustc-dev-guide/pull/11) is how such a PR looks like.

### DIFF
--- a/.github/workflows/rustc-pull.yml
+++ b/.github/workflows/rustc-pull.yml
@@ -1,0 +1,55 @@
+name: rustc-pull
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Run at 04:00 UTC every Monday
+    - cron: '0 4 * * 1'
+
+jobs:
+  pull:
+    if: github.repository == 'rust-lang/rustc-dev-guide'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          # We need the full history for josh to work
+          fetch-depth: '0'
+      - name: Install stable Rust toolchain
+        run: rustup update stable
+      - uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: "josh-sync"
+          # Cache the josh directory with checked out rustc
+          cache-directories: "/home/runner/.cache/rustc-dev-guide-josh"
+      - name: Install josh
+        run: RUSTFLAGS="--cap-lints warn" cargo +stable install josh-proxy --git https://github.com/josh-project/josh --tag r24.10.04
+      - name: Setup bot git name and email
+        run: |
+          git config --global user.name 'The rustc-dev-guide Cronjob Bot'
+          git config --global user.email 'github-actions@github.com'
+      - name: Perform rustc-pull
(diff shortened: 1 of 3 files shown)


.github/workflows/rustc-pull.yml
```diff
@@ -0,0 +1,55 @@
...
+        with:
+          workspaces: "josh-sync"
+          # Cache the josh directory with checked out rustc
+          cache-directories: "/home/runner/.cache/rustc-dev-guide-josh"
+      - name: Install josh
+        run: RUSTFLAGS="--cap-lints warn" cargo +stable install josh-proxy --git https://github.com/josh-project/josh --tag r24.10.04
+      - name: Setup bot git name and email
+        run: |
+          git config --global user.name 'The rustc-dev-guide Cronjob Bot'
+          git config --global user.email 'rustc-dev-guide@cron.bot'
```
camelid, line 33: Maybe best not to use an email that we don't control.

Kobzol: I copied this from [miri](https://github.com/rust-lang/miri/blob/master/.github/workflows/ci.yml#L133), it has been using a similar e-mail address for a similar workflow for a long time. Do you see any attack vector? It's just a string that gets appended to the commit, it doesn't really mean anything.

camelid: I don't see an obvious one, but to me it seems like good practice to avoid using addresses from unrelated domains. Perhaps you could use `@example.com` instead? At least then it's a known fake domain.

camelid:
I'm not sure if I'm the best person to review the overall PR, though, because I don't have experience with josh and our josh workflow.

jieyouxu:
cc @BoxyUwU (I'll try to take a look, but I can't say I'm super familiar either lol)

BoxyUwU:
I don't really know how github CI files work but +1 to doing this and the rustc pull command seems like what I do locally so thats good :3

.github/workflows/rustc-pull.yml
JohnTitor: Or we could use `github-actions@github.com` which is widely used in GHA automation (like [this](https://github.com/rust-lang/rust/blob/7648efaf1a26aaec38d3babfdd26df1bc8648222/.github/workflows/dependencies.yml#L126-L127)).

Kobzol: Good point, changed it to the `github.com` address.

jieyouxu (approved):
I can't confirm this 100% works, but I looked at the workflow definition, and it seems reasonable to me. We'll have to actually use this in the real environment to see if we run into any hiccups in action (heh), so let's merge this and see.

PR #2205, opened 2025-01-10
Document how to find the configuration used in CI

ehuss:
This documents how to determine which settings are used in CI, since I see this question come up regularly. We currently don't have a great way to answer the question, but at least there is something.


jieyouxu (approved):
Thanks

PR #2207, opened 2025-01-11
update outdated LLVM stamp filename

onur-ozkan:
Follow-up for https://github.com/rust-lang/rust/pull/135281


jieyouxu:
@onur-ozkan btw, since now that r-d-g  is a josh subtree in r-l/r, u can also update r-d-g in the same r-l/r PR!

PR #2218, opened 2025-01-22
Run schedule CI more often due to link cache

marxin:
This attempt runs CI job more often so that the chunk of expired links to github.com is relatively small and so it's going to prevent: `429 Too Many Requests` error.

CC: @camelid @Kobzol

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,8 +6,8 @@ on:
       - master
   pull_request:
   schedule:
-    # Run at 18:00 UTC every day
-    - cron: '0 18 * * *'
+    # Run 4 times a day so that the link expiration turn over is relatively small
+    - cron: '0 */6 * * *'
 
 jobs:
   ci:


camelid:
Thanks! I wonder if it would be better to just make the cache last longer instead though. Any thoughts on that? On a related note, do you know if the cache also includes links that were in the previous cache?

marxin:
Oh, I finally understand what's wrong with the cache reuse:

```
Failed to save: Unable to reserve cache with key linkcheck--0.8.1, another job may be creating this cache. More details: Cache already exists. Scope: refs/heads/master, Key: linkcheck--0.8.1, Version: 33f8fd511bed9c91c40778bc5c27cb58425caa894ab50f9c5705d83cb78660e0
Warning: Cache save failed.
```

 Let me prepare a separate PR, where I am going to fix it.

marxin:
Fix: #2223

PR #2229, opened 2025-01-28
Add some extra pointers for rustdoc frontend devs

notriddle:
- https://github.com/rust-lang/rust/pull/135353
- https://github.com/rust-lang/rust/pull/136161


jieyouxu (approved):
Thanks! This can always be further modified, so.

PR #2234, opened 2025-01-30
Make rustc pulls on CI more frequent

Kobzol:
This PR runs the `rustc-pull` workflow every day, and changes the CI script so that it only sends a message to Zulip when something was actually pulled (regardless of whether an existing PR was updated or a new one was created).


Kobzol:
This will be best tested by simply running it :)

PR #2235, opened 2025-01-30
Rewrite section on executing Docker tests

Kobzol:
Suggested in https://github.com/rust-lang/rust/issues/136249#issuecomment-2624790246.

CC @jieyouxu


src/tests/docker.md
```diff
@@ -1,36 +1,44 @@
...
+
+You can run `src/ci/docker/run.sh <image-name>` directly. A few important notes regarding the `run.sh` script:
+- There is some configuration used on CI that you may need to recreate. In particular, set `submodules = false` in your `config.toml` so that it doesn't attempt to modify the read-only directory.
+- `<image-name>` corresponds to a single directory located in one of the `src/ci/docker/host-*` directories. Note that image name does not necessarily correspond to a job name, as some jobs execute the same image, but with different environment variables or Docker build arguments (this is a part of the complexity that makes it difficult to run CI jobs locally).
+- If you are executing a "dist" job (job beginning with `dist-`), you should set the `DEPLOY=1` environment variable.
```
onur-ozkan, line 26: If this is mandatory, why aren't we doing it automatically?

src/tests/docker.md
```diff
@@ -1,36 +1,44 @@
...
 
-Some additional notes about using the Docker images:
+If the above script does not work for you, you would like to have more control of the Docker image execution, or you want to understand what exactly happens during Docker job execution, then continue reading below.
 
+## The `run.sh` script
+The [`src/ci/docker/run.sh`] script is used to build a specific Docker image, run it,
+build Rust within the image, and either run tests or prepare a set of archives designed for distribution. The script will mount your local Rust source tree in read-only mode, and an `obj` directory in read-write mode. All the compiler artifacts will be stored in the `obj` directory. The shell will start out in the `obj`directory. From there, it will execute `../src/ci/run.sh` which starts the build as defined by the Docker image.
+
+You can run `src/ci/docker/run.sh <image-name>` directly. A few important notes regarding the `run.sh` script:
+- There is some configuration used on CI that you may need to recreate. In particular, set `submodules = false` in your `config.toml` so that it doesn't attempt to modify the read-only directory.
```
onur-ozkan, line 24: If we are running a docker image, it already sets `submodules = false` if I am not mistaken.

We should instead say "make sure submodules are fetched with `git ...`" I think.

src/tests/docker.md
Kobzol: Well, we are doing it automatically, on CI with a bash script executed in GitHub actions, and locally through the `ci.py` script. `run.sh` should really be seen as an "implementation detail", and users who want to run a given workflow locally on Linux should use `ci.py` instead.

src/tests/docker.md
Kobzol: This always confuses me, I could swear that the Docker build sometimes took data from my root config.toml :) But it seems like it doesn't happen, and yeah, the build expects submodules to be downloaded, because that's what happens on CI.

I reworded the paragraph, let me know what you think.

onur-ozkan (approved):

PR #2241, opened 2025-02-03
overlong line

tshepang:
Also, this follows the common (in this repo) practice of avoiding use of inline links


jieyouxu (approved):
I think we don't really have any line length guides for r-d-g, but at least this is self-consistent within the file so

PR #2246, opened 2025-02-06
Run CI multiple times a day

marxin:
Right now, the cache is properly loaded and updated during each run of the CI. However, we still hit `429`s every time the big portion of cache entries (pointing to GH) are invalidated. Plus, sometimes there are volatile links that tend to fail occasionally. Thus I recommend running the CI multiple times a day, it would run fast (only invalidated cache entries + failing will be checked).

@camelid What do you think?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,8 +6,8 @@ on:
       - master
   pull_request:
   schedule:
-    # Run at 18:00 UTC every day
-    - cron: '0 18 * * *'
+    # Run multiple times a day as the successfull cached links are not checked every time.
+    - cron: '0 */3 * * *'
 
 jobs:
   ci:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ on:
   pull_request:
   schedule:
     # Run multiple times a day as the successfull cached links are not checked every time.
-    - cron: '0 */3 * * *'
+    - cron: '0 */8 * * *'
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -6,8 +6,8 @@ on:
       - master
   pull_request:
   schedule:
-    # Run at 18:00 UTC every day
-    - cron: '0 18 * * *'
+    # Run multiple times a day as the successfull cached links are not checked every time.
+    - cron: '0 */4 * * *'
```
camelid, line 10: Definitely willing to try running more than once per day, but I feel like 2 or 3 times per day should be enough. This will run six times a day, which seems excessive to me.

marxin: Sure, let's run it then 3x a day.

.github/workflows/ci.yml
```diff
@@ -6,8 +6,8 @@ on:
       - master
   pull_request:
   schedule:
-    # Run at 18:00 UTC every day
-    - cron: '0 18 * * *'
+    # Run multiple times a day as the successfull cached links are not checked every time.
+    - cron: '0 */3 * * *'
```
camelid, line 10: I think you wrote the cron schedule wrong. This will run it every 3rd hour, i.e. 8x per day. https://crontab.guru/#0_*/3_*_*_*

marxin: Yeah, sorry ,fixed.

camelid (approved):
Thanks!

PR #2254, opened 2025-02-14
Start using latest release where -f checks all local links

marxin:
CC: @camelid

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       MDBOOK_VERSION: 0.4.21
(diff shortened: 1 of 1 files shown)


camelid (approved):
Thanks!

camelid:
FYI, the new description for the `-f` flag isn't the clearest imho:

> Check web links only the given files (check all files if omitted).

I wouldn't know what this means if we hadn't discussed it before. Might be better to say something like:

> Limit checking of web links to only the given files (check all files if omitted). Internal links to other Markdown files are always checked in all files, even when this flag is passed.

marxin:
Sure, very good point, fixed in the next release: #2257.

PR #2255, opened 2025-02-15
Rename config.toml to bootstrap.toml

Shourya742:
Currently, both Bootstrap and Cargo uses same name as their configuration file, which can be confusing. This PR is based on a discussion to rename `config.toml` to `bootstrap.toml` for Bootstrap. It is part of issue https://github.com/rust-lang/rust/issues/126875.


jieyouxu:
@Shourya742 FWIW, you can coalesce this change into the r-l/r PR because they certainly make sense together. Usually we would ask to file the change separately if the r-d-g changes isn't accompanied by a r-l/r side change as well.

PR #2256, opened 2025-02-15
Fix examples to work with nightly-2025-02-13

torfsen:
While there were comments indicating which nightly versions the examples were tested with, those versions did not work for me: neither did the examples compile, nor did they produce the expected output.

This commit fixes the compilation issues, using nightly-2025-02-13 for all examples (previously the version differed between the examples) and, in the case of the `rustc_driver` examples, also fixes the argument passing: rustc ignores the first argument, so we need to pass the filename as the second (otherwise we only get the help text printed).

Note that the `rustc-interface-getting-diagnostics.rs` example still does not produce any output, which I assume is not how it is intended. However, I don't know enough to fix it. I'd be happy to include a fix if somebody can point me towards it.

To avoid inconsistencies between the documented version and the actually required version I've moved the version comment from the Markdown into the Rust code where it hopefully won't be forgotten as easily.

Finally I've clarified in the examples' README that you also need to use the proper nightly version when compiling the examples, not just when running them.

### DIFF
--- a/examples/README
+++ b/examples/README
@@ -4,7 +4,10 @@ For each example to compile, you will need to first run the following:
 
 To create an executable:
 
-    rustc rustc-driver-example.rs
+    rustup run nightly rustc rustc-driver-example.rs
+
+You might need to be more specific about the exact nightly version. See the comments at the top of
+the examples for the version they were written for.
 
 To run an executable:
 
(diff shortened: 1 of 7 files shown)


jieyouxu (approved):
Thanks

PR #2258, opened 2025-02-20
Rewrite effects checking chapter

fee1-dead:
[rendered](https://github.com/fee1-dead-contrib/rustc-dev-guide/blob/constck/src/effects.md)

cc @rust-lang/project-const-traits

### DIFF
--- a/src/effects.md
+++ b/src/effects.md
@@ -1,66 +1,130 @@
-# Effects and effect checking
+# Effects and const condition checking
 
-Note: all of this describes the implementation of the unstable `effects` and
-`const_trait_impl` features. None of this implementation is usable or visible from
-stable Rust.
+## The `HostEffect` predicate
 
-The implementation of const traits and `~const` bounds is a limited effect system.
-It is used to allow trait bounds on `const fn` to be used within the `const fn` for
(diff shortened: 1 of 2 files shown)


src/effects.md
```diff
@@ -1,66 +1,130 @@
...
+[`Checker::revalidate_conditional_constness`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_const_eval/check_consts/check/struct.Checker.html#method.revalidate_conditional_constness
+
+## Proving `HostEffectPredicate`s
+
+`HostEffectPredicate`s are implemented both in the [old solver] and the [new
+trait solver]. In general, we can prove a `HostEffect` predicate when either of
+these conditions are met:
+
+* The predicate can be assumed from caller bounds;
+* The type has a `const` `impl` for the trait, *and* that const conditions on
```
compiler-errors, line 122: You should also mention the implied const bounds, which are like item bounds but only hold if an associated type or RPIT's const conditions hold.

oli-obk (approved):

src/effects.md
fee1-dead: Done.

PR #2264, opened 2025-02-27
chore: cc-rs MUST be forked for new targets

nhtyy:


ChrisDenton:
This is changing. A PR to fix it is currently being drafted https://github.com/rust-lang/cc-rs/pull/1413

nhtyy:
I see... I will close this then.

PR #2266, opened 2025-02-28
Introduce a normalization chapter

BoxyUwU:
I kind of just asborbed the new solver normalization chapter into this one because it was super out of date and seemed better to move what bits of information it had that were still correct somewhere alongside all the other norm information.

There's some lost information around the exactly impl details of the new solver :thinking: In general I would like to cover more impl details in this chapter butr I think this is a sufficiently large first step :laughing:


src/SUMMARY.md
```diff
@@ -144,6 +144,7 @@
     - [ADTs and Generic Arguments](./ty_module/generic_arguments.md)
     - [Parameter types/consts/regions](./ty_module/param_ty_const_regions.md)
 - [`TypeFolder` and `TypeFoldable`](./ty-fold.md)
+- [Normalization and Aliases](./normalization/normalization.md)
```
tshepang, line 147: maybe we should not nest it this deep, unless the plan is to add other files under normalization/ dir

src/normalization.md
```diff
@@ -0,0 +1,310 @@
...
+Leaving the alias unnormalized would also be wrong as the old solver expects all aliases to be rigid. This was a soundness bug before the new solver was stabilized in coherence: [relating projection substs is unsound during coherence](https://github.com/rust-lang/rust/issues/102048).
+
+Ultimately this means that it is not always possible to ensure all aliases inside of a value are rigid.
+
+[universes]: https://rustc-dev-guide.rust-lang.org/borrow_check/region_inference/placeholders_and_universes.html#what-is-a-universe
+[deeply_normalize]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_trait_selection/traits/normalize/trait.NormalizeExt.html#tymethod.deeply_normalize
+
+## Handling uses of diverging aliases
+
+> NOTE: This section should talk about how we handles uses of diverging aliases. writeback normalizes everything explicitly, lots of places are adhoc and accidentally error or don't manage to error even though they ought to, const aliases are a bit more "principled" about checking for termination.
```
BoxyUwU, line 310: Imagine learning that writeback does not infact normalize everything explicitly. what a concept

src/normalization.md
```diff
@@ -0,0 +1,314 @@
...
+    let _: <T as Iterator>::Item;
+}
+
+fn bar<T: Iterator<Item = u32>>() {
+    // This alias is *not* rigid as it can be normalized to `u32`
+    let _: <T as Iterator>::Item;
+}
+```
+
+This definition also implies that illformed aliases can be considered rigid. For example the type `<u32 as Iterator>::Item` is a rigid alias as it can never be normalized as there is no `u32: Iterator` impl (ignoring `feature(trivial_bounds)`).
```
lcnr, line 35: that doesn't seem right. `<u32 as Iterator>::Item` is not rigid in the new solver and iirc the old solver also uses `normalizes_to_error` in these cases https://github.com/rust-lang/rust/blob/efcbb94dccbe0fcc1076f031f88332db4a6a2a52/compiler/rustc_trait_selection/src/traits/project.rs#L459

src/normalization.md
```diff
@@ -0,0 +1,314 @@
...
+```rust
+fn foo<T: Iterator, U: Iterator>() {
+    // This alias is considered to be "ambiguous"
+    let _: <_ as Iterator>::Item;
+}
+```
+
+The reason we call them "ambiguous" aliases is because its *ambiguous* whether this is a rigid alias or not.
+
+The source of the `_: Iterator` trait impl is *ambiguous* (i.e. unknown), it could be some `impl Iterator for u32` or it could be some `T: Iterator` trait bound, we don't know yet. Depending on how `_: Iterator` holds the alias could be an unnormalized alias or it could be a rigid alias; it's *ambiguous* what kind of alias this is.
```
lcnr, line 47:
```suggestion
The source of the `_: Iterator` trait impl is *ambiguous* (i.e. unknown), it could be some `impl Iterator for u32` or it could be some `T: Iterator` trait bound, we don't know yet. Depending on why `_: Iterator` holds the alias could be an unnormalized alias or it could be a rigid alias; it's *ambiguous* what kind of alias this is.
```

src/normalization.md
```diff
@@ -0,0 +1,314 @@
...
+
+Finally, an alias can just be unnormalized, `<Vec<u32> as IntoIterator>::Iter` is an unnormalized alias as it can already be normalized to `std::vec::IntoIter<u32>`, it just hasn't been done yet.
+
+---
+
+It is worth noting that Free and Inherent aliases cannot be rigid or ambiguous as   naming them also implies having resolved the definition of the alias, which specifies the underlying type of the alias:
```
lcnr, line 53:
```suggestion
It is worth noting that Free and Inherent aliases cannot be rigid or ambiguous as  naming them also implies having resolved the definition of the alias, which specifies the underlying type of the alias:
```

src/normalization.md
```diff
@@ -0,0 +1,314 @@
...
+    // `ret_alias` returns a value of type `Foo<?x>::Inherent`
+    t = ret_alias(t);
+    
+    // we have equated `?x` with `Foo<?x>::Inherent` this can
+    // only succeed if we're able to normalize inherent aliases
+    // with ambiguous where clauses on the impl.
+    //
+    // Or, put another way, it is *not* an ambiguous alias.
+}
+```
```
lcnr, line 84: What are you trying to show with this example? It feels confusing to me (and too "large" for showing that inherent aliases are always normalizeable)

src/normalization.md
```diff
@@ -0,0 +1,314 @@
...
+    type DivergingAssoc<U: Trait>;
+}
+impl<T: ?Sized> Trait for T {
+    // This alias always diverges but we don't emit an error because
+    // the compiler can't "see" that.
+    type DivergingAssoc<U: Trait> = <U as Trait>::DivergingAssoc<U>;
+}
+```
+
+Ultimately this means that we have no guarantee that aliases in the type system terminate. As aliases may only diverge for some specific arguments to generic parameters, it also means that only fully concrete aliases can be checked to terminate, which means that codegen/const evaluation is the only time that aliases can reliably be checked to terminate:
```
lcnr, line 113:
```suggestion
Ultimately this means that we have no guarantee that aliases in the type system don't diverge. As aliases may only diverge for some specific arguments to generic parameters, it also means that we only know whether an alias diverges once it is fully concrete. This means that codegen/const evaluation also has to handle diverging aliases can reliably be checked to terminate:
```
I personally dislike the perspective "only codegen can check". It's more that "codegen has to check for them *as well*"

src/normalization.md
```diff
@@ -0,0 +1,314 @@
...
+
+## Normalization
+
+There are two main "parts" to how normalization exists in the compiler:
+- The core logic responsible for turning an alias into its underlying type
+- The various entry points to the core normalization logic that expose real APIs
+
+### Core normalization logic
+
+Normalizing aliases is a little bit more nuanced than replacing the alias with whatever it was defines as being equal to in its definition; the result of normalizing an alias should either be a rigid type or an inference variable (which will later be inferred to a rigid type). To accomplish this we do two things:
```
lcnr, line 167:
```suggestion
Normalizing aliases is a little bit more nuanced than replacing the alias with whatever it is defined as being equal to in its definition; the result of normalizing an alias should either be a rigid type or an inference variable (which will later be inferred to a rigid type). To accomplish this we do two things:
```

src/normalization.md
```diff
@@ -0,0 +1,314 @@
...
+
+First, when normalizing an ambiguous alias it is normalized to an inference variable instead of leaving it as-is, this has two main effects: 
+- Even though an inference variable is not a rigid type, it will always wind up inferred *to* a rigid type so we ensure that the result of normalization will not need to be normalized again
```
lcnr, line 170: shallow vs deep normalization. idk how to best introduce this, but it only guarantees that the "outer type constructor" is rigid.

you could normalize to `fn(<&() as Trait>::Assoc>)` after all which contains a normalizeable alias.

I think before explaining *how* normalization works, it may make sense to introduce "structural normalization" and "deep normalization"

src/normalization.md
```diff
@@ -0,0 +1,314 @@
...
+- `deeply_normalize`: Folds over the provided value normalizing all aliases, erroring on ambiguous aliases[^1].
+- `structurally_normalize`: Normalizes an alias to a rigid type or an inference variable, does nothing if the type is not an alias.
+
+When the new solver is stabilized the `infcx.at.normalize` function will be removed and everything will have been migrated to either deeply normalize or structurally normalize. In cases where normalization is required in both old and new solvers, the `structurally_normalize` function is often the right choice. Otherwise, if normalization is only needed in the old solver then `infcx.at.normalize` is often the right choice.
+
+`structurally_normalize` should be thought of as only normalizing the "outer layer" of a type to a rigid type or inference variable. Any types inside of the normalized form may still be unnormalized (even if an alias *was* actually normalized[^2]). `structurally_normalize` is named such as it is typically called before looking at the "structure" of a type (e.g. by matching on `TyKind`/`ConstKind`).
```
lcnr, line 223: i think you probably should move this up without yet talking about impl details.

shallow vs deep normalization is an important distinction

src/normalization.md
BoxyUwU: oh neat okay :3

src/normalization.md
```diff
@@ -0,0 +1,314 @@
...
+[structurally_normalize]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_trait_selection/traits/trait.StructurallyNormalizeExt.html#tymethod.structurally_normalize_ty
+[infcx]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_trait_selection/infer/struct.InferCtxt.html
+[fcx]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir_typeck/fn_ctxt/struct.FnCtxt.html
+[ocx]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_trait_selection/traits/struct.ObligationCtxt.html
+[structurally_resolve]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir_typeck/fn_ctxt/struct.FnCtxt.html#method.structurally_resolve_type
+
+#### `tcx.normalize_erasing_regions`
+
+[`normalize_erasing_regions`][norm_erasing_regions] is generally used by parts of the compiler that are not doing type system analysis' as this normalization entry point does not handle inference variables, lifetimes, or any diagnostics. Lints and codegen make heavy use of this entry point as they typically are working with fully inferred aliases that can be assumed to be well formed (or at least, are not responsible for erroring on).
```
lcnr, line 245:
```suggestion
[`normalize_erasing_regions`][norm_erasing_regions] is generally used by parts of the compiler that are not doing type system analysis. This normalization entry point does not handle inference variables, lifetimes, or any diagnostics. Lints and codegen make heavy use of this entry point as they typically are working with fully inferred aliases that can be assumed to be well formed (or at least, are not responsible for erroring on). 
```

src/normalization.md
BoxyUwU: Yeah it feels like its missing a section specifically calling out shallow vs deep norm :thinking: It's awkward because the old solver kinda likes to pretend that normalization is deep, yet `deeply_normalize` exists and is arguably just as "deep" as shallow norm in the old solver.

Even in the new solver "shallow" normalization is largely deep norm with the exception of aliases in binders </3

src/normalization.md
```diff
@@ -0,0 +1,314 @@
...
+
+### Normalizing parameter environments
+
+Another problem was that it was not possible to normalize `ParamEnv`s correctly in the old solver as normalization itself would expect a normalized `ParamEnv` in order to give correct results. See the chapter on `ParamEnv`s for more information: [`Typing/ParamEnv`s: Normalizing all bounds](./typing_parameter_envs.md#normalizing-all-bounds)
+
+### Ambiguous aliases in higher ranked types
+
+Given a type such as `for<'a> fn(<?x as Trait<'a>::Assoc>)`, it is not possible to correctly handle this with the old solver's approach to normalization.
+
+If we were to normalize it to `for<'a> fn(?y)` and register a goal to normalize `for<'a> <?x as Trait<'a>>::Assoc -> ?y`, this would result in errors in cases where `<?x as Trait<'a>>::Assoc` normalized to `&'a u32`. This is because the inference variable `?y` would be in a lower [universe][universes] than the placeholders made when instantiating the `for<'a>` binder.
```
lcnr, line 299:
```suggestion
If we were to normalize it to `for<'a> fn(?y)` and register a goal to normalize `for<'a> <?x as Trait<'a>>::Assoc -> ?y`, this would result in errors in cases where `<?x as Trait<'a>>::Assoc` normalized to `&'a u32`. The inference variable `?y` would be in a lower [universe][universes] than the placeholders made when instantiating the `for<'a>` binder.
```

src/normalization.md
BoxyUwU: Or well, deeply_normalize *is* arguably more deep in the sense that you never have to worry about encountering non-rigid aliases inside of the term anymore. But that feels so subtle given it only *really* matters in the edge case of hr aliases :sweat_smile:

src/normalization.md
BoxyUwU: Yeah idk removed it

src/normalization.md
BoxyUwU: I'm not really sure I understand the distinction :thinking: Codegen has to check for them because *only* codegen can check some aliases.

src/normalization.md
```diff
@@ -0,0 +1,314 @@
...
+    // only succeed if we're able to normalize inherent aliases
+    // with ambiguous where clauses on the impl.
+    //
+    // Or, put another way, it is *not* an ambiguous alias.
+}
+```
+
+### Diverging Aliases
+
+An alias is considered to "diverge" if its definition does not actually specify an underlying non-alias type to normalize to. On the flip side, we consider an alias to "terminate" if it does have an underlying value it can be normalized to (i.e. it does not diverge). A concrete example of diverging aliases:
```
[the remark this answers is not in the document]
BoxyUwU: Looking at some private convo chat logs I definitely see a lot of talking about "alias terminates", atleast enough to justify explaining the term here as people will probably encounter it in the wild.

src/normalization.md
BoxyUwU: I updated the wording anyway :thinking:

src/normalization.md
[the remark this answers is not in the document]
BoxyUwU: Thinking about it more, we only really use "terminates" terminology like twice, so it's probably More confusing to keep it here instead of using diverging everywhere.

jackh726:
@BoxyUwU I just read through this and I wanted to just let you know that this is awesome.

BoxyUwU:
Thanks :D

PR #2269, opened 2025-03-05
Don't suggest explicitly `cfg`-gating `trace!` calls in bootstrap

moxian:
I was trying to learn more about bootstrap, read about the tracing macros, thought they were cool, but also writing `#[cfg(tracing)] trace!` each time gets annoying.

So I thought I'd PR a shim macro that hides the `cfg` internally (and contribute to rustc finally),  but turns out @jieyouxu has already done that in https://github.com/rust-lang/rust/pull/136392

Well, guess I'll be content with just a doc change for now then 🙃


jieyouxu (approved):
Thanks

PR #2271, opened 2025-03-08
fix text

tshepang:
- There is more than just target and stage
- There is only 3 stages, so don't mention them specially


jieyouxu:
Thanks

PR #2279, opened 2025-03-10
those should not get shell highlighting

tshepang:
Also, add some copy paste convenience


tshepang:
did another pr that break things apart

PR #2284, opened 2025-03-13
Remove the doc for `no-system-llvm`

cuviper:
This compiletest directive was removed in rust-lang/rust#120265.


jieyouxu:
Thanks!

PR #2295, opened 2025-03-18
add new section on the `rustdoc` test suite

lolbinarycat:
Currently this stuff is only documented in a comment at the start of `htmldocck.py`, and that commend is very outdated and often missing important details, like the fact that `has` and `matches` are accept different kinds of patterns.


src/tests/compiletest.md
```diff
@@ -78,7 +78,7 @@ The following test suites are available, with links for more information:
 
 ### Rustdoc test suites
 
-See [Rustdoc tests](../rustdoc.md#tests) for more details.
+See [Rustdoc § Tests](../rustdoc.md#tests) for more details.
```
fmease, line 81: The rest of the dev guide doesn't use the section sign at all. It's a bit odd to deviate from the status quo

src/rustdoc-internals/htmldocck.md
```diff
@@ -0,0 +1,101 @@
...
+as if it were to be parsed by `eval`.
+(note that the list is actually parsed by `shlex.split`,
+so it cannot contain arbitrary python expressions).
+
+Example: `//@ files "foo/bar" '["index.html", "sidebar-items.js"]'`
+
+[^1]: Whitespace normalization means that all spans of consecutive whitespace are replaced with a single space.  The files themselves are also whitespace-normalized.
+
+## Limitations
+All `XPATH` arguments must start with `//` due to a flaw in the implemention.
```
fmease, line 98: Other limitation: This uses the XPath impl provided by the Python stdlib which — if I remember correctly — is far from up to date and generally only supports a subset of feature.

(We eventually want to rewrite HtmlDocCk to Rust (https://github.com/rust-lang/rust/pull/125780) which would allow us to have more control over the XPath impl.)

src/rustdoc-internals/htmldocck.md
```diff
@@ -0,0 +1,101 @@
+# `rustdoc` tests
+
+This page is specifically about the `rustdoc` test suite, for other test suites used for testing rustdoc, see [Rustdoc § Tests](../rustdoc.md#tests).
+
+`htmldocck.py` is a custom checker script that uses [XPath] to verify the HTML output of rustdoc.
+
+[XPath]: https://en.wikipedia.org/wiki/XPath
+
+## Directives
+Directives to htmldocck are similar to those given to `compiletest` in that they take the form of `//@` comments.
```
fmease, line 10: This paragraph doesn't make it clear that all regular compiletest directives are supported, too.

src/rustdoc-internals/htmldocck.md
```diff
@@ -0,0 +1,101 @@
+# `rustdoc` tests
+
+This page is specifically about the `rustdoc` test suite, for other test suites used for testing rustdoc, see [Rustdoc § Tests](../rustdoc.md#tests).
```
fmease, line 3: While it's mentioned in `src/rustdoc.md` that `tests/rustdoc/` is specifically for testing the output of rustdoc's HTML backend; people arriving from `SUMMARY.md` won't necessarily know that given the (unfortunately) generic test suite name "`rustdoc`".

Therefore I'd mention that again here.

src/rustdoc-internals/htmldocck.md
```diff
@@ -0,0 +1,101 @@
+# `rustdoc` tests
```
fmease, line 1: It's slightly unusual to have section heading != file name.

src/rustdoc-internals/htmldocck.md
```diff
@@ -0,0 +1,101 @@
...
+
+`htmldocck.py` is a custom checker script that uses [XPath] to verify the HTML output of rustdoc.
+
+[XPath]: https://en.wikipedia.org/wiki/XPath
+
+## Directives
+Directives to htmldocck are similar to those given to `compiletest` in that they take the form of `//@` comments.
+
+All `PATH`s in directives are relative to the the rustdoc output directory (`build/TARGET/test/rustdoc/TESTNAME`),
+so it is conventional to use a `#![crate_name = "foo"]` attribute to avoid writing paths.
```
fmease, line 13: That's not really the reason? In my understanding people use that if the crate name is overly long since "TESTNAME" is just the crate name as derived from the file name (if no `#![crate_name]` is present).

src/rustdoc-internals/htmldocck.md
```diff
@@ -0,0 +1,101 @@
...
+To avoid repetion, `-` can be used in any `PATH` argument to re-use the previous `PATH` argument.
+
+All arguments take the form of quoted strings,
```
fmease, line 16: You might want to mention that both single quotes as well as double quotes are supported.

fmease (commented):
Thanks for working on this. I have a couple of comments!

src/rustdoc-internals/htmldocck.md
fmease: (Not by HtmlDocCk directly but by compiletest since the rustdoc/ test suite is both checked by compiletest *and* HtmlDocCk by virtue of compiletest invoking/driving HtmlDocCk)

src/tests/compiletest.md
lolbinarycat: How else should I differentiate between "The page titled 'rustdoc tests'" and "The 'tests' section of the 'rustdoc' page"?

src/rustdoc-internals/htmldocck.md
fmease, line 1: Coming back to the file name, I don't quite think this should be named after HtmlDocCk (htmldoc check**er**) but simply after the test suite `rustdoc/` (maybe `rustdoc-test-suite.md`).

In my eyes, HtmlDocCk is just an impl detail of the `rustdoc/` test suite (similiar to what JsonDocCk is to `rustdoc-json/`). The actual interface is the set of directives which is a union of compiletest and htmldocck ones as previously mentioned.

Ofc, we should still mention HtmlDocCk and `src/etc/htmldocck.py` somewhere on this page.

fmease: (Ideally `rustdoc/` would be called `rustdoc-html/`)

src/rustdoc-internals/htmldocck.md
lolbinarycat: I did this to avoid using a filepath we might want to use later for a general overview of all `rustdoc*` test suites.

while moving files can be done smoothly with redirects, swapping them will always lead to confusing.

src/rustdoc-internals/htmldocck.md
lolbinarycat: That's basically what I meant to say, but I seem to have jumbled my words a bit.

lolbinarycat:
@fmease I think I've managed to address all your comments in a way I'm satisfied with, does this look good?

notriddle (approved):

jieyouxu:
Merging since rustdoc maintainers approved, can always be improved in follow-ups.

jieyouxu:
(https://github.com/rust-lang/team/pull/1704 will allow rustdoc maintainers to have write access to this repo, since it makes no sense if the rustdoc experts can't merge rustdoc doc changes.)

fmease:
Noo, my review wasn't done :/ I'll create a follow-uo

src/rustdoc-internals/rustdoc-test-suite.md
```diff
@@ -0,0 +1,111 @@
...
+
+Usage: `//@ matchesraw PATH PATTERN`
+
+Checks that the contents of the file `PATH` matches the regex `PATTERN`.
+
+### `hasraw`
+
+Usage: `//@ hasraw PATH PATTERN`
+
+Same as `matchesraw`, except `PATTERN` is a whitespace-normalized[^1] string instead of a regex.
```
fmease, line 63: ditto

jieyouxu:
Oh sorry :(

src/rustdoc-internals/rustdoc-test-suite.md
```diff
@@ -0,0 +1,111 @@
...
+
+### `has`
+
+Usage 1: `//@ has PATH`
+Usage 2: `//@ has PATH XPATH PATTERN`
+
+In the first form, `has` checks that a given file exists.
+
+In the second form, `has` is an alias for `matches`,
+except `PATTERN` is a whitespace-normalized[^1] string instead of a regex.
```
fmease, line 45: This only mentions in the footnote that file's contents are also whitespace normalized. However, that's arguably more important (in 99% cases the PATTERN the test author writes is normalized already). This should be the other way around.

jieyouxu:
> Noo, my review wasn't done (this still has several smallish issues) :/ I'll create a follow-up PR

I can address some of the stuff you mentioned unless you already have changes locally, I'll also add a bit more on how this is wired up between bootstrap <-> rustdoc test suites.

(Sorry)

src/rustdoc-internals/rustdoc-test-suite.md
fmease, line 1: https://rustc-dev-guide.rust-lang.org/rustdoc-internals.html#dotting-is-and-crossing-ts already mentions htmldocck albeit more briefly. Furthermore it mentions useful compiletest directives like `//@ build-aux-docs`.

We should remove this section and import some of the additional pieces of information into this document. We should also mention compiletest directive `//@ doc-flags`.

fmease:
I already have my editor open and will open a PR in <30 minutes I guess. Feel free to add extra information after that :)

src/rustdoc-internals/rustdoc-test-suite.md
```diff
@@ -0,0 +1,111 @@
...
+[XPath]: https://en.wikipedia.org/wiki/XPath
+
+## Directives
+Directives to htmldocck are similar to those given to `compiletest` in that they take the form of `//@` comments.
+
+In addition to the directives listed here,
+`rustdoc` tests also support most
+[compiletest directives](../tests/directives.html).
+
+All `PATH`s in directives are relative to the the rustdoc output directory (`build/TARGET/test/rustdoc/TESTNAME`),
```
fmease, line 18: Unless I'm confusing myself, if you say the path is relative to `build/$TARGET/test/rustdoc/$CRATE_NAME` I would think a path like `foo/fn.bar.html` would expand to `build/$TARGET/test/rustdoc/foo/foo/fn.bar.html` assuming the crate name is `foo` (notice the double `foo/`).

I would describe it as relative to `build/$TARGET/test/rustdoc/` in which case this sentence merely describes an implementation detail of compiletest which the test writer shouldn't need to worry about.

jieyouxu: Is `tests/rustdoc` supposed to support revisions?

fmease: They're currently rejected by compiletest (via an assert) since htmldocck.py has no notion of  that

PR #2296, opened 2025-03-19
Don't linkcheck external web links in PR CI

jieyouxu:
This can block PR CI yet remain unactionable for the contributor, since this is subject to network conditions and availability of the servers hosting the external web links.

~~I don't see a way for PR CI to override the linkcheck config to skip external web links yet let the cron job CI to perform external web link checks, so this PR does the sledgehammer approach of disabling external web link checks for both.~~ EDIT: apparently `MDBOOK_OUTPUT__LINKCHECK__FOLLOW_WEB_LINKS` is a thing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,7 @@ jobs:
       MDBOOK_LINKCHECK2_VERSION: 0.9.1
       MDBOOK_MERMAID_VERSION: 0.12.6
       MDBOOK_TOC_VERSION: 0.11.2
+      MDBOOK_OUTPUT__LINKCHECK__FOLLOW_WEB_LINKS: ${{ github.event_name != 'pull_request' }}
       DEPLOY_DIR: book/html
       BASE_SHA: ${{ github.event.pull_request.base.sha }}
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


book.toml
```diff
@@ -27,31 +27,11 @@ level = 0
...
-    "github\\.com/issues\\?q=.*",
-    # Similarly 500 is sometimes returned here.
-    "github\\.com/rust-lang/rust/pulls\\?q=.*",
-    # 401 is returned here for unknown reason
-    "github\\.com/wesleywiser/rustc-bootstrap-wpa-analysis",
-    # Handle: connection closed before message completed
-    "microsoft\\.com/en-us/research/publication/",
-    # 500 is returned for HEAD request
-    "code\\.visualstudio\\.com/docs/editor/tasks",
-]
```
jieyouxu, line 54: This long list of excludes to me suggests that it's not good in terms of signal-to-noise ratio, because we're already skipping the (arguably most import) github links (issues, PRs, commits, etc.).

Kobzol:
We can set `MDBOOK_OUTPUT__LINKCHECK__FOLLOW_WEB_LINKS=false` in the non-cron job to disable external web link checks.

ci/linkcheck.sh
```diff
@@ -42,7 +46,7 @@ fi
 echo "exec $LINKCHECK_BINARY $FLAGS"
 if [ "$USE_TOKEN" = 1 ]; then
   config=$(set_github_token)
-  exec $LINKCHECK_BINARY $FLAGS <<<"$config"
+  exec env MDBOOK_OUTPUT__LINKCHECK__FOLLOW_WEB_LINKS=$FOLLOW_WEB_LINKS $LINKCHECK_BINARY $FLAGS <<<"$config"
 else
-  exec $LINKCHECK_BINARY $FLAGS
+  exec env MDBOOK_OUTPUT__LINKCHECK__FOLLOW_WEB_LINKS=$FOLLOW_WEB_LINKS $LINKCHECK_BINARY $FLAGS
```
jieyouxu, line 51: I honestly don't know if this shell incantation is correct... Does this look right to you?

jieyouxu:
Actually, I'll skip for PR CI only for now.

ci/linkcheck.sh
Kobzol: It has to be true/false, not 0/1, otherwise it fails :laughing: Otherwise looks fine.

jieyouxu: ... Oh.

jieyouxu: Hopefully I fixed it :D

Kobzol:
Can you try to commit a change to some random MD that contains a non-existent third-party URL? It has to start with http/https.

jieyouxu:
💀

jieyouxu:
(Looks like that works)

PR #2298, opened 2025-03-20
rustdoc: Further improve chapters and sections on testing

fmease:
Follow-up to rust-lang/rustc-dev-guide#2295 (applies my remaining review comments).

This is WIP / a draft because I g2g. Will impl all FIXME(fmease)s when I come back.

cc @lolbinarycat @jieyouxu @notriddle

### DIFF
--- a/src/rustdoc-internals.md
+++ b/src/rustdoc-internals.md
@@ -270,35 +270,6 @@ in `test.rs` is the function `make_test`, which is where hand-written
 Some extra reading about `make_test` can be found
 [here](https://quietmisdreavus.net/code/2018/02/23/how-the-doctests-get-made/).
 
-## Dotting i's And Crossing t's
-
-So that's `rustdoc`'s code in a nutshell, but there's more things in the
-compiler that deal with it. Since we have the full `compiletest` suite at hand,
-there's a set of tests in `tests/rustdoc` that make sure the final `HTML` is
-what we expect in various situations. These tests also use a supplementary
-script, `src/etc/htmldocck.py`, that allows it to look through the final `HTML`
-using `XPath` notation to get a precise look at the output. The full
-description of all the commands available to `rustdoc` tests (e.g. [`@has`] and
-[`@matches`]) is in [`htmldocck.py`].
-
-To use multiple crates in a `rustdoc` test, add `//@ aux-build:filename.rs`
-to the top of the test file. `filename.rs` should be placed in an `auxiliary`
-directory relative to the test file with the comment. If you need to build
-docs for the auxiliary file, use `//@ build-aux-docs`.
-
-In addition, there are separate tests for the search index and `rustdoc`'s
-ability to query it. The files in `tests/rustdoc-js` each contain a
-different search query and the expected results, broken out by search tab.
-These files are processed by a script in `src/tools/rustdoc-js` and the `Node.js`
-runtime. These tests don't have as thorough of a writeup, but a broad example
-that features results in all tabs can be found in `basic.js`. The basic idea is
-that you match a given `QUERY` with a set of `EXPECTED` results, complete with
-the full item path of each item.
-
-[`@has`]: https://github.com/rust-lang/rust/blob/master/src/etc/htmldocck.py#L39
-[`@matches`]: https://github.com/rust-lang/rust/blob/master/src/etc/htmldocck.py#L44
-[`htmldocck.py`]: https://github.com/rust-lang/rust/blob/master/src/etc/htmldocck.py
-
 ## Testing Locally
 
 Some features of the generated `HTML` documentation might require local
--- a/src/rustdoc-internals/rustdoc-test-suite.md
+++ b/src/rustdoc-internals/rustdoc-test-suite.md
@@ -1,112 +1,190 @@
 # The `rustdoc` test suite
 
-This page is specifically about the test suite named `rustdoc`.
-For other test suites used for testing rustdoc, see [Rustdoc tests](../rustdoc.md#tests).
+This page is about the test suite named `rustdoc` used to test the HTML output of `rustdoc`.
+For other rustdoc-specific test suites, see [Rustdoc test suites].
 
-The `rustdoc` test suite is specifically used to test the HTML output of rustdoc.
+Each test file in this test suite is simply a Rust source file `file.rs` sprinkled with
+so-called *directives* located inside normal Rust code comments.
+These come in two flavors: *Compiletest* and *HtmlDocCk*.
 
-This is achieved by means of `htmldocck.py`, a custom checker script that leverages [XPath].
+To learn more about the former, read [Compiletest directives].
+For the latter, continue reading.
 
-[XPath]: https://en.wikipedia.org/wiki/XPath
+Internally, [`compiletest`] invokes the supplementary checker script [`htmldocck.py`].
 
-## Directives
-Directives to htmldocck are similar to those given to `compiletest` in that they take the form of `//@` comments.
+[Rustdoc test suites]: ../tests/compiletest.md#rustdoc-test-suites
+[`compiletest`]: ../tests/compiletest.md
+[`htmldocck.py`]: https://github.com/rust-lang/rust/blob/master/src/etc/htmldocck.py
 
-In addition to the directives listed here,
-`rustdoc` tests also support most
-[compiletest directives](../tests/directives.html).
+## HtmlDocCk Directives
 
-All `PATH`s in directives are relative to the rustdoc output directory (`build/TARGET/test/rustdoc/TESTNAME`),
-so it is conventional to use a `#![crate_name = "foo"]` attribute to avoid
-having to write a long crate name multiple times.
-To avoid repetition, `-` can be used in any `PATH` argument to re-use the previous `PATH` argument.
+Directives to HtmlDocCk are assertions that place constraints on the generated HTML.
+They look similar to those given to `compiletest` in that they take the form of `//@` comments
+but ultimately, they are completey distinct and processed by different programs.
 
-All arguments take the form of quoted strings
-(both single and double quotes are supported),
+[XPath] is used to query parts of the HTML document tree.
+
+**Introductory example**:
+
+```rust,ignore (illustrative)
+//@ has file/type.Alias.html
+//@ has - '//*[@class="rust item-decl"]//code' 'type Alias = Option<i32>;'
+pub type Alias = Option<i32>;
+```
+
+Here, we check that documentation generated for crate `file` contains a page for the
+public type alias `Alias` where the code block that is found at the top contains the
+expected rendering of the item. The `//*[@class="rust item-decl"]//code` is an XPath
+expression.
+
+Conventionally, you place these directives directly above the thing they are meant to test.
+Technically speaking however, they don't need to be as HtmlDocCk only looks for the directives.
+
+All directives take a `PATH` argument.
+To avoid repetition, `-` can be passed to it to re-use the previous `PATH` argument.
+Since the path contains the name of the crate, it is conventional to add a
+`#![crate_name = "foo"]` attribute to the crate root to shorten the resulting path.
+
+All arguments take the form of shell-style (single or double) quoted strings,
 with the exception of `COUNT` and the special `-` form of `PATH`.
 
-Directives are assertions that place constraints on the generated HTML.
-
-All directives (except `files`) can be negated by putting a `!` in front of their name.
+All directives (except `files`) can be *negated* by putting a `!` in front of their name.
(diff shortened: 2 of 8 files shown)


src/rustdoc-internals/rustdoc-test-suite.md
```diff
@@ -4,21 +4,19 @@ This page is specifically about the test suite named `rustdoc`, for other test s
...
 [XPath]: https://en.wikipedia.org/wiki/XPath
 
-## Directives
-Directives to htmldocck are similar to those given to `compiletest` in that they take the form of `//@` comments.
+## Htmldocck-Specific Directives
 
-In addition to the directives listed here,
-`rustdoc` tests also support most
-[compiletest directives](../tests/directives.html).
+Directives to htmldocck are similar to those given to `compiletest` in that they take the form of `//@` comments.
```
jieyouxu, line 15: Suggestion: even though it's an implementation detail, I think it's worth hammering in the distinction that htmldocck directives are handled by *htmldocck* and *not* compiletest.

src/rustdoc-internals/rustdoc-test-suite.md
```diff
@@ -34,33 +32,41 @@ In this case, the start of the next line should be `//`, with no `@`.
 
 For example, `//@ !has 'foo/struct.Bar.html'` checks that crate `foo` does not have a page for a struct named `Bar` in the crate root.
 
+<!-- FIXME(fmease): Mention that the  regexes match case-sensitively and in single-line mode? -->
```
jieyouxu, line 35: Discussion: may also be worth mentioning if the regexes have any string and/or regex escape rules. For instance, I know some compiletest regex directives don't handle this very well.

src/rustdoc-internals/rustdoc-test-suite.md
```diff
@@ -78,34 +84,61 @@ determined by the XPath, and compares it to a pre-recorded value
 in a file. The file's name is the test's name with the `.rs` extension
 replaced with `.NAME.html`, where NAME is the snapshot's name.
 
-htmldocck supports the `--bless` option to accept the current subtree
+Htmldocck supports the `--bless` option to accept the current subtree
```
jieyouxu, line 87: Question: IIRC the `--bless` flow requires html-tidy to be installed and available somewhere. Not sure if this is mentioned already. Maybe worth explicitly remarking?

src/rustdoc-internals/rustdoc-test-suite.md
```diff
@@ -78,34 +84,61 @@ determined by the XPath, and compares it to a pre-recorded value
...
+## Compiletest Directives
+
+In addition to the directives listed here,
+`rustdoc` tests also support most
+[compiletest directives](../tests/directives.html).
+
+<!-- FIXME(fmease):
+Should definitely also mention `//@ aux-crate` and `//@ proc-macro`
+UNLESS we nuke this paragraph entirely and refer to the compiletest section(s)?
+-->
```
jieyouxu, line 124: Remark (not-a-review-comment): I think this is where I would like compiletest to support some kind of directive handling "plugin" and allow htmldocck to register plugins. Then we could phrase this as "htmldocck registers additional directive handling to compiletest" either here, or even put this bit under main compiletest chapter.

src/rustdoc-internals/rustdoc-test-suite.md
jieyouxu, line 1: Suggestion: may or may not be helpful for someone looking to adjust how `tests/rustdoc` works / change htmldocck/compiletest for `tests/rustdoc` purposes... but here's a very coarse TL;DR on the test infra bits involved on running `tests/rustdoc`:

---

## Test infra setup

The test infra setup for `tests/rustdoc` involves the following bits:

- [Bootstrap](../building/bootstrapping/writing-tools-in-bootstrap.md) (the build system) and its test step that builds:
    - A suitably-staged sysroot (including the compiler and standard library).
    - A rustdoc built with the sysroot.
    - [`compiletest`] itself, and running it with the suitable configuration under the `rustdoc` test mode.
    - The implementation can be traced back from https://github.com/rust-lang/rust/blob/4e2b096ed6c8a1400624a54f6c4fd0c0ce48a579/src/bootstrap/src/core/build_steps/test.rs#L1349-L1355.
- [`compiletest`] handles the test discovery, compiletest directive parsing/handling (not to be confused with htmldocck directives) and test running via libtest. See `src/tools/compiletest/src/runtest/rustdoc.rs`.
- [`compiletest`] invokes `src/etc/htmldocck.py` against the test file.

[`compiletest`]: ../tests/compiletest.md

src/tests/directives.md
```diff
@@ -261,14 +261,25 @@ Consider writing the test as a proper incremental test instead.
...
-| `doc-flags` | Flags passed to `rustdoc` when building the test or aux file | `rustdoc`, `rustdoc-js`, `rustdoc-json` | Any valid `rustdoc` flags |
+| `doc-flags` | Flags passed to `rustdoc` when building the test or aux file | `rustdoc`, `rustdoc-js`, `rustdoc-json`  | Any valid `rustdoc` flags |
 
 <!--
 **FIXME(rustdoc)**: what does `check-test-line-numbers-match` do?
 Asked in
 <https://rust-lang.zulipchat.com/#narrow/stream/266220-t-rustdoc/topic/What.20is.20the.20.60check-test-line-numbers-match.60.20directive.3F>.
 -->
 
+#### Test-suite-specific directives
```
fmease, line 272: I've added this subsection because you may very likely land in section `Compiletest directives § Rustdoc` (the parent section) via fuzzy searching (I noticed that earlier) when (from the perspective of a new user) all you wanted was to get to know what on earth those tool-specific directives meant.

On master, this section wouldn't help you in the slightest.

Re. my change: I wouldn't classify this as duplicating information (which master rustdoc docs do a bunch and it's maddening!) but as *necessary redundancy*. It's also helpful IMO if you're just scrolling this section trying to get an overview over directive handling (tool-specific directives are an important part of it).

src/rustdoc.md
```diff
@@ -67,43 +67,29 @@ does is call the `main()` that's in this crate's `lib.rs`, though.)
...
-  browser-UI-test](https://github.com/GuillaumeGomez/browser-UI-test/) that uses
-  puppeteer to run tests in a headless browser and check rendering and
-  interactivity.  For information on how to write this form of test,
-  see [`tests/rustdoc-gui/README.md`][rustdoc-gui-readme]
-  as well as [the description of the `.goml` format][goml-script]
-* Tests on the structure of rustdoc HTML output are located in `tests/rustdoc`,
-  where they're handled by the test runner of bootstrap and
-  the supplementary script `src/etc/htmldocck.py`.
-  [These tests have several extra directives available to them](./rustdoc-internals/rustdoc-test-suite.md).
-* Additionally, JavaScript type annotations are written using [TypeScript-flavored JSDoc]
```
fmease, line 100: This bullet point had *nothing* to do with tests (but with static analysis). I've moved it into *Code Structure* where the term *JavaScript* was already mentioned.

Ofc, that's still not great of a place, ideally we would have a separate section about internal linting / static analysis or something of similar nature.

src/rustdoc.md
```diff
@@ -67,43 +67,29 @@ does is call the `main()` that's in this crate's `lib.rs`, though.)
...
 * Frontend CSS and JavaScript are stored in `html/static/`.
+  * Re. JavaScript, type annotations are written using [TypeScript-flavored JSDoc]
+comments and an external `.d.ts` file.
+    This way, the code itself remains plain, valid JavaScript.
+    We only use `tsc` as a linter.
 
-## Tests
+[TypeScript-flavored JSDoc]: https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html
 
-* Tests on search engine and index are located in `tests/rustdoc-js` and `tests/rustdoc-js-std`.
```
fmease, line 84: This unordered list of tests suites was incredibly illegible. If you're trying to find out about a specific test suite, you basically needed to linearly scan through it and skip all that prose.

A table is way better. First I tried to turn it into a table, then I found out that we already have such a table over at `tests/compiletest.md#rustdoc-test-suites`. There's no need to duplicate this information, it's even counterproductive as everything gets easily outdated and it's hard to find a "single source of truth" because they don't match 100%.

So I've just removed this subsection (because this subpage shouldn't actually talk about *testing*, we have an entire directory full of documents for that[^1]) in favor of a cross-link. I've slightly fleshed out that table, so no information gets lost.

[^1]: I really dislike how rustdoc content is split weirdly between `rustdoc.md` and `rustdoc-internals.md` as it's unclear *where* to put content. We need to reassess and restructure. Find a better flow.

src/rustdoc-internals.md
```diff
@@ -270,35 +270,6 @@ in `test.rs` is the function `make_test`, which is where hand-written
 Some extra reading about `make_test` can be found
 [here](https://quietmisdreavus.net/code/2018/02/23/how-the-doctests-get-made/).
 
-## Dotting i's And Crossing t's
```
fmease, line 273: *Severely outdated* (the `rustdoc-js{,-std}` test suites are nowadays fully documented in a separate document) and we also shouldn't talk about tests here, we have whole subdirectory dedicated for that.

src/rustdoc.md
```diff
@@ -67,43 +67,29 @@ does is call the `main()` that's in this crate's `lib.rs`, though.)
...
+  The functions responsible for creating them from the `HIR` and the `rustc_middle::ty` IR
+  live in `clean/mod.rs`.
 * The bits specific to using rustdoc as a test harness are in
   `doctest.rs`.
 * The Markdown renderer is loaded up in `html/markdown.rs`, including functions
   for extracting doctests from a given block of Markdown.
 * Frontend CSS and JavaScript are stored in `html/static/`.
+  * Re. JavaScript, type annotations are written using [TypeScript-flavored JSDoc]
+comments and an external `.d.ts` file.
+    This way, the code itself remains plain, valid JavaScript.
```
lolbinarycat, line 84: We should explain that this is so that `tsc` is not required as a build-time dependancy of rustdoc, not because we don't transform the source code at all (we are minifying it, so the jsdoc comments will not appear in the final documentation bundle unless minification is disabled, not sure how many of those details we should include)

isn't this duplicating some text that already exists somewhere? should we deduplicate that and make a subpage for how we use tsc? or is it small enough that the duplication isn't a problem?

src/rustdoc.md
lolbinarycat: the rationale was that both are run in CI, but I guess there is a fairly meaningful distinction between static analysis and unit/integration tests, with the line only becoming blurry again once you bring in proof assistants like `kani`.

src/rustdoc-internals/rustdoc-test-suite.md
fmease: No, it's not necessary. In compiletest, html-tidy is only used to render html diffs, so it's only cosmetic ^^. It's unrelated to `--bless`.

I for one don't use it (IIRC three years ago when I last tried it I found the html-tidy-generated diff to be awfully verbose and uninstalled it immediately (I'm sure you can super duper tweak its settings and whatnot but I couldn't be bothered)).

fmease: (it's apparently also used in `src/tools/html-checker` which is completely unrelated to HtmlDocCK)

src/rustdoc-internals/rustdoc-test-suite.md
fmease: Hmm, I mean Python-flavored regex (which is mentioned) have all your typical Perl-style regex escape sequences, but that's to be expected. Regarding interaction with HtmlDocCk directive syntax, well, PATTERN is a shell-style quoted string (which is mentioned) whose escaping rules are pretty relaxed and thus there shouldn't really be a need to double-escape / worry about that.

I don't have the energy anymore to phrase this succinctly in-document and I think it can be deduced from "shell-style" + "Python-flavored".

src/rustdoc-internals/rustdoc-test-suite.md
fmease: I will add this in a separate PR in a subsection named "Internals" because this PR has grown too much (but lol @ "rustdoc-internals > rustdoc > internals").

src/rustdoc.md
fmease: Yep, we can revisit this when creating the static analysis / xyz section

fmease:
Latest changes — they only concern the "rustdoc test suite" chapter itself:

* I've made sure that the flow makes sense when you start at the top and make it down to the bottom (concepts/.. get introduced/defined in order etc.)
* I've added an introductory example
* A lot of small tweaks and small additions

I'm about to merge this PR now — at long last

PR #2311, opened 2025-03-30
Fix partial clone link

clubby789:
This link currently points to the current section, instead of the section on partial clones


jieyouxu:
Thanks

PR #2318, opened 2025-04-07
improve flow

tshepang:
am unsure about "exceptional cases"... feels too prescrictive
but I did not want to change any meaning on this pr


jieyouxu (approved):
Seems fine, thanks

jieyouxu:
The exceptional cases wording is right to me. In general, we don't want to skip annotations unless it's necessary.

PR #2319, opened 2025-04-08
compiletest: document `{ignore,only}-{$target_triple,$target_arch}` impl limitation

jieyouxu:
cc @RalfJung 
Closes https://github.com/rust-lang/rust/issues/139516.

### DIFF
--- a/src/tests/directives.md
+++ b/src/tests/directives.md
@@ -129,6 +129,19 @@ means the test won't be compiled or run.
 
 Some examples of `X` in `ignore-X` or `only-X`:
 
+> NOTE:
+>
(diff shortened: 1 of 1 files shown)


RalfJung:
> computed based on all the builtin target specs,

That is not correct. My issue is all about the fact that there are builtin targets that are not recognized by compiletest.

RalfJung:
Also, what about OS and "env"? I assume those have the same problem? Do any other directives also have that problem? Someone who knows compiletest should ensure this list of caveats is complete.

jieyouxu:
Let me see if I can adjust the impl so that this caveat is no longer necessary.

PR #2323, opened 2025-04-10
Align advice to discourage issue-number-only names

smanilov:
https://rustc-dev-guide.rust-lang.org/tests/best-practices.html#test-naming specifically asks devs to "Avoid using only issue numbers as test names".

This commit introduces a minor change to https://rustc-dev-guide.rust-lang.org/tests/compiletest.html#crashes-tests that changes the advice to use only issue numbers as names to the advice to append them at the end of the file name.

### DIFF
--- a/src/tests/best-practices.md
+++ b/src/tests/best-practices.md
@@ -70,6 +70,11 @@ related tests.
 > //!
 > //! Regression test for <https://github.com/rust-lang/rust/issues/123456>.
 > ```
+>
+> One exception to this rule is [crashes tests]: there it is canonical that
+> tests are named only after issue numbers because its purpose is to track
+> snippets from which issues no longer ICE/crash, and they would either be
+> removed or converted into proper ui/other tests in the fix PRs.
 
 ## Test organization
 
@@ -194,3 +199,4 @@ See [LLVM FileCheck guide][FileCheck] for details.
 [compiletest directives]: ./directives.md
 [`run-make`]: ./compiletest.md#run-make-tests
 [FileCheck]: https://llvm.org/docs/CommandGuide/FileCheck.html
+[crashes tests]: ./compiletest.md#crashes-tests


BoxyUwU:
Do we have consensus on this from somewhere? I kind of also prefer people to not write tests named `what-my-test-is-183842.rs`

BoxyUwU:
oh wait this is crashes tests

BoxyUwU:
I think the advice for crashes tests tends to actually be to just name the test after the issue no., if you look at the `tests/crashes` directory every single test is named something like `289743.rs` (except for two...)

smanilov:
> I think the advice for crashes tests tends to actually be to just name the test after the issue no. (so the current docs seem in line with what is done in practice), if you look at the `tests/crashes` directory every single test is named something like `289743.rs` (except for two...)

Oh, I see, thanks for point this out.

So more or less I would close the issue... if it weren't for another "conflicting" advice: https://rustc-dev-guide.rust-lang.org/tests/ui.html#test-organization says "Avoid including the issue number in the test name.", but then points to the best practices, which says "Avoid using issue numbers alone as test names." and "Avoid starting the test name with issue-xxxxx prefix as it degrades auto-completion.".

Perhaps a note for "crashes" tests that it explicitly goes agains the general best practices is still warranted?

Is the convention for ui tests also specific (and ok to conflict with best practices), in your opinion?

jieyouxu:
Hi, the page there is explicitly for ui tests, which notably does not include crashes tests. Probably just remark that for crashes tests it is canonical that tests are named only after issue numbers because its purpose is to track which issues no longer ICE/crash, and they would either be removed or converted into proper ui/other tests in the fix PR.

smanilov:
WIP, incorporating latest comment by @jieyouxu

smanilov:
Sorry for the delay. @rustbot review

smanilov:
Hmm, how about @rustbot label -S-waiting-on-author +S-waiting-on-review

smanilov:
Last try: @rustbot label +S-waiting-on-review

BoxyUwU:
thanks :)

PR #2333, opened 2025-04-12
Coding conventions chapter cleaning

tshepang:


jieyouxu (approved):
Changes look good. Can you please squash the commits into one? r=me after

tshepang:
I separated those since each change could be its own pr, and so one could easily see in history why each change was made. Squashing loses that information.

PR #2339, opened 2025-04-17
Add a first version of autodiff docs

ZuseZ4:
Upstreaming of std::autodiff got approved, so I want to get rid of my docs under https://enzyme.mit.edu/rust/
These are the first ones, I'll add a few more docs here and to the nightly docs.

r? @jieyouxu


jieyouxu (approved):
Thanks, I only did some formatting changes, content looks good

PR #2357, opened 2025-04-27
Update `rustc_on_unimplemented` docs

mejrs:
Note that these docs reflect the current implementation but some fixes and updates have not percolated into the bootstrap compiler yet:
- `"{This}"` is not supported
- the "several values to match on" part works with `_Self` but not `Self`.

So I guess we should avoid merging this until the middle of May?

[Rendered](https://github.com/mejrs/rustc-dev-guide/blob/abe59658d0416969a43e09dce1606623d6ec00cb/src/diagnostics.md#rustc_on_unimplemented)

### DIFF
--- a/src/diagnostics.md
+++ b/src/diagnostics.md
@@ -866,19 +866,17 @@ struct](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_errors/json/struct
 (and sub-structs) for the JSON serialization. Don't confuse this with
 [`errors::Diag`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_errors/struct.Diag.html)!
 
-## `#[rustc_on_unimplemented(...)]`
+## `#[rustc_on_unimplemented]`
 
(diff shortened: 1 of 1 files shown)


jieyouxu:
For future reference, since rustc-dev-guide is now a josh subtree in rust-lang/rust, you can simply update rustc-dev-guide in the same PR as the compiler change :D

Please ping me after the next beta bump, I'll merge this then.

jieyouxu:
Let me know if this is now mergeable.

mejrs:
Not yet, I want to make a couple of changes.

jieyouxu:
Huh, right, we should probably enable waiting-on-{review,author} flips

src/diagnostics.md
```diff
@@ -930,45 +922,61 @@ Would generate the following output:
 
 ```text
 error[E0277]: message
-  --> <anon>:14:5
+  --> <file>>:10:19
```
jieyouxu, line 925: Nit:

```suggestion
  --> <file>:10:19
```

PR #2359, opened 2025-04-28
use repo name in push pr title

tshepang:
~Also, remove `@ghost` in pr description, to avoid confusion... thought it was a real user.~
Heh, I see [it has utility](https://forge.rust-lang.org/triagebot/pr-assignment.html#ghost).


fmease:
cc @rust-lang/wg-rustc-dev-guide

jieyouxu (approved):
Yeah, the ghost is intentional because otherwise it would by default rotate subtree sync PRs between compiler reviewers on active rotation, whereas usually only a subset of people are familiar with the sync workflow.

PR #2365, opened 2025-04-29
Fix footnotes

BoxyUwU:
For some reason `mdbook serve` locally gives different output than in the built version of the dev guide... oh well


tshepang:
was it maybe different mdbook versions... the footnote functionality of mdbook was recently changed

BoxyUwU:
> As part of this change, footnotes are now only rendered at the bottom of the page

Ah yeah that would be it... that's an unfortunate change ah well. I'll update my local mdbook I guess :-)

tshepang:
oh, what's unfortunate about it... did you prefer previous behavior

BoxyUwU:
Yeah I was trying to have footnotes split between each "big header" so that readers would stay at "the same general context", but I guess it doesn't really matter since there's a nice button that takes you back to where you were anyway :)

PR #2366, opened 2025-04-29
for a more friendly output

tshepang:
Also, these are normal Rust things (crates/packages), so remove the word *normal*.


jieyouxu (approved):
Thanks

PR #2368, opened 2025-05-01
Adds 'with' to the bootstrapping docs to help clarify how to build a new compiler

martinomburajr:
The sentence

> However it was abandoned long ago and the only way to build a modern version of rustc is **a** slightly less modern version.

feels like it's grammatically missing the preposition 'with', as well as some commas around "However" etc.

### DIFF
--- a/src/building/bootstrapping/intro.md
+++ b/src/building/bootstrapping/intro.md
@@ -7,7 +7,7 @@ of the same compiler.
 This raises a chicken-and-egg paradox: where did the first compiler come from?
 It must have been written in a different language. In Rust's case it was
 [written in OCaml][ocaml-compiler]. However it was abandoned long ago and the
-only way to build a modern version of rustc is a slightly less modern
+only way to build a modern version of rustc is with a slightly less modern
 version.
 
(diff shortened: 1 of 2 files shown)


tshepang:
this is a good change, and it exposes text that is almost exactly duplicated... are you happy to remove/reduce that, or would you rather that happen in a different pr

tshepang:
will defer that to a future pr... no need to delay this

martinomburajr:
@tshepang thanks for the review 🙏 . I had noticed the paragraph is duplicated word for word across both files. I had thought about removing the second occurrence, but wasn't sure if this was intended.

If we were to go forward with this, the two options that make sense to me are:

1. Summarize what was in `src/building/bootstrapping/intro.md` into `src/building/bootstrapping/what-bootstrapping-does.md`
2. Completely remove the duplicate paragraph in `what-bootstrapping-does.md`

Let me know what makes sense and I can make the quick change.

tshepang:
or 2a, add a link to that bit of text
I am curious what a summary would look like though... the text is short already
feel free to decide

PR #2378, opened 2025-05-09
Is the parser undergoing refactoring?

smanilov:
The suggested change is more of conversation starter. I could track the introduction of the message to https://github.com/rust-lang/rustc-dev-guide/commit/a50b8f144f554e8d56210ed064f77f2d7ec7d41f which is from Nov 2019.

Is the refactor still undergoing? Is there a tracking issue for it?

If not, the N.B. can just be deleted... but maybe the document would need to be updated with the changes from the refactor if that's the case?

### DIFF
--- a/src/macro-expansion.md
+++ b/src/macro-expansion.md
@@ -2,7 +2,7 @@
 
 <!-- toc -->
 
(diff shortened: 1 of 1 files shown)


jieyouxu:
AFAIK not really. I would just remove that.

jieyouxu (approved):
Thanks

PR #2384, opened 2025-05-13
Add git command quick reference for git.md

xizheyin:
I've added some cheat sheets for the git command, but they're not perfect.


jieyouxu (commented):
I'm not actually sure if we should add these... Because for instance I use different git (sub-)commands and/or flags, and it might be better to delegate to actual git docs for git commands...

xizheyin:
I think what you said is reasonable. I mainly use this pr to test `behind-upsteam`.

PR #2385, opened 2025-05-13
documentation for Enzyme Type Trees 

KMJ-007:


src/autodiff/type-trees.md
```diff
@@ -0,0 +1,118 @@
+# Type Trees in Enzyme
+
+This document describes type trees as used by Enzyme for automatic differentiation.
+
+## What are Type Trees?
+
+Type trees in Enzyme are a way to represent the types of variables, including their activity (e.g., whether they are active, duplicated, or contain duplicated data) for automatic differentiation. They provide a structured way for Enzyme to understand how to handle different data types during the differentiation process.
```
ZuseZ4, line 7: Are there any docs or code which shows them including activity?

src/autodiff/type-trees.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Representing Rust Types as Type Trees
+
+Enzyme needs to understand the structure and properties of Rust types to perform automatic differentiation correctly. This is where type trees come in. They provide a detailed map of a type, including pointer indirections and the underlying concrete data types.
+
+The `-enzyme-rust-type` flag in Enzyme helps in interpreting types more accurately in the context of Rust's memory layout and type system.
```
ZuseZ4, line 13: The flag just tells enzyme to parse (rust) debug "dwarf" information.
A lot of type information is not encoded in such debug metadata, and the flag hasn't been re-evaluated or used in years. It's good to mention it here (with the corrected description), but I wouldn't mention it in the following sections. We should generate typetrees based on Rust types even without debug metadata.

src/autodiff/type-trees.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Primitive Types
+
+#### Floating-Point Types (`f32`, `f64`)
+
+Consider a Rust reference to a 32-bit floating-point number, `&f32`.
+
+In LLVM IR, this might be represented, for instance, as an `i8*` (a generic byte pointer) that is then `bitcast` to a `float*`. Consider the following LLVM IR function:
```
ZuseZ4, line 21: Ah, you took that from `rustmutpointer.ll` I guess? Unfortunately they are too outdated, I just realized.
Typed ptr were removed (see my comment about this flag being very outdated), so i8* isn't a thing anymore.
Instead, we now have `ptr` ("opaque pointers") in LLVM. 

You can look for the PRs which introduced these tests a few years ago. They should have instructions on how to reproduce them, so you can re-generate newer tests once you have a working setup.

src/autodiff/type-trees.md
```diff
@@ -0,0 +1,118 @@
...
+%x.dbg.spill = bitcast i8* %x to float*: {[-1]:Pointer, [-1,0]:Float@float}
+```
+
+**Understanding the Type Tree: `{[-1]:Pointer, [-1,0]:Float@float}`**
+
+This string is the type tree representation. Let's break it down:
+
+*   **`{ ... }`**: This encloses the set of type information for the variable.
+*   **`[-1]:Pointer`**:
+    *   `[-1]` is an index or path. In this context, `-1` often refers to the base memory location or the immediate value pointed to.
```
ZuseZ4, line 45: `-1` has a slightly different meaning, something along the lines of `everything accessible from here, without dereferencing a pointer`.
So [f64;32]` could be represented as `[-1]:Float@double`, or as `[0]:Float@double, [8]:Float@double, ...`
Afaik we usually prefer `-1` in such cases since it's shorter, but IIRC there were some gotchas.
@wsmoses  can you share the private youtube video with him (if you prefer in a zulip dm), such that he has more information on how to write these docs?

ZuseZ4: Or e.g. `x: *const [f64;32]` could be represented as
`[[-1]:Pointer, [-1:-1]Float@double]`, or as `[[0]:Pointer, [0:0]:Float@double, [0:8]:Float@double, ...`
or as `[[0]:Pointer, [0:-1]:Float@double` or as `[[-1]:Pointer, [-1:0]:Float@double, [-1:8]:Float@double, ...`
(I think)

That also all is under the assumption that we won't access e.g. `x[34]` which is out of bounds of the original array.
I am not 100% sure if there are cases where this could be valid, I think with raw pointers it might be valid to access other elements. E.g. `struct { x: [f64;32], y: i32 }`. I you derive a raw pointer to x, you might be able to use it to access y (legally), in which case `-1:Float@double` would be wrong. I'd need a refresher on pointer provenance and other things, hopefully Oli will know more about it. You can add this as an open question at the end.

src/autodiff/type-trees.md
KMJ-007: my bad i saw this and got confused they both are in the same, later when i was going through enzyme codebase i realised they are different, fixing this
image

src/autodiff/type-trees.md
KMJ-007: got this, the flag was there in most of test files, which were related to rust

src/autodiff/type-trees.md
KMJ-007: yes, 

found the PR, this will be a great reference and understanding of how things were done

https://github.com/EnzymeAD/Enzyme/pull/307

wsmoses:
I looked and didn't see any private youtube video.

src/autodiff/type-trees.md
```diff
@@ -0,0 +1,163 @@
...
+```
+
+## Example: `fn compute(x: &f32, data: &[f32]) -> f32`
+
+**Input 0: `x: &f32`**
+```rust
+TypeTree(vec![Type {
+    offset: -1, size: 8, kind: Pointer,
+    child: TypeTree(vec![Type {
+        offset: -1, size: 4, kind: Float,
```
ZuseZ4: I think this should be `offset: 0`, since it is only one value. But I'm also open to ask Billy for opinions.

Both are kind of ok, but probably confusing to readers since now different Rust types both lower to the same typetree.
I would only generate `-1` when it's actually worth it, thus more than one value. Otherwise, 0 is more precise.

src/autodiff/type-trees.md
```diff
@@ -0,0 +1,163 @@
...
+```rust
+TypeTree(vec![
+    Type { offset: 0, size: 4, kind: Float },   // x at byte 0
+    Type { offset: 4, size: 4, kind: Float },   // y at byte 4
+    Type { offset: 8, size: 4, kind: Integer }  // id at byte 8
+])
+```
+
+Generates LLVM:
+```llvm
```
ZuseZ4, line 115: Why is it generating this IR though?
Shouldn't it be 
`"enzyme_type"="{[0]:Float@float, [4]:Float@float, [8]:Integer@int, [9]:Integer@int, [10]:Integer@int, [11]:Integer@int}"`

src/autodiff/type-trees.md
```diff
@@ -0,0 +1,179 @@
...
+#### Example 1: Array `[f32; 100]`
+```rust
+TypeTree(vec![Type {
+    offset: -1, // ALL positions
+    size: 4,    // each f32 is 4 bytes
+    kind: Float, // every element is float
+}])
+```
+
+Generates LLVM: `"enzyme_type"="{[-1,-1]:Float@float}"`
```
ZuseZ4: If there is no ptr indirection, this one should only be `[-1]:Float@float`

src/autodiff/type-trees.md
```diff
@@ -0,0 +1,179 @@
...
+TypeTree(vec![
+    Type { offset: 0, size: 4, kind: Float },   // x at byte 0
+    Type { offset: 4, size: 4, kind: Float },   // y at byte 4
+    Type { offset: 8, size: 4, kind: Integer }  // id at byte 8
+])
+```
+
+Generates LLVM:
+```llvm
+"enzyme_type"="{[-1]:Pointer, [-1,0]:Float@float, [-1,4]:Float@float, [-1,8]:Integer}"
```
ZuseZ4: I think that Integers are one byte at a time, so it would be 
```
"enzyme_type"="{[-1]:Pointer, [-1,0]:Float@float, [-1,4]:Float@float, [-1,8]:Integer, [-1,9]:Integer, [-1,10]:Integer, [-1,11]:Integer}"
```
or did you see something different in Enzyme tests?

src/autodiff/type-trees.md
KMJ-007: added one more example for this to explain

src/autodiff/type-trees.md
KMJ-007: fixed it

ZuseZ4:
Thanks. That now describes Enzyme's typetrees to the best of my knowledge, if we misunderstood something we can fix it in the future.

PR #2386, opened 2025-05-14
Fix link to GatherBorrows

Lysxia:


tshepang:
maybe better to have 2 links: one to the type, and another to the method

Lysxia:
Which part should link to what method?

tshepang:
looks like the link you removed links to a method on GatherBorrows

Lysxia:
In this context, there's nothing about the `visit_local` method that's worth mentioning. It's clear that this is a simple case of copying the wrong link.

PR #2393, opened 2025-05-19
Remove unused references and simplify one

smanilov:


tshepang:
@smanilov how did you notice that those references were unused

PR #2396, opened 2025-05-20
Small typo and style fixes in  binders.md

smanilov:
Normally I refrain from nit picking, but this seamed worth it.

### DIFF
--- a/src/ty_module/binders.md
+++ b/src/ty_module/binders.md
@@ -1,6 +1,6 @@
 # `Binder` and Higher ranked regions
 
-Sometimes we define generic parameters not on an item but as part of a type or a where clauses. As an example the type `for<'a> fn(&'a u32)` or the where clause `for<'a> T: Trait<'a>` both introduce a generic lifetime named `'a`. Currently there is no stable syntax for `for<T>` or `for<const N: usize>` but on nightly `feature(non_lifetime_binders)` feature can be used to write where clauses (but not types) using `for<T>`/`for<const N: usize>`.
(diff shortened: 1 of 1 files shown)


src/ty_module/binders.md
```diff
@@ -43,7 +44,7 @@ Binder(
     &[BoundVariableKind::Region(...)],
 )
 ```
-This would cause all kinds of issues as the region `'^1_0` refers to a binder at a higher level than the outermost binder i.e. it is an escaping bound var. The `'^1` region (also writeable as `'^0_1`) is also ill formed as the binder it refers to does not introduce a second parameter. Modern day rustc will ICE when constructing this binder due to both of those regions, in the past we would have simply allowed this to work and then ran into issues in other parts of the codebase. 
+This would cause all kinds of issues as the region `'^1_0` refers to a binder at a higher level than the outermost binder i.e. it is an escaping bound var. The `'^1` region (also writeable as `'^0_1`) is also ill formed as the binder it refers to does not introduce a second parameter. Modern day rustc will ICE when constructing this binder due to both of those reasons, in the past we would have simply allowed this to work and then ran into issues in other parts of the codebase.
```
Noratrieb, line 47: I don't think this was really a mistake, but the new one works too

Noratrieb (approved):

Noratrieb:
thanks! btw, while we're correcting typos, it's "seemed", not "seamed" :)

tshepang:
@smanilov am happy to review your future nit picks... I appreciate making text better, even in *small* ways

PR #2398, opened 2025-05-20
Update links between ty-module and binders

smanilov:
The order might have been reversed at some point, leading to the two chapters talking about each other in the wrong order.

### DIFF
--- a/src/ty_module/instantiating_binders.md
+++ b/src/ty_module/instantiating_binders.md
@@ -105,7 +105,8 @@ the `RePlaceholder` for the `'b` parameter is in a higher universe to track the 
 
 ## Instantiating with `ReLateParam`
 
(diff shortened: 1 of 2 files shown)


smanilov:
Just to make it clearer for review: at the moment, instantiating_binders comes first (chapter 48.1). It refers to param_ty_const_regions (chapter 50.2) indirectly via "As discussed in a previous chapter, RegionKind has two variants for representing generic parameters, ReLateParam and ReEarlyParam.".

I searched in the docs and saw no other references to ReLateParam and ReEarlyParam, so I proposed this change.

BoxyUwU:
Possibly we should move the "the ty module" chapter before the binders chapter, but I'm not sure and regardless it seems like an improvement to make the wording be chapter order insensitive

PR #2400, opened 2025-05-20
Make it clear we talk about early bound params

smanilov:
While this is implicit, since "early bound params" are the ones that are bound when naming a function, for the non-expert (myself; most people motivated to read the manual) it helps to be explicit here.

Not feeling strongly about this one. Feel free to ignore/tell me to close PR.

Also let me know if I should rather not propose similar changes in the future.


BoxyUwU:
I like this clarification :+1: Thank you for filing the PR. Indeed, the target audience here is people who are not already experts on the material so explicitly calling this out makes sense to me.

PR #2418, opened 2025-05-27
Link to description of opaque types

smanilov:
It was really hard to read this chapter without knowing what opaque types are. When I reached the chapter that describes them later I felt a huge relief.

Adding this link to share the relief and make it come sooner for others.


Noratrieb (approved):

smanilov:
> @smanilov you can also link like in #2419, which makes reading the markdown more easy

@tshepang , okay, noted. Thanks for fixing it!

PR #2429, opened 2025-05-30
Update opaque-types-type-alias-impl-trait.md

smanilov:
1. Clarify that there is a single concrete type for an opaque type
2. Clarify that defining a type alias to an opaque type is an unstable feature
3. Add complete example

I struggled quite a bit with this one and chatbots failed to help, so I'm pretty confident a complete example is warranted here 😅


jieyouxu:
Maybe Boxy you know better?
r? @BoxyUwU

src/opaque-types-type-alias-impl-trait.md
```diff
@@ -61,3 +83,28 @@ impl Baz for Quux {
     fn foo() -> Self::Foo { ... }
 }
 ```
+
+For this you would also need to use `nightly` and the (different) `#![feature(impl_trait_in_assoc_type)]` annotation.
+Note that you don't need a `#[define_opaque(Foo)]` on the method anymore.
```
BoxyUwU:
```suggestion
Note that you don't need a `#[define_opaque(Foo)]` on the method anymore as the opaque type is mentioned in the function signature (behind the associated type).
```

PR #2435, opened 2025-06-01
Add opaque type attributes

smanilov:
This allows for the code to compile on `nightly`.


smanilov:
Tested locally in CLI, as well as running `mdbook build --open` and manually verifying in the browser.

PR #2441, opened 2025-06-02
use consistent title capitalization

tshepang:


Noratrieb:
I'd prefer using Title Case instead

tshepang:
That would need a bigger diff, for we been going with sentence case (which is also something that was decided on), so this makes things more consistent.

PR #2443, opened 2025-06-02
Simplify long sentences

smanilov:


src/coroutine-closures.md
```diff
@@ -10,7 +10,9 @@ As a consequence of the code being somewhat general, this document may flip betw
 
 ## HIR
 
-Async closures (and in the future, other coroutine flavors such as `gen`) are represented in HIR as a `hir::Closure` whose closure-kind is `ClosureKind::CoroutineClosure(_)`[^k1], which wraps an async block, which is also represented in HIR as a `hir::Closure`) and whose closure-kind is `ClosureKind::Closure(CoroutineKind::Desugared(_, CoroutineSource::Closure))`[^k2].
+Async closures (and in the future, other coroutine flavors such as `gen`) are represented in HIR as a `hir::Closure`.
+Its closure-kind is `ClosureKind::CoroutineClosure(_)`[^k1], which wraps an async block, which is also represented in HIR as a `hir::Closure`.
```
tshepang: "its" can be thought to refer to "Async closures"
```suggestion
The closure-kind of the `hir::Closure` is `ClosureKind::CoroutineClosure(_)`[^k1], which wraps an async block, which is also represented in HIR as a `hir::Closure`.
```

smanilov: Sorry, but I am going to gently push back on this one: I don't think "its" can be thought to refer to "Async closures", because of the plurality of "Async closures". Only "their" can be used where I've written "its", which makes the "its" unambiguous.

tshepang: I see... that it tripped me indicates that the text here is not so simple. My suggestion removes an ambiguity, even if repetitive.

smanilov: Yeah, even as I was writing my objection that came to mind 😅 

I applied your suggestion. Thanks for the review!

tshepang: thanks for loving them docs

PR #2451, opened 2025-06-03
Manual rustc pull

jieyouxu:
This is to sync back the stage 0 redesign bootstrap doc updates.


tshepang:
@jieyouxu note that you could also do this with "Run workflow" from https://github.com/rust-lang/rustc-dev-guide/actions/workflows/rustc-pull.yml

PR #2465, opened 2025-06-13
Adjust some doc for Query System

xizheyin:
This PR improves the documentation for the rustc query system.

- It is emphasized that the provider needs to be registered manually, not looked for by the compiler (this term is too vague).
- Move local and extern provider explanation to the section `Providers`
- Making the provider registration process easier to understand, with step-by-step explanations and clearer examples.


xizheyin:
r? rustc-dev-guide

src/query.md
```diff
@@ -108,61 +125,51 @@ the macros here, but it is basically a big list of function pointers:
...
-N.B. Most of the `rustc_*` crates only provide **local
-providers**. Almost all **extern providers** wind up going through the
-[`rustc_metadata` crate][rustc_metadata], which loads the information
-from the crate metadata. But in some cases there are crates that
-provide queries for *both* local and external crates, in which case
-they define both a `provide` and a `provide_extern` function, through
-[`wasm_import_module_map`][wasm_import_module_map], that `rustc_driver` can invoke.
+1. Implement the provider function:
+    ```rust,ignore
+    fn fubar<'tcx>(tcx: TyCtxt<'tcx>, key: DefId) -> Fubar<'tcx> { ... }
```
tshepang, line 160: will do this change in a future pr
```suggestion
    fn fubar<'tcx>(tcx: TyCtxt<'tcx>, key: LocalDefId) -> Fubar<'tcx> { ... }
```

tshepang: oh, I see we now have `ExternProvider` as well (since https://github.com/rust-lang/rust/pull/85830)... would you be able to adjust docs to account for it @xizheyin

xizheyin: No problem, I'm just working on this pr https://github.com/rust-lang/rust/pull/142450, which involves extern providers.

tshepang: alright... feel free to modify rdg there as well, to avoid having to open a separate pr here

xizheyin: I thought the changes to rdg could only be made here. It seems that I can submit other improvements in rust-lang/rust in the future.

But for this improvement, I think it would be better to submit it here, because rust-lang/rust may take time to synchronize.

tshepang: - larger PRs generally take longer to be reviewed here, so I would not worry about the lag
- doing docs changes on the other side has more chances of being reviewed by someone more familiar with what is being documented on the other side (since they would be the one reviewing the non-doc part)

xizheyin: I see. In the future, I will submit pr related to a specific code part in rust-lang/rust. Thanks.

PR #2485, opened 2025-06-30
update AST-to-HIR lowering examples for conditionals and loops

dianne:
- `for` loops now use two `match`es for all of their bindings. I'm not sure this is the most helpful way of conveying that, but it's about as informative as before while staying brief. I'd be happy to make this clearer, if anyone has ideas.
- `while let` and `if let` don't use `match`; they use `let` expressions in their conditions. Since `if let` no longer has a significantly different desugaring and having a whole bullet point for `while` would feel redundant with `for`, I've removed those examples.


dianne:
r? rustc-dev-guide

PR #2489, opened 2025-07-04
Rustc pull update

Kobzol:
Merge ref 'c96a69059ecc' from rust-lang/rust

Pull recent changes from rust-lang/rust via Josh.

Upstream ref: c96a69059ecc618b519da385a6ccd03155aa0237
Filtered ref: 7b9552d4c39c31aabf6749629da2d4a7e6e1cd60

This PR was created using https://github.com/rust-lang/josh-sync. It also initializes its configuration file (which is done only once, then it's pretty much immutable).


Kobzol:
Once/if this merges, I'll do a push using `josh-sync`. If it works, I'll remove the custom sync code from this repository.

PR #2491, opened 2025-07-04
Finish transition to `rustc-josh-sync`

Kobzol:
Write documentation about `rustc-josh-sync` in RDG, remove the previous Rust crate and update CI to use `rustc-josh-sync`.


jieyouxu (approved):
Thanks

PR #2497, opened 2025-07-09
tweak some git clone commands

tshepang:
- --depth=1 is more useful for once-off uses, like on ci
- .git postfix on github repo url is not needed

### DIFF
--- a/src/autodiff/installation.md
+++ b/src/autodiff/installation.md
@@ -6,7 +6,7 @@ In the near future, `std::autodiff` should become available in nightly builds fo
 
 First you need to clone and configure the Rust repository:
 ```bash
-git clone --depth=1 git@github.com:rust-lang/rust.git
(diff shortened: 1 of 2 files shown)


jieyouxu:
`--depth=1` can be intentional sometimes, if history doesn't matter for your particular workflow instance.
r? @ZuseZ4 (autodiff workflow, I'm not sure)

Noratrieb:
wasn't there a problem that depth 1 doesn't work properly with bootstrap?

jieyouxu:
For rust-lang/rust almost certainly, I was moreso wondering about the enzyme/llvm cases

ZuseZ4:
I think depth=1 worked well for me even when contributing and saved some time, but then users should also know or learn related commands like unshallow. I guess having  slower but simpler and more robust default is better, so lgtm. Thanks!

PR #2502, opened 2025-07-10
Migrate rustc-pull to CI workflow from `josh-sync`

Kobzol:
Replace the CI workflow in this repo with a generalized workflow from `josh-sync`. I tested this on my forks, but there's no other good way to test this than just merge it and see what happens on CI =D


jieyouxu (approved):
Thanks. We'll have to see if this works!

PR #2511, opened 2025-07-18
Mistake in section serialization in rustc

ffengyu:
`TyDecoder` and `TyEncoder` are the **subtraits** of `Decoder` and `Encoder`. The guide says they are the **supertraits**.


ffengyu:
r? rustc-dev-guide

PR #2516, opened 2025-07-22
glossary: add entry for rustbuild

lolbinarycat:


Noratrieb:
I see it more as the old name, I'd call the rust part bootstrap too

lolbinarycat:
Yes, I would say bootstrap is the whole system of bootstrap.py + rustbuild.

The name "rustbuild" still gets used a fair bit by t-bootstrap when the distinction between the two parts is meaningful.  I honestly had no idea the history of the name, I just noticed it was a term people were using that didn't seem to be written down anywhere really.

jieyouxu:
I don't think this is actually used that way? At least, I don't use rustbuild to refer to the rust part, I call that `bootstrap`, and disambiguate from the python part by calling the python part `bootstrap.py` or "the python wrapper" etc. etc.

Noratrieb:
https://github.com/rust-lang/rust/pull/127434

jyn514:
i think having the entry is ok as a reference. i agree we should be consistent in the docs themselves, but i think a glossary entry doesn't hurt, since it will come up occasionally in old issues.

aDotInTheVoid:
How about having this entry, but mentioning that it’s an old deprecated term?

lolbinarycat:
Mentioned the term is deprecated.

src/appendix/glossary.md
```diff
@@ -68,6 +68,7 @@ Term                                                  | Meaning
 <span id="rib">rib</span>                      |  A data structure in the name resolver that keeps track of a single scope for names. ([see more](../name-resolution.md))
 <span id="rpit">RPIT</span>                    |  A return-position `impl Trait`. ([see the reference](https://doc.rust-lang.org/reference/types/impl-trait.html#abstract-return-types)).
 <span id="rpitit">RPITIT</span>                |  A return-position `impl Trait` in trait. Unlike RPIT, this is desugared to a generic associated type (GAT). Introduced in [RFC 3425](https://rust-lang.github.io/rfcs/3425-return-position-impl-trait-in-traits.html). ([see more](../return-position-impl-trait-in-trait.md))
+<span id="rustbuild">rustbuild</span>          |  The part of bootstrap that is written in rust (old deprecated term)
```
tshepang:
```suggestion
<span id="rustbuild">rustbuild</span>          |  A deprecated term for the part of bootstrap that is written in Rust
```

PR #2529, opened 2025-08-01
add a chapter documenting candidate preference

lcnr:
cc @BoxyUwU @compiler-errors


lcnr:
gonna merge so that I can point to it, we can always clean this up later

PR #2538, opened 2025-08-06
Link from "Overview of the compiler § Queries" to the Queries chapter

lolbinarycat:
There's already a link in the other direction, so this seems fairly logical.

### DIFF
--- a/src/overview.md
+++ b/src/overview.md
@@ -321,6 +321,10 @@ the name `'tcx`, which means that something is tied to the lifetime of the
 
 [`TyCtxt`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.TyCtxt.html
 
+For more information about queries in the compiler, see [the queries chapter][queries].
+
+[queries]: https://rustc-dev-guide.rust-lang.org/query.html
+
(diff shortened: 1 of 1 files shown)


src/overview.md
```diff
@@ -321,6 +321,10 @@ the name `'tcx`, which means that something is tied to the lifetime of the
 
 [`TyCtxt`]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.TyCtxt.html
 
+For more information about queries in the compiler, see [the queries chapter][queries].
+
+[queries]: https://rustc-dev-guide.rust-lang.org/query.html
```
tshepang: there is some intro text in same page
```suggestion
[queries]: #queries
```

lolbinarycat: That's... the section I'm linking from?  The section linking to itself isn't useful at all, it needs to link to the chapter.

Anyways this did make me realize I should be using relative a relative URL, so that's been updated.

tshepang: > That's... the section I'm linking from? The section linking to itself isn't useful at all, it needs to link to the chapter.

heh, of course I missed that :)

PR #2549, opened 2025-08-19
Losslessly optimize PNG files

lumiscosity:
Losslessly optimizes all of the PNG files in the repo. Done with:

```
oxipng -o max -a -s
oxipng -o max --zopfli -a -s
```


Noratrieb:
oh that's pretty cool, I wasn't even aware that PNGs can be optimized by this much :3

tshepang:
I did pause when I saw this pr, since it increases the size of the repo, but now I just thought of the great likelihood that the clones will be far less frequent than remote accesses.

lumiscosity:
Yeah, my reasoning is that it'll save on fetch time on the rendered docs. Push comes to shove, you can always clone the repo with reduced history.

PR #2556, opened 2025-08-28
dates: refresh diagnostics/ and tests/ annotated dates to August 2025

ali90h:
Updated explicit date annotations to 2025-08 where guidance can drift.

Addresses #2528.


ali90h:
r? rustc-dev-guide

PR #2557, opened 2025-08-28
dates: refresh infra/tooling date annotations to Aug 2025

ali90h:
Updated explicit date annotations to 2025-08 

Addresses https://github.com/rust-lang/rustc-dev-guide/issues/2528.

### DIFF
--- a/src/backend/updating-llvm.md
+++ b/src/backend/updating-llvm.md
@@ -1,6 +1,6 @@
 # Updating LLVM
 
-<!-- date-check: Aug 2024 -->
+<!-- date-check: Aug 2025 -->
 Rust supports building against multiple LLVM versions:
 
 * Tip-of-tree for the current LLVM development branch is usually supported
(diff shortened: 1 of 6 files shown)


ali90h:
r? rustc-dev-guide

jyn514:
you don't need reviews for small stuff like this.

oh i see you're not in the org—@rust-lang/wg-rustc-dev-guide any objection to adding ali to the team so they can get merge rights?

ali90h:
Thanks @jyn514 I am very happy about this.

tshepang:
@jyn514 I think we would want to see some sustained contributions before we move there... often it happens people contribute for a bit then leave, otherwise we will have lots of churn in team membership (where some members do not even respond to questions of if they want to continue being in the team). To be clear, this is not a blocking concern (I would not want to gatekeep), so maybe others can decide that the churn is not something to be concerned about. Also, we don't have a charter I know of (I should actually just ask/search how others were accepted into the team), and we probably must create one, to avoid the ad-hoc.

ali90h:
Thanks for the guidance and continued support. I’ll keep contributing and hope to live up to your expectations.

PR #2558, opened 2025-08-28
dates: refresh query-system date annotations to 2025-08 (preserve style)

ali90h:
Updated date-check annotations to **2025-08**, preserving the local format at each site.

**Addresses rust-lang/rustc-dev-guide#2528**


ali90h:
r? rustc-dev-guide

PR #2559, opened 2025-08-28
dates: refresh type system & traits date annotations to Aug/2025

ali90h:
Updated date-check annotations to the current month.

**Addresses rust-lang/rustc-dev-guide#2528**


ali90h:
r? rustc-dev-guide

PR #2561, opened 2025-08-29
make sentence more simple

tshepang:
I don't understand what "not (only) `rustdoc`" means

### DIFF
--- a/src/tests/compiletest.md
+++ b/src/tests/compiletest.md
@@ -90,7 +90,8 @@ The following test suites are available, with links for more information:
 | `rustdoc-ui`                         | Check terminal output of `rustdoc` ([see also](ui.md))                   |
 
 Some rustdoc-specific tests can also be found in `ui/rustdoc/`.
-These check rustdoc-related or -specific lints that (also) run as part of `rustc`, not (only) `rustdoc`.
+These check lints that are either related or specific to rustdoc,
+that also run as part of `rustc`, not (only) `rustdoc`.
 Run-make tests pertaining to rustdoc are typically named `run-make/rustdoc-*/`.
 
 [rustdoc-html-tests]: ../rustdoc-internals/rustdoc-test-suite.md


src/tests/compiletest.md
```diff
@@ -90,7 +90,8 @@ The following test suites are available, with links for more information:
 | `rustdoc-ui`                         | Check terminal output of `rustdoc` ([see also](ui.md))                   |
 
 Some rustdoc-specific tests can also be found in `ui/rustdoc/`.
-These check rustdoc-related or -specific lints that (also) run as part of `rustc`, not (only) `rustdoc`.
+These check lints that are either related or specific to rustdoc,
+that also run as part of `rustc`, not (only) `rustdoc`.
```
jyn514: this was not a great sentence even before your change.

some background:
1. bootstrap allows abbreviating Step paths named x/y/z as z. if there are multiple paths that end with z, it effectively picks a random path (it’s deterministic, but not predictable without reading the code).

this sentence is using `rustdoc` in a way that makes it look like it refers to `x test tests/ui/rustdoc`. i am not convinced that is even the suite path that runs when you say `x test rustdoc`, and i have no idea what `x test rustc` does, maybe unit tests for rustc-main (i.e. effectively a no-op).

i *think* the intent is actually to refer to the CLI tools themselves, not the paths passed to bootstrap, but that’s very unclear from context.

2. `tests/ui/rustdoc` is running rustc, not rustdoc. `tests/ui/rustdoc/README.md` says this clearly IMO:

> This directory is for tests that have to do with rustdoc, but test the behavior
of rustc. For example, rustc should not warn that an attribute rustdoc uses is
unknown.

 i would suggest something like this:
```suggestion
These check lints and behaviors *of rustc* that are related to rustdoc. Note that [not all lints emitted by rustc are emitted by rustdoc](https://rustc-dev-guide.rust-lang.org/rustdoc.html#constraints).
```

fmease: I guess I wrote these sentences and they're indeed quite poorly worded, oh my! I must've written them with a tired brain powering through the endless iterations of PR https://github.com/rust-lang/rustc-dev-guide/pull/2298 I simply wanted to have merged and be done with :'D

fmease: This isn't meant to refer to specific compiletest test suites, with "*lints that are run as part of $program*" I was trying to express the fact that some lints are [emitted] / some lint [impls and/or passes] are run by *both* the binaries, rustc *and* rustdoc and we want to ensure that they (i.e., a specific subset of lints) are not only emitted / run when executing rustdoc but also when executing rustc.

tshepang:
@fmease are you happy with latest change

src/tests/compiletest.md
```diff
@@ -90,7 +90,8 @@ The following test suites are available, with links for more information:
 | `rustdoc-ui`                         | Check terminal output of `rustdoc` ([see also](ui.md))                   |
 
 Some rustdoc-specific tests can also be found in `ui/rustdoc/`.
-These check rustdoc-related or -specific lints that (also) run as part of `rustc`, not (only) `rustdoc`.
+These tests ensure that lints that are emitted as part of executing rustdoc
```
fmease: Not sure how to rephrase this, but it's not (all) lints, it's only some lints. There are lints that are only emitted by rustdoc, not rustc, like the ones prefixed with `rustdoc::`.

fmease:
> [@]fmease are you happy with latest change

That's already *miles* better, thank you! It's still not quite correct (see inline review comment).

src/tests/compiletest.md
fmease: maybe "certain" lints?

tshepang:
am happy if you can push directly to the pr... I imagine you can

jieyouxu:
r? fmease

src/tests/compiletest.md
BoxyUwU:
```suggestion
These tests ensure that certain lints that are emitted as part of executing rustdoc
```

BoxyUwU:
do we really block merging to dev guide based off any changes requested... thats...

PR #2583, opened 2025-09-08
remove untrue/contentious statement

tshepang:
One can use unstable features on stable too, like via an env var

r? ghost

### DIFF
--- a/src/contributing.md
+++ b/src/contributing.md
@@ -52,9 +52,7 @@ channels: stable, beta, and nightly.
 
 - **Stable**: this is the latest stable release for general usage.
 - **Beta**: this is the next release (will be stable within 6 weeks).
-- **Nightly**: follows the `master` branch of the repo. This is the only
-  channel where unstable, incomplete, or experimental features are usable with
-  feature gates.
+- **Nightly**: follows the `master` branch of the repo.
 
 See [this chapter on implementing new features](./implementing_new_features.md) for more
 information.


BoxyUwU:
That env var is not supposed to be used by anyone. It is very much true that for pretty much all purposes nightly is the only channel where you can use unstable features

Noratrieb:
yeah it's probably more useful to people for understanding the toolchains to leave it as is

tshepang:
the statement is misleading though, and it's not true... I will add _intended_

jieyouxu:
The current statement I find is accurate, and less confusing than trying to caveat it with "intended".

tshepang:
it would be accurate if a RUSTC_BOOTSTRAP did not exist, and that is something widely used

BoxyUwU:
RUSTC_BOOTSTRAP is very much not widely used

tshepang:
I should of chosen a different word

jieyouxu:
I actually find removing the intended usage to be more confusing[^alt], so I think we should retain current wording.

[^alt]: Or less helpful.

tshepang:
I see now how I made it less helpful. How about now...
- I restored the explanation
- I removed the false claim
- I added the intent (to imply that people should avoid using feature gates in beta and stable)

src/contributing.md
```diff
@@ -52,9 +52,9 @@ channels: stable, beta, and nightly.
 
 - **Stable**: this is the latest stable release for general usage.
 - **Beta**: this is the next release (will be stable within 6 weeks).
-- **Nightly**: follows the `master` branch of the repo. This is the only
-  channel where unstable, incomplete, or experimental features are usable with
-  feature gates.
+- **Nightly**: follows the `master` branch of the repo.
+  This is the only channel where unstable, incomplete, or experimental features
+  should be used (with feature gates).
```
jieyouxu: Suggestion: can you say something more like "are intended/designed to be usable through opt-in via feature gates"? (Well, ideally with a less confusing wording than what I wrote.) "Should" sounds a bit to me like a "weak recommendation/vibe", but this really is the intended stability versus experimentation design setup. `RUSTC_BOOTSTRAP` has nothing to do with the design intention, it is merely an internal implementation detail to make bootstrapping possible. I would also just say "unstable", since that's why we have different release channels. Whether said unstable feature is incomplete/experimental doesn't really matter.

tshepang:
hoping that is good now... feel free to improve furthe in a future pr

PR #2586, opened 2025-09-08
Fix broken Microsoft URL missing slash

ali90h:
The PowerShell error example had `https:/go.microsoft.com` missing a slash, making the URL invalid.

Fixed to `https://go.microsoft.com` so Windows users can access the documentation.


ali90h:
r? rustc-dev-guide

PR #2587, opened 2025-09-10
Add missing documentation for running tests with GCC backend

GuillaumeGomez:
Since we're now going to make GCC backend mandatory for CI (https://github.com/rust-lang/rust/pull/146414), we need to make it as easy as possible for contributors to be able to test their code with this backend as well.

r? @Kobzol

### DIFF
--- a/src/tests/codegen-backend-tests/cg_gcc.md
+++ b/src/tests/codegen-backend-tests/cg_gcc.md
@@ -1,3 +1,34 @@
 # GCC codegen backend tests
 
-TODO: please add some more information to this page.
+To test the GCC codegen backend, you need to add `"gcc"` into the `rust.codegen-backends`
+setting in `bootstrap.toml`:
+
+```toml
+rust.codegen-backends = ["llvm", "gcc"]
+```
+
+If you don't want to change your `bootstrap.toml` file, you can alternatively run your `x.py`
+commands with `--set rust.codegen-backends=["llvm", "gcc"]'`. For example:
+
+```bash
+x.py test --set 'rust.codegen-backends=["llvm", "gcc"]'
+```
+
+If you don't want to build `gcc` yourself, you also need to set:
+
+```toml
+gcc.download-ci-gcc = true
+```
+
+Then when running tests, add the `--test-codegen-backend gcc` option. For example:
+
+```bash
+./x.py test tests/ui --test-codegen-backend gcc
+```
+
+If you want to build the sysroot using the GCC backend, you need to set it first
+in `rust.codegen-backends`:
+
+```toml
+rust.codegen-backends = ["llvm", "gcc"]
+```
--- a/src/tests/codegen-backend-tests/cg_gcc.md
+++ b/src/tests/codegen-backend-tests/cg_gcc.md
@@ -1,34 +1,56 @@
-# GCC codegen backend tests
+# GCC codegen backend
 
-To test the GCC codegen backend, you need to add `"gcc"` into the `rust.codegen-backends`
-setting in `bootstrap.toml`:
+If you ran into an error related to tests executed with the GCC codegen backend on CI,
+you can use the following command to run tests locally using the GCC backend:
+
+```bash
+./x test tests/ui --set 'rust.codegen-backends = ["llvm", "gcc"]' --test-codegen-backend gcc
+```
+
+Below, you can find more information about how to configure the GCC backend in bootstrap.
+
+## Choosing which codegen backends are built
+
+The `rust.codegen-backends = [...]` bootstrap option affects which codegen backends will be built and
+included in the sysroot of the produced `rustc`. To use the GCC codegen backend, `"gcc"` has to
+be included in this array in `bootstrap.toml`:
 
 ```toml
 rust.codegen-backends = ["llvm", "gcc"]
 ```
 
-If you don't want to change your `bootstrap.toml` file, you can alternatively run your `x.py`
+If you don't want to change your `bootstrap.toml` file, you can alternatively run your `x`
 commands with `--set rust.codegen-backends=["llvm", "gcc"]'`. For example:
 
 ```bash
-x.py test --set 'rust.codegen-backends=["llvm", "gcc"]'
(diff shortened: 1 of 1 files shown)


src/tests/codegen-backend-tests/cg_gcc.md
```diff
@@ -1,3 +1,20 @@
...
+If you don't want to build `gcc` yourself, you also need to set:
+
+```toml
+gcc.download-ci-gcc = true
+```
+
+Then when running tests, add the `--test-codegen-backend gcc` option. For example:
+
+```bash
+./x.py test tests/ui --test-codegen-backend gcc
```
antoyo: With this flag, it is not necessary to add `gcc` to `codegen-backends` above, no?

antoyo:
cc @RalfJung

src/tests/codegen-backend-tests/cg_gcc.md
```diff
@@ -1,3 +1,20 @@
 # GCC codegen backend tests
 
-TODO: please add some more information to this page.
+To test the GCC codegen backend, you need to add `"gcc"` into the `rust.codegen-backends`
+setting in `bootstrap.toml`:
+
+```toml
+rust.codegen-backends = ["llvm", "gcc"]
```
RalfJung, line 19: Since this can hit any contributor in any PR, I think it should be possible to run these tests in a one-shot way without permanently altering one's configuration.

I don't know what exactly altering the config in this way does, but if it means during my *next* PR it will build this backend, then that's not great -- it's unlikely to be relevant again then.

Please put yourself into the shoes of a rustc contributor that has nothing to do with the GCC backend, and is unlikely to encounter it for most of their PRs -- and tell that contributor what they should do to fix the PR where this weird failure they didn't expect shows up. :)

src/tests/codegen-backend-tests/cg_gcc.md
Kobzol: It is still needed. The user could specify an external rustc sysroot that we don't build, and still ask for GCC to be used in tests.

`--test-codegen-backend` => which backend to use in tests
`codegen-backends` => which backends should be built and included in rustc's sysroot

src/tests/codegen-backend-tests/cg_gcc.md
Kobzol: This can be done with `./x test ... --test-codegen-backend gcc --set 'rust.codegen-backends=["llvm", "gcc"]'`. A bit of a mouthful, but it will likely be copy-pasted from the docs anyway, or stored in some local helper file.

src/tests/codegen-backend-tests/cg_gcc.md
antoyo: > `codegen-backends` => which backends should be built and included in rustc's sysroot

Does that mean that the sysroot will be built with LLVM with the config above?

Kobzol: If by sysroot you mean the standard library, then yes. To build the standard library with GCC, you'd need to set `codegen-backends = ["gcc"]`.

GuillaumeGomez:
I added more examples and also cases where you don't need to update `bootstrap.toml`.

src/tests/codegen-backend-tests/cg_gcc.md
RalfJung: Somewhere on this page, in a super obvious place that the reader will find immediately when going here the first time (i.e., probably at the top), should be a command to copy-paste to run the tests. That's still not the case, isn't it? I have to copy multiple things together to get the actual command I need to use.

I don't know how much more clear I can be about optimizing this for an inexperienced contributor who doesn't want to do much work on the GCC backend, they just want to fix their PR. Or even an experienced contributor who doesn't normally touch GCC but has their tests fail in GCC every 4 months (i.e., too rarely to remember the command). *Please*, when writing this, put yourself in the shoes of such a contributor, not in the shoes of someone who's been working on rustc and its GCC backend for years.

I think that command is
```
x.py test --set 'rust.codegen-backends=["llvm", "gcc"]' tests/ui --test-codegen-backend gcc
```
but I am not actually sure.

src/tests/codegen-backend-tests/cg_gcc.md
```diff
@@ -1,3 +1,34 @@
...
+commands with `--set rust.codegen-backends=["llvm", "gcc"]'`. For example:
+
+```bash
+x.py test --set 'rust.codegen-backends=["llvm", "gcc"]'
+```
+
+If you don't want to build `gcc` yourself, you also need to set:
+
+```toml
+gcc.download-ci-gcc = true
```
RalfJung: Do I also have to add this to `--set` to avoid waiting hours until GCC is built?

Why is `true` not the default?

GuillaumeGomez: > Do I also have to add this to` --set` to avoid waiting hours until GCC is built?

Yes.

> Why is `true` not the default?

Good question. But not something to be discussed in here I think. :wink:

Noratrieb: Yeah it should really be the default for any non-dist profile. No one wants to build GCC except for the few people working on it.

Kobzol: I agree: https://github.com/rust-lang/rust/pull/146435.

Kobzol:
Hmm, the documentation was kind of all over the place, and it seemed to me that it was a bit hard to understand which command does what. I tried to restructure it, let me know what you think.

FWIW, I think that it would make more sense to have a dedicated page for each backend, rather than a "testing with XYZ backend", but for now I think this is good enough.

Kobzol:
GCC is now downloaded by default, and the libgccjit path issue has been fixed. @GuillaumeGomez Unless you want to change something here, I'm fine with merging this to unblock https://github.com/rust-lang/rust/pull/146414.

GuillaumeGomez:
Let's merge as is, I'll send follow-ups as I continue to work on https://github.com/rust-lang/rust/pull/146414.

PR #2599, opened 2025-09-29
how-to-build-and-run.md: update "building compiler" section

tshepang:
These were not updated after a bootstrap change where std cannot be built with stage0 rustc

### DIFF
--- a/src/building/how-to-build-and-run.md
+++ b/src/building/how-to-build-and-run.md
@@ -226,16 +226,17 @@ Once you've created a `bootstrap.toml`, you are now ready to run
 `x`. There are a lot of options here, but let's start with what is
 probably the best "go to" command for building a local compiler:
 
-```bash
-./x build library
+```console
+./x build rustc
 ```
 
-This may *look* like it only builds the standard library, but that is not the case.
-What this command does is the following:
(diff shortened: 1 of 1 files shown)


src/building/how-to-build-and-run.md
```diff
@@ -226,16 +226,17 @@ Once you've created a `bootstrap.toml`, you are now ready to run
 `x`. There are a lot of options here, but let's start with what is
 probably the best "go to" command for building a local compiler:
 
-```bash
-./x build library
+```console
+./x build rustc
 ```
```
jieyouxu, line 231: Remark: hm, I wonder if we should suggest `./x build library` for producing a stage 1 rustc/std toolchain instead. But anyway, not for this PR.

tshepang: hm, an earlier version of this pr had that command... I think I should do that, since what we have here does not quite match the text that follows

Zalathar: Unfortunately, the current text (`./x build rustc`) is now incorrect, because that command doesn't produce a fully-functional compiler (since it doesn't have libraries in its sysroot).

So `./x build library` is actually the necessary command.

Zalathar: Fixed in #2610.

PR #2602, opened 2025-10-04
how-to-build-and-run.md: fix some text regarding bootstraping

tshepang:
See https://github.com/rust-lang/rustc-dev-guide/pull/2599/files#r2403841373

r? @jieyouxu


src/building/how-to-build-and-run.md
```diff
@@ -231,11 +231,12 @@ probably the best "go to" command for building a local compiler:
 ```
 
 What this command does is build `rustc` using the stage0 compiler and stage0 `std`.
+This produces the stage1 compiler.
 
-To build `rustc` with the in-tree `std`, use this command instead:
+To build the in-tree `std`, use this command instead:
```
jieyouxu: Nit: I would say to build the in-tree compiler _and_ std. Which is exactly why in practice this is usually more useful for contributors if they actually need to produce std-using programs or run tests.

jieyouxu: I probably also would use `./x build library`, I think I never used `./x build std` myself. Though I guess I might even drop `./x build rustc` or `./x build compiler` for that matter, and only say something like:

- `./x build library` -> produce stage 1 rustc/std toolchain
- `./x build library --keep-stage-std=1` -> if there's only compiler changes that do not need to rebuild std

tshepang: I prefer rustc/std over compiler/library since the latter feels less precise, especially since they are also directory names. Am also concerned of expanding this intro section more (and it already does point to `--keep` section, though I see now it was removed 257e73f3a30d70ef51509733258b36948a903039).

Noratrieb: I don't think "std" is an alias for "library". std will not build proc_macro and test, for example.

tshepang: > `./x build library --keep-stage-std=1` -> if there's only compiler changes that do not need to rebuild std

why would you not `./x build compiler` for this

jieyouxu: Because that only builds the compiler. If you want to run ui tests or build any non-`no_std` programs, having only the compiler is approximately useless. Or do you mean not using `--keep-stage-std` + `./x build library`?

tshepang: it looked like `./x build compiler` is equivalent to `./x build library --keep-stage-std 1`, but maybe not

tshepang:
@jieyouxu I made further changes which include your suggestion (and that of @Noratrieb)

src/building/how-to-build-and-run.md
```diff
@@ -224,37 +224,33 @@ You may want to have upwards of 10 or 15 gigabytes available to build the compil
...
-probably the best "go to" command for building a local compiler:
+probably the best "go to" command:
 
 ```console
-./x build rustc
+./x build library
 ```
 
-What this command does is build `rustc` using the stage0 compiler and stage0 `std`.
+What this command does is the following:
```
amanjeev:
```suggestion
This command does the following:
```

More direct and terse.

amanjeev:
I did approve this but maybe my approval should not be taken as seriously and someone more knowledgeable should weigh in before merging.

tshepang:
made obsolete by #2610

PR #2603, opened 2025-10-05
update libEnzyme file name, and add missing strict-aliasing flag

ZuseZ4:


src/autodiff/debugging.md
```diff
@@ -28,7 +28,7 @@ The actual numbers will depend on your code.
 To confirm that your previous step worked, we will use llvm's `opt` tool. find your path to the opt binary, with a path similar to `<some_dir>/rust/build/<x86/arm/...-target-tripple>/build/bin/opt`. also find `llvmenzyme-19.<so/dll/dylib>` path, similar to `/rust/build/target-tripple/enzyme/build/enzyme/llvmenzyme-19`. Please keep in mind that llvm frequently updates it's llvm backend, so the version number might be higher (20, 21, ...). Once you have both, run the following command:
 
 ```sh
-<path/to/opt> out.ll -load-pass-plugin=/path/to/llvmenzyme-19.so -passes="enzyme" -s
+<path/to/opt> out.ll -load-pass-plugin=/path/to/build/<target-tripple>/stage1/lib/libEnzyme-21.so -passes="enzyme" -enzyme-strict-aliasing=0  -s
```
wsmoses: tripple -> triple

also LLVMEnzyme-XX is the correct one here, not libEnzyme

ZuseZ4: Interestingly enough libEnzyme works! From my local shell:
```sh
~/path/to/rust2/build/x86_64-unknown-linux-gnu/llvm/build/bin/opt mod.ll -load-pass-plugin=/path/to/rust2/build/x86_64-unknown-linux-gnu/stage1/lib/libEnzyme-21.so -passes="enzyme" -enzyme-strict-aliasing=0 -o rel3.ll -debug-pass-manager -S
```

Which is good, because we don't build LLVMEnzyme anymore in our default rust+autodiff setup.

wsmoses: that's not guaranteed to work and may break if the llvm plugin linking interfaces/libraries change [and may also be system dependent]

probably not wise to put in the docs

ZuseZ4: I added a warning. The problem about LLVMEnzyme is that it's not build by default and I think it also requires LLVM's plugin infrastructure, which we don't enable in our LLVMs by default. I'll leave it for now and bring it up on the rust side when we discuss how to distribute Enzyme in CI.

wsmoses: I'm still really skeptical of putting something into the docs that is not supposed to work (but may only work by accident).

PR #2611, opened 2025-10-13
Overhaul GCC codegen backend section

Kobzol:
1) Merge the two sections about testing (with) cg_gcc into a single page
2) Show how to exclude tests from being executed when cg_gcc is used
3) Show how to ping members of wg-gcc-backend


GuillaumeGomez:
Thanks!

PR #2613, opened 2025-10-20
debugging.md: Remove wrong claim that LLVM bitcode is not the same as LLVM IR

Enselic:
They are the same thing since they can be converted back and forth without loss.

We hint at that in the second bullet point two paragraphs down:

>  * To get human readable versions of the LLVM bitcode, one just needs to convert
>    the bitcode (`.bc`) files to `.ll` files using `llvm-dis`, which should be in
>    the target local compilation of rustc.

That is also claimed in the last paragraph in this SO answer: https://stackoverflow.com/a/12954198/287761

### DIFF
--- a/src/backend/debugging.md
+++ b/src/backend/debugging.md
@@ -77,7 +77,7 @@ llvm-ir`). `--build-type=debug` emits code for debug builds. There are also
 other useful options. Also, debug info in LLVM IR can clutter the output a lot:
 `RUSTFLAGS="-C debuginfo=0"` is really useful.
 
-`RUSTFLAGS="-C save-temps"` outputs LLVM bitcode (not the same as IR) at
+`RUSTFLAGS="-C save-temps"` outputs LLVM bitcode at
 different stages during compilation, which is sometimes useful. The output LLVM
 bitcode will be in `.bc` files in the compiler's output directory, set via the
 `--out-dir DIR` argument to `rustc`.


tshepang:
so, these are different formats of the same data, so I would not call them the same thing

your change still looks like an improvement though, other than the commit message

Noratrieb:
"LLVM IR" is the abstract idea, bitcode being one way to format IR

ZuseZ4:
Just fyi, not all LLVM tooling can handle both. 
LLVM bc has magic bytes set, which provide information to certain tools.
I tried to pass llvm IR files to them and the GPU offload tools failed, which back then took me a bit to figure out since I also first thought they'd be interchangeable. I'm not sure if that's worth mentioning.

Enselic:
Strange because on https://llvm.org/docs/LangRef.html#introduction we can read

> The three different forms of LLVM are all equivalent. 

Maybe what you see is a bug?

ZuseZ4:
I think it's not a bug. The reason I was given was that parsing the full llvm-IR to understand it is an unreasonable amount of work for arbitrary llvm tools, checking the first magic bytes of a .bc file is a much easier and more reasonable requirement. 

cc @Kevinsala can you confirm that you can not always pass both .ll and .bc files to llvm tools? The openmp slack unfortunately already deleted the discussion. I think Joseph answered back then.

PR #2655, opened 2025-11-25
Update contributing.md with Rust Book reference

reddevilmidzy:
Added a reference to The Rust Book for more information on Rust's release channels.

r? rustc-dev-guide


tshepang:
there is an incorrect claim on the target page
> If you’re using a beta or stable release of Rust, you can’t use any feature flags.

jyn514:
@tshepang that book is targeted at beginners and is one of the first resources people see when they use rust. it is not an appropriate place to introduce RUSTC_BOOTSTRAP.

reddevilmidzy:
How about adding "by default" to make this technically accurate?

tshepang:
that would be an improvement @reddevilmidzy

PR #2657, opened 2025-11-27
Give an overview of our stability guarantees

Enselic:
In https://github.com/rust-lang/rust/pull/147952#issuecomment-3536204283 I learned `remote-test-client` is considered fully internal and that we can change its "public interface" whenever we want to.

I couldn't find this documented anywhere and figured it would be useful to add a section to rustc-dev-guide that gives an overview of our stability guarantees.

cc @jieyouxu


Enselic:
A whole new page requires review.

r? rustc-dev-guide

PR #2662, opened 2025-11-29
add coercions chapter and split out non hir typeck stuff

BoxyUwU:
This chapter feels a bit incomplete without accompanying docs for adjustments and expectations but that can definitely be a future PR :sparkles: cc rust-lang/rustc-dev-guide#2663

The structure of this chapter is kinda meh, it's mostly just a pile of information in not too much of a thought out reading order

### DIFF
--- a/src/hir-typeck/coercions.md
+++ b/src/hir-typeck/coercions.md
@@ -2,7 +2,7 @@
  
 Coercions are implicit operations which transform a value into a different type. A coercion *site* is a position where a coercion is able to be implicitly performed. There are two kinds of coercion sites: 
 - one-to-one
-- LUB (Least-Uppper-Bound)
+- LUB (Least-Upper-Bound)
 
 ```rust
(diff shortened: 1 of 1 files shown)


src/hir-typeck/coercions.md
```diff
@@ -0,0 +1,284 @@
+# Coercions
+ 
+Coercions are implicit operations which transform a value into a different type. A coercion *site* is a position where a coercion is able to be implicitly performed. There are two kinds of coercion sites: 
+- one-to-one
+- LUB (Least-Uppper-Bound)
```
spastorino:
```suggestion
- LUB (Least-Upper-Bound)
```

spastorino (approved):
This chapter looks very good :heart:. Very nice use of examples. Not sure if you want some other feedback but my usual way is, we can merge things that are better than the current state and then improve later.

PR #2672, opened 2025-12-03
adding wip benchmark example

ZuseZ4:
Still needs some debugging. Based on https://github.com/LLNL/RAJAPerf/blob/develop/src/apps/DEL_DOT_VEC_2D.hpp
@Sa4dUs @kevinsala


jyn514:
hi @ZuseZ4 ! what's the state of this PR? I know it's marked as a draft, but do you think it's better than no docs at all, can we merge it as-is? Or do you think it will be misleading in its current state?

ZuseZ4:
Directly starting the work here? Thanks!
We have an open frontend pr, so it's not worth adding outdated syntax anymore, we'll add some with the new PR instead.

Issue #2663, opened 2025-11-29
Type System Documentation Overhaul


BoxyUwU:
This issue is intended to be used as a bit of a central thread for my ongoing work to write a bunch of docs for type system stuff and make sure existing docs we have are up to date. A corresponding zulip thread can be found here: [#t-compiler/rustc-dev-guide > Type System Docs Rewrite](https://rust-lang.zulipchat.com/#narrow/channel/196385-t-compiler.2Frustc-dev-guide/topic/Type.20System.20Docs.20Rewrite/with/543632377)

This is part of the "Type System Documentation" project goal: rust-lang/rust-project-goals#405

This is an **unordered** list of general type system concepts that should have documentation in the dev guide. Some may already have pages in the dev guide and so their presence here just means they should be looked over for missing/incorrect/confusing information. Thanks to lcnr for helping put together the initial list:
- [ ] opaque types
    - [ ] non defining vs defining uses
    - [ ] member constraints (borrowck overlap)
    - [ ] checking item bounds
    - [ ] high level normalization/opaque type storage approach (new solver)
    - [ ] normalization incompleteness
    - [ ] method/function incompleteness
    - [ ] how does `use<...>` work
    - [ ] `'erased` regions causes problems with outlives item bounds in liveness
    - [ ] consistency across defining scopes
    - [ ] RPITIT inference? does this have special stuff
    - [ ] capturing of bound vars in opaques under binders, `Fn` bounds are somewhat special in relation to this
    - [ ] opaques inheriting late bound function parameters
- [ ] non opaque type, `impl Trait`
    - [ ] RPITIT in traits desugaring
    - [ ] `impl Trait` in bindings
    - [ ] APIT desugaring impl details
- [ ] const generics
    - [ ] anonymous constants
    - [ ] ConstArgHasType
    - [ ] TSVs vs RVs and generally upstream doc from lang meeting to dev guide
    - [ ] deterministic CTFE requirement
- [ ] HIR typeck
    - [ ] expectations (and how used incorrectly :3)
    - [ ] method lookup + assorted code cleanups
    - [ ] coercions
    - [ ] auto-deref/reborrows (in coercions/method selection)
    - [ ] closure signature inference
    - [ ] fudge_inference_if_ok :>
    - [ ] diverging block handling :3
    - [ ] fallback :3
- [ ] MIR borrowck
    - [ ] MIR typeck
        - [ ] why do we want two typecks
[...]

PR #2677, opened 2025-12-10
Initial docs for const generics

BoxyUwU:
This is a fairly minimal set of const generics docs. It doesn't talk much about `min_generic_const_args` but definitely should. It also doesn't talk much about `adt_const_params` but definitely should :3 though we mostly already have docs for that living in hackmd for one of the lang meetings. Just need to upstream those which I intend to do as part of drafting an RFC for adt_const_params

cc @lcnr  added a new type system invariant for the whole const eval determinism thing, in some sense this is a more specific version of other stuff like "normalizing assoc types should yield a single type" but it feels interesting enough to list it independently.

cc rust-lang/rustc-dev-guide#2663


src/const-generics.md
```diff
@@ -0,0 +1,222 @@
...
+    }
+    ```
+
+It is currently unclear what the right way to make `generic_const_parameter_types` work nicely with the rest of const generics is. 
+
+`generic_const_exprs` would have allowed for anon consts with types referencing generic parameters, but that design wound up unworkable. 
+
+`min_generic_const_args` will allow for some expressions (for example array construction) to be representable without an anon const and therefore without running into these issues, though whether this is *enough* has yet to be determined.
+
+## Checking types of Const Arguments
```
BoxyUwU, line 166: should talk about why `ty::Const` doesnt store a `Ty`

BoxyUwU: also give an intuition for what the Ty means on `Value`

src/solve/invariants.md
```diff
@@ -168,5 +168,10 @@ We lookup types using structural equality during codegen, but this shouldn't nec
 Semantically different `'static` types need different `TypeId`s to avoid transmutes,
 for example `for<'a> fn(&'a str)` vs `fn(&'static str)` must have a different `TypeId`.
 
+## Evaluation of const items is deterministic ✅
```
BoxyUwU, line 171: should talk about the asterisks here, mainly Location right now I think?

BoxyUwU: ok actually the PR fixing that unsoundness has already been r+'d so  :3

src/const-generics.md
```diff
@@ -0,0 +1,222 @@
...
+
+This also makes it much more likely that the compiler will ICE or atleast incidentally emit some kind of error if we *do* accidentally allow generic parameters in an anon const, as the anon const will have none of the necessary information in its environment to properly handle the generic parameters.
+
+```rust
+fn foo<T: Sized>() {
+    let a = [1_u8;  size_of::<*mut T>()];
+}
+```
+
+The one exception to all of the above is repeat counts of array expressions. As a *backwards compatibility hack* we allow the repeat count const argument to use generic parameters.
```
lcnr, line 120:
````suggestion
The one exception to all of the above is repeat counts of array expressions. As a *backwards compatibility hack* we allow the repeat count const argument to use generic parameters.

```rust
fn foo<T: Sized>() {
    let a = [1_u8;  size_of::<*mut T>()];
}
```
````
example is surprising when reading this, also, maybe use a sub-header here

File: src/const-generics.md [hunk 4]
```diff
@@ -0,0 +1,222 @@
...
+    ```rust
+    #![feature(adt_const_params, generic_const_parameter_types)]
+
+    fn foo<const N: usize, const M: [u8; N]>() {}
+
+    fn bar<const N: usize>() {
+        // There is no way to specify the const argument to `M`
+        foo::<N, { [1_u8; N] }>();
+    }
+    ```
```
lcnr, line 144: incorrect indentation

src/const-generics.md
```diff
@@ -0,0 +1,222 @@
...
+`min_generic_const_args` will allow for some expressions (for example array construction) to be representable without an anon const and therefore without running into these issues, though whether this is *enough* has yet to be determined.
+
+## Checking types of Const Arguments
+
+In order for a const argument to be well formed it must have the same type as the const parameter it is an argument to. For example a const argument of type `bool` for an array length is not well formed, as an array's length parameter has type `usize`.
+
+```rust
+type Alias<const B: bool> = [u8; B];
+//~^ ERROR: 
+```
```
reddevilmidzy, line 173: I have a question.
The code compiles when I only define it without using `Alias`. Is this intended?

BoxyUwU: no :pensive: realised this while writing this section. i wrote `ERROR: ` then opened playground to get the error message and realised it was broken then didnt update the section. need to fix this in the compiler rust-lang/rust#149774

Issue #2651, opened 2025-11-21
How to run fmt docs insufficient?


ZuseZ4:
Might be totally on me missing something obvious, but I've just spend some time with testing and asking chatgpt on how to format my pr here: https://github.com/rust-lang/rust/pull/149126

https://rustc-dev-guide.rust-lang.org/conventions.html?highlight=fmt#formatting-and-the-tidy-script
mentions a few ways, none of which actually formatted anything. I tried it in three ways, with the new folder and files committed, with them staged but not yet committed, and with them not added to git at all. In the third case I at least got notes from rustfmt/fmt that the folder was skipped.

@jieyouxu since you care about docs and tests (please let me know if you're more focused on other things nowadays and wan't me to stop pinging you)

BoxyUwU:
I think the main point of clarification here is that we don't format our test suite

jieyouxu:
The ui test suite is intentionally not formatted by tidy. You'll need to either manually wrap those test lines, or `// ignore-tidy-linelength`.

ZuseZ4:
Ah true, I forgot about that. I guess that should be in the docs then, I can write a sentence later.

PR #2681, opened 2025-12-11
clarify test suite is not formatted

BoxyUwU:
Fixes rust-lang/rustc-dev-guide#2651


BoxyUwU:
Oh really? That's news to me :laughing: I don't think I've ever ran into a test suite that gets formatted

jieyouxu:
It (run-make) was initially *not* formatted, but was changed to be formatted. You likely would not have noticed unless you were actually reviewing the PR :ferrisClueless:

jieyouxu:
> Oh really? That's news to me 😆 I don't think I've ever ran into a test suite that gets formatted

Yeah, see https://github.com/rust-lang/rust/blob/main/rustfmt.toml

PR #2685, opened 2025-12-12
Make test suite formatting remark more precise

jieyouxu:
r? BoxyUwU

### DIFF
--- a/src/conventions.md
+++ b/src/conventions.md
@@ -21,8 +21,15 @@ Instead, formatting should be done using `./x fmt`. It's a good habit to run
 Formatting is checked by the `tidy` script. It runs automatically when you do
 `./x test` and can be run in isolation with `./x fmt --check`.
 
-We do not check that tests under the `tests/` directory are formatted, and
-similarly the `./x fmt` command will not format any files under that directory.
+> **Note: Formatting and test suites**
+>
+> Most Rust source files under `tests/` directory are not formatted for reasons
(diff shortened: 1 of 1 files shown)


jieyouxu:
cc @ZuseZ4 does this seem clearer to you?

src/conventions.md
```diff
@@ -21,8 +21,15 @@ Instead, formatting should be done using `./x fmt`. It's a good habit to run
...
-similarly the `./x fmt` command will not format any files under that directory.
+> **Note: Formatting and test suites**
+>
+> Most Rust source files under `tests/` directory are not formatted for reasons
+> such as whitespace sensitivity, nature of snapshot tests, location-sensitive
+> comments and more.
+>
+> Consult the `ignore` entries in
+> <https://github.com/rust-lang/rust/blob/main/rustfmt.toml> for which test
+> files are not formatted.
```
jieyouxu, line 32: Remark: Not listing any test suites here because `rustfmt.toml` is the source of truth, specific tests can also get ignored, not just on a test suite granularity.

ZuseZ4:
I have a local, not checked in, GPU test that keeps triggering the fmt check, so I would have been a bit confused by docs saying that `/tests` is not formatted at all, though it's better than no docs. The new wording and link would have been more than enough to get me to not waste time on this, thanks!

jieyouxu:
Don't recall which test suite that would fall under, but yes, at times you may have to add specific tests if it doesn't make sense to be formatted.

PR #2696, opened 2025-12-14
Fix #1692: Add example for profiling diff locally

### LINKED ISSUE #1692: document how to get a cachegrind diff from a perf run
it shows a useful command in the 'detailed-query' view: https://perf.rust-lang.org/detailed-query.html?commit=4b87ed9d0f18166a26aadb5289bac004e058b90f&base_commit=2034b6d23ca18311a23273f2f08c381a04fa05d9&benchmark=serde_derive-1.0.136-check&scenario=incr-unchanged
and there's instructions on zulip for how to build from source to get debuginfo: https://rust-lang.zulipchat.com/#narrow/stream/247081-t-compiler.2Fperformance/topic/annotations.20for.20cachegrind.20diff.3F

Mattias-Petersson:
Added an example for profiling an external crate diff locally. rust-lang/rustc-dev-guide#1692 also mentions that the debuginfo-level = 1 should be mentioned, but that has been solved by a different commit and as such was not included here.

### DIFF
--- a/src/profiling/with_rustc_perf.md
+++ b/src/profiling/with_rustc_perf.md
@@ -34,15 +34,16 @@ To start, in the `rustc-perf` repo, build the collector, which runs the Rust com
 ```
 cargo build --release -p collector
 ```
-After this the collector can be located in `.\target\release\collector`, can be run locally with `bench_local` and expects the following arguments:
+The collector can then be run using cargo, specifying the collector binary. It expects the following arguments:
 - `<PROFILE>`: Profiler selection for how performance should be measured. For this example we will use Cachegrind.
 - `<RUSTC>`: The Rust compiler revision to benchmark, specified as a commit SHA from `rust-lang/rust`.
-Optional arguments allow running profiles and scenarios as described above. `--include` in `x perf` is instead `--exact-match`. More information regarding the mandatory and
+Optional arguments allow running profiles and scenarios as described above. More information regarding the mandatory and 
 optional arguments can be found in the [rustc-perf-readme-profilers].
 
-Then, for the case of generating a profile diff for the crate `serve_derive-1.0.136`, for two commits `<SHA1>` and `<SHA2>` in the `rust-lang/rust` repository, run the following
+Then, for the case of generating a profile diff for the crate `serve_derive-1.0.136`, for two commits `<SHA1>` and `<SHA2>` from the `rust-lang/rust` repository, 
+run the following in the `rustc-perf` repo:
 ```
-./target/release/collector profile_local cachegrind +<SHA1> --rustc2 +<SHA2> --exact-match serde_derive-1.0.136 --profiles Check --scenarios IncrUnchanged
+cargo run --release --bin collector profile_local cachegrind +<SHA1> --rustc2 +<SHA2> --exact-match serde_derive-1.0.136 --profiles Check --scenarios IncrUnchanged
 ```
 
 


Mattias-Petersson:
r? rustc-dev-guide

src/profiling/with_rustc_perf.md
tshepang, line 1: could you show this command instead

    cargo run --release --bin collector

this would remove the need to indicate where the exe is located

Mattias-Petersson: Done now, apologize for taking so long. I have also updated the branch to be on the latest main. (Took a few commits to get right but now it should be proper :)) Does it look okay now? :)

tshepang: I would not have bothered with updating the branch, unless there was a conflict with `main`. I have submitted https://github.com/rust-lang/rustc-dev-guide/pull/2712 to remove the bot messages.

Mattias-Petersson: Perfect, thank you for the update :) There was no conflict with main, just how I am used to working to rebase onto the latest main branch. Good to know that it is not necessary!

src/profiling/with_rustc_perf.md
```diff
@@ -28,6 +28,24 @@ You can use the following options for the `x perf` command, which mirror the cor
...
+## Example profiling diff for external crates
+It can be of interest to generate a local diff for two commits of the compiler for external crates. 
+To start, in the `rustc-perf` repo, build the collector, which runs the Rust compiler benchmarks as follows.
+```
+cargo build --release -p collector
+```
+The collector can then be run using cargo, specifying the collector binary. It expects the following arguments:
+- `<PROFILE>`: Profiler selection for how performance should be measured. For this example we will use Cachegrind.
+- `<RUSTC>`: The Rust compiler revision to benchmark, specified as a commit SHA from `rust-lang/rust`.
+Optional arguments allow running profiles and scenarios as described above. `--include` in `x perf` is instead `--exact-match`. More information regarding the mandatory and
```
tshepang: what does this mean
> `--include` in `x perf` is instead `--exact-match`

src/profiling/with_rustc_perf.md
```diff
@@ -28,6 +28,24 @@ You can use the following options for the `x perf` command, which mirror the cor
...
+optional arguments can be found in the [rustc-perf-readme-profilers].
+
+Then, for the case of generating a profile diff for the crate `serve_derive-1.0.136`, for two commits `<SHA1>` and `<SHA2>` from the `rust-lang/rust` repository, run the following in the `rustc-perf` repo (where the collector was built in the previous step):
```
tshepang: regarding
> where the collector was built in the previous step

I would not show the command to build it, since the command to run it also builds it

tshepang:
sorry this is taking so long, but had to learn how this works so I can confidently approve... just really [one more thing](https://github.com/rust-lang/rustc-dev-guide/pull/2696#discussion_r2656316613) before I can do so

src/profiling/with_rustc_perf.md
Mattias-Petersson: Here I try to mention how in https://github.com/rust-lang/rustc-perf/blob/master/collector/README.md this README the benchmark options after running locally is `--exact-match`. This option does not exist when running /x perf, instead you can include what is necessary. Maybe that is unclear and should be revised, what do you think?

Mattias-Petersson: I removed the part entirely in a new commit, I can see how it was confusing! :)

src/profiling/with_rustc_perf.md
Mattias-Petersson: True, missed this, removed in the latest commit.

Mattias-Petersson:
> sorry this is taking so long, but had to learn how this works so I can confidently approve... just really [one more thing](https://github.com/rust-lang/rustc-dev-guide/pull/2696#discussion_r2656316613) before I can do so

No worries at all. It was the same for me making this commit :) I appreciate you taking thorough time to look through it ensuring things are correct!

PR #2707, opened 2025-12-23
Fix function formatting in src/overview.md

reddevilmidzy:
rustc_expand::module::parse_external_mod a is a method, so `()` seems to be needed.


reddevilmidzy:
r? rustc-dev-guide

BoxyUwU:
Im kinda surprised the existing paths here use `()`. It'd make total sense to not do that, if anything it makes it more consistent with how we tend to write docs with intradoc links :thinking: Can you instead change the existing links in this paragraph to not use `()` ?

BoxyUwU:
thx

PR #2352, opened 2025-04-25
Enable [behind-upstream] triagebot option

xizheyin:
Let's try running it on a repository first.

Doc: https://forge.rust-lang.org/triagebot/behind-upstream.html

r? @Urgau 

cc @Kobzol

PR #2712, opened 2025-12-28
stop checking if PRs are behind main

tshepang:
rust-lang/rustc-dev-guide#2352 was done as a test, and it was known at the time that it would not be so useful on a repo such as this


jieyouxu (approved):
Yeah, agreed, it's not super relevant for docs

PR #2718, opened 2026-01-04
Add a description of how FileCheck is used for mir-opt tests

meithecatte:
This had me quite confused until I figured out that, [as documented in the README of `tests/mir-opt`](https://github.com/rust-lang/rust/blob/HEAD/tests/mir-opt/README.md), the FileCheck runs on the runtime MIR, rather than the output of the `EMIT_MIR` directive. This PR documents this in the dev guide as well, to make this information easier to find.

### DIFF
--- a/src/tests/compiletest.md
+++ b/src/tests/compiletest.md
@@ -411,6 +411,17 @@ problematic in the presence of pointers in constants or other bit width
 dependent things. In that case you can add `// EMIT_MIR_FOR_EACH_BIT_WIDTH` to
 your test, causing separate files to be generated for 32bit and 64bit systems.
 
+On top of the snapshot-based checks performed on the output requested by `EMIT_MIR`,
+the *final runtime MIR*, as emitted by `--emit=mir`, is checked with the LLVM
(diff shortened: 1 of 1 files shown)


meithecatte:
r? rustc-dev-guide

jyn514 (approved):
this seems fine. do you mind also documenting `//@ test-mir-pass`?

i notice that the in-tree docs at tests/mir-opt say that the format for verifying the output of a pass is `x.after.mir`, but the dev-guide is inconsistent; sometimes it says `x.mir` and sometimes it says `x.after.mir`. do you mind changing it to `x.after.mir` everywhere for consistency?

i'm wondering now if it makes sense to just inline everything in the in-tree README into the guide and change it to just link here ...

Noratrieb:
having those docs duplicated sounds bad, yeah

jyn514:
I'm going to close this in favor of https://github.com/rust-lang/rustc-dev-guide/issues/2974. I do think this is on the right track, I just don't think this PR should land as-is. Feel free to re-open when you have more time to work on this :)

Issue #2687, opened 2025-12-12
nits for const generics docs


BoxyUwU:
````suggestion
The one exception to all of the above is repeat counts of array expressions. As a *backwards compatibility hack* we allow the repeat count const argument to use generic parameters.

```rust
fn foo<T: Sized>() {
    let a = [1_u8;  size_of::<*mut T>()];
}
```
````
example is surprising when reading this, also, maybe use a sub-header here

_Originally posted by @lcnr in https://github.com/rust-lang/rustc-dev-guide/pull/2677#discussion_r2614465517_

khyperia:
@BoxyUwU this issue has been fixed, wanna close this? :3

the PR linked above didn't say the magic github funny words so it didn't auto-close this issue

(sorry, I'm just scrolling through doing some exploration/learning and stumbled across this, haha)

PR #2719, opened 2026-01-08
docs:improve const generics

xonx4l:
This PR improves const generics docs as discussed in https://github.com/rust-lang/rustc-dev-guide/issues/2687


BoxyUwU:
thank you

PR #2720, opened 2026-01-09
Retire outdated stage0 std remarks

jieyouxu:
After the stage0 rework, our bootstrap chain is moreso the flat

```
precompiled                     in-tree
(stage0 rustc => stage0 std) => (stage1 rustc => stage1 std)
```

(where `=>` here means "builds" or "is used to build")

as opposed to the chain previously

```
precompiled     in-tree (`cfg(bootstrap)`)             in-tree (`cfg(not(bootstrap))`)
stage0 rustc => stage0 std                          => (stage1 rustc => stage1 std)
```

so these library dances should no longer be needed, as we always build in-tree library with in-tree compiler, and the in-tree compiler will know about the new target.

### DIFF
--- a/src/building/new-target.md
+++ b/src/building/new-target.md
@@ -80,32 +80,12 @@ will then add a corresponding file for your new target containing a
 
 Look for existing targets to use as examples.
 
-After adding your target to the `rustc_target` crate you may want to add
-`core`, `std`, ... with support for your new target. In that case you will
-probably need access to some `target_*` cfg. Unfortunately when building with
-stage0 (a precompiled compiler), you'll get an error that the target cfg is
-unexpected because stage0 doesn't know about the new target specification and
-we pass `--check-cfg` in order to tell it to check.
-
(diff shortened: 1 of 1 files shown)


jieyouxu:
cc @fee1-dead

fee1-dead:
I will want to investigate whether the infra around `STAGE0_MISSING_TARGETS` is actually needed...

jieyouxu:
> I will want to investigate whether the infra around `STAGE0_MISSING_TARGETS` is actually needed...

For bootstrap it's a sanity check, since in parts of bootstrap we unfortunately have to make assumptions based on the existing target set (so other parts of the build system / test infra don't have to).

Ideally, the target specs would be moreso "pure definitions" that can be shared between bootstrap/test infra/compiler/library, but you know, it's a lot of work (design/impl wise). And ideally, you'd have "structured target spec query APIs" like being able to know which plausible target archs there are, etc. etc. (Custom targets throw a spanner in the works there.)

PR #2728, opened 2026-01-15
Added section on using GitHub Dev for PR inspection

reddevilmidzy:
Added some simple review tips

r? rustc-dev-guide

### DIFF
--- a/src/git.md
+++ b/src/git.md
@@ -495,6 +495,14 @@ command to check it out locally. See <https://cli.github.com/> for more info.
 
 ![`gh` suggestion](./img/github-cli.png)
 
+### Using GitHub dev
+
+As an alternative to the GitHub web UI, GitHub Dev provides a web-based editor for browsing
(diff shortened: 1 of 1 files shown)


src/git.md
```diff
@@ -495,6 +495,13 @@ command to check it out locally. See <https://cli.github.com/> for more info.
 
 ![`gh` suggestion](./img/github-cli.png)
 
+### Using GitHub dev
+
+For quick and lightweight inspection of PR changes, you can use the GitHub Dev web-based editor.
```
BoxyUwU: I don't know that i'd call it "quick and lightweight". The quickest and lightweightest thing is to just click "files changes" :sweat_smile: This github.dev thing requires manually cahnging the url, waiting for it to take ages to load, and then you've got a bunch of pop ups and setup to do to get a whole IDE.

I would cut this and just say its an option

reddevilmidzy: You're right, the wording was too strong😄 I've updated it.  thanks!

Issue #1544, opened 2023-01-05
Document how to use rust-analyzer with rustc_private crates out-of-tree


jyn514:
https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/.E2.9C.94.20How.20to.20get.20rust-analyzer.20in.20vscode.20for.20OOT.20rustc.20plugins/near/319659236

jonboh:
@jyn514 I just found myself having this problem setting up the environment to develop a lint for clippy, the fix (as stated on the zulip thread) is pretty easy to implement, but it took me a while to get here. 
I could make a PR with the write up needed to document how to configure rust-analyzer with rustc_private.
I'm still in the (early) process of reading the rustc-dev-guide so I'm not sure where it would be a good place to leave this tip. 
Let me know if a specific section comes to mind and whether the explanation should cover what the option "discover" in "rust-analyzer.rustc.source" does or how it works. If nothing comes to mind I can try on myself and have a reviewer feedback afterwards.

PR #2729, opened 2026-01-16
Add rust-analyzer setup for out-of-tree rustc_private crates

reddevilmidzy:
close: rust-lang/rustc-dev-guide#1544
Is the document location okay?

r? jyn514

cc: @bjorn3

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -140,7 +140,7 @@
 
 - [Command-line arguments](./cli.md)
 - [rustc_driver and rustc_interface](./rustc-driver/intro.md)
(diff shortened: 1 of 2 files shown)


jyn514 (approved):
whoa, this is really cool! thank you for writing this up :D

the location here is ok but the title of the page makes it hard to find. could you rename it to something like "External rustc_drivers"?

reddevilmidzy:
> the location here is ok but the title of the page makes it hard to find. could you rename it to something like "External rustc_drivers"?

I've applied your suggestion! thank you for review :)

PR #2733, opened 2026-01-19
replace _ with - in path names (for consistency)

tshepang:
r? ghost


BoxyUwU:
Is there a recommended style for mdbooks?

tshepang:
I have not seen any

BoxyUwU:
im not sure what my preference is of `a-b` vs `a_b`, regardless of what we do though we should docuemnt it somewhere and add a CI check so the consistency doesn't regress

tshepang:
I wrote a tool to help with the conversion, but it's too limited to place here... I will make it better and add it to ci

tshepang:
> im not sure what my preference is of `a-b` vs `a_b`

there is a general preference for `a-b` in URLs

PR #2749, opened 2026-01-29
all but one test suite support revisions directive

tshepang:
r? rustc-dev-guide


tshepang:
forgot about that... I think it does, but let me check

fmease:
`rustdoc-html` and `rustdoc-json` don't support revisions.

tshepang:
looks like I conflated `TestMode` and `TestSuite`... will fix

Issue #1158, opened 2021-07-01
Document feature gate checking 


rylev:
There is a [blank chapter](https://github.com/rust-lang/rustc-dev-guide/blob/master/src/feature-gate-ck.md) on feature gate checking that should either be filled out or removed. 

There is a small mention of how to check whether a [feature is enabled](https://github.com/rust-lang/rustc-dev-guide/blob/master/src/implementing_new_features.md#stability-in-code), but this might not be sufficient.

reddevilmidzy:
I'd like to work on this issue. I'll document feature gate checking.
@rustbot claim

PR #2750, opened 2026-01-30
Document feature gate checking

reddevilmidzy:
close: rust-lang/rustc-dev-guide#1158

I tried to document it by looking at the code, and I welcome any feedback 😀

r? rustc-dev-guide

### DIFF
--- a/src/feature-gate-check.md
+++ b/src/feature-gate-check.md
@@ -5,6 +5,8 @@ nightly-only `#![feature(...)]` opt-in. This chapter documents the implementatio
 of feature gating: where gates are defined, how they are enabled, and how usage
 is verified.
 
+<!-- data-check: Feb 2026 -->
+
 ## Feature Definitions
 
 All feature gate definitions are located in the `rustc_feature` crate:
@@ -30,7 +32,10 @@ Before AST validation or expansion, `rustc` collects crate-level
   `removed` tables:
   - **Removed** features cause an immediate error.
   - **Accepted** features are recorded but do not require nightly. On
-    stable/beta they trigger the "already stabilized" diagnostic.
+    stable/beta, `maybe_stage_features` in
+    [`rustc_ast_passes/src/feature_gate.rs`] emits the non-nightly
+    diagnostic and lists stable features, which is where the "already
+    stabilized" messaging comes from.
   - **Unstable** features are recorded as enabled.
   - Unknown features are treated as **library features** and validated later.
 - With `-Z allow-features=...`, any **unstable** or **unknown** feature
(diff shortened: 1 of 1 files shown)


src/feature-gate-ck.md
```diff
@@ -1,3 +1,133 @@
...
+
+### `check_crate`
+
+`check_crate` performs high-level validation:
+
+- `maybe_stage_features`: Rejects `#![feature]` on stable/beta.
+- `check_incompatible_features`: Ensures incompatible feature combinations
+  (declared in `rustc_feature::INCOMPATIBLE_FEATURES`) are not used together.
+- `check_new_solver_banned_features`: Bans features incompatible with
+  certain compiler modes (e.g., the next trait solver).
```
tshepang: looks like this is specifically about next_solver
```suggestion
- `check_new_solver_banned_features`: Bans features incompatible with
  compiler mode for the next trait solver.
```

src/feature-gate-check.md
```diff
@@ -1,3 +1,133 @@
...
+
+- The visitor uses helper macros (`gate!`, `gate_alt!`, `gate_multi!`) to check:
+  1. Is the feature enabled?
+  2. Does `span.allows_unstable` permit it (for internal compiler macros)?
+- Examples include `trait_alias`, `decl_macro`, `extern types`, and various
+  `impl Trait` forms.
+
+### Checking `GatedSpans`
+
+`check_crate` iterates over `sess.psess.gated_spans`:
```
tshepang, line 75: this make it seem like it's not a continuation of line 65... maybe line 65 could have a link to this, or that the content of this section should be added under line 65

src/feature-gate-ck.md
```diff
@@ -1,3 +1,133 @@
...
+
+- [`rustc_ast_passes::check_attribute`] inspects attributes against
+  `BUILTIN_ATTRIBUTE_MAP`.
+- If the attribute is `AttributeGate::Gated` and the feature isn’t enabled,
+  `feature_err` is emitted.
+
+### `cfg` options
+
+- [`rustc_attr_parsing/src/attributes/cfg.rs`] defines
+  `find_gated_cfg`/`gate_cfg` to reject gated `cfg`s.
```
tshepang: I see `find_gated_cfg` is defined elsewhere

src/feature-gate-ck.md
```diff
@@ -1,3 +1,133 @@
...
+
+Before AST validation or expansion, `rustc` collects crate-level
+`#![feature(...)]` attributes to build the active `Features` set.
+
+- The collection happens in [`rustc_expand/src/config.rs`] in [`features`].
+- Each `#![feature]` entry is classified against the `unstable`, `accepted`, and
+  `removed` tables:
+  - **Removed** features cause an immediate error.
+  - **Accepted** features are recorded but do not require nightly. On
+    stable/beta they trigger the "already stabilized" diagnostic.
```
tshepang: how do I follow the logic to verify that the diagnostic only happens on stable/beta

tshepang:
I would add a date-check and also change the filename, perhaps `-check` instead of `-ck`

src/feature-gate-ck.md
reddevilmidzy: Ah, I changed it to `rustc_feature::find_gated_cfg`. Thank you.

src/feature-gate-check.md
reddevilmidzy: Oh, that's right, I changed the location of the section and added a link just in case

src/feature-gate-ck.md
reddevilmidzy: Added explanation!

reddevilmidzy (commented):
thank you for the review

tshepang:
thanks... learned a lot reviewing this

just needs a git conflict fix

reddevilmidzy:
> thanks... learned a lot reviewing this
> 
> just needs a git conflict fix

Resolved the conflict by rebasing,

I learned a lot while documenting it too. 😃

tshepang:
There is another chapter on feature gates, as well as other mentions of feature gates in the book. I wonder how we can harmonize them, because they, at the moment, mostly don't know about each other.

reddevilmidzy:
> There is another chapter on feature gates, as well as other mentions of feature gates in the book. I wonder how we can harmonize them, because they, at the moment, mostly don't know about each other.

I was thinking about that early in the process, but I seem to have forgotten about it 😅

These two topics are related to feature gates, but they cover different topics. So, I'll keep the current document location and add sentences and links to connect them!

PR #2761, opened 2026-02-02
link to a policy on using new external crates

tshepang:
Ticks a box in rust-lang/rustc-dev-guide#2759

### DIFF
--- a/src/crates-io.md
+++ b/src/crates-io.md
@@ -6,8 +6,8 @@ Rust Forge has [official policy for vetting new dependencies].
 
 ## Permitted dependencies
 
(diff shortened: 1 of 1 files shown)


Noratrieb:
this is not really a replacement, it's about extracting compiler code into crates

tshepang:
oh, so it looks like I must link to https://forge.rust-lang.org/compiler/third-party-out-of-tree#third-party-crates instead then

Noratrieb:
> and that, where possible, a compiler team member is added as a maintainer

huh this is weird and definitely not done

tshepang:
not sure that is a blocking comment... will merge this for now though

PR #2762, opened 2026-02-02
sembr a few files

tshepang:
r? ghost

### DIFF
--- a/src/compiler-debugging.md
+++ b/src/compiler-debugging.md
@@ -55,8 +55,7 @@ If this is not desirable, you can prevent the ICE file from being created with `
 
 When you have an ICE (panic in the compiler), you can set
 `RUST_BACKTRACE=1` to get the stack trace of the `panic!` like in normal Rust programs.
-IIRC backtraces **don't work** on MinGW,
-sorry.
+IIRC backtraces **don't work** on MinGW, sorry.
 If you have trouble or the backtraces are full of `unknown`,
 you might want to find some way to use Linux, Mac, or MSVC on Windows.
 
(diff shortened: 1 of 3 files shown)


src/backend/debugging.md
```diff
@@ -145,29 +145,29 @@ $ ./build/$TRIPLE/llvm/bin/llvm-extract \
 
 If you are seeing incorrect behavior due to an optimization pass, a very handy
 LLVM option is `-opt-bisect-limit`, which takes an integer denoting the index
-value of the highest pass to run.  Index values for taken passes are stable
+value of the highest pass to run.
+Index values for taken passes are stable
```
BoxyUwU, line 149: why did this do a line break?

BoxyUwU:
this does generally seem nicer :) I like splitting at the end of sentences instead of half way through :sweat_smile:

src/backend/debugging.md
tshepang: it splits on `.`,
and it only reflows the next sentence if combining with the one that follows it would fit in 100 chars

BoxyUwU: why is there a newline after "denoting the index" then, there's no `.` there

tshepang: changing that would mean having a line longer than 100 chars
```
LLVM option is `-opt-bisect-limit`, which takes an integer denoting the index value of the highest pass to run.
```
was thinking to also break on `,` unless the next `.` would easily fit in the 100 chars
```
LLVM option is `-opt-bisect-limit`,
which takes an integer denoting the index value of the highest pass to run.

BoxyUwU: hmm. I feel like I would prefer to have a line longer than 100 lines than split it arbitrarily in the middle of a line

BoxyUwU: though splitting on commas does also seem potentially reasonable to me :)

tshepang: Noting that the line ended like that, so it was not arbitrary splitting... the tool does not do that. It does look not ideal though, so I will work on splitting on commans to reduce the awkward.

PR #2778, opened 2026-02-18
Update `try_green_mark` part

reddevilmidzy:
~~`try_green_mark` method has been moved~~

While I was looking at the code, I found that the existing one was from 7 years ago and is slightly different from the current structure, so I will update this as well.

Ah, the link has been fixed here https://github.com/rust-lang/rust/pull/152703


src/queries/incremental-compilation-in-detail.md
reddevilmidzy, line 111: I thought this part of the code was old and would be good to change, but it seems appropriate for pseudo code, so I'll just leave it.

PR #2780, opened 2026-02-19
rustdoc.md: consistently use `./x` rather than `./x.py`

DanielEScherzer:
Most of the commands were specified with `./x`, but three of the commands were using `./x.py` without a clear reason - align with the guide's convention of using `./x`.


DanielEScherzer:
r? rustc-dev-guide

PR #2785, opened 2026-02-28
Mention `@bors squash` in squashing documentation

Kobzol:
It is probably the easiest way of squashing.

### DIFF
--- a/src/git.md
+++ b/src/git.md
@@ -383,6 +383,11 @@ Both the upside and downside of this is that it simplifies the history.
 On the one hand, you lose track of the steps in which changes were made, but
 the history becomes easier to work with.
 
+The easiest way to squash your commits in a PR on the `rust-lang/rust` repository is to use the `@bors squash` command in a comment on the PR. By default, [bors] combines all commit messages of the PR into the squashed commit message. To customize the commit message, use `@bors squash msg=<commit message>`.
+
+
+If you want to squash commits using local git operations, read on below.
+
 If there are no conflicts and you are just squashing to clean up the history,
 use `git rebase --interactive --keep-base main`.
 This keeps the fork point of your PR the same, making it easier to review the diff of what happened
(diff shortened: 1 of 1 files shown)


BoxyUwU:
does bors squash have a way of setting the commit description?

Kobzol:
By description, do you mean multi-line commit messages? Otherwise, this PR states how to override the commit message :sweat_smile:

tshepang:
@Kobzol you probably missed https://github.com/Kobzol/rustc-dev-guide/commit/e334fa0e4cb3d9a576726ba0cab3243b254cebb9

Kobzol:
Lol, I did indeed. I think that we should mention it at the top of the section though, otherwise people will just start with the local operations (case in point, I didn't notice the `@bors squash` documentation is already in the same section that I modified :) ), which is the more annoying option and can more likely lead to git issues. WDYT?

BoxyUwU:
> By description, do you mean multi-line commit messages?

oh is that all a description in a commit message is?  :laughing: I guess so :) Specifically I was thinking of this commit I approved today which has all the extra junk in addition to the main commit message:
https://github.com/rust-lang/rust/pull/153050/changes/efc150e5b3de648871a9ce88b2ca0acf38b066fd

It sounds like just setting the commit message will replace *everything* with the one thing you specify?

Kobzol:
Yeah.

src/git.md
reddevilmidzy: ~~I'm mentioning this in case you missed it, but it's being introduced here! 
If the description seems appropriately placed at the top, I think it would be best to remove this section.
(It seems like there's exactly the same content in the same section.)~~

EDIT: Ah, I missed tshepang's comment 😅

jieyouxu (approved):
Thanks

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2785/38b9dd24b3ea52a32bff76410742ca41d3c8d138..7088bc52c8805ffc9a5958af2c85d411125ef3ce)*

Kobzol:
I combined my new paragraph and the previous paragraph, and kept the new text at the top. Thanks for letting me know!

PR #2833, opened 2026-04-13
show instead of explain

tshepang:
This should be clear with less words

r? @reddevilmidzy


reddevilmidzy:
I thought about doing that, but the [bors help page](https://bors.rust-lang.org/help) also uses <commit message>. Since quotes aren’t strictly needed for a single word (even though that’s rare for commit messages😅), I left it as is. But I agree it looks cleaner this way 👍

tshepang:
yeah, that bors help page could also do with `"<commit-message>"`

PR #2838, opened 2026-04-17
effects.md: update const trait syntax

DanielEScherzer:
`~const` was replaced with `[const]`, making uses of the former confusing


DanielEScherzer:
Cross-link rust-lang/rust#139858
No write access, need a review
r? rustc-dev-guide

fmease:
Thanks!

PR #2864, opened 2026-05-06
Fix name of first compiled stage, as stage0 is downloaded

cammeresi:


reddevilmidzy (approved):
This looks correct to me. `stage0` is the downloaded beta compiler itself, while `stage1` is built using that compiler, so `stage1` is more accurate here.


*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2864/cefef0e25cce5eddbaa2c84b66cb1e0438863233..392049d720ec486754b55bab43df83cf8b9a1f71)*

PR #2877, opened 2026-05-20
Update quickstart.md with stage2 note

Its-Just-Nans:
Added note about stage2 dependency in rustup toolchain linking.

### DIFF
--- a/src/building/quickstart.md
+++ b/src/building/quickstart.md
@@ -37,6 +37,7 @@ After building the compiler and standard library, you now have a working
 compiler toolchain. You can use it with rustup by linking it.
 
 ```sh
+# note that sometimes you can only have the stage2 depending on the build config
 rustup toolchain link stage1 build/host/stage1
 ```
 
--- a/src/building/quickstart.md
+++ b/src/building/quickstart.md
@@ -40,8 +40,11 @@ compiler toolchain. You can use it with rustup by linking it.
 rustup toolchain link stage1 build/host/stage1
 ```
 
-Depending on your configuration and the stage picked, you may only have
-the `stage2` instead of the `stage1`. Change the commands accordingly.
+**NOTE**: If you use `./x setup tools`, the default stage will be set to 2 instead of 1.
+Adjust your command accordingly:
+```sh
+rustup toolchain link stage2 build/host/stage2
+```
 
 Now you have a toolchain called `stage1` linked to your build. You can use it to
 test the compiler.
--- a/src/building/quickstart.md
+++ b/src/building/quickstart.md
@@ -37,9 +37,11 @@ After building the compiler and standard library, you now have a working
 compiler toolchain. You can use it with rustup by linking it.
 
 ```sh
-# note that sometimes you can only have the stage2 depending on the build config
 rustup toolchain link stage1 build/host/stage1
 ```
+
+Depending on your configuration and the stage picked, you may only have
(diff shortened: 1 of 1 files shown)


Its-Just-Nans:
r? rustc-dev-guide

src/building/quickstart.md
```diff
@@ -37,6 +37,7 @@ After building the compiler and standard library, you now have a working
 compiler toolchain. You can use it with rustup by linking it.
 
 ```sh
+# note that sometimes you can only have the stage2 depending on the build config
```
jyn514: I think without context, and especially in the quickstart guide, this is just confusing. Is there a default `setup` profile where stage 1 doesn't exist (maybe `tools`)? Can we name that profile explicitly rather than this cryptic comment?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2877/f147f7708f496453c35a5f60201b57fc0a831b7a..e615aec97c7f1b3334a583c2a7bb61cae98d5cfc)*

jyn514 (changes requested):

src/building/quickstart.md
Its-Just-Nans: hi

Yes, thanks for the review. It's just that I followed the guide and don't have the stage1 and only the stage2

I think this is linked to https://rustc-dev-guide.rust-lang.org/building/bootstrapping/what-bootstrapping-does.html?highlight=stage%202#building-the-stages

image

So this command and the subsequent using `stage1` were not working for me



I updated my MR with a separated note

jyn514: Your separate note still doesn't address my concern, it's not clear when it's relevant. Can you post your config.toml and the exact bootstrap command you ran, please?

Its-Just-Nans:
```
 cat bootstrap.toml 
# See bootstrap.example.toml for documentation of available options
#
profile = "tools"  # Includes one of the default files in src/bootstrap/defaults
change-id = 154587
```

Yes you are right, I chosen the tools

src/building/quickstart.md
```diff
@@ -40,6 +40,9 @@ compiler toolchain. You can use it with rustup by linking it.
 rustup toolchain link stage1 build/host/stage1
 ```
 
+Depending on your configuration and the stage picked, you may only have
+the `stage2` instead of the `stage1`. Change the commands accordingly.
```
jyn514:
````suggestion
**NOTE**: If you use `./x setup tools`, the default stage will be set to 2 instead of 1.
Adjust your command accordingly:
```sh
rustup toolchain link stage2 build/host/stage2
```
````

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2877/f147f7708f496453c35a5f60201b57fc0a831b7a..5e22ebb6e17589e4c8e544c20ea628d8c2781920)*

jyn514 (changes requested):

src/building/quickstart.md
Its-Just-Nans: Thanks!

jyn514 (approved):
thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2877/f147f7708f496453c35a5f60201b57fc0a831b7a..d36ef9731342263d2d48a44d00b3086c1613f7f4)*

PR #2889, opened 2026-06-03
Well-formedness dev guide page

fallible-algebra:
Initial work off of discussions with boxy + chasing hyperlinks. Part of rust-lang/rustc-dev-guide#2663.

Few unsolved items here:

- [x] **Conceptual**: Items "contain" Terms? Is this the right way to say this? I imagine so, syntactically, but I need a vibe check.
- [x] Remove the "boxy said so" citations
- [x] "Term"[^term] needs to be better specified as a concept. Maybe added to the glossary.
    - May end up as "Type-Level Term" again.
- [x] Wfck doesn't appear on its own, but as `item-wfck` or `term-wfck` (alt: term well-formedness checking for ths one specifically).
- [ ] Run the formatter that deals with maximum line length.
- [x] Remove unused links

I've also ended up doing some syntax here "from scratch", which I could use some input on. Specifically for showing off obligations I've done:

```
<Items or Terms>
---
<Obligations and (relevant) Substitutions>
```

If this works fine, cool. Otherwise, open to being told to do it a different way :)

Finally, I'm putting this in an `analysis` subfolder as I've been told we need to do some structural reordering of the mdbook directory, and this should *eventually* live in an `analysis` subfolder so might as well have it there now.

[^term]: This was "Type-Level Term" in these files before I find-replaced it all to Term, which is maybe too vague to our target audience.

### DIFF
--- a/src/analysis/well-formed.md
+++ b/src/analysis/well-formed.md
@@ -1,0 +1,293 @@
+# Well-Formedness
+
+## What is Well-Formedness?
+
+"Well-formed" means "correctly built."[^wf-history] Something is _well-formed_ when its structure follows rules. When we use this term in the Rust compiler we are concerned with establishing some kind of _internal consistency_.
+
+## Well-Formedness in Rust
+
+To check that something is well-formed is to perform a "Well-formedness check."
+
+In the Rust compiler there are two different forms of well-formedness checking:
+
+- **Type-Level Term**[^terms][^terms-abbreviated] well-formedness check.
+    - Also called "Term well-formedness" or "Term well-formedness checking."
+    - Not a distinct analysis stage, this gets performed throughout analysis. 
+- **Item**[^items] well-formedness check (item-wfck.)
+    - "Item-wfck" will often wind up requiring Terms be well-formed, but skips some areas.
+    - Inner "Terms" can (incorrectly) get normalized first.
+    - More coherent as a stage in the compiler than "term well-formedness" (which is performed in many places.)
+
+See: [What Well-Formedness Isn't](#what-well-formedness-isnt).
+
+## Well-Formedness of Type-Level Terms
+
+Term well-formedness checking begins with building a list of things that need to be true for a term to be well-formed. We call these "Obligations"[^obligations].
+
+Type-Level Terms are considered Well-Formed when their associated obligations are satisfied by the trait solver.
+
+### Obligations for Well-Formedness
+
+Specific obligations are things like `String: Clone`, `A: usize`, or `<T as Iterator>::Item: Debug`.
+
+On this page we show the split between obligations and terms/items as:
+
+```rust,ignore
+<terms or items>
+---
+<obligations>
+```
+
+Here is an example of a well-formed type-level term:
+
+```rust,ignore
+Vec<String>
+---
+// Obligations to fulfill
+Vec<T> where T: Sized
+// Trait solver says `String: Sized` is true, so this is well-formed.
+Vec<String> where String: Sized
+```
+
+When we compute the obligations for `Vec<String>`, we'll find that `Vec<T>` generates the obligation `T: Sized`. We substitute `T` with `String` in `Vec<String>`, so we find the obligation `String: Sized` which the trait solver will determine to be satisfied.
+
+The following **is not** well-formed:
+
+```rust,ignore
+Vec<str>
+---
+// Obligations to fulfill
+Vec<T> where T: Sized
+// Trait solver says `str: Sized` is not true, so this is not well-formed.
+Vec<str> where str: Sized
+```
+
+The above computes the obligation `T: Sized`, like before, but we substitute `T` for `str` in the instance of `Vec<str>` finding the obligation `str: sized`. This obligation will be determined by the trait solver to be _unsatisfied_.
+
+#### Determining Obligations
+
+In the compiler, obligations of terms are found through the [`obligations`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_trait_selection/traits/wf/fn.obligations.html) function in the [term well-formedness module](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_trait_selection/traits/wf/index.html).
+
+#### Other Obligations
+
+Obligations are more than just trait and const generic bounds, but we've only mentioned these specific obligations so far as they are what we care about when we do "well-formedness checking" of terms. See: [`PredicateKind`](https://doc.rust-lang.org/beta/nightly-rustc/rustc_type_ir/predicate_kind/enum.PredicateKind.html) and [`ClauseKind`](https://doc.rust-lang.org/beta/nightly-rustc/rustc_type_ir/predicate_kind/enum.ClauseKind.html) for a full list of obligations.
+
+### We Don't Need Normalization (Yet)
+
+[Normalization](../normalization.md) is the process of resolving [type aliases](../normalization.md#aliases) into their underlying type. 
+
+A type alias is considered well-formed if its where clauses are satisfied. The underlying type undergoes well-formedness checking at most definition and instantiation sites, but there are exceptions.
+
+### Const Generic Arguments
+
+Term well-formedness is responsible for getting "type checking" obligations of const generic terms[^tyck-const-generics]. Let's look at the following use of const generics:
+
+```rust,ignore
+fn use_const_generics<const U: usize>() { /* ... */ }
+// call site
+use_const_generics::<6>();
+---
+// call site wfck obligations
+const 6: usize
+```
+
+The call site will provide us with the obligation `6: usize` during well-formedness checking. This obligation will be passed off to the trait solver just like any trait-style obligation, as the trait solver has more responsibilities than its name suggests.
+
+## Well-Formedness of Items
+
+Items are, generally speaking, "Things that get defined." Item-wfck happens at the signature level for types and functions, methods, and definitions/implementations of traits.
+
+```rust,ignore
+// The `Vec<str>` is checked during item wfck
+fn foo(_: Vec<str>) {
+    // The `Vec<[u8]>` is not handled by item wfck as it's not in the signature
+    let _: Vec<[u8]>
+}
+---
+Vec<str>: Sized // Generated
+Vec<[u8]>: Sized // Not done at item-wfck. Done elsewhere.
+```
+
+Item-wfck has more responsibilities than only collecting the obligations of its internal type-level terms and passing them to the trait solver. We do not talk about all of these here, but they can be found at the individual `check_*` functions in [**the item-wfck module**](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir_analysis/check/wfcheck/index.html).
+
+<!-- FIXME: Expand more on item well-formedness that isn't const generic / trait bound obligation based. These are not special cases, but important points! -->
+
+### Global and Trivial Bounds
+
+<!-- TODO later: Cut this into its own page -->
+
+Trait bounds are a common Obligation. Global and Trivial trait bounds are kinds of trait bounds where we already have enough information to determine if they are true or false. Item-wfck is responsible for finding and checking these bounds.
+
+- **Global bounds** are, in the old solver, post-normalization bounds that don't contain any generic parameters (like `<T>` or `'a`) or bound variables (like `for<'b>`).
+- **Trivial bounds** are bounds that do not need further normalization to determine if they're well-formed or not. <!-- TODO: check with lcnr if this is genuinely what a trivial bound is. -->
+
+Consider the following function definition:
+
+```rust,ignore
+fn apartment_complex<T>(block: T, name: String) where String: Clone { /* ... */ }
+---
+String: Clone // Trivial & Global bound! There's no aliases to resolve.
+// There could be bligations on T but we don't care about them here.
+```
+
+This produces a trait bound obligation `String: Clone` that is _Global_ (no generic parameters) and _Trivial_ (didn't require normalization to be well-formedness checked). The trait solver doesn't need to be given any additional information for it to be able to make a judgment on the well-formedness of `String: Clone`.
+
+False trivial bounds are simply trivial bounds that do not hold. The following is a basic example:
+
+```rust,ignore
+fn apartment_simple<T>(block: T, name: String) where String: Copy { /* ... */ }
+---
+String: Copy // Trivial bound again, but this one is false!
+```
+
+Here we have a trivial bound that does not hold, because `String` is not `Copy`.
+
+#### Trivial Bounds Are Not Always Global
+
+Trivial Bounds are not a subset of Global Bounds. A trivial bound that isn't Global is `for<'a> String: Clone` (trivially true, has a bound variable) or `&'a str: Copy` (trivially false, has a generic parameter).
+
+#### Item-Wfck and Trivial/Global Bounds
+
+<!-- When cutting out the subsection on global/trivial bounds, keep this part on the well-formedness page. -->
+
+When checking items are well-formed we will check that there are no trivially false global bounds.
+
+## When We Don't Fully Do Well-Formedness Checking
+
+Well-formedness checking is not a coherent "stage" of type checking. There are many areas where well-formedness checking is performed, and some areas where we skip over well-formedness checking due to limitations in what kinds of analysis we can currently perform. Ideally, we would never skip or defer well-formedness checking.
+
+### We (Sometimes) Need Normalization
+
+There are places where normalization of an Item happens before its Terms have gone through well-formedness checking. This is considered problematic as doing so allows some terms to [bypass term well-formedness checking entirely](https://github.com/rust-lang/rust/issues/100041).
+
+### Trait Objects
+
+We do not require the where clauses of trait objects to be well-formed when determining if that trait object is well-formed. These where clauses are proven when coercing into a trait object, but this remains a hole in well-formedness checking.
+
+As an example, the following will compile because we don't have a point where we're constructing the trait object from a concrete type:
+
+```rust,ignore
+trait Trait
+where 
+    for<'a> [u8]: Sized {}
+
+fn foo(_: &dyn Trait) {}
+---
+// This doesn't end up being generated, because it happens within a trait object.
+[u8]: Sized
+```
+
+The above should not compile because `[u8]: Sized`, but this won't be checked until actual use:
+
+```rust
+trait Trait
+where 
+    for<'a> [u8]: Sized {}
+    
+fn foo(_: &dyn Trait) {}
+
+// We still need to specify the bound here, otherwise `[u8]: Sized` _is_
+// checked as an obligation. 
+impl Trait for u8 where for<'a> [u8]: Sized {}
+
+fn main() {
+    // No matter what we do, this boundary between concrete type and trait
+    // object will produce the obligation `[u8]: Sized`, which will fail when
+    // handed over to the trait solver.
+    let object: Box<dyn Trait> = Box::new(42u8);
+    foo(&object);
+}
+```
+
+This exception does not apply to Const Generic Arguments in trait objects:
+
+```rust,ignore
+trait Trait<const N: usize> {}
+fn foo<const B: bool>(_: &dyn Trait<B>) {}
+---
+const N: usize
+const B: bool
+N = B // Substitution
+const B: usize + bool 
+```
+
+The above doesn't compile, unlike the previous example we gave. We're doing _some_ well-formedness checking here when it comes to the const generic arguments.
+
+### Binders / Higher-Ranked Types
+
+Binders / Higher-Ranked Types reduce the amount well-formedness checking we do on a term, leaving well-formedness checking to when the bound is instantiated:
+
+```rust,ignore
+let _: for<'a> fn(Vec<[&'a ()]>);
+---
+// This doesn't end up being generated, because it happens within a HRB
+[&'a ()]: Sized // slices aren't sized, this would fail!
+```
+
+Specifically, obligations involving variables from binders (`for<'a>`) are only checked when the binder is instantiated. Some things are stilled checked under the `for<'a>`, but we still skip a lot of things.
+
+A lot of unsoundness surrounds this behavior. See: [#25860](https://github.com/rust-lang/rust/issues/25860), [#84591](https://github.com/rust-lang/rust/issues/84591). 
+
+Let's consider the following:
+
+```rust,ignore
+for<'a, 'b> fn(&'a &'b ())
+```
+
+The above HRB implies `'b: 'a` (a lifetime bound), rather than two completely separate lifetimes. This is normal lifetime behavior, but during well-formedness checking we cannot prove that this bound is generally true[^horrible], so we skip it.
+
+### Free Type Aliases
+
+The right-hand side of Free Type Aliases[^fta] is not fully checked to be well-formed at the definition site, only the types of const generic arguments in the RHS are checked.
+
+The following free type alias passes type checking, at time of writing:
+
+```rust,ignore
+type WorksButShouldNot = Vec<str>;
+---
+// This should fail! But we skip the RHS of free type aliases
+str: Sized // Not generated
+```
+
+This shouldn't work, as both `T: Sized`, `str: Sized` are implied by `Vec<T>`. This "passes" item-wfck because the RHS of a free type alias doesn't go through well-formedness checking _until it's used_. Item-wfck is **deferred until use** for this specific case.
+
+For Const Generics we still do a small amount of well-formedness checking at the definition site of a free type alias. This is consistent with our current special-casing of const generic well-formedness checking when we skip over things like where bounds.
+
+This means that the following, despite being of a similar form to the above example, fails as it should:
+
+```rust,ignore
+pub struct Consty<const A: bool>;
+type Alias = Consty<42>;
+---
+// This *is* generated as an obligation, so this (correctly) fails.
+42: bool // This is generated!
+```
+
+<!-- TODO: Link to something explaining the underlying "why" of the difference between const and trait well-formedness checking in FTAs, or eliminate that difference. Whatever comes first. -->
+
+## "Well-Formed" or "Wellformed"?
+
+Prefer "well-formed" over "wellformed," as this is consistent with logic literature. This also gets abbreviated to WF in other parts of the dev guide / docs.
+
+## Informal Usage
+
+In conversation, contributors may refer to something as "well-formed" and not necessarily mean what we cover here because "well-formedness" is a general phrase associated with the correctness of formal structures. This isn't necessarily in error, but it should be looked out for.
+
+## What Well-Formedness Isn't
+
+Well-formedness checking is not "number of parameters" or "parameter type" checking[^kind-checking]. Neither term well-formedness checking nor item-wfck is concerned with if a type with 2 parameters has 1 or 3 types applied to it (assuming no defaults), or if a const generic parameter has a type applied to it. These kinds of problems will get handled during HIR-ty Lowering[^hir-ty-lower], not wfck.
+
+Well-formedness doesn't check or validate lifetimes, this is handled in [MIR](../borrow-check.md).
+
+Well-formedness in the Rust compiler doesn't correspond to "correct syntax" as it does in logic. The term has a history of general use in a mathematical context of "follows a given set of rules." In Rust, our original usage was closer to "this thing is internally consistent" with respect to the bounds on a type in places such as the original [clarification on projections and well-formedness RFC](https://github.com/rust-lang/rfcs/blob/master/text/1214-projections-lifetimes-and-wf.md).
+
+[^obligations]: These get referred to as Obligations, Requirements, or Constraints in the documentation. Preferred term is "obligations", as this matches the suffix of the type and the names of relevant functions. In future, this may be superseded by the new solver's term "Goal."
+[^wf-history]: In linguistics this is "grammatically correct," in logic it is "syntactically correct," and in casual mathematician use it can be read as a more general "follows the rules we set for this domain."
+[^horrible]: Instead, this bound is checked during "MIR borrowck" when the lifetimes are instantiated.
+[^fta]: Type aliases not associated with anything, i.e. a module-level `type Alias = Vec<u8>;`.
+[^items]: "Definition" style things in rust, See the [glossary](../appendix/glossary.md).
+[^terms]: AKA Type expressions and subexpressions in the general sense, not a specific struct or enum in the rust compiler. See the [glossary](../appendix/glossary.md).
+[^terms-abbreviated]: Abbreviated as "Terms" on this page in some areas.
+[^kind-checking]: AKA "kind checking", as we might see in languages like Haskell.
+[^hir-ty-lower]: <https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir_analysis/hir_ty_lowering/index.html>
+[^tyck-const-generics]: <https://rustc-dev-guide.rust-lang.org/const-generics.html#checking-types-of-const-arguments>
--- a/src/appendix/glossary.md
+++ b/src/appendix/glossary.md
@@ -102,6 +102,7 @@ Term                                           | Meaning
 <span id="trans">trans 👎</span>                  |  Short for _translation_, the code to translate MIR into LLVM IR. **Renamed to** [codegen](#codegen).
 <span id="ty">`Ty`</span>                      |  The internal representation of a type. ([see more](../ty.md))
 <span id="tyctxt">`TyCtxt`</span>              |  The data structure often referred to as [`tcx`](#tcx) in code which provides access to session data and the query system.
+<span id="type-level-term">Type-Level Term</span> | An expression at the type level, such as a Type or Const Generic.
 <span id="ufcs">UFCS 👎</span>                 |  Short for _universal function call syntax_, this is an unambiguous syntax for calling a method. **Term no longer in use!** Prefer _fully-qualified path / syntax_. ([see more](../hir-typeck/summary.md), [see the reference](https://doc.rust-lang.org/reference/expressions/call-expr.html#disambiguating-function-calls))
 <span id="ut">uninhabited type</span>          |  A type which has _no_ values. This is not the same as a ZST, which has exactly 1 value. An example of an uninhabited type is `enum Foo {}`, which has no variants, and so, can never be created. The compiler can treat code that deals with uninhabited types as dead code, since there is no such value to be manipulated. `!` (the never type) is an uninhabited type. Uninhabited types are also called _empty types_.
 <span id="upvar">upvar</span>                  |  A variable captured by a closure from outside the closure.
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -191,6 +191,7 @@
         - [Sharing the trait solver with rust-analyzer](./solve/sharing-crates-with-rust-analyzer.md)
     - [`Unsize` and `CoerceUnsized` traits](./traits/unsize.md)
     - [Having separate `Trait` and `Projection` bounds](./traits/separate-projection-bounds.md)
+- [Well-Formedness](./analysis/well-formed.md)
 - [Variance](./variance.md)
 - [Coherence checking](./coherence.md)
 - [HIR Type checking](./hir-typeck/summary.md)


src/analysis/well-formed.md
```diff
@@ -0,0 +1,166 @@
...
+Vec<str>
+---
+// Obligations to fulfill
+Vec<T> where T: Sized
+Vec<str> where str: Sized // This is not true, therefore the term is not well-formed.
+```
+
+During wfck we encounter the obligation for `Vec<T>` that `T: Sized`. For `Vec<str>` we encounter the obligation `str: Sized`, and as `str` is not `Sized` the term is not well-formed.
+
+### We Don't Need Normalization (Yet)
```
BoxyUwU, line 75: when i talk about normalization im referring to the process of replacing aliases with the aliased type (see https://rustc-dev-guide.rust-lang.org/normalization.html), which is orthogonal from whether a term has generic parameters in it or not.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2889/47a57f628fbf80e2b309ee6577d8fffaadbdc2fd..bb9a51591a6ee9e30afac8f6d1a6a4fd6a6e2ba0)*

src/analysis/well-formed.md
```diff
@@ -0,0 +1,166 @@
...
+- Term[^terms] wfck.
+- Items[^items] wfck. 
+    - "Items wfck" can call into "Terms wfck" as Items contain Terms.
+    - Sometimes normalize their inner Terms first.
+
+Wfck is not "Kindedness" checking, as we might see in languages like Haskell. Wfck is not concerned with if a type with 2 parameters has 1 or 3 types applied to it (assuming no defaults), or if a const generic parameter has a type applied to it. These kinds of errors will get handled during HIR-ty Lowering[^hir-ty-lower], not wfck.
+
+Wfck doesn't check or validate lifetimes, this is handled in [MIR](../borrow-check.md).
+
+## Generating Obligations
```
BoxyUwU: I think this makes more sense as being part of the wfness of terms section, item wfck doesn't really start by doing this 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2889/47a57f628fbf80e2b309ee6577d8fffaadbdc2fd..bb9a51591a6ee9e30afac8f6d1a6a4fd6a6e2ba0)*

src/analysis/well-formed.md
```diff
@@ -0,0 +1,166 @@
...
+
+The first step of wfck is generating the list of things for a subject of the wfck that need to be true for that subject to be well-formed. 
+
+These end up being referred to as Obligations, Requirements, or Constraints. Prefer to call them obligations for now, as this matches the suffix of the type and the names of relevant functions. In future, this may be superseded by the Polonius term "Goal"[^boxy].
```
BoxyUwU: new solver is distinct from polonius :sparkles: polonius is new borrowck :relieved:

```suggestion
These end up being referred to as Obligations, Requirements, or Constraints. Prefer to call them obligations for now, as this matches the suffix of the type and the names of relevant functions. In future, this may be superseded by the new solver's term "Goal"[^boxy].
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2889/47a57f628fbf80e2b309ee6577d8fffaadbdc2fd..bb9a51591a6ee9e30afac8f6d1a6a4fd6a6e2ba0)*

src/analysis/well-formed.md
```diff
@@ -0,0 +1,166 @@
...
+struct Struct<T>(T);
+struct Set<A>(A) where A: Ord;
+Set<Struct<T>> where T: Ord
+---
+Struct<T>: Ord // This is true / well-formed if T: Ord (which it is here)
+```
+
+This produces an obligation that still has a generic in it. While more normalized versions of `Struct<T>` may not be `Ord`, we can say that `Set<Struct<T>>` is well-formed when `T: Ord`.
+
+### Const Generic Arguments
```
BoxyUwU, line 81: worth linking also to https://rustc-dev-guide.rust-lang.org/const-generics.html#checking-types-of-const-arguments 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2889/47a57f628fbf80e2b309ee6577d8fffaadbdc2fd..bb9a51591a6ee9e30afac8f6d1a6a4fd6a6e2ba0)*

src/analysis/well-formed.md
```diff
@@ -0,0 +1,166 @@
...
+
+```rust
+fn apartment_simple<T>(block: T, name: String) where String: Copy { /* ... */ }
+---
+String: Copy // Trivial bound again, but this one is false!
+```
+
+Here we have a trivial bound that does not hold, because `String` is not `Copy`.
+
+## When We Don't Fully Do Wfck
```
BoxyUwU: i think as a general rule we should never just say "wfck" on its own in this, we should always explicitly say if we're talking about wfness of terms or checking wf of items.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2889/47a57f628fbf80e2b309ee6577d8fffaadbdc2fd..bb9a51591a6ee9e30afac8f6d1a6a4fd6a6e2ba0)*

src/analysis/well-formed.md
```diff
@@ -0,0 +1,166 @@
+# Well-Formedness of Terms and Items
+
+The area of the analysis pipeline that deals with questions like "does `T: Debug` hold true for some data structure that uses T?" or "is this const generic parameter `const B: bool` being handed a value of the right type?"
```
BoxyUwU: this feels slightly off to me, i would say the trait solver deals with those questions. though i guess that kind of makes sense since the trait solver is also responsible for proving wf obligations (i.e. checking terms are wf when we require them to be)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2889/47a57f628fbf80e2b309ee6577d8fffaadbdc2fd..bb9a51591a6ee9e30afac8f6d1a6a4fd6a6e2ba0)*

src/analysis/well-formed.md
fallible-algebra: Right, got wires crossed with monomorphization. This makes why this is problematic make a lot more sense ^^

src/analysis/well-formed.md
```diff
@@ -0,0 +1,163 @@
...
+String: Copy // Trivial bound again, but this one is false!
+```
+
+Here we have a trivial bound that does not hold, because `String` is not `Copy`.
+
+## When We Don't Fully Do Well-Formedness Checking
+
+Well-formedness checking is not a coherent "stage" of type checking. It gets called from various contexts, and there are places where it gets skipped partially or entirely.
+
+### Trait Objects
```
BoxyUwU, line 163: Another example to illustrate this:
```rust
trait Trait
where
    for<'a> [u8]: Sized {}
    
fn foo(_: &dyn Trait) {}
```

this compiles even though `[u8]: Sized` doesn't hold. But if you try to actually construct an `&dyn Trait` we'll wind up checking all the where clauses hold and get an error.

probably a nicer example to lead with so you can separate out discussion of trait object wf jank from the const generics jank working around trait object wf jank lol

src/analysis/well-formed.md
```diff
@@ -0,0 +1,163 @@
...
+
+```rust,ignore
+fn apartment_simple<T>(block: T, name: String) where String: Copy { /* ... */ }
+---
+String: Copy // Trivial bound again, but this one is false!
+```
+
+Here we have a trivial bound that does not hold, because `String` is not `Copy`.
+
+## When We Don't Fully Do Well-Formedness Checking
```
BoxyUwU, line 155: Structurally it feels weird to lump this all under one header instead of wellformedness of terms and wellformedness of items having its own separate lists of "jank". Though I realise some things (e.g. const generics jank) would appear in both sections

src/analysis/well-formed.md
```diff
@@ -0,0 +1,163 @@
...
+    - This is the primary subject of "Well-Formedness Checking."
+- Items[^items] wfck (item-wfck). 
+    - "Items wfck" can call into "Term-wfck" as Items contain Terms.
+    - Sometimes normalize their inner Terms first.
+
+Wfck is not "number of parameters" or "parameter type" checking (AKA "kind checking"), as we might see in languages like Haskell. Neither term-wfck nor item-wfck is concerned with if a type with 2 parameters has 1 or 3 types applied to it (assuming no defaults), or if a const generic parameter has a type applied to it. These kinds of problems will get handled during HIR-ty Lowering[^hir-ty-lower], not wfck.
+
+Wfck doesn't check or validate lifetimes, this is handled in [MIR](../borrow-check.md).
+
+## Type-Level Terms
```
BoxyUwU: would in theory like this header to be "Well-Formedness of Type-Level Terms" to mirror the following "Well-Formedness of Items" header 🤔 I realise you use this name as a subheader of Type-Level terms but that feels a bit weird to me.

I don't think the primary point of this page is to explain what a "type level term" is so I wouldn't structure it too heavily around that, rather than what it means for one to be well formed

src/analysis/well-formed.md
```diff
@@ -0,0 +1,163 @@
...
+
+There are two different forms of well-formedness checking:
+
+- Type-Level Term[^terms] wfck (term-wfck).
+    - This is the primary subject of "Well-Formedness Checking."
+- Items[^items] wfck (item-wfck). 
+    - "Items wfck" can call into "Term-wfck" as Items contain Terms.
+    - Sometimes normalize their inner Terms first.
+
+Wfck is not "number of parameters" or "parameter type" checking (AKA "kind checking"), as we might see in languages like Haskell. Neither term-wfck nor item-wfck is concerned with if a type with 2 parameters has 1 or 3 types applied to it (assuming no defaults), or if a const generic parameter has a type applied to it. These kinds of problems will get handled during HIR-ty Lowering[^hir-ty-lower], not wfck.
```
BoxyUwU: I wonder if it'd make sense to have a more general header at the end like "what isn't covered by well formedness" so that this sort-of tangent doesn't get in the way, and also we can talk about other things which one might expect is covered by wfck but actually isnt

i guess problem statement: not a huge fan of having this big paragraph here at the very start of the page :3 it feels like it takes up a lot of space and is kinda scary (talks about haskell and "kind checking") while not being all that important for probably the average reader

src/analysis/well-formed.md
```diff
@@ -0,0 +1,163 @@
+# Well-Formedness of Items and Type-Level Terms
+
+The area of analysis that produces questions like "does `T: Debug` hold true for some data structure that uses T?" or "is this const generic parameter `const B: bool` being handed a value of the right type?" for the trait solver to answer.
```
BoxyUwU: something about this paragraph feels odd to me and im not sure I can place why. it might just be that it feels like it makes more sense in the "Well-Formedness of Type-Level Terms" section, but also that would feel a bit weird too.

I think leading with the second paragraph "Items and Type-Level Terms are 'well formed' when they ..." feels like it would work fine and also feels better to me for some reason?

I guess concretely leading with this feels weird because:
- item wfck does more than just produce obligations
- it implies that item wfck and wellformedness of terms are one singular analysis rather than two semi-separate things
- refers to an area of analysis that hasn't previously been referred to I guess :3

src/analysis/well-formed.md
```diff
@@ -0,0 +1,163 @@
...
+const B: usize + bool 
+```
+
+The above shouldn't compile, and yet it does. `foo`s const generic argument is a `bool`, while `Trait`'s is a `usize`. But because the wfck of trait objects doesn't happen until coercion into a concrete type, the above compiles just fine.
+
+### Binders / Higher-Ranked Bounds
+
+HRBs also skip the wfck on their subjects.
+
+TODO: bit of background of why this doesn't happen.
```
BoxyUwU: u need us to chat about this or just haven't gotten around to writing down what's in your notes?

src/analysis/well-formed.md
```diff
@@ -0,0 +1,163 @@
...
+// call site wfck obligations
+const 6: usize
+```
+
+Applying term-wfck to the call site will provide us with the obligation `6: usize`. This obligation will be passed off to the trait solver just like any trait-style obligation, as the trait solver has more responsibilities than its name suggests.
+
+## Well-Formedness of Items
+
+Items[^items] are, generally speaking, "Things that get defined." Item-wfck[^item-wf-module] only happens at the signature level for types and functions, including the methods and implementations. This doesn't happen for Free Type Aliases other than Const Generic Argument type checking.
+
```
BoxyUwU, line 99: probably a concrete example here of a function and going "things here are item wfck'd" and "things here are handled by normal type checking" would be useful

BoxyUwU (commented):
another round of feedback, I think slightly more nitpicky than last time? though still mostly just based around the general structure rather than too much about individual sentences/paragraphs/whatever

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2889/47a57f628fbf80e2b309ee6577d8fffaadbdc2fd..d9f8e3a37f572193bce6e72e301fc0d9ceaf838b)*

src/analysis/well-formed.md
```diff
@@ -0,0 +1,163 @@
+# Well-Formedness of Items and Type-Level Terms
+
+The area of analysis that produces questions like "does `T: Debug` hold true for some data structure that uses T?" or "is this const generic parameter `const B: bool` being handed a value of the right type?" for the trait solver to answer.
+
+Items and Type-Level Terms[^terms][^terms-abbreviated] are "well formed" when they "follow rules" AKA "fulfill obligations" or "meet the necessary constraints." When we're doing a well-formedness check (wfck) we're usually concerned about if Trait obligations are met, but this also covers obligations of the types broadly, including making sure that the types of const generic terms type check.
+
+There are two different forms of well-formedness checking:
+
+- Type-Level Term[^terms] wfck (term-wfck).
```
BoxyUwU: I think we discussed this a little 1:1 but to note it down here too, I feel unsure about the term "Term wfck". I feel like it has strong implications of it being a particular analysis/pass/something.

This *is* the case for item wfck but not as much for wellformedness of terms which is more of a general "thing you can get the trait solver to check" and is done all over the place 🤔 

I think things would be a lot closer to how I "expect" them to be if instead of saying, for example, "Item-wfck is one of the entry points where most term-wfck will get performed" we instead wrote like, "Item-wfck is one of the places where well-formedness of terms will get checked" 

definitely slightly nitpicky and not entirely structural, but also I guess it's kind of a semi-big-picture framing choice so feels somewhat important :3

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2889/47a57f628fbf80e2b309ee6577d8fffaadbdc2fd..d9f8e3a37f572193bce6e72e301fc0d9ceaf838b)*

src/analysis/well-formed.md
fallible-algebra: That's reasonable! I figured "problems I ran into while narrowing down what well-formedness is" just had to live somewhere, and that an earlier correction made more sense than a late one. But I'll move it down and replace this with a smaller "what this isn't" link.

src/analysis/well-formed.md
fallible-algebra: This is a helpful nudge ^^ I think I misinterpreted this originally as "always specify item-wfck vs term-wfck", rather than "'ck' has specific suffix implications and isn't just an abbreviation of check"

src/analysis/well-formed.md
fallible-algebra: Yeah agreed, this is a weird glossary out of nowhere. Restructuring 🫡

src/analysis/well-formed.md
fallible-algebra: It's a bit difficult. I like having them all here so there's a stronger emphasis on them as collective jank, but I'll see if I can rework them into the two major sections.

src/analysis/well-formed.md
fallible-algebra: We gotta chat about this :') This is lightly mentioned but any details are missing from my notes.

src/analysis/well-formed.md
fallible-algebra: I think I can see it? But also I'd like to put the high-level "input/output" of the processes described here somewhere close to the top. I think the following paragraph isn't concrete enough in comparison. I'll wiggle things around.

src/analysis/well-formed.md
BoxyUwU: didn't misinterpret :3 that was the point I was making but then seeing it written out made me unsure about term-wfck as a piece of terminology :sweat_smile: this is wholly nitpicky so feel free to ignore this and we can figure it out at a later point, just wanted to make sure i wrote it down and didnt forget

fallible-algebra: Gotcha 😌 I've been rotating this in my mind today regardless so I'm kinda fine with either I think?

src/analysis/well-formed.md
```diff
@@ -0,0 +1,194 @@
...
+    - "Item-wfck" can call into "Term well-formedness checking" as Items contain Terms.
+    - Inner "Terms" can get normalized first.
+
+See: [What Well-Formedness Is Not](#what-well-formedness-isnt)
+
+## Well-Formedness of Type-Level Terms
+
+Type-Level Terms are the data structures that well-formedness checking is focused on analyzing. Item-wfck is one of the entry points where term well-formedness checking will get performed.
+
+### Obligations
```
BoxyUwU: something feels vaguely off with this header 🤔 it states that "terms are wf when the obligations are met" like 3 times, and then we also follow it up with a new header ("When Type-Level Terms Are Well Formed") that answers this again.

i guess that it feels like it has a lot of redundancy within the header, and then also feels somewhat redundant and out of place when considering the wider context it's been placed in

src/analysis/well-formed.md
```diff
@@ -0,0 +1,194 @@
...
+
+Items[^items] are, generally speaking, "Things that get defined." Item-wfck[^item-wf-module] only happens at the signature level for types and functions, including the methods and implementations. This doesn't happen for Free Type Aliases other than Const Generic argument type checking.
+
+Items are a major entry point for term well-formedness. Because Items contain Terms, item-wfck can invoke term well-formedness checking.
+
+### We (Sometimes) Need Normalization
+
+Currently, there are places where normalization of an Item happens before its Terms have gone through well-formedness checking. This is considered problematic as doing so allows some terms to [bypass term well-formedness checking entirely](https://github.com/rust-lang/rust/issues/100041).
+
+### Global & Trivial Bounds
```
BoxyUwU: normalization is orthogonal from this stuff :3 well, there's some jank around normalization and whether we consider a bound to be global or not buuuut that's weird edge case jank and not relevant for this doc 😅 

in general should be somewhat careful with this heading as it's not really the place for this doc to explain what trivial & global bounds mean (ideally that probably gets its own document) and don't want it to wind up taking up too much space. I think the current size is like, fine. I think giving a short high level explanation here is also like, fine for now.

structure wise the point of talking about global/trivial bounds would be to say "hey, item wfck will implicitly error on false global bounds" and rn we don't do so which I think makes this header feel disconnected

src/analysis/well-formed.md
```diff
@@ -0,0 +1,194 @@
...
+String: Copy // Trivial bound again, but this one is false!
+```
+
+Here we have a trivial bound that does not hold, because `String` is not `Copy`.
+
+#### Trivial Bounds Are Not Always Global
+
+Trivial Bounds are not a subset of Global Bounds. A trivial bound that isn't Global is `for<'a> String: Clone` (trivially true, has a bound variable) or `&'a str: Copy` (trivially false, has a generic parameter).
+
+## When We Don't Fully Do Well-Formedness Checking
```
BoxyUwU, line 155: I think it'd make sense to move the discussion of how item wfck will normalize terms before checking wf of them, into this section.

BoxyUwU (commented):
hell yeah :) this feels about right structurally to me

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2889/47a57f628fbf80e2b309ee6577d8fffaadbdc2fd..b18833893178609947d4ad197c76fbfcc3be3d14)*

src/analysis/well-formed.md
fallible-algebra: That's reasonable, this kind of got away from me ^^

src/analysis/well-formed.md
```diff
@@ -0,0 +1,245 @@
...
+
+```rust,ignore
+for<'a, 'b> fn(&'a &'b ())
+```
+
+The above HRB implies `'b: 'a` (a lifetime bound), rather than two completely separate lifetimes. This is normal lifetime behavior, but during well-formedness checking we cannot prove that this bound is generally true[^horrible], so we skip it.
+
+### Free Type Aliases
+
+The right-hand side of Free Type Aliases[^fta] do not go through a full well-formedness check at the definition site, with the exception of shallowly[^shallow] "type checking" const generic parameters of the RHS.
```
BoxyUwU:
```suggestion
The right-hand side of Free Type Aliases[^fta] do not go through a full well-formedness check at the definition site, with the exception of "type checking" const generic arguments in the RHS.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2889/47a57f628fbf80e2b309ee6577d8fffaadbdc2fd..d257f8067b2ca042b96a0d9f7325458326f1ce76)*

BoxyUwU:
diff of the headers lol

```diff
+- Well-Formedness
-- Well-Formedness of Items and Type-Level Terms
+   - What is Well-Formedness
+     - talk about the abstract concept. wfness isnt a specific thing its just the nebulous concept of things being 'correct'. we generally use it informally all over the place just to talk about us having checked something in the type system.
+     - there are some actual places in the compiler that use the wf terminology specifically. item wfck and well formed obligations
   - Well-Formedness of Type-Level Terms
    - Obligations for Well-Formedness
    - When Type-Level Terms Are Well-Formed
    - We Don’t Need Normalization (Yet)
    - Const Generic Arguments
 -  Well-Formedness of Items
    - We (Sometimes) Need Normalization
    - Global and Trivial Bounds
 - When We Don’t Fully Do Well-Formedness Checking
    - Trait Objects
    - Free Type Aliases
 - “Well-Formed” or “Wellformed”?
 - What Well-Formedness Isn’t
```

vague thoughts about adding a new header at the top. briefly talked about this to u on zulip but not much. I think that the right framing is probably to start by going "hey wfness is a nebulous informal concept" and then go "there are a couple concrete places in the compiler that are named after this nebulous informal concept".

src/analysis/well-formed.md
```diff
@@ -0,0 +1,247 @@
+# Well-Formedness of Items and Type-Level Terms
+
+"Well-formedness" is the area of analysis that produces questions like "does `T: Debug` hold true for some data structure that uses `T`?" or "is this const generic parameter `const B: bool` being handed a value of the right type?" These questions are then handed off to the trait solver to answer.
```
BoxyUwU: i guess relating to my above comment about new header, this also i think is why this paragraph feels vaguely off to me. because wellformedness is a higher level concept than the specific application of it to items and terms or to even obligations 

in theory this paragraph is just obviated by the new header 🤔 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2889/47a57f628fbf80e2b309ee6577d8fffaadbdc2fd..ce64f2f2fb721d6e4fd4d840665e76a9f82c3836)*

src/analysis/well-formed.md
```diff
@@ -0,0 +1,247 @@
...
+- **Item**[^items] well-formedness check (item-wfck.) 
+    - "Item-wfck" can call into "Term well-formedness checking" as Items contain Terms.
+    - Inner "Terms" can get normalized first.
+    - Can be considered a more coherent "pass" in the compiler than "term well-formedness" (which is performed in many places.)
+
+See: [What Well-Formedness Isn't](#what-well-formedness-isnt)
+
+## Well-Formedness of Type-Level Terms
+
+Type-Level Terms are the fundamental subject of well-formedness checking. We are also concerned with [Items](#well-formedness-of-items), but as a downstream consumer of type-level term well-formedness.
```
BoxyUwU: vibes here are kind of off, item wfck does stuff other than require terms to be wf and that's not an edge case/oddity thats a large part of its point 🤔 I think also framing wfness of terms as "the main wfness" feels wrong

i would say sth like "checking type level terms are well formed is one of the main places we talk about wellformedness in the compiler"

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2889/47a57f628fbf80e2b309ee6577d8fffaadbdc2fd..ce64f2f2fb721d6e4fd4d840665e76a9f82c3836)*

src/analysis/well-formed.md
```diff
@@ -0,0 +1,247 @@
...
+
+## Well-Formedness of Items
+
+Items are, generally speaking, "Things that get defined." Item-wfck[^item-wf-module] only happens at the signature level for types and functions, including the methods and implementations. This doesn't happen for Free Type Aliases other than Const Generic argument type checking.
+
+Items are a major entry point for term well-formedness. Because Items contain Terms, item-wfck can invoke term well-formedness checking.
+
+### We (Sometimes) Need Normalization
+
+There are places where normalization of an Item happens before its Terms have gone through well-formedness checking. This is considered problematic as doing so allows some terms to [bypass term well-formedness checking entirely](https://github.com/rust-lang/rust/issues/100041).
```
BoxyUwU: this might make more sense in the term wfness section 🤔 items don't get normalized, terms do. the notable thing abotu itemwfck is that it's the place we wind up normalizing terms before checking they're well formed

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2889/47a57f628fbf80e2b309ee6577d8fffaadbdc2fd..ce64f2f2fb721d6e4fd4d840665e76a9f82c3836)*

src/analysis/well-formed.md
```diff
@@ -0,0 +1,247 @@
...
+
+### Global and Trivial Bounds
+
+<!-- TODO later: Cut this into its own page -->
```
BoxyUwU, line 117:
```suggestion
<!-- TODO later: Cut this into its own page -->
When checking items are well-formed we will check that there are no trivially false global where clauses.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2889/47a57f628fbf80e2b309ee6577d8fffaadbdc2fd..ce64f2f2fb721d6e4fd4d840665e76a9f82c3836)*

src/analysis/well-formed.md
```diff
@@ -0,0 +1,264 @@
+# Well-Formedness
+
+## What is Well-Formedness?
+
+At a high level, "well-formed" just means "correctly built."[^wf-history] Something is _well-formed_ when its structure follows rules, and _ill-formed_ when those rules are broken.
+
+## Well-Formedness in Rust
+
```
BoxyUwU, line 8: wf in rust is *also* a similar kind of high level thing and I think we should be explicit about that. e..g say sth like: 

people will informally use "well formedness" to talk about things that aren't referred to by well-formed in the compiler. e.g. it makes sense to talk about a method call being well-formed even we just check method calls during type checking and at no point does the compiler talk about "well formedness" there

src/analysis/well-formed.md
```diff
@@ -0,0 +1,264 @@
...
+Items and Type-Level Terms are "well-formed" when they "fulfill obligations" or "meet the necessary constraints." When we're performing well-formedness checks we're usually concerned about if Trait obligations are met, but this also includes making sure that the types of const generic parameters "type check."
```
BoxyUwU: this feels like a slight lie since item wfck isnt just concerned with obligations 🤔 i still somewhat feel like it'd be fine to just drop this paragraph, the information is in the respective headers for item wfck and type level term wfness, and the information here rn is somewhat misleading 🤔

src/analysis/well-formed.md
```diff
@@ -0,0 +1,264 @@
...
+
+In the Rust compiler there are two different forms of well-formedness checking:
+
+- **Type-Level Term**[^terms][^terms-abbreviated] well-formedness check.
+    - Primary subject of "Well-Formedness Checking."
+    - Abbreviated here to "Term well-formedness" or "Term well-formedness checking." 
+    - This isn't a distinct "analysis pass," this gets performed throughout the compiler. 
+- **Item**[^items] well-formedness check (item-wfck.) 
+    - "Item-wfck" can call into "Term well-formedness checking" as Items contain Terms.
```
BoxyUwU:
```suggestion
    - "Item-wfck" will often wind up requiring Terms be well-formed.
```

src/analysis/well-formed.md
```diff
@@ -0,0 +1,264 @@
...
+    - Inner "Terms" can get normalized first.
```
BoxyUwU:
```suggestion
    - Inner "Terms" can (incorrectly) get normalized first.
```

src/analysis/well-formed.md
```diff
@@ -0,0 +1,264 @@
...
+
+See: [What Well-Formedness Isn't](#what-well-formedness-isnt)
+
+## Well-Formedness of Type-Level Terms
+
+Well-formedness of type-level terms is the area of analysis that produces questions like "does `T: Debug` hold true?" or "is this const generic parameter `const B: bool` being handed a value of the right type?" These questions are then handed off to the trait solver to answer.
+
+### Obligations for Well-Formedness
+
+Term well-formedness begins with generating a list of things that need to be true for a term to be well-formed.
```
BoxyUwU: i think this is less of a "begins" and more of a "this is all it is", term well formedness in its entirety is pretty much just "we produce a list of obligations for a term and then hand them off to the trait solver" 🤔

there's the actual trait solving process but that's more general and doesn't feel like it's really "part of" well formedness

src/analysis/well-formed.md
```diff
@@ -0,0 +1,264 @@
...
+const 6: usize
+```
+
+The call site will provide us with the obligation `6: usize` during well-formedness checking. This obligation will be passed off to the trait solver just like any trait-style obligation, as the trait solver has more responsibilities than its name suggests.
+
+## Well-Formedness of Items
+
+Items are, generally speaking, "Things that get defined." Item-wfck[^item-wf-module] only happens at the signature level for types and functions, including the methods and implementations. This doesn't happen for Free Type Aliases other than Const Generic argument type checking.
+
+Items are a major entry point for performing term well-formedness. Because Items contain Terms, item-wfck can invoke term well-formedness checking.
```
BoxyUwU:
```suggestion
Items are a major source of checking well-formedness of terms. Because Items contain Terms, item-wfck must check that those terms are well formed.
```

src/analysis/well-formed.md
```diff
@@ -0,0 +1,264 @@
...
+
+We find the obligation for `Vec<T>` that `T: Sized`. For `Vec<str>` we find the obligation `str: Sized`, which cannot be satisfied/is false.
+
+### We Don't Need Normalization (Yet)
+
+[Normalization](../normalization.md) is the process of resolving [type aliases](../normalization.md#aliases) into their underlying type. A type alias is considered well-formed if its underlying type is well-formed. The underlying type undergoes well-formedness checking at most definition and instantiation sites, but there are exceptions.
+
+### We (Sometimes) Need Normalization
+
+There are places where normalization of an [Item](#well-formedness-of-items) happens before its Terms have gone through well-formedness checking. This is considered problematic as doing so allows some terms to [bypass term well-formedness checking entirely](https://github.com/rust-lang/rust/issues/100041).
```
BoxyUwU: i still think this would make more sense as part of the "places we dont fully check well formedness" header :3

src/analysis/well-formed.md
```diff
@@ -0,0 +1,264 @@
...
+
+#### Item-Wfck and Trivial/Global Bounds
+
+<!-- When cutting out the subsection on global/trivial bounds, keep this part on the well-formedness page. -->
+
+When checking items are well-formed we will check that there are no trivially false global bounds.
+
+## When We Don't Fully Do Well-Formedness Checking
+
+Well-formedness checking is not a coherent "stage" of type checking. It gets called from various contexts, and there are places where it gets skipped partially or entirely.
```
BoxyUwU: this reads weird to me since you don't really "call" term wf checking because that's not really a Thing:tm: I think the main thing we care about here is jsut "there are places we dont fully check things are well formed", not the stuff about "its not a coherent stage of type checking" (which is fairly orthogonal)

src/analysis/well-formed.md
```diff
@@ -0,0 +1,264 @@
...
+
+### Trait Objects
+
+Trait objects do not undergo well-formedness checking except at the boundaries of coercion into being a trait object or being downcast into a concrete type.
```
BoxyUwU: this is slightly wrong in that we do check some stuff about trait objects, for example we check that the types/consts used in the trait are well formed sort of "standalone" . e.g. `dyn Trait<Vec<str>>` is not well formed. 

The precise thing to say here would be that well formedness of trait object types doesn't require the where clauses of the trait to hold.

src/analysis/well-formed.md
```diff
@@ -0,0 +1,264 @@
...
+
+```rust,ignore
+for<'a, 'b> fn(&'a &'b ())
+```
+
+The above HRB implies `'b: 'a` (a lifetime bound), rather than two completely separate lifetimes. This is normal lifetime behavior, but during well-formedness checking we cannot prove that this bound is generally true[^horrible], so we skip it.
+
+### Free Type Aliases
+
+The right-hand side of Free Type Aliases[^fta] do not go through a full well-formedness check at the definition site, with the exception of "type checking" const generic arguments in the RHS.
```
BoxyUwU:
```suggestion
The right-hand side of Free Type Aliases[^fta] is not fully checked to be well-formed at the definition site, only the types of const generic arguments in the RHS are checked.
```

src/analysis/well-formed.md
```diff
@@ -0,0 +1,264 @@
...
+
+```rust,ignore
+Vec<str>
+---
+// Obligations to fulfill
+Vec<T> where T: Sized
+Vec<str> where str: Sized // This is not true, therefore the term is not well-formed.
+```
+
+We find the obligation for `Vec<T>` that `T: Sized`. For `Vec<str>` we find the obligation `str: Sized`, which cannot be satisfied/is false.
```
BoxyUwU: dont fully understand this sentence, "we find the obligation for `Vec<T>` that `T: Sized`" doesnt parse to me

src/analysis/well-formed.md
```diff
@@ -0,0 +1,264 @@
...
+- **Item**[^items] well-formedness check (item-wfck.) 
+    - "Item-wfck" can call into "Term well-formedness checking" as Items contain Terms.
+    - Inner "Terms" can get normalized first.
+    - Can be considered a more coherent "pass" in the compiler than "term well-formedness" (which is performed in many places.)
+
+See: [What Well-Formedness Isn't](#what-well-formedness-isnt)
+
+## Well-Formedness of Type-Level Terms
+
+Well-formedness of type-level terms is the area of analysis that produces questions like "does `T: Debug` hold true?" or "is this const generic parameter `const B: bool` being handed a value of the right type?" These questions are then handed off to the trait solver to answer.
```
BoxyUwU: I'm not a huge fan of this paragraph and I can't really tell why 🤔 I feel like the information I expect to be here is just the stuff in"When type-level terms are well formed"

I would probably just replace this with the "when type level terms are well formed" section and then I think it'd read more nice to me

src/analysis/well-formed.md
```diff
@@ -0,0 +1,264 @@
...
+// Obligations to fulfill
+Vec<T> where T: Sized
+Vec<str> where str: Sized // This is not true, therefore the term is not well-formed.
+```
+
+We find the obligation for `Vec<T>` that `T: Sized`. For `Vec<str>` we find the obligation `str: Sized`, which cannot be satisfied/is false.
+
+### We Don't Need Normalization (Yet)
+
+[Normalization](../normalization.md) is the process of resolving [type aliases](../normalization.md#aliases) into their underlying type. A type alias is considered well-formed if its underlying type is well-formed. The underlying type undergoes well-formedness checking at most definition and instantiation sites, but there are exceptions.
```
BoxyUwU: is this supposed to be a section about what it means for an alias to be well formed? if so I think I would write this as
- an alias is well formed if all of its where clauses are satisfied, and the self type and generic arguments are themselves well formed
- we don't need to check that the normalized form of the alias is well formed as item-wfck will check that term is well formed separately. an alias being well formed *implies* that it's normalized form will be well formed

notably this is quite different than "a type alias is well formed if its underlying type is well formed". a type alias can have stronger wf requirements than its underlying type, and we also just sort of get the fact that the underlying type is wf for "free" without ever checking it when checking the alias is well formed

BoxyUwU (commented):
its somewhat arbitrary but im not a huge fan of saying stuff like "perform term well-formedness checking" and would rather we say things like "check terms are well-formed". im not sure I can justify why this is the case but it feels quite off to be saying the former instead of the latter 😅 

i think this is the nitpickiest of all my reviews but i think we're pretty much There with this doc at this point?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2889/47a57f628fbf80e2b309ee6577d8fffaadbdc2fd..a05c960de079a81011dbef2dd332436f4def89e9)*

src/appendix/glossary.md
```diff
@@ -102,6 +102,7 @@ Term                                           | Meaning
 <span id="trans">trans 👎</span>                  |  Short for _translation_, the code to translate MIR into LLVM IR. **Renamed to** [codegen](#codegen).
 <span id="ty">`Ty`</span>                      |  The internal representation of a type. ([see more](../ty.md))
 <span id="tyctxt">`TyCtxt`</span>              |  The data structure often referred to as [`tcx`](#tcx) in code which provides access to session data and the query system.
+<span id="type-level-term">Type-Level Term</span> | Type terms we're reasoning about and manipulating in an analysis context. Comparable/related to type expressions etc.
```
BoxyUwU: types and const generics 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2889/47a57f628fbf80e2b309ee6577d8fffaadbdc2fd..f6ab4dd2888759e6e36c432cbc94be12bec1898e)*

src/analysis/well-formed.md
```diff
@@ -0,0 +1,278 @@
...
+    - Inner "Terms" can (incorrectly) get normalized first.
+    - More coherent as a stage in the compiler than "term well-formedness" (which is performed in many places.)
+
+See: [What Well-Formedness Isn't](#what-well-formedness-isnt).
+
+## Well-Formedness of Type-Level Terms
+
+Term well-formedness checking begins with building a list of things that need to be true for a term to be well-formed. We call these "Obligations"[^obligations].
+
+Type-Level Terms are considered Well-Formed when obligations within them are satisfied by the trait solver.
```
BoxyUwU:
```suggestion
Type-Level Terms are considered Well-Formed when their associated obligations are satisfied by the trait solver.
```

src/analysis/well-formed.md
```diff
@@ -0,0 +1,278 @@
...
+```rust,ignore
+Vec<String>
+---
+// Obligations to fulfill
+Vec<T> where T: Sized
+// Trait solver says `String: Sized` is true, so this is well-formed.
+Vec<String> where String: Sized
+```
+
+When we run a search for obligations on `Vec<String>`, we'll find that `Vec<T>` generates the obligation `T: Sized`. We substitute `T` with `String` in `Vec<String>`, so we find the obligation `String: Sized`.
```
BoxyUwU:
```suggestion
When we compute the obligations for `Vec<String>`, we'll find that `Vec<T>` generates the obligation `T: Sized`. We substitute `T` with `String` in `Vec<String>`, so we find the obligation `String: Sized`.
```

src/analysis/well-formed.md
BoxyUwU: still relevant :3

src/analysis/well-formed.md
BoxyUwU: still relevant :3 would word this the same as the previous example as I think that made more sense to me

src/analysis/well-formed.md
```diff
@@ -0,0 +1,278 @@
...
+
+#### Determining Obligations
+
+In the compiler, obligations of terms are found through the [`obligations`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_trait_selection/traits/wf/fn.obligations.html) function in the [term well-formedness module](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_trait_selection/traits/wf/index.html).
+
+#### Other Obligations
+
+Obligations are more than just trait and const generic bounds, but we've only mentioned these specific obligations so far as they are what we care about when we do "well-formedness checking" of terms. For example, lifetime bounds (`'b: 'a` / `'b` outlives `'a`) can be part of the obligation output but _is not relevant to well-formedness of terms_. This information is saved to part of the compiler's internal state for later use, rather than used during well-formedness checks. See: [`PredicateKind`](https://doc.rust-lang.org/beta/nightly-rustc/rustc_type_ir/predicate_kind/enum.PredicateKind.html) and [`ClauseKind`](https://doc.rust-lang.org/beta/nightly-rustc/rustc_type_ir/predicate_kind/enum.ClauseKind.html) for a full list of obligations.
```
BoxyUwU: I think the lifetime stuff here is somewhat untrue as sometimes (specifically involving foralls) we do check lifetimes inside of the trait solver. I think it's also generally the case that we wouldnt consider a term wf if its region constrains don't hold

src/analysis/well-formed.md
```diff
@@ -0,0 +1,278 @@
...
+
+When checking items are well-formed we will check that there are no trivially false global bounds.
+
+## When We Don't Fully Do Well-Formedness Checking
+
+Well-formedness checking is not a coherent "stage" of type checking. There are many areas where well-formedness checking is performed, and some areas where we skip over well-formedness checking due to limitations in what kinds of analysis we can currently perform. Ideally, we would never skip or defer well-formedness checking.
+
+### Trait Objects
+
+We do not require the where clauses of trait objects to be well-formed when determining if that trait object is well-formed. These where clauses still need to be well-formed when coercing into/out of a trait object, but this remains a hole in well-formedness checking.
```
BoxyUwU:
```suggestion
We do not require the where clauses of trait objects to be well-formed when determining if that trait object is well-formed. These where clauses are proven when coercing into/out of a trait object, but this remains a hole in well-formedness checking.
```

generally a where clause isn't wf when it holds, it's wf when its constituent terms are wf. i'd typically say proven or checked to hold if i wanted to say this

src/analysis/well-formed.md
```diff
@@ -0,0 +1,278 @@
...
+
+As an example, the following will compile because we don't have a point where we're constructing the trait object or coercing it back to a concrete type:
```
BoxyUwU: what's "coercing it back to a concrete type" is that a thing rust supports

src/analysis/well-formed.md
```diff
@@ -0,0 +1,278 @@
...
+---
+const N: usize
+const B: bool
+N = B // Substitution
+const B: usize + bool 
+```
+
+The above doesn't compile, unlike the previous example we gave. We're doing _some_ well-formedness checking here when it comes to the const generic arguments.
+
+### Higher-Ranked Bounds
```
BoxyUwU: "Higher-Ranked Bounds" feels off because `for<'a> fn(...)` isnt a bound it's a higher ranked type 🤔 It's kind of jargon but maybe just "Wf Under Binders" is a possibility

src/analysis/well-formed.md
```diff
@@ -0,0 +1,278 @@
...
+
+Higher-Ranked Bounds skip well-formedness checking, leaving well-formedness checking to when the bound is instantiated:
```
BoxyUwU: to be somewhat precise, it's. "obligations involving variables from binders (`for<'a>`) are only checked when the binder is instantiated", we do actualyl (somewhat) check things are wf under `for<'a>`s

src/analysis/well-formed.md
BoxyUwU:
```rust
// The `Vec<str>` is checked during item wfck
fn foo(_: Vec<str>) {
    // The `Vec<[u8]>` is not handled by item wfck as it's not in the signature
    let _: Vec<[u8]>
}
```

src/analysis/well-formed.md
BoxyUwU: :3 ?

BoxyUwU (commented):
get nitzoned lmao 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2889/47a57f628fbf80e2b309ee6577d8fffaadbdc2fd..32eca49068994613c07e1b578441f69338226d1e)*

src/analysis/well-formed.md
fallible-algebra: \>:)c !

src/analysis/well-formed.md
fallible-algebra: Oh this got eaten / I missed this lol.

src/analysis/well-formed.md
fallible-algebra: I'm going to reword your comment into a post-code-block elaboration to keep the flow of the section decent :p

src/analysis/well-formed.md
fallible-algebra: I think we're fully in the jargon pit at this point. I'm going to do the cowards manoeuvre and merge both Higher-Ranked Types and Binders as the title here.

src/analysis/well-formed.md
BoxyUwU: :3c !!!! x3 uwuwuwuwuwuw

BoxyUwU (approved):

tshepang:
are the intermediate commits worth keeping... should we squash

fallible-algebra:
I'm comfortable with a squash here

PR #2890, opened 2026-06-04
Avoid date-checking contributing examples

Herrtian:
Fixes part of rust-lang/rustc-dev-guide#2887.

This updates the `src/contributing.md` date-check examples to use placeholders instead of real stale dates, so the examples do not keep appearing in monthly date triage output.

Check: `cargo run --manifest-path ci/date-check/Cargo.toml . | Select-String -Pattern 'src/contributing.md|Date Reference Triage'`

### DIFF
--- a/src/contributing.md
+++ b/src/contributing.md
@@ -444,20 +444,20 @@ Just a few things to keep in mind:
     For the action to pick the date, add a special annotation before specifying the date:
 
     ```md
-    <!-- date-check --> Nov 2025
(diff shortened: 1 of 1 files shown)


Herrtian:
r? rustc-dev-guide

tshepang:
I used these as a reminder to also keep the contribution section current

Herrtian:
Thanks, that makes sense. Closing this since those dates are intentional reminders.

PR #2910, opened 2026-06-23
update autodiff install docs for nix

activexray:
Now that the Enzyme component is in rustup, usage in Nix is much simpler and just requires declaring the component.
This PR updates the docs to reflect that. Alternatively, we could drop this block entirely as this is now "standard practice" for nix + rust with rustup components.

### DIFF
--- a/src/autodiff/installation.md
+++ b/src/autodiff/installation.md
@@ -18,30 +18,30 @@ Please run:
 rustup +nightly component add enzyme
 ```
 
-## Installation guide for Nix user.
+## Installation guide for Nix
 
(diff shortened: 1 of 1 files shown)


activexray:
r? ZuseZ4

ZuseZ4:
Thanks. I'd keep the nix guides for now, only if we at some point end up supporting more of these alternatives I'd consider shortening it.

Can you adjust the links to have the URL at the end of the file, like it's already done in the other cases? Otherwise lgtm.

activexray:
Thanks! Just fixed (and a typo).

PR #2911, opened 2026-06-26
Update TypeTree chapter to reflect newer understanding of it

ZuseZ4:
@scottmcm @workingjubilee does that match your understanding after the last discussions?

### DIFF
--- a/src/autodiff/type-trees.md
+++ b/src/autodiff/type-trees.md
@@ -1,7 +1,35 @@
 # TypeTrees for Autodiff
 
 ## What are TypeTrees?
-Memory layout descriptors for Enzyme. Tell Enzyme exactly how types are structured in memory so it can compute derivatives efficiently.
+Memory layout descriptors for Enzyme. They tell Enzyme what "type" bytes are, with the main categories being Float, Integer, or Pointer. In Rust, memory is conceptually untyped, so it is possible to store a float into 4 bytes, and later read the bytes back as an integer. This is generally true in Rust even in the absence of `enum` or `union` types. We therefore can not directly put typetree metadata on allocations. We can also not accept Enzyme's default behaviour, which incorrectly assumes that LLVM-IR follows `strict aliasing` rules (known from C/C++). As a solution, we disable Enzyme's strict-aliasing behaviour and only generate TypeTree metadata in selected locations.
+
+## Where we generate TypeTree
+The underlying idea is that memory "at rest" is untyped, but plenty of usages interprete bytes in a way that we can communicate to Enzyme. For example, when we call a function, the memory passed to it is interpreted according to the function's signature, so we can add TypeTrees to the LLVM-IR function definitions. We currently only do that for the outermost functions differentiated (those that have a `#[autodiff]` macro on them), but we plan to extend it to all functions which are called from them. We currently also generate TypeTree information for all calls to mem{cpy|move|set}. Finally, we started to add TypeTrees to the input or return values of certain instructions, for now that mainly is `extractvalue`.
+
+## How we add TypeTrees
+If we determined that a value has a meaningfull type, then we walk the MIR `Ty` of that value in the middle-end and generate a Rust TypeTree out of it. In the codegen\_llvm backend we lower our Rust TypeTree to LLVM/Enzyme TypeTrees. We then attach them to one of three locations:
+
+Parameters in function definitions:
+```llvm
+define internal void @_RNvCs7tI50jyFEig_3foo1f(ptr align 8 "enzyme_type"="{[-1]:Pointer, [-1,-1]:Float@double}" %0, ptr align 8 "enzyme_type"="{[-1]:Pointer, [-1,-1]:Float@double}" %1, ptr align 8 "enzyme_type"="{[-1]:Pointer, [-1,-1]:Float@double}" %2) unnamed_addr #0 !dbg !1089 {
+```
+
+Argument to calls:
+```llvm
+  call void @llvm.memcpy.p0.p0.i64(ptr align 8 "enzyme_type"="{[0]:Pointer, [0,0]:Pointer, [0,0,-1]:Float@double}" %6, ptr align 8 "enzyme_type"="{[0]:Pointer, [0,0]:Pointer, [0,0,-1]:Float@double}" %0, i64 24, i1 false), !dbg !669
+```
+
+Input or return values of instructions, via debug metadata:
+```llvm
+  %14 = extractvalue { ptr, i64 } %13, 0, !dbg !906, !enzyme_type !907
+  %15 = extractvalue { ptr, i64 } %13, 1, !dbg !906, !enzyme_type !910
+...
+!907 = !{!"Unknown", i32 -1, !908}
+!908 = !{!"Pointer", i32 -1, !909}
+!909 = !{!"Float@double"}
+!910 = !{!"Unknown", i32 0, !911, i32 1, !911, i32 2, !911, i32 3, !911, i32 4, !911, i32 5, !911, i32 6, !911, i32 7, !911}
+!911 = !{!"Integer"}
+```
 
 ## Structure
 ```rust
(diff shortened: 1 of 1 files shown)


src/autodiff/type-trees.md
```diff
@@ -1,7 +1,35 @@
 # TypeTrees for Autodiff
 
 ## What are TypeTrees?
-Memory layout descriptors for Enzyme. Tell Enzyme exactly how types are structured in memory so it can compute derivatives efficiently.
+Memory layout descriptors for Enzyme. They tell Enzyme what "type" bytes are, with the main categories being Float, Integer, or Pointer. In Rust, memory is conceptually untyped, so it is possible to store a float into 4 bytes, and later read the bytes back as an integer. This is generally true in Rust even in the absence of `enum` or `union` types. We therefore can not directly put typetree metadata on allocations. We can also not accept Enzyme's default behaviour, which incorrectly assumes that LLVM-IR follows `strict aliasing` rules (known from C/C++). As a solution, we disable Enzyme's strict-aliasing behaviour and only generate TypeTree metadata in selected locations.
+
+## Where we generate TypeTree
```
workingjubilee, line 6:
```suggestion
## Where we generate TypeTrees
```

src/autodiff/type-trees.md
```diff
@@ -1,7 +1,35 @@
 # TypeTrees for Autodiff
 
 ## What are TypeTrees?
-Memory layout descriptors for Enzyme. Tell Enzyme exactly how types are structured in memory so it can compute derivatives efficiently.
+Memory layout descriptors for Enzyme. They tell Enzyme what "type" bytes are, with the main categories being Float, Integer, or Pointer. In Rust, memory is conceptually untyped, so it is possible to store a float into 4 bytes, and later read the bytes back as an integer. This is generally true in Rust even in the absence of `enum` or `union` types. We therefore can not directly put typetree metadata on allocations. We can also not accept Enzyme's default behaviour, which incorrectly assumes that LLVM-IR follows `strict aliasing` rules (known from C/C++). As a solution, we disable Enzyme's strict-aliasing behaviour and only generate TypeTree metadata in selected locations.
```
workingjubilee, line 4: This can be split across multiple lines since Markdown (*proper* Markdown, like the book's) gets concatenated across newlines (but not double-newlines). This allows diffing individual sentences of a paragraph.

You do not have to do this if you do not like how it looks, it is just a suggestion.

```suggestion
Memory layout descriptors for Enzyme. They tell Enzyme what "type" bytes are, with the main categories being Float, Integer, or Pointer. In Rust, memory is conceptually untyped, so it is possible to store a float into 4 bytes, and later read the bytes back as an integer. This is generally true in Rust even in the absence of `enum` or `union` types. We therefore can not directly put typetree metadata on allocations. We can also not accept Enzyme's default behaviour, which incorrectly assumes that LLVM-IR follows `strict aliasing` rules (known from C/C++).
As a solution, we disable Enzyme's strict-aliasing behaviour and only generate TypeTree metadata where Rust actively asserts a type.
```

src/autodiff/type-trees.md
```diff
@@ -1,7 +1,35 @@
...
+
+## Where we generate TypeTree
+The underlying idea is that memory "at rest" is untyped, but plenty of usages interprete bytes in a way that we can communicate to Enzyme. For example, when we call a function, the memory passed to it is interpreted according to the function's signature, so we can add TypeTrees to the LLVM-IR function definitions. We currently only do that for the outermost functions differentiated (those that have a `#[autodiff]` macro on them), but we plan to extend it to all functions which are called from them. We currently also generate TypeTree information for all calls to mem{cpy|move|set}. Finally, we started to add TypeTrees to the input or return values of certain instructions, for now that mainly is `extractvalue`.
```
workingjubilee, line 7:
```suggestion
The underlying idea is that while the memory of a place is untyped, plenty of usages impose a type assertion on bytes in ways that we can communicate to Enzyme.
For example, when we call a function, its arguments and return values are passed by typed copies matching the function's signature, so we can add TypeTrees to the LLVM-IR function definitions.
We currently only do that for the outermost functions differentiated (those that have a `#[autodiff]` macro on them), but we plan to extend it to all functions which are called from them. We currently also generate TypeTree information for all calls to mem{cpy|move|set}. Finally, we started to add TypeTrees to the input or return values of certain instructions, for now that mainly is `extractvalue`.
```
So this one is not *just* nitpicking the spelling of "interpret"... the interpretation is a very specific kind, and it applies to the values that receive what are often referred to as "typed copies".

Saying "memory" risks being vague for people because "memory" can mean both the values that receive typed copies and the memory in a place, and an argument can be a pointer to a place.

Much of the idea of the validity here can be considered as matching https://doc.rust-lang.org/std/mem/fn.transmute.html

src/autodiff/type-trees.md
```diff
@@ -1,7 +1,35 @@
...
+
+## How we add TypeTrees
+If we determined that a value has a meaningfull type, then we walk the MIR `Ty` of that value in the middle-end and generate a Rust TypeTree out of it. In the codegen\_llvm backend we lower our Rust TypeTree to LLVM/Enzyme TypeTrees. We then attach them to one of three locations:
```
workingjubilee, line 10:
```suggestion
If we determine that a value has a meaningful type, then we walk the MIR `Ty` of that value in the middle-end and generate a Rust TypeTree out of it. In the `codegen_llvm` backend we lower our Rust TypeTree to Enzyme TypeTrees. We then attach them to one of three locations:
```

Hm. Calling them LLVM/Enzyme TypeTrees confuses the matter, I think? What makes them "LLVM/Enzyme TypeTrees"? Is it because they are directly embedded in the LLVMIR? I think it could just say that, then?

This should probably explain that first, actually. "A TypeTree is an Enzyme concept that gets smuggled through LLVMIR metadata" https://llvm.org/docs/LangRef.html#metadata

src/autodiff/type-trees.md
```diff
@@ -1,7 +1,35 @@
...
+
+Parameters in function definitions:
+```llvm
+define internal void @_RNvCs7tI50jyFEig_3foo1f(ptr align 8 "enzyme_type"="{[-1]:Pointer, [-1,-1]:Float@double}" %0, ptr align 8 "enzyme_type"="{[-1]:Pointer, [-1,-1]:Float@double}" %1, ptr align 8 "enzyme_type"="{[-1]:Pointer, [-1,-1]:Float@double}" %2) unnamed_addr #0 !dbg !1089 {
+```
+
+Argument to calls:
```
workingjubilee, line 17:
```suggestion
Arguments to calls:
```

src/autodiff/type-trees.md
```diff
@@ -47,11 +75,12 @@ TypeTree(vec![Type {
...
 ```
 
-## Why Needed?
-- Enzyme can't deduce complex type layouts from LLVM IR
-- Prevents slow memory pattern analysis
-- Enables correct derivative computation for nested structures
-- Tells Enzyme which bytes are differentiable vs metadata
+## Why are they needed?
+- Plenty of LLVM types are opaque (e.g. `ptr`), but types are needed to compute the correct derivatives.
+- They tell Enzyme which bytes are differentiable (e.g. the pointer to float within a slice) vs metadata (e.g. the integer length of a slice)
```
workingjubilee, line 80: or "a float", but the pointer of a slice can also be to zero floats, so...

```suggestion
- They tell Enzyme which bytes are differentiable (e.g. the pointer to floats within a slice) vs. metadata (e.g. the integer length of a slice)
```

src/autodiff/type-trees.md
```diff
@@ -47,11 +75,12 @@ TypeTree(vec![Type {
...
+- Enzyme can't deduce all types from LLVM IR, but can (to some extend) deduce them from usage (Type Analysis).
```
workingjubilee, line 81:
```suggestion
- Enzyme can't deduce all types from LLVM IR, but can (to some extent) deduce them from usage (Type Analysis).
```

src/autodiff/type-trees.md
```diff
@@ -1,7 +1,35 @@
...
+```llvm
+define internal void @_RNvCs7tI50jyFEig_3foo1f(ptr align 8 "enzyme_type"="{[-1]:Pointer, [-1,-1]:Float@double}" %0, ptr align 8 "enzyme_type"="{[-1]:Pointer, [-1,-1]:Float@double}" %1, ptr align 8 "enzyme_type"="{[-1]:Pointer, [-1,-1]:Float@double}" %2) unnamed_addr #0 !dbg !1089 {
+```
+
+Argument to calls:
+```llvm
+  call void @llvm.memcpy.p0.p0.i64(ptr align 8 "enzyme_type"="{[0]:Pointer, [0,0]:Pointer, [0,0,-1]:Float@double}" %6, ptr align 8 "enzyme_type"="{[0]:Pointer, [0,0]:Pointer, [0,0,-1]:Float@double}" %0, i64 24, i1 false), !dbg !669
+```
+
+Input or return values of instructions, via debug metadata:
```
workingjubilee, line 22: that's what these are, right? https://llvm.org/docs/LangRef.html#metadata-nodes-mdnode
```suggestion
Input or return values of instructions rustc uses for typed copies, via metadata nodes:
```

workingjubilee (commented):
some typos and some concept-level thoughts

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2911/e57cc5df56c50ba7bfa7e5171d331c067ed4fc9d..f0cfd5b468c958a31345974299d39e9d5589c31b)*

jyn514:
@workingjubilee @ZuseZ4 👋 what's the state of this? is it ok to merge as-is and improve in follow-ups?

PR #2921, opened 2026-07-09
Add install notes for older rustup versions

ZuseZ4:
Just encountered during a tutorial.

### DIFF
--- a/src/autodiff/installation.md
+++ b/src/autodiff/installation.md
@@ -18,6 +18,12 @@ Please run:
 rustup +nightly component add enzyme
 ```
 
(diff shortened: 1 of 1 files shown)


src/autodiff/installation.md
```diff
@@ -18,6 +18,12 @@ Please run:
 rustup +nightly component add enzyme
 ```
 
+Older rustup versions are not aware of this component, so if you run into issues try updating rustup itself:
+```console
+rustup update
```
ChrisDenton, line 23: You mean
```suggestion
rustup self update
```

PR #2924, opened 2026-07-10
HIR "Eager Type Inference" / Expectations dev guide page

fallible-algebra:
Page on "Eager Type Inference" / Expectations. Currently deep in drafts.

- [x] Is eager type inference the right phrase here? How do we need to represent it to the reader?
- [ ] Answer why the following need to use eager type inference / expectations
    - [x] closures (kinda done)
    - [x] coercions
    - These do not reference expectations in terms of logic, but do pass around an expectations arg for diagnostics.
      - [ ] indexes
      - [ ] fields
      - [ ] methods
- [ ] Establish if the whole bidirectional type system paper needs mentioning, and how to do that in a way that makes sense to people other than type theory sickos.
  - it's the source of "do some more type checking earlier"(?) 
- [ ] Post in the zulip


fallible-algebra:
@BoxyUwU

BoxyUwU:
@fallible-algebra  :)

src/hir-typeck/expectations.md
```diff
@@ -0,0 +1,51 @@
+# Expectations and Eager Type Inference
+
+In an ideal world, we would only perform type inference at one point in the compiler: once all constraints have been collected.
+
+In reality, we perform it (at least) twice: First, at eager evaluation points, and secondly "at the end" when those constraints have been collected.
+
+`Expectations` are a piece of type inference state we maintain for the cases where we need to eagerly infer the types of expressions rather than leave them to the end. They allow us to "ask questions" of the form "hey, we're expecting this term to have this type, is this true?"
```
BoxyUwU: i feel like its more of a we can ask "what are we expecting the type of this expression to be?"

src/hir-typeck/eager-inference.md
```diff
@@ -0,0 +1,68 @@
...
+
+_Closures_ need to have type inference eagerly applied to them because they are functions that are rarely fully annotated:
+
+```rust
+let closure = |a, b| if a < b {vec![1, 2, 3]} else {vec![5, 6, 7]};
+```
+
+If we didn't do eager type inference we would instead have closures whose types were filled with [inference variables](../appendix/glossary.md#inf-var). 
+
+This would be able to be solved in some situations, but because we do not have Higher-Ranked Inference Variables[^higher-ranked-inference] this would make the higher-ranked bounds for lifetimes that rust can have unusable without more annotation.
```
BoxyUwU: the other reason we care about this is that if you have `fn(?x) -> ?y` as your closure signature then you can't do field accesses/method calls/etc on `?x` which is a quite common thing to want to do

src/hir-typeck/eager-inference.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Coercions 
+
+[Coercions](./coercions.md) can happen in many places. We check to see if a coercion can happen, and if it can we perform the coercion. 
+
+When we successfully find a coercion, we need to eagerly perform type inference/checking on it as future inference will require or benefit from this information to be known ahead of time.
```
BoxyUwU, line 106: I don't really know what this sentence is saying 🤔 it also feels a bit backwards to me because once we've "successfully found a coercion" then we've already done "eager type inference" because finding a coercion is the thing that cares about the current state of inference :3

src/hir-typeck/eager-inference.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Method calls, Fields, and Indexes.
```
BoxyUwU, line 114: structuring this as
- Closures
- Coercions
- Other stuff

feels slightly off to me as I feel like it places too much importance on the expectation vs non expectation distinction between cases of eager type inference

I feel like I would prefer just a flat list of all the places that are eager type inference spots

src/hir-typeck/eager-inference.md
```diff
@@ -0,0 +1,68 @@
...
+
+These are areas which technically take expectations, but in practice use them for diagnostics only.
+
+#### Methods
+
+Maybe Not. Maybe just point to [method lookup](./method-lookup.md).
```
BoxyUwU, line 122: definitely want to talk about methods :3 

I think for each item (e.g. coercions, closures, fields, etc) here we want to cover:
- Why it is an eager inference spot
- How we handle it being an eager inference spot (we would probably talk about "hey we use expectations" here if we do)
- Why we handle it the way we do

for methods this is probably:
- method lookup is an eager inference spot because we need to look at the type of the receiver to determine what methods exist
- we handle this by erroring if there are inference variables which prevent us from figuring out all the methods we might be able to call
- we could move this into the trait solver but it'd complicate things quite a lot and hasn't really been necessary to do so (so mostly historical reasons).

src/hir-typeck/eager-inference.md
```diff
@@ -0,0 +1,68 @@
...
+
+? Method calls engage in Coercion and therefore need to engage in Eager Type Inference.
+
+#### Fields?
```
BoxyUwU, line 126: definitely want to talk about field accesses :3 its mostly the same reasoning as methods

src/hir-typeck/eager-inference.md
BoxyUwU: I think more generally we want to talk about how inferring higher ranked things is eager, and there are two main places this happens:
- inferring higher ranked closure signatures
- subtyping of higher ranked types

src/hir-typeck/eager-inference.md
```diff
@@ -0,0 +1,68 @@
+# Eager Type Inference
+
+There are places during compilation where we need to depend on the current state of inference. This means we need to establish the type of a {term, item, etc, figure out exact wording in review} earlier than we otherwise would.
+
+Eager evaluation of the type of a term (type inference) is required at some specific points to either make later type inference more consistent or (in the case of higher-ranked bounds / Higher Ranked Lifetime bounds) make it even possible.
+
+Eager type inference is when we do type inference earlier than we otherwise would. To do this, we bring in Expectations as an additional piece of context.
+
```
BoxyUwU, line 10: general structural thing I think it would be nice to talk about like "the ways we can handle places that care about the current inference state" and sort of talk abstractly about:
- we can defer it to arbitrary points during type inference (trait solving works like this)
- we can eagerly error on inference variables
- we can defer it to the end of typeck after everything has been inferred (array repeat expr checks do this)

src/hir-typeck/eager-inference.md
```diff
@@ -0,0 +1,68 @@
...
+### Closures and Higher-Ranked Variables
```
BoxyUwU, line 69: i kind of want to add trait solving to the list of "eager inference points" in that they're still part of the general "sometimes we do things which cares about the current inference state" its just that "trait" solving is how we handle that correctly instead of jankily

BoxyUwU (commented):
beep boop

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2924/0c20a0f3140af9eb6233e668172ad3e47b557d69..63acfd011e6a98bcb8668f3181d11a71a0185a56)*

src/hir-typeck/eager-inference.md
fallible-algebra: - re: trait solver doing this: Does this behave differently between the current and next trait solver? And is that a relevant thing for me to bring up :p

jyn514:
@BoxyUwU @fallible-algebra 👋 what's the state of this? is it ok to merge as-is and improve in follow-ups?

jyn514:
(note that CI is failing)

fallible-algebra:
Still working on it, and I'm away until next week.

jyn514:
No worries at all :) thanks for the update

src/hir-typeck/eager-inference.md
```diff
@@ -0,0 +1,130 @@
...
+```rust
+let x = ...;
+---
+// At initialization.
+x: ?a;
+?a = not known;
+// Maybe later?
+?a = Vec<?b>;
+// And later still
+?a = Vec<String>;
```
BoxyUwU, line 30: in practice we won't ever record a second constraint for an inference variable. instead we'll zonk it and would have `Vec<?b> eq Vec<String>` which yields a `?b = String` constraint

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2924/0c20a0f3140af9eb6233e668172ad3e47b557d69..855f60cbf682b7b7edba3ed86111c9fedef2e01b)*

fallible-algebra: Straight up

src/hir-typeck/eager-inference.md
```diff
@@ -0,0 +1,130 @@
...
+
+Trait solving can be run on a [`TyKind`][tykind] at any point. Failure is recoverable, so we can repeat this operation, but if at the end of an Item we can't establish if a type implements a trait then we error.
+
+### Coercions 
+
+[Coercions](./coercions.md) can happen in many places. We check to see if a coercion can happen, and if it can we perform the coercion. 
+
+TODO: Stub, relationship to eager inference is not well established.
+
+TODO: Following are stubs, need to determine if these are relevant to bring up.
```
BoxyUwU, line 115: they are :3 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2924/0c20a0f3140af9eb6233e668172ad3e47b557d69..855f60cbf682b7b7edba3ed86111c9fedef2e01b)*

src/hir-typeck/eager-inference.md
```diff
@@ -0,0 +1,130 @@
...
+let closure = |a, b| if a < b {vec![1, 2, 3]} else {vec![5, 6, 7]};
+---
+// Doesn't capture any information so it's a bare function pointer
+// but we still don't know much about it.
+closure: fn(?x, ?y) -> ?z
+```
+
+If we didn't do eager type inference we would instead have closures whose types were filled with [inference variables](../appendix/glossary.md#inf-var) (like the `?x, ?y, ?z` variables above).
+
+Eagerly inferring the type of `closure` here serves a couple of purposes. Firstly, it's a decent heuristic that if we define a closure we'll use it later and having its type be known will mean a less intense inference solve. Having eagerly inferred the type of a closure lets us establish [Expectations](#expectations) that don't have inference variables in them.
```
BoxyUwU, line 99: vibes off bestie

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2924/0c20a0f3140af9eb6233e668172ad3e47b557d69..855f60cbf682b7b7edba3ed86111c9fedef2e01b)*

fallible-algebra: truke

src/hir-typeck/eager-inference.md
```diff
@@ -0,0 +1,130 @@
...
+expect result: bool;
+// Then we type check and infer
+```
+
+These pieces of information have not yet been canonized as "things an inference variable _is_" the same way we might be able to look up what `?x` currently maps to in the current state of inference.
+
+### Constraints
+
+This is the set of constraints that need to be solved. We collect these constraints as we move through the codebase being type checked.
+
```
BoxyUwU, line 67: ### Obligations

sometimes we look at the set of current obligations/goals/predicates the trait solver is required to prove.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2924/0c20a0f3140af9eb6233e668172ad3e47b557d69..855f60cbf682b7b7edba3ed86111c9fedef2e01b)*

PR #2925, opened 2026-07-13
Refresh diagnostic items date check

Bortlesboat:
Refresh the date check for the diagnostic-items symbol-registration instructions after verifying them against current rustc.

The compiler still defines diagnostic-item symbols in `compiler/rustc_span/src/symbol.rs` through the `symbols!` list and exposes them through `rustc_span::symbol::sym`, so the surrounding guidance remains current.

Part of rust-lang/rustc-dev-guide#2914.

Validation:

- ran `cargo run --manifest-path ci/date-check/Cargo.toml .` and confirmed `src/diagnostics/diagnostic-items.md` is no longer reported
- `git diff --check`

### DIFF
--- a/src/diagnostics/diagnostic-items.md
+++ b/src/diagnostics/diagnostic-items.md
@@ -48,7 +48,7 @@ A new diagnostic item can be added with these two steps:
     For the naming conventions of diagnostic items, please refer to
     [*Naming Conventions*](#naming-conventions).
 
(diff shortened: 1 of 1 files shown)


Bortlesboat:
r? rustc-dev-guide

Bortlesboat:
r? tshepang

tshepang:
note that you need not add the Validation section to PR comment

PR #2936, opened 2026-07-21
docs: Update profiling inlining guidance

aland-omed:
Refreshes the profiling date-reference and corrects its outdated statement that inlining was missing from Cranelift. Current `rustc_codegen_cranelift` sysroot builds explicitly enable MIR inlining with `-Zinline-mir`; LLVM and GCC may also inline during code generation.

Source checked: https://github.com/rust-lang/rustc_codegen_cranelift/blob/master/build_system/build_sysroot.rs

Part of rust-lang/rustc-dev-guide#2914.

Validation: `git diff --check`

### DIFF
--- a/src/profiling.md
+++ b/src/profiling.md
@@ -109,6 +109,6 @@ The llvm-lines output is affected by several options.
 MIR optimizations have little impact.
 Compared to the default `RUSTFLAGS="-Z
 mir-opt-level=1"`, level 0 adds 0.3GB and level 2 removes 0.2GB.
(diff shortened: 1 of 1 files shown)


aland-omed:
r? rustc-dev-guide

jyn514:
@aland-omed from what i can tell, this and all your other PRs were LLM generated, which means you need to disclose that usage. I'm going to close this until you follow our [LLM policy](https://forge.rust-lang.org/policies/llm-usage.html).

PR #2938, opened 2026-07-22
docs: Clarify Neovim workspace path setup

### LINKED ISSUE #2473: neovim setup using .vscode/settings.json incomplete
https://rustc-dev-guide.rust-lang.org/building/suggested.html#neovim describes three ways to use neovim.
1. Using neoconf.nvim and the vscode settings.
2. using coc.nvim
3. using the manual config.

Option 1 mostly seem to work just fine, except neovim not recognizing or replacing ${workspaceFolder}
(see also) https://github.com/rust-lang/rustc-dev-guide/pull/2259#discussion_r1972729429
I ended up replacing it with the absolute path. I'm not super experienced with these configs, so not sure if neovim has another name for it, if we should ask Neovim devs to add support, or (probably a good addition either way) add a note to the docs which tells users to use the absolute path. 

At least option 3. also defines a function to recognize and handle the workspacefolder, but it's not clear that users have to handle it manually when using neoconf.

aland-omed:
Fixes rust-lang/rustc-dev-guide#2473.

Clarify that Neovim/neoconf does not expand VS Code’s `${workspaceFolder}` variable in the generated `.vscode/settings.json`, and instruct users to replace it with the absolute path to their rust repository.

Validation:
- `git diff --check`
- Verified the current generated `rust_analyzer_settings.json` still contains `${workspaceFolder}` in rustfmt, proc-macro, rustc, and cargo paths

The local environment does not have Cargo or mdBook installed; repository CI will run the complete book and link checks.

### DIFF
--- a/src/building/suggested.md
+++ b/src/building/suggested.md
@@ -163,6 +163,8 @@ Steps for this can be [found here][r-a nvim lsp].
 2. Run `./x setup editor`, and select `vscode` to create a `.vscode/settings.json` file.
    `neoconf` is able to read and update
    rust-analyzer settings automatically when the project is opened when this file is detected.
+   Neovim does not expand VS Code's `${workspaceFolder}` variable, so replace each occurrence
+   in the generated file with the absolute path to your rust repository.
 
 #### coc.nvim
 


aland-omed:
r? rustc-dev-guide

tshepang:
r? rustc-dev-guide

jyn514 (approved):
this looks right, yup. good catch :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2938/34e9d98cfe78dc7b766631bceda93dda16089145..8ad485425324a002c1aaaf2110cfd5d1e925d756)*

PR #2950, opened 2026-08-02
Update parallel compilation status for August 2026

eeminionn:
## What changed

I went through the four date markers in `src/parallel-rustc.md` and checked the surrounding text against the current compiler source instead of only changing the month.

This updates the markers to August 2026 and also:

- documents the current `--jobs*` options and the one-job front-end default;
- replaces the removed `parallel-compiler` split with the current runtime synchronization behavior;
- updates `WorkerLocal`, the parallel helper names, and representative call sites;
- points from the completed query-parallelization tracker to the active parallel front-end tracker; and
- records that rustdoc HTML rendering is still serial.

## Why

Several claims around the dates had fallen behind the implementation. In particular, `-Z threads` now directs users to `--jobs-frontend`, `par_iter` is no longer exported, the HIR helper names changed, and the old query-parallelization tracker was completed in June 2026. Updating only the dates would have made the page look freshly verified while leaving those details wrong.

The result should give compiler contributors current flags and source names while keeping the existing warning that this area is still evolving.

## Verification

- Checked the documented helpers and call sites against `rust-lang/rust` main at `73dc9167f1cd099e525c9ade2e068d1907b78564`.
- Confirmed that rust-lang/rust#113349 and rust-lang/rust#82741 are open, while rust-lang/rust#48685 is closed as completed.
- Confirmed that all new documentation links return HTTP 200.
- Ran `git diff --check`.

This addresses the four `src/parallel-rustc.md` entries in rust-lang/rustc-dev-guide#2948.

@jyn514, I rewrote the description in my own words and included the checks I used for each dated claim. Thanks again for pointing out that the verification needed to be explicit.

### DIFF
--- a/src/parallel-rustc.md
+++ b/src/parallel-rustc.md
@@ -1,25 +1,26 @@
 # Parallel compilation
 
 <div class="warning">
-As of <!-- date-check --> November 2024,
-the parallel front-end is undergoing significant changes,
-so this page contains quite a bit of outdated information.
+As of <!-- date-check --> August 2026, the parallel front-end remains under
+active development, so implementation details on this page may change.
 
 Tracking issue: <https://github.com/rust-lang/rust/issues/113349>
 </div>
 
-As of <!-- date-check --> November 2024, most of the rust compiler is now
-parallelized.
+As of <!-- date-check --> August 2026, `rustc` is always built with support for
+parallel front-end work. The front-end covers compilation from lexing through
+the generation of backend IR. It uses one job by default; `-j`/`--jobs` or
+`--jobs-frontend` can be used to request more. See the
+[compiler job options][compiler-job-options] for details.
 
-- The codegen part is executed concurrently by default.
-  You can use the `-C
-  codegen-units=n` option to control the number of concurrent tasks.
-- The parts after HIR lowering to codegen such as type checking, borrowing
-  checking, and mir optimization are parallelized in the nightly version.
-  Currently, they are executed in serial by default, and parallelization is
-  manually enabled by the user using the `-Z threads = n` option.
-- Other parts, such as lexical parsing, HIR lowering, and macro expansion, are
-  still executed in serial mode.
+- Within the front-end, query evaluation and several intra-query loops can run
+  concurrently. Individual stages that have not been parallelized still run
+  serially.
+- Code generation is executed concurrently by default. The `-C
+  codegen-units=n` option controls how many codegen units a crate is split into,
+  while `--jobs-backend` limits concurrent backend work.
+- The linker has a separate `--jobs-linker` limit when the selected linker
+  supports configuring its parallelism.
 
 <div class="warning">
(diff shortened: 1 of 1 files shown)


src/parallel-rustc.md
```diff
@@ -1,25 +1,26 @@
...
-  checking, and mir optimization are parallelized in the nightly version.
-  Currently, they are executed in serial by default, and parallelization is
-  manually enabled by the user using the `-Z threads = n` option.
-- Other parts, such as lexical parsing, HIR lowering, and macro expansion, are
-  still executed in serial mode.
+As of <!-- date-check --> August 2026, `rustc` is always built with support for
+parallel front-end work. The front-end covers compilation from lexing through
+the generation of backend IR. It uses one job by default; `-j`/`--jobs` or
+`--jobs-frontend` can be used to request more. See the
+[compiler job options][compiler-job-options] for details.
```
jyn514, line 14: could you note that these flags (and the ones below) are unstable?

src/parallel-rustc.md
```diff
@@ -44,90 +45,62 @@ This process occurs in the [`rustc_codegen_ssa::base`] module.
...
-  threads increases beyond 4. So we audit the use of these data structures
-  which leads to either a refactoring so as to reduce the use of shared state,
-  or the authoring of persistent documentation covering the specific of the
-  invariants, the atomicity, and the lock orderings.
-
-- On the other hand, we still need to figure out what other invariants
-  during compilation might not hold in parallel compilation.
+
+The compiler no longer has separate `parallel-compiler` and non-parallel build
+configurations. Instead, `Lock<T>` selects its synchronization mode at runtime.
```
jyn514, line 50: please write this in terms of how things work now, not compared to how they used to be.

src/parallel-rustc.md
```diff
@@ -1,25 +1,26 @@
 # Parallel compilation
 
 <div class="warning">
-As of <!-- date-check --> November 2024,
-the parallel front-end is undergoing significant changes,
-so this page contains quite a bit of outdated information.
+As of <!-- date-check --> August 2026, the parallel front-end remains under
+active development, so implementation details on this page may change.
 
 Tracking issue: <https://github.com/rust-lang/rust/issues/113349>
```
jyn514, line 7: could you change this to reuse the new `[tracking]` link you added?

jyn514 (changes requested):
@eeminionn thank you for writing the description in your own words, this is a lot better now! i do want to note that we try to discourage LLM-generated documentation — i've read through it and in practice i think it's fine, but please write documentation in your own words going forwards.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc-dev-guide/2950/f5333a6ab8e34b445794cf7d572806765482598b..833b4cf3dc2cc37eb9d2498e2f61912fa829a607)*

PR #2954, opened 2026-08-06
Fix link in llm-use rejection template

ohAitch:


ohAitch:
r? rustc-dev-guide

tshepang:
hm, is it not better to leave the link target as it is... the link text is asking for the policy, not a particular section of it

jieyouxu:
> hm, is it not better to leave the link target as it is... the link text is asking for the policy, not a particular section of it

I agree, changing it to a particular section seems strange.

PR #2961, opened 2026-08-12
Add revised assumptions on binders shorthand to the glossary

fallible-algebra:
Removing the old one as it would otherwise be highly-visible and was already deprecated.


BoxyUwU:
long live abby

PR #2962, opened 2026-08-12
Fix mentions of "`ai-assisted`" label in llm guidance docs

steffahn:
The actual label we created is called `llm-assisted` now.

It's a dedicated PR for just a small change to the document, but IMO this was sufficiently confusing in practice that it's worth fixing sooner rather than later.

I've also opened a PR to fix the policy document itself: rust-lang/rust-forge#1089

### DIFF
--- a/src/llm-guidance/reviewing.md
+++ b/src/llm-guidance/reviewing.md
@@ -17,12 +17,12 @@
 
 ## Reviewing LLM-created code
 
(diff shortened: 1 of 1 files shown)


steffahn:
r? rustc-dev-guide

jyn514 (approved):

PR #3000, opened 2026-09-03
Fix rust.override-allocator mentions

laundmo:
PR https://github.com/rust-lang/rust/pull/160100 changed it from rust.override-allocator to build.allocator

This changes mentions to the new field


BoxyUwU:
thx :3

